PR #28, opened 2020-08-19
Find and evaluate solutions for hosting images / assets

### LINKED ISSUE #18: [Misc. Research] Find and evaluate solutions for hosting images / assets
## Summary

There are ~3,000 images currently in Drupal for the docs site. There are some engineering challenges that are going to come with these images. One is that we will likely not want to hold all of our images in our repository, so we have to find and evaluate solutions for hosting images.

## Procedure

Open up a PR merging into `research`. If there is no changes in the code (and only research details), you can just make an edit in a text file, since there is no plan to merge this into `main`.  Use the PR template [research_PR_template.md](https://github.com/newrelic/docs-website/blob/main/.github/PULL_REQUEST_TEMPLATE/research_PR_template.md) either by coping the markdown into the PR or by appending the query string `?template=research_PR_template.md` to the URL. Fill out the template by answering the research questions listed below. 

## Research Goals

* [ ] Evaluate solutions for extracting the images 
* [ ] Evaluate solutions for hosting the images 
* [ ] Discuss solutions with engineers, content creators and PMs to determine best path forward

caylahamann:
# Find and evaluate solutions for hosting images / assets
## Research Goals

* [ ] Evaluate solutions for extracting the images 
* [X] Evaluate solutions for hosting the images 
* [ ] Discuss solutions with engineers, content creators and PMs to determine best path forward
Closes #18 

### Github 
There is actually no hard limit actually on the repository size, but it is likely that they will send us an email when it is over 5 GB, so it seems unlikely we would want to host everything in Github. After talking to Clinton, it seems that there is only 475 mb of images (and some of them are old images that are currently not being used?). 
#### Pros 
* Migration most likely a bit easier 
* We don't have to manage storage on S3 
* We don't have to figure out how to store the images and reference them again in our repository
* For content creators, we can easily put everything into one place to manage rather than asking them to use an external source 
#### Cons
* Overtime, the total repo size might push the limits of what is allowed in a Github repository (they prefer less than 1GB and will send warnings at around 5GB) 
* If people are cloning our repo, it might take a bit longer because of the size of the repo

### S3
We can host our images in an S3 bucket. There is an plugin for Gatsby to stitch our images into the GraphQL schema. While migrating content over, we would also want to make sure that we take the images, store them in a S3 bucket, and then reference them in our `` component for markdown. That would be an extra layer to add onto our migration technique. 

There is a plugin to stitch S3 images into the Gatsby GraphQL schema called [gatsby-source-s3-image](https://www.gatsbyjs.com/plugins/gatsby-source-s3-image/). This might make it easier to reference the images and extract them out, as well at utilize some of the capabilities of Gatsby's imaging handling. 

#### Pros 
* We don't have to worry about repo size limits onGgithub 
* We don't have to worry about file organization of these images in our repository 

#### Cons 
* Migration will be more difficult to transfer these images into an S3 bucket and successfully reference them in each of our markdown files 
* Content creators will have to learn another tool for contributing to this repo
* How do we handle outside contributions?

## Recommendation

It seems for now, just using **Github** may be the easiest and best option. I'm not entirely sure we would need to move to S3, and this might be an issue that if (we need to) we can tackle later in the process when we're perhaps dealing with too many images, but for right now the total size of the images works for a Github repository. 

In terms of _other assets_, I'd recommend to use an S3 bucket for including other types of content, like PDFs. I'm not sure how PDFs are currently linked into content and how we might want to migrate this over. 

## Open Questions

* Part of migrating content over for any page will have to include migrating images as well, so do we want to figure this out before we actually start migrating content (should this be apart of the content migration mmf?)
* Is there an easy way to extract all of the images from Drupal? Is there a unique naming convention and reference convention used in Drupal? Should we utilize this in our own repository 

## Resources

* [Disk quotas for Github](https://docs.github.com/en/github/managing-large-files/what-is-my-disk-quota)
* [Importing media assets in Gatsby](https://www.gatsbyjs.com/docs/importing-media-content/#images-svg-and-pdfs)


jpvajda:
I'm fine with GH image storage, since we are only at 475MB , and the limit is 5GB we have plenty of room to grow. Also, S3 adds complexity and we want to reduce complexity as much as we can. question, does GH allow you to pay for more image storage @caylahamann like as an add on feature?

caylahamann:
> I'm fine with GH image storage, since we are only at 475MB , and the limit is 5GB we have plenty of room to grow. Also, S3 adds complexity and we want to reduce complexity as much as we can. question, does GH allow you to pay for more image storage @caylahamann like as an add on feature?

@jpvajda there isn't a pay feature (to my understanding) but there is an open-source plugin for storing large files (https://git-lfs.github.com/) which I think we should utilize

jpvajda:
I added @zstix  / @jerelmiller  as there are several code changes intertwined in here. I'm ok with github for image hosting.

jerelmiller:
I like the idea of using GH. One of the problems I've been thinking about with S3 is how we get new images in place, and when we do, how do we reference the right one in the markdown file itself. Using GH lets us keep it everything in git, which is much easier to manage. Let's see if we can stretch GH this as far as it goes and re-evaluate if it looks like we might get close to hitting those limits.

caylahamann:
Like the sound of that! I'm going to wait for @zstix, @roadlittledawn, and @austin-schaefer  to weigh in, but I'm feeling as if it would be good to stick with GH.

caylahamann:
> I added @zstix / @jerelmiller as there are several code changes intertwined in here. I'm ok with github for image hosting.

Code changes don't matter! I just merged in master, but this is merging into the research branch which will never be merged into main or develop, so it doesn't matter.

zstix:
@caylahamann I know we have a rough estimate about the number of images, but do we have any data about how large those images are? Also, do we have a sense for what other assets we may be needing to store (i.e. PDFs)?

I'm happy to help look into this.

One of my concerns with GitHub is that by increasing our repo size, we are also increasing the time it takes to clone / fork. If an outside contributor needs to wait for 10GB of repo to download before they can make a spelling fix, they may not bother.

Also, we probably don't need to worry about keeping the `research` branch up to date. Seems like it's just adding static to the PR.

caylahamann:
@zstix yes Clinton helped me by doing some digging into that! It's only going to be around 500 MB of images, but we may need to implement some system or check to figure out if images are being used in the repository (just so we don't have a bunch of images that are not used, so this number would go down). In terms of other files, we can utilize the library that I mentioned earlier (Github Large Storage) to hold some of our bigger files, such as PDFs.

PR #29, opened 2020-08-20
Clinton/research categories solution

roadlittledawn:
# IA: Identify a solution for categories

## Research Goals
- [X] Investigate solutions for the taxonomy (categories) that can be reflected in the URL
- [x] Discuss solutions with engineers, content creators and PMs to determine best path forward

_Research for issue #19_
_Strong overlap with issue #11_

## Findings
The categories on the site that are used for forming URLs / organizing content is a taxonomy called Primary Navigation. This taxonomy essentially simulates a directory structure. Those categories are used to form URLs for pieces of authored content. On each of those category pages, by default, is a menu that's generated via the Views module. That Views module uses the URLs to know where in the taxonomy tree it was and display navigation/menus accordingly.

There is sometimes authored content on the category pages called Landing Pages. And API doc menus are a slightly different format than the standard menu.

The left sidebar nav present on single content pages uses an enormous MySQL query to get the term ID from the current page, get the term tree from two levels up and all content links therein, and displays the links/terms hierarchically.

Additionally, content authors would really like the ability to specify what other categories a piece of content should show up in. The current site uses a hacky workaround where authors create a separate doc (called a dummy doc), add it to the location they want the source doc to also show up in, and add redirect to the source doc.

There are couple options for how URL paths are formed. A content author can specify the URL path in frontmatter OR we can tell Gatsby to form the URL based on the directory structure via `gatsby-node.js`. When I last dug into this, I made that piece of frontmatter optional by telling Gatsby to look for `path` in frontmatter and if not there create path via the directory structure. But when I implemented redirects via `@thundersquared/gatsby-redirect-from` plugin, it required having that data in the frontmatter.

We do index each category term hierarchy in each swiftype record (if applicable) to support filtering content to a specific category. The functionality is used on [https://docs.newrelic.com/search] to filter results to the first level only.

## Recommendation
The rigid 3 levels of categories is a non-issue in this setup. Content can easily live in any part of a directory / URL path. And we can add hierarchical data (for directories and markdown files) on build in `gatsby-config.js` to the `allDirectory` GraphQL type to capture and query against those relationships (I have a PoC for that).

I'd strongly recommend not tying any functionalty to a URL. Depending on how we implement redirects, we may be able to specify default URL path pattern if author doesn't set it. If its required, user has to set it, but it could really be anything as long as no other functionality is tied to it.

Where it gets tricky is providing content authors the ability to specify that a piece of content should show up in other directory listings / categories. There's probably a couple different ways of designing that (that likely include using frontmatter), but it's contingent on other decisions:
* if we still use index pages for all categories (a page that essentially is just a menu of links), 
* what the index template looks like, 
* do we use search instead to provide those results (similar to Related resources on developer)

From an SEO perspective, those pages aren't particularly valuable. Search bots like Google will have what it needs via sitemap.xml and traversing the site. Same with swiftype (in fact it would reduce noise and help keep # of pages indexed under our max). Open question from a content author / user of the site if it's valuable to still be able to browse content in a directory-like structure (this will be available via github repo).

Having authored direcory index pages (index.mdx) will likely be useful, especially if we have one main nav for all pages that only links to specific pages and not the entire site directory. That essentially comes out of the box when MDX is set up.

### Navigation / menu designs
Having designs for main nav menu and any other nav menus will help inform a lot of these open questions / how we build site and organize content.

### Category pages / AKA directory index pages
Do we keep these? Do we redesign them? 
Do we keep authored index pages, but do not provide directory listing pages? 
Or is having it available in github sufficient?
Do we provide content authors a way to specify what other directories a piece of content should show up in?

### URLs
Should all segments in a URL path resolve to a page?
Should specifying the URL path be required in frontmatter?
How much, if any, functionality is tied to URL paths?

## Resources
[Example: Category landing page (authored content)](https://docs.newrelic.com/docs/agents/php-agent)
[Example: Category index page](https://docs.newrelic.com/docs/agents/php-agent/getting-started)
[Example: Category index page for API doc content type](https://docs.newrelic.com/docs/agents/php-agent/php-agent-api/view-all-methods)
[PoC of automatic directory index listing via gatsby-config.js](https://github.com/roadlittledawn/gatsby-netlify-cms-example/blob/master/gatsby-node.js)


caylahamann:
Some thoughts reading through this: 
* I agree with @jerelmiller on having all segments resolve to a path, but I think that's probably more of a decision for @djsauble and from my understanding of his IA presentation he intends to have this be the case with the new docs structure 
* I also agree with having the URL path be dependent on the directory structure. From my perspective, I believe this would be the easiest way to manage content. I think we would have to figure out a way to configure redirects for when a file gets moved around (some way to detect when this has been changed and to create the proper redirects). 
* In regards to determining how to link to other content in the site when the URL can change at any moment: would we have to create a redirect for everytime the URL changes? Will we have to scan and replace content on build with links where there are redirects?

jerelmiller:
> I think we would have to figure out a way to configure redirects for when a file gets moved around (some way to detect when this has been changed and to create the proper redirects).

I think we could start simple and have the user's manage that. We will need to be diligent in PRs to check to see if a file has moved and make sure an entry has been added for the redirect. If we find this doesn't scale well, we can find a way to automate/detect it. 

> would we have to create a redirect for everytime the URL changes?

Per my above statement, yes. You do raise a good question though about needing to scan and replace links. Ideally those links target the actual URL and not an old one. That is likely more from an SEO perspective though and less from a UX perspective since linking to a page that redirects will still get you to the right page. It should be a simple find and replace to change links around, but I think the challenge here is remembering to do so.

djsauble:
Yes, keeping the existing behavior where all segments resolve to a path makes a ton of sense from a usability perspective. I'd rather flatten the hierarchy than have interstitial path segments which go nowhere.

Regarding category pages in particular, I think having them act as directory listings by default makes sense, but we should have the capacity to incrementally augment them with landing pages over time. For example:

Initially, `/path/to/category` opens a directory listing page.

Over time, we decide to create a rich landing page experience, and visiting `/path/to/category` takes you to this new page.

However, the directory listing still serves an important purpose for content authors who want to quickly scan all the pages available under a particular category. In this case I'd suggest we design the sidebar nav in such a way that it can serve this important function without our needing to standup parallel pages in cases where there would otherwise be contention for directory listing and landing page at the same path.

And yes, this last is a bit different from what I proposed in the meeting with the team last week, I've been mulling it over and it seems like it would be a bit simpler and less confusing from a user perspective.

roadlittledawn:
*re: auto-configuring redirects*. agree w/ @jerelmiller that we could start by having content authors manage redirects. the tech docs team has done it that way for a long time, so it's something they're used to. i can also imagine scenarios where a file is removed and a redirect is added to another file elsewhere, so auto-detecting may not work in that case.

*re: scan and replace old links with new URLs*. i have some notes re: that in https://github.com/newrelic/docs-website/pull/31. there is a plugin that can check for broken links and hashes. so that might be one way to catch and fix old links before build. though now i'm curious what that plugin does if it encounters and old link that is redirected.

jerelmiller:
> though now i'm curious what that plugin does if it encounters and old link that is redirected.

Even if it only tells you about it, having the _visibility_ is super important. If we can easily see whats broken, or redirected, it gives us an easy way to tell what needs to be fixed, even if that process of fixing it is manual. I still really like your idea of using that plugin to help us there.

zstix:
Just read through everything here and I don't have much else to add other than that I'm in agreement.

I like using the directory structure to create URLs and think that's a much more approachable solution over requiring content creators to add something to frontmatter. It also makes the migration process a little more straightforward. Redirects are a little sticky, but I would rather optimize for creating and maintaining pages rather than for moving them around within the site.

+1 to the conversation about "category index pages". I suspect it would be straightforward to generate a page of links and we can always create a more content-rich index page to replace them as needed.

PR #31, opened 2020-08-24
Clinton/research internal links

roadlittledawn:
# Migration: How to handle internal links post-migration

## Research Goals
* [X] Investigate methods of parsing internal links 
* [X] Investigate methods of testing for broken links within a site 

_Research for issue #16_

## Findings
The current docs site makes extensive use of relative links (so links work on any domain/in any environment). A quick test in another Gatsby project similar to ours looks like relative links will work as expected. Also relative links with a hash value also worked as expected.

It's important to note that many links on the current site eventually go stale. There are links to pages that have been redirected to another page or removed from the site. When I tested that in the other project, it redirected the old, relative URL to the new page.

The current site also has a link checker that looks for broken links and creates a report on a weekly basis. I believe it relies on the proper HTTP status code being thrown though (301, 302, 404). It checks both internal and external links.

The `src` value on `` tags are generally also relative and use a Drupal specific path pattern.

### Relative links
Relative links to other pages on the site should work as expected. We will want to ensure we have redirects set for all pages to properly redirect stale links.

### Check for broken links
We may want to consider using the [gatsby-remark-check-links](https://www.gatsbyjs.com/plugins/gatsby-remark-check-links/) plugin to check for broken links to pages and headings on build.

## Open Questions
How should we handle old links that target a page that's since been removed and no redirect available?
* Maybe `gatsby-remark-check-links` helps avoid this?
* On current site if the page is not found it takes the URL path segments and asks if the user wants to search the site using the words from the segments.

How does `gatsby-remark-check-links` behave when it encounters an old link that has a redirect?

## Resources

[gatsby-remark-check-links](https://www.gatsbyjs.com/plugins/gatsby-remark-check-links/)


caylahamann:
Looking into `gatsby-remark-check-link` would work for this use case, we would just need to make sure our caches are cleared to run it. Otherwise, this looks good to me! @jerelmiller @zstix any thoughts on this?

jerelmiller:
I'll all for that plugin if it helps us detect broken links. I know we will need the computer's help to scale this properly. Thanks for investigating this!

zstix:
+1 to this. That plugin looks rad.

If we wanted to go above and beyond, we could run a GitHub Action on a cron to check for this and report the results (if any) back to us. Might be overkill, but it would appear like an option with this plugin and our current setup.

Issue #34, opened 2020-08-26
Determine where markdown files will reside within the repository


zstix:
## Description
In order to migrate content from the existing site to the new repository, we need to know where each page's markdown file will live within the repository. The original developer proposal was to have a directory structure that mimics the page's category. A page with this category structure `foo > bar > page` would be stored a directory like `src/md/foo/bar/page.mdx`.

Before we start the migration work, we need to confirm that this will work. If another structure is preferred, we should document that here to help inform the upcoming engineering work.

## Related Issue(s)
After reviewing the remaining research tasks, we have decided to break the work in #12 down into multiple smaller tickets. This is one of those tickets.

The developer proposal for this has been identified in #11. This ticket is to track that we still have an open decision to be made.

## Success Criteria
* [ ] Come to a consensus with engineering, PM, design, and other stakeholders for a first-pass solution to where we store content within the repository

zstix:
There is some ongoing conversation about where we should put the files. Some of the options discussed include:
* Put the markdown files in a directory structure that maps to the existing category hierarchy
* Put the markdown files in a "shallow" version of the category hierarchy (i.e. all files in a directory for the top-level category)
* Put the markdown files in a completely flat directory structure (all in one or two directories)

The trade-off of a rigid system (like the first option) is that it's easier for contributors to find files to edit, but it's less flexible for where to put them.

Whatever we decide to do, we will always be flexible post-migration. Our setup in Gatsby doesn't really care where the files live within the `markdown-pages` directory.

We are planning on talking about this during our Monday sync and may schedule a follow-up meeting.

zstix:
Upon further review, this seems like a duplicate of #11

PR #37, opened 2020-08-27
Clinton/research migrate plain text

roadlittledawn:
# [Migration Research] Find and evaluate solutions for migrating Plain Text

## Research Goals
- [ ] Investigate methods of plaintext to mdx file migration
- [ ] Determine/estimate the amount of work for the implementation of each method
- [ ] Determine the amount of manual work (i.e. are there any pages that we can't transfer over programmatically? why?)
- [ ] Create POC for one of the methods of plaintext to mdx file migration for the representative sample of documents

_Research for issue #13_

## Findings
I believe this issue refers to the attribute definition content type. The most important field in this content type is a field called `field_short_description`. Its value is in plain text. This field is what's displayed in NR1 Query Builder attribute definitions.

### Directory structure
This content is not included in the primary navigation on the current site. It's discoverable through a single, form-based page at https://docs.newrelic.com/attribute-dictionary.

The way attribute definitions are organized/modeled makes it tricky to organize in a filesystem. One attribute can belong to multiple events. Most events belong to one data source, but there's a longstanding issue where some (cloud integrations) events can belong to multiple data sources. Those are currently unpublished on the site, but the team did want to eventually fix that. And the current site only attaches attributes to events, and events to data sources. 

**Example: 1 attribute in many events**
This is a common pattern on the site. Many attributes, like `appName`, are in multiple events. And most events come from one data source.
    
    New Relic Browser (data source)
    - PageView (event)
      - appName (attribute)
    - PageAction (event)
      - appName (attribute)
    APM (data source)
    - Transaction (event)
      - appName (attribute)
    -TransactionError (event)
      - appName (attribute)

**Example: 1 event in many data sources**
Attributes with this schema are not published as attribute definition content types on the docs site, nor are they included in JSON for the data dictionary service to consume. This mostly just applies to cloud integrations (ex. [AWS integrations](https://docs.newrelic.com/docs/integrations/amazon-integrations/aws-integrations-list)) attributes. . During a certain period in time, some integrations were set up so that multiple data sources (ex: AWS managed datastores like Elasticsearch, S3, DynamoDB, etc) send their attributes to a generic event and identify the source with a `provider` attribute.
    
    Integrations (data source)
      - Cloud integrations (data source)
        - AWS integrations (data source)
          - DynamoDB (data source)
            - DatastoreSample (event)
              - consumedWriteCapacityUnits (attribute)
          - Elasticsearch (data source)
            - DatastoreSample (event)
              - ClusterStatus.green (attribute)

## Recommendation
* Only migrate the published attributes on the current site (no cloud integrations for time being)
* Migrate attribute definition content that looks like [this](https://gist.github.com/roadlittledawn/a211e47fc0596107b7b7b6bf44f0958e), where the plain text `shortDescription`(and other relevant fields) is stored in frontmatter and longDesription (if it exists) is stored in the body of the file.
* .mdx files should be created successfully if it has just frontmatter and no content.
* File follow up ticket to reevaluate if we can store in JSON instead long term.

### Directory structure
To organize these in a filesystem you're left with two less-than-ideal options:
* duplicate attribute files and event directories where there's 1:many relationships (1 attribute in multiple events; 1 event in multiple data sources). this makes editing a little more painful.
* put all ~3k attribute files in one directory

### One dynamic page for browsing, searching
Similar to the existing [/attribute-dictionary page](https://docs.newrelic.com/attribute-dictionary), we need a way for users to search / filter for attributes. That could be done via:
* `src/pages/attributeDictionary.js`
* `src/attribute-definitions/index.mdx` that uses a template like `src/templates/attributeDefinitionPage.js`

### JSON
We need to ensure the attribute/event definitions are still available via JSON.

## Open Questions
* What directory structure should we go with?
* Does the tech docs team want to try to address the issue with cloud integrations pre-migration, post-migration, never?
* Do we try to also set the dataSource on each attribute?

### Long term considerations
* Should this eventually be stored in JSON file instead or elsewhere?
* Do they each need to have their own page (like current site), or can we just build one view where you can search / filter through one big JSON file? (would need to ensure we maintain JSON endpoints that the data dictionary consumes for NR1 Query Builder)

## Resources
* [Gist: exampleAttribute.mdx](https://gist.github.com/roadlittledawn/a211e47fc0596107b7b7b6bf44f0958e)
* [JSON that data dictionary reads](https://docs-dev.newrelic.com/attribute-dictionary/v1/attributes.json) (requires KB Dev creds)
  * use `eventType` query param to filter results to a specific event.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-[![Community Project header](https://github.com/newrelic/opensource-website/raw/master/src/images/categories/Community_Project.png)](https://opensource.newrelic.com/oss-category/#community-project)
+[![Community Project header](https://github.com/newrelic/opensource-website/raw/master/src/images/categories/Community_Project.png)](https://opensource.newrelic.com/oss-category/#community-project) 
 
 # docs-website
 


zstix:
This is very helpful information, thanks for looking into this! A couple of thoughts based on the open questions ya have:

`What directory structure should we go with?`
This is somewhat TBD while we talk through #34, but I think it probably makes sense for these to all live in a separate directory. The attribute definitions seem quite different from the other content types and having a specific location in the repository for them seems logical to me. As far as how we organize _within_ that directory...I'm flexible.

`Does the tech docs team want to try to address the issue with cloud integrations pre-migration, post-migration, never?`
Not sure if I understand the problem space well enough to weigh in on this. I like your suggestion of holding off on these for now, since they are not published. Perhaps this is something that the IA work @djsauble is doing can help resolve.

`Do we try to also set the dataSource on each attribute?`
If we have access to this information, I don't see why not. As I'm writing this comment, I realize that these might not even make sense as markdown files. If we're not making use of MDX, we might want to even consider these get stored in a different format (JSON, yaml, etc). Just thinking out loud.

roadlittledawn:
**re: setting dataSource on each attribute**. it's not readily available. i think it would require good deal of effort to add that either on Drupal side or when during scripted migration. it's something that i think may be required in order to solve the longstanding problem with those cloud integrations.

caylahamann:
It seems to me that rather than store this in a `mdx` file, it might make more sense to store it in a JSON. Since this is relatively structured data and there is really no need for it to be in a mdx file, I think it would also help our performance because when building our files we need to query graphql in order to build these pages. If it is stored in a JSON, we might have more control. What are the reasons that we wouldn't want to store it in a JSON? @zstix @roadlittledawn

roadlittledawn:
@caylahamann ideally, i think we would store it in JSON since that's the primary output to the data dictionary service anyway. if we can find a way to expose the event/attribute definitions in JSON so data dictionary service can consume it, i'd say it's viable. and yeah, that removes the need to generate ~2200k HTML pages, so that should decrease build times.

if we store it in JSON in repo, is it accessible via an HTTP call?

roadlittledawn:
or is it possible to set up github action to store that JSON in an S3 bucket on build?

zstix:
@roadlittledawn Gatsby builds the site into static HTML and JS files. It _may_ be possible that we can also host raw JSON files which could be accessed with a plain GET call.

That said, this feels like a different use case than the rest of the site. Is there a reason that the attribute definitions were part of the docs site to begin with? Is there a reason this wasn't made into it's own API? Sorry if I'm having a hard time following along!

roadlittledawn:
@zstix yeah, good question. long story short (can chat more in detail if you'd like the full download): historically, attribute definitions sat in a repo or repos where there weren't great processes for updating them. those definitions were also separately defined in scattershot on the docs site where they were kept more up to date. there was a decision made to have Drupal be the central "repo" for them where non-technical users could easily add/edit definitions AND there was a full tech docs team behind that to review / edit / vet that content.

yep, they are really are a different beast. i think long term (or if we can figure out something prior to migration), it's definitely a good idea to rethink where these live, how they're managed, how they're served to NR1, etc.

caylahamann:
Another alternative is just source the content from the JSON during build time. We can have multiple sources and build those into the static HTML / JS files https://www.gatsbyjs.com/docs/sourcing-content-from-json-or-yaml/

jerelmiller:
I also like the idea of using JSON to store this data. It would be easy to consume and update. Gatsby gives us the `createPage` API, so it would be trivial for us to create pages for each event/attribute definition in that case.

djsauble:
@roadlittledawn @zstix What's the issue with cloud integrations? I read the description but I don't understand it.

zstix:
@djsauble I think the issue with that attribute is that it belongs to multiple data sources. This seems to be an issue with the old system, but should not be a problem with how we are handling it in Gatsby (a more flexible setup for how things can be related).

PR #38, opened 2020-08-27
Cayla/research redirects

### LINKED ISSUE #15: [Migration Research] Find and evaluate solutions for migrating Redirects
## Summary

In the current docs site, there are over 5,550 redirects from old pages to current docs content. We need to figure out methods for handling these redirects, whether it be through Gatsby, Amplify, or some other technology. Currently, the redirects are listed out in Drupal (the CMS the docs site uses) as referring to other node numbers which represents pages. We need to figure out how to match the redirects from the nodes to the URLs and then configure the redirects.

## Procedure

Open up a PR merging into `research`. If there is no changes in the code (and only research details), you can just make an edit in a text file, since there is no plan to merge this into `main`.  Use the PR template [research_PR_template.md](https://github.com/newrelic/docs-website/blob/main/.github/PULL_REQUEST_TEMPLATE/research_PR_template.md) either by coping the markdown into the PR or by appending the query string `?template=research_PR_template.md` to the URL. Fill out the template by answering the research questions listed below. 

## Research Goals

* [ ] Investigate methods of transferring the redirects 
* [ ] Determine/estimate the amount of work for the implementation of each method 
* [ ] Create POC for one of the methods

caylahamann:
# Migration for redirects

## Research Goals

* [x] Investigate methods of transferring the redirects 
* [x] Determine/estimate the amount of work for the implementation of each method 
* [x] Create POC for one of the methods

closes #15 

## Findings

This is heavily tied with the work done for migration other types of content. As dicussed with @jerelmiller  it would be best to use an [abstract syntax tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree) (AST) in order to transfer over the content. An AST is a tree representation of source code for a markup or programming language. For example, the tree for HTML would look something like this.

Untitled

For a markdown file, it might have a structure where it specifies root, paragraph, YAML (where the frontmatter is), emphasis, ect. We can use this method of ASTs in order to transform one form of content to another. When transforming this content, we will be able to stitch in the redirects. As discussed by @ Clinton two days ago in our research meeting, we will likely be able to add the redirects in the JSON from the exported HTML/Markdown/PlainText files. In our migration, we can add the redirects to the frontmatter, and then the redirects will be handled by Gatsby. We also looked at placing HTML in the body of markdown, which can also be useful for this migration and for later uses of trying to convert components of the HTML / markdown files.

## Recommendation

Using this set of libraries ([mdast](https://github.com/syntax-tree/mdast) for markdown, [remark-mdx](https://www.npmjs.com/package/remark-mdx) for mdx [hast](https://github.com/syntax-tree/hast) for html, [unified](https://github.com/unifiedjs/unified) as the root library) we will be able to manipulate our content in order to transform or add anything to our content. 

I added a script (called [redirect.js](./redirect.js) but it's a little bit of a misnomer because it just parses one of our mdx files) to demonstrate what some of this parsing would look like for one of our mdx files. I worked too long on actually trying to change the frontmatter and I decided it wasn't worth the load this time around.

## Concerns

* Gaining an understanding of how these libraries work could give us a lot of capabilities in migrating the content, but it could be a little bit of a lift so we need to figure out if we want to invest in this method. 

## Resources

All linked above!


zstix:
@caylahamann / @roadlittledawn  Out of curiosity, how are the redirects currently stored?

I think the AST stuff has a ton of potential when it comes to converting the different content types into MDX, but I'm a little unclear how this applies to redirects.

caylahamann:
@zstix so we can use this to stitch the redirects into the frontmatter, and gatsby is able to handle the redirects from the frontmatter! @roadlittledawn said that we might be able to add the redirects to the json, and we can use this in the migration

roadlittledawn:
@caylahamann . yep, i believe we can include old URLs that redirect to a page in its JSON. 

@zstix i believe we have 5k + redirects that are actively accessed on the site

roadlittledawn:
actually, looks like the redirect data has to be in a separate JSON endpoint, FWIW

zstix:
@caylahamann Got it. Using the AST stuff to _add_ the links, not to extract them. That makes a lot more sense.

If they're stored in JSON then it's probably a straightforward process to fetch them. This all seems great to me 👍

caylahamann:
awesome! I'm going to close this PR and this issue because it seems that we can use this structure in our migration :)

PR #64, opened 2020-09-17
Dynamically fetch docs

zstix:
## Description

Rather than using a hard-coded set of IDs, the script can now fetch a batch of docs by content type. This is done by using a new endpoint available in the dev instance (`/list`).

The script now brings over 10 of each content type. Gatsby will start, but there is some additional cleanup that needs to be done to these new files (out of the scope of this PR).

### DIFF
--- a/scripts/constants.js
+++ b/scripts/constants.js
@@ -1,6 +1,6 @@
 const path = require('path');
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


caylahamann (approved):
this looks good to me!

PR #66, opened 2020-09-18
Migration Script Refactor

zstix:
## Description

This PR moves the logic for our migration script into separate files, sets up a `yarn migrate` script, and makes it easier to convert docs to _other_ formats.

This is a hefty PR, but it's mostly just moving the existing logic around.

### DIFF
--- a/scripts/constants.js
+++ b/scripts/constants.js
@@ -13,4 +13,24 @@ module.exports = {
     'nr1_announcement',
     'attribute_definition',
   ],
+
+  GATSBY_CONTENT_TYPES: {
+    page: 'page',
+    api_doc: 'apiDoc',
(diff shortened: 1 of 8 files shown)


scripts/constants.js
```diff
@@ -1,16 +1,38 @@
...
-    'api_doc',
-    'release_notes',
-    'release_notes_platform',
-    'troubleshooting_doc',
-    'nr1_announcement',
-    'attribute_definition',
-  ],
+  TYPES,
+
+  GATSBY_CONTENT_TYPES: {
```
zstix, line 19: I'm going to move this to to the frontmatter-specific file since it's not used anywhere else.

caylahamann (approved):

PR #83, opened 2020-09-22
Codemod framework

### LINKED ISSUE #55: [Prototype] Create a framework for transforming migrated content
## Summary

We may discover issues with the HTML when migrating the content from the existing docs site to MDX. Furthermore, there may be HTML snippets we need to convert from one format to another. We need to create the tools and the framework for generating codemods that we can run on all the MDX files.

## Acceptance criteria

- [ ] Can run a codemod on an MDX file and transform it
- [ ] Allow others to create custom codemods that can transform content in some way

jerelmiller:
Closes #55 

## Description

Adds a codemod framework that uses unified and babel to parse the MDX files into
an AST and run transforms. Defines a node script called `codemod` that will
allow us to run the codemods on our MDX files.

This is currently in a work-in-progress and only runs the codemods on a single
hardcoded file. Eventually this will need to be replaced to run on a glob of the
src/content directory. This PR also adds some sample codemods that we can remove
once we define real ones.


zstix (approved):

codemods/index.js
```diff
@@ -0,0 +1,8 @@
+const coolButton = require('./cool-button');
```
caylahamann, line 1: It's good that we committed these now, but we will probably want to remove them later on. Maybe we should have an example code mod?

PR #85, opened 2020-09-22
Simplify JSX codemod transform

jerelmiller:
## Description

Simplifies the JSX transform by using @babel/core instead of the separate babel packages. This also means the codemods will be fully compatible as a babel plugin.


zstix (approved):
This looks dope, nice work

PR #86, opened 2020-09-23
Convert HTML to JSX during migration

### LINKED ISSUE #33: What adjustments needed to be made to raw HTML when migrating to MDX?
## Description
During the migration process, we will have access to the raw HTML that represents a page. Whether we get this from a JSON output or from MySQL, we will need to process the HTML content before we can add it to a `.mdx` file.

During our initial research, we copied HTML for a page into a `.mdx` file in the dev site repository as a simple test. It became clear that some adjustments needed to be made before Gatsby could properly serve the page. Some of the initial findings include:

* Specific elements (such as the `dl` tag) have requirements for how they are being indented
* Special characters (i.e. `{` and `}`) need to be escaped or removed.

We should take some time to identify the most common adjustments / santization steps. The goal is to have a list of functions we will need to implement in the engineering stage and to uncover any potential issues _before_ we get started developing.

## Related Issue(s)
After reviewing the remaining research tasks, we have decided to break the work in #12 down into multiple smaller tickets. This is one of those tickets.

## Success Criteria
* [ ] Create a list of adjustments that need to be made to raw HTML documents
* [ ] Document findings in this issue so that they can be used to inform the automation code (upcoming MMF)

zstix:
## Description
Keep `<div>`, `<table>`, and `<dl>` tags in the markdown files during migration. These will be used to convert into components later on.

## Related Issue(s)
* Closes #33

## Commits
* feat: converting problematic HTML to JSX
* chore: keeping `dl` tags for future component work

### DIFF
--- a/scripts/converters/to-markdown.js
+++ b/scripts/converters/to-markdown.js
@@ -1,9 +1,12 @@
 const path = require('path');
 const TurndownService = require('turndown');
+const HTMLtoJSX = require('htmltojsx');
 
 const getCategories = require('../utils/get-categories');
 const getFrontmatter = require('../frontmatter/get-frontmatter');
 const { TYPES, BASE_DIR } = require('../constants');
+
+const htmlToJSXConverter = new HTMLtoJSX({ createClass: false });
 
 const toMarkdown = (doc) => {
   const dir = path.join(BASE_DIR, ...getCategories(doc.docUrl));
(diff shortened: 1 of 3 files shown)


jerelmiller (approved):

scripts/converters/to-markdown.js
```diff
@@ -1,10 +1,13 @@
 const path = require('path');
 const TurndownService = require('turndown');
+const HTMLtoJSX = require('htmltojsx');
 
 const getCategories = require('../utils/get-categories');
 const getFrontmatter = require('../frontmatter/get-frontmatter');
 const { TYPES, BASE_DIR } = require('../constants');
 
+const htmlToJSXConverter = new HTMLtoJSX({ createClass: false });
```
caylahamann, line 9: having trouble understanding the documentation for this site, what does setting createClass to false do here?

jerelmiller: This is the only html -> jsx converter I could find. Unfortunately its super old and still uses the `React.createClass` API (it was the primary API before JS classes were mainstream). Here we just want the raw JSX and don't want it wrapped in a component via `React.createClass`. Disabling this ensures we get the raw JSX output back.

PR #94, opened 2020-09-28
Use MDX v2

jerelmiller:
## Description

Use MDX v2 packages. v2 has a much more robust AST parsing engine to handle
embedded JSX. This will help us write much more robust codemods as well as
handle embedded markdown within JSX.

## Commits

- chore: install mdx v2 packages
- refactor: all codemods now operate on the MDXAST
- chore: remove old babel codemods that are no longer needed
- chore: remove jsx and jsxComment unified plugins since v2 handles this now


roadlittledawn (approved):
:shipit:

PR #96, opened 2020-09-29
Improve table parsing + codemods

jerelmiller:
## Description

Handles table output/parsing/codemods to ensure the output is rendered as expected.

* Ensures paragraphs nested inside table cells are parsed/compiled correctly
* Ensures paragraphs following lists inside table cells are parsed/compiled correctly
* Fixes an issue with indented code blocks 
* Adds some helpers for updating the MDXAST
* Updates fenced code blocks to use `\`` as the fence character
* Adds the option to run `DEBUG=true yarn migrate` to write the raw HTML files, which we can then use to compare the migration output against
* Re-runs the migration script against the latest turndown rules


gatsby-config.js
```diff
@@ -67,6 +69,7 @@ module.exports = {
     {
       resolve: 'gatsby-plugin-mdx',
       options: {
+        remarkPlugins: [indentedCodeBlock],
```
caylahamann, line 72: why is this added here?

scripts/convert-docs.js
```diff
@@ -22,6 +29,20 @@ const convertDocs = (docs) => {
     fs.writeFile(fileName, content, (err) => {
       if (err) logger.error(`Could not create ${fileName}.`);
     });
+
+    if (process.env.DEBUG) {
```
caylahamann, line 33: is this turning it back into html?

scripts/utils/htmlToMarkdown.js
```diff
@@ -0,0 +1,124 @@
...
+    filter: ['pre'],
+    replacement: (_content, node, options) => {
+      const code = node.textContent;
+      const fenceChar = options.fence.charAt(0);
+      const fenceInCodeRegex = new RegExp(`^${fenceChar}{3,}`, 'gm');
+
+      let fenceSize = 3;
+      let match;
+
+      while ((match = fenceInCodeRegex.exec(code))) {
```
caylahamann, line 69: could you explain what's going on here? I'm a bit confused

codemods/paragraphsInsideTableCells.js
```diff
@@ -0,0 +1,52 @@
...
+
+        removeAttribute('data-fixnewline', p);
+        removeChild(p, parent);
+      }
+    );
+
+    visit(td, isMdxElement('p'), (p) => {
+      p.type = 'paragraph';
+
+      delete p.name;
```
caylahamann, line 46: what is happening exactly here? why are the name and attributes being deleted?

caylahamann (commented):
just a few questions!

scripts/convert-docs.js
caylahamann: ooops nevermind I read the description! Ignore this comment :)

gatsby-config.js
jerelmiller: Unfortunately [there is an issue](https://github.com/mdx-js/mdx/issues/1283) with MDX v2 that causes indented code blocks from parsing correctly. Whether or not this is intentional, it means that indented code blocks output to the screen with the backticks in the output.

Because its a _parsing_ issue, this means that a simple codemod won't work to solve this issue. Our `indentedCodeBlock` codemod also doubles as a remark plugin, so I've added this hear to patch the issue so that we can have indented code blocks rendering as we expect. Check out that GitHub issue if you want a deeper dive into the issue.

scripts/convert-docs.js
jerelmiller: Ish. It is writing the raw HTML back to a file with a `.html` extension. I used this a couple times and thought it might be helpful to compare the output we get from turndown with the original HTML.

scripts/utils/htmlToMarkdown.js
jerelmiller: I meant to go back and add a comment here that points to the original source (which I will do as a followup). This comes from the [original implementation](https://github.com/domchristie/turndown/blob/master/src/commonmark-rules.js#L111-L134) in Turndown. This handles the case of a nested fenced block inside of another. For example, our [developer website component guide](https://github.com/newrelic/developer-website/blob/develop/COMPONENT_GUIDE.md#usage-4) has examples of using fenced code blocks:

````
```jsx
console.log('Hello')
```
````

Note here how I had to wrap that with 4 backticks to get it to render the 3 backticks in the code block. This `while` essentially handles that edge case.

Does that make sense?

codemods/paragraphsInsideTableCells.js
jerelmiller: As per the MDAST's interface for a [`Paragraph`](https://github.com/syntax-tree/mdast#paragraph) type, it only includes the `type` and `children` properties. Since I'm converting a JSX `<p>` tag to a markdown `paragraph`, I figured it best to remove the additional attributes from the AST node.

Likely it wouldn't hurt anything to leave it, but for completeness sake, I removed them to match the `Paragraph` interface.

jerelmiller:
@caylahamann I'm going to follow-up with a few more code comments in response to some of your questions.

gatsby-config.js
caylahamann: oooh gotcha! thank you

scripts/utils/htmlToMarkdown.js
caylahamann: ahhh yes that makes perfect sense

caylahamann (approved):
thank you for adding all the comments!!!! it really helped clarify things :) looks all good to me

PR #97, opened 2020-09-29
chore: use vfile-glob to read files in codemod processing

roadlittledawn:
Use vfile-glob to more flexibly read files for codemod processing

### DIFF
--- a/scripts/codemod.js
+++ b/scripts/codemod.js
@@ -1,16 +1,13 @@
 const createProcessor = require('./utils/codemod/createProcessor');
-const { read, write } = require('to-vfile');
+const { write } = require('to-vfile');
+const vfileGlob = require('vfile-glob');
 const codemods = require('../codemods');
 
-const FILE =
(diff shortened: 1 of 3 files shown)


scripts/codemod.js
```diff
@@ -1,16 +1,13 @@
...
-  const file = await read(path);
+const process = async (file) => {
   const processor = createProcessor({ codemods });
 
   await processor.process(file);
   await write(file, 'utf-8');
 };
 
-process(FILE);
+vfileGlob('./src/content/**/*.mdx').subscribe((file) => process(file));
```
caylahamann, line 13: what does `vfileGlob` do?

jerelmiller: Much like the `read` from `to-vfile` (which we were using before), this allows you to pass a glob pattern and get vfile's for each file in the glob. This allows us to process each file in the glob with the codemod (here being all `.mdx` files inside of `src/content`)

jerelmiller: https://github.com/shinnn/vfile-glob

caylahamann (approved):

PR #98, opened 2020-09-30
Add collapser component

roadlittledawn:
Add collapser component that's similar in look to NR1 Help Center style. You can pass an `id` string so the row is directly linkable via a hash in URL. You can specify if the row should be open by default on page load.

Kept it fairly basic. See Notes for things we may want to eventually add.

## Notes
* May want to use `react-spring` for better animating the height of the hidden/shown content.
* Added GlobalHeader component via the theme in order to toggle light/dark mode and update component style.
* May eventually want to add a keybinding via the theme to open all rows when typing `f` (when user finds on page).
* May eventually want to add a button at top of Collapser to open all rows on click.

## Screenshot
image


src/components/Collapser.css
```diff
@@ -0,0 +1,62 @@
+@import url('https://fonts.googleapis.com/css?family=Open+Sans:400,600&display=swap');
```
jerelmiller, line 1: Since the CSS is all in emotion now, you should be able to just delete this file.

src/content/kitchen-sink.mdx
```diff
@@ -0,0 +1,25 @@
...
+contentType: page
+template: basicDoc
+topics:
+  - Mobile apps
+  - New Relic Mobile apps
+  - Android app
+japaneseVersion: ''
+---
+
+<Collapser>
```
jerelmiller, line 12: Taking a look at how this component is put together, it looks like each `CollapserRow` is also wrapped by its own `Collapser` component. Because of that, does it even make sense to wrap these rows from a user perspective?

I could however see this being useful for making sure a "collapser group" has the proper spacing in between each collapser row. What if we rename these slightly?

`Collapser` -> `CollapserGroup`
`CollapserRow` -> `Collapser`

The thinking is this: I should be able to use a single `Collapser` on its own. If I only have a single collapser, I shouldn't need to embed it in a `CollapserGroup` (hence the name) because there are no other collapser's I'd need to add space between. If I want multiple `Collapser`s next to each other (like this example), then I'd wrap it in a `Collapser` group to make sure there is proper space between them.

Thoughts on this idea?

src/components/CollapserRow.js
```diff
@@ -0,0 +1,104 @@
...
+  const [setHeight, setHeightState] = useState(isOpen ? 'auto' : '0px');
+
+  const content = useRef(null);
+  const toggleCollapser = () => {
+    toggleOpen(!isOpen);
+    setHeightState(isOpen ? '0px' : `${content.current.scrollHeight}px`);
+  };
+
+  return (
+    <Collapser>
```
jerelmiller, line 27: I see each row is wrapped by its on collapser. Do we want this? From a component design perspective, components tend to scale a bit better if components don't add their own surrounding whitespace (the callout components are the one exception to this). We want the parent components to layout their children so that they can space them accordingly. This ensures you can use a component in multiple contexts without having to rejigger the spacing each time you do.

Because of that, I think this surrounding `<Collapser>` should actually be removed and instead have the the consumer of the API use it.

src/components/CollapserRow.js
```diff
@@ -0,0 +1,104 @@
...
+  transition: transform 0.6s ease;
+  color: var(--accent-text-color);
+  ${isOpen &&
+  `
+transform: rotate(180deg);`}
+`;
+
+const CollapserRow = ({ title, id, openByDefault, children }) => {
+  const [isOpen, toggleOpen] = useState(openByDefault);
+  const [setHeight, setHeightState] = useState(isOpen ? 'auto' : '0px');
```
jerelmiller, line 18: [nit] What about naming this `height` instead of `setHeight`. Then your `setHeightState` could become `setHeight`, which to me reads a bit more natural.

src/components/CollapserRow.js
```diff
@@ -0,0 +1,104 @@
...
+              .dark-mode & {
+                background-color: var(--color-dark-100);
+              }
+            }
+          `}
+        >
+          <h5
+            id={id}
+            css={css`
+              font-family: 'Open Sans', sans-serif;
```
jerelmiller, line 61: We shouldn't need this since the font family is already defined site-wide.

src/components/CollapserRow.js
```diff
@@ -0,0 +1,104 @@
...
+              font-weight: 600;
+              font-size: 14px;
+              text-align: left;
+              margin-top: 0;
+              margin-bottom: 0;
+              font-size: 16px;
```
jerelmiller, line 67: Should we try and maintain the same defaults for an `h5` as we do the rest of the site? In that case, you'd likely want to remove the `font-size`, and `font-weight` properties. I think the `margin-top: 0` and `margin-bottom: 0` make sense here since you're trying to remove it for the purposes of this component. Do we need to override the default size and weight for the `h5`?

src/components/CollapserRow.js
```diff
@@ -0,0 +1,104 @@
...
+              }
+            }
+          `}
+        >
+          <h5
+            id={id}
+            css={css`
+              font-family: 'Open Sans', sans-serif;
+              font-weight: 600;
+              font-size: 14px;
```
jerelmiller, line 63: You can remove this since its a duplicate of line 67

src/components/CollapserRow.js
```diff
@@ -0,0 +1,104 @@
...
+              text-align: left;
```
jerelmiller, line 64: This is the default, and it doesn't look like anything overrides it, so you should be safe to remove this.

src/components/CollapserRow.js
```diff
@@ -0,0 +1,104 @@
...
+  };
+
+  return (
+    <Collapser>
+      <div
+        css={css`
+          display: flex;
+          flex-direction: column;
+          border-radius: 3px;
+          margin-bottom: 8px;
```
jerelmiller, line 33: Seeing as this `div` is the container of the whole component, I think we should let the parent component define the space between the rows rather than have this define it. Doing so would help us use this component in more places without having to rejigger the space each time.

src/components/CollapserRow.js
```diff
@@ -0,0 +1,104 @@
...
+          </h5>
+          <Icon
+            name={Icon.TYPE.CHEVRON_DOWN}
+            size="1.25rem"
+            css={collapserIcon(isOpen)}
+          />
+        </button>
+        <div
+          ref={content}
+          style={{ maxHeight: `${setHeight}` }}
```
jerelmiller, line 80: Since `setHeight` is already a string, wrapping it in a string doesn't do much here. You can just set the value directly:

```
style={{ maxHeight: setHeight }}
```

That being said, one really cool thing about emotion is that you can interpolate dynamic values right into the styles. Rather than using an inline style here (which translates to the `style` property on the element in the DOM), you can add this to the `css` prop below:

```js
css={css`
  overflow: hidden;
  transition: max-height 0.6s ease;
  max-height: ${setHeight};
`}
```

jerelmiller (changes requested):
Had a few suggestions for you. Nice work on this! Would you mind also posting a screenshot of the component in the PR description? Thanks!

src/components/Collapser.css
roadlittledawn: oops, thought i did 😬

src/content/kitchen-sink.mdx
roadlittledawn: my thinking was eventually in the `<Collapser />` component we may add things like a button to open all `<CollapserRow>` rows on click. buuut that and maybe some styling on the container were the only use cases i could think of, really.

src/components/CollapserRow.js
roadlittledawn: when i removed the font-size i feel like it made the text just a bit too small. 16px or 1rem seemed about right. i get what you're saying though. would be more ideal if we didn't override default heading style.

we could always use another tag though too. i used `<h5>` because the current site primarily uses `<h2>` and `<h3>`, and on rare occasion `<h4>`.

src/content/kitchen-sink.mdx
roadlittledawn: Thinking on this more, I think I get what you're puttin' down. Will work on that today, may ping you for some input :)

roadlittledawn:
@jerelmiller Just pushed changes based on your feedback to this branch. One question I had was how to add margin if we now have a standalone `<Collapser />` component. I was thinking margin should be set there. Wasn't sure what style needed to be set for `<CollapserGroup>` right now

src/content/kitchen-sink.mdx
jerelmiller: I probably should have provided an example to better explain this. Ideally I'd be able to use the collapser 1 of 2 ways:

```js
// on its own
<Collapser title="My title">
  Stuff
</Collapser>

// When I have multiple
<CollapserGroup>
  <Collapser title="My title">
    Stuff
  </Collapser>
  <Collapser title="My title 2">
    Stuff 2
  </Collapser>
</CollapserGroup>
```

`CollapserGroup` would be responsible here for adding the spacing between each of the `Collapser`s. Does that help clarify?

jerelmiller: > we may add things like a button to open all <CollapserRow> rows on click

That should still be possible with this design if we put it in the `CollapserGroup`. It may not make sense to show a "open all" button anyways if there is a single collapser.

jerelmiller (approved):
This looks 💯 . Thanks so much for making those updates!!!

jerelmiller (approved):

PR #99, opened 2020-09-30
Run existing codemods on existing docs

jerelmiller:
## Description

Runs the codemods on the existing docs to keep the docs up-to-date with the current changes.


roadlittledawn (approved):
Looking good! Only thing I'm noticing thus far is that we're treating inline `<code>` elements like code blocks (via `MDXContainer` component). I imagine we'll likely address that in the code block codemod ticket soon.

:shipit:

jerelmiller:
Ya that will be addressed soon as we update the existing clamshell components

PR #104, opened 2020-09-30
Keep icons inside the `i` tag

jerelmiller:
## Description

Ensures icons remain inside the `i` tag with the `className` to ensure we can migrate these more easily in the future.


roadlittledawn (approved):
reviewed via pairing. :shipit:

PR #105, opened 2020-09-30
Patch turndown

jerelmiller:
## Description

Patches `turndown` to include the changes that fix the issue where `iframe`s are removed from the output due to how the library parses blank nodes.

This uses [patch-package](https://github.com/ds300/patch-package) to patch the changes from https://github.com/domchristie/turndown/pull/326 into the code. This will be safe to remove once turndown releases a new version with this fix.


roadlittledawn (approved):
:shipit:

PR #107, opened 2020-10-01
Refactor scripts to use vfiles

jerelmiller:
## Description

Refactors scripts to use `vfile` objects throughout the scripts. Also updates the migration script to run codemods before saving the files.

By using `vfile` objects throughout, we have a consistent interface to make all of the changes for a file. This allows us a few luxuries:

* We can perform all the write operations at the end of processing
* We unlock the ability to report on warnings/errors during all migration operations, including codemods (since `unified` utilizes `vfile`)
* We can execute the codemods on each file independent of the codemod script


roadlittledawn (approved):
When I ran `yarn migrate` it successfully used the codemods to convert the content. Looks good!

At some point it would be useful to get quick walkthrough of the things that were moved around.

:shipit:

PR #110, opened 2020-10-02
Convert youtube iframes to Video component also

roadlittledawn:
Modified the video codemod a bit to account for youtube video `<iframe>` elements as well. This codemod should throw error if it isn't able infer the `id` and `type` from the `src` attribute.

## Assumptions
I'm 99% sure we don't use `<iframe>` elements on the site for anything other than embedding these two types of videos. So in `html-to-markdown.js` I changed the rule filter to only look for `<iframe>` elements (the youtube tag didn't have a class).

I also assume the domain for wistia and youtube are static. I'm confident the wistia domain is static, but not sure about youtube. Luckily we only have one youtube video, so probably safe to assume for time being.


scripts/utils/migrate/html-to-markdown.js
```diff
@@ -145,9 +145,7 @@ turndown
   })
   .addRule('videos', {
     filter: (node) => {
-      return (
-        node.nodeName === 'IFRAME' && node.classList.contains('wistia_embed')
-      );
+      return node.nodeName === 'IFRAME';
```
jerelmiller, line 148: [nit] Since this is just comparing against the `nodeName`, you can use the shortcut: `filter: 'iframe'`

codemods/videos.js
```diff
@@ -1,22 +1,36 @@
...
+    propsObj.type = 'wistia';
+    propsObj.id = src.match(/iframe\/([a-zA-Z0-9]+)\??/)[1];
+  }
+  if (domain === 'www.youtube.com') {
+    propsObj.type = 'youtube';
+    propsObj.id = src.match(/embed\/([a-zA-Z0-9]+)\??/)[1];
+  } else {
+    throw new Error('Video type not recognized.');
+  }
+  return propsObj;
```
jerelmiller, line 21: [nit] Rather than creating an object, then filling in its values in the `if` blocks, and returning it, you can just `return` in the blocks themselves. I typically find that a less noisy for these kinds of things.

That might look something like:
```js
const domain = src.split('/')[2];

if (domain === 'fast.wistia.net') {
  return {
    id: src.match(/iframe\/([a-zA-Z0-9]+)\??/)[1],
    type: 'wistia',
  }
}

if (domain === 'www.youtube.com') {
  return {
    id: src.match(/embed\/([a-zA-Z0-9]+)\??/)[1],
    type: 'youtube',
  }
}

throw new Error('Video type not recognized.');
```

codemods/videos.js
```diff
@@ -1,22 +1,36 @@
...
 const {
   addAttribute,
-  hasClassName,
   findAttribute,
+  isMdxBlockElement,
 } = require('./utils/mdxast');
 
+const getVideoProps = (src) => {
+  const propsObj = {};
+  const domain = src.split('/')[2];
```
jerelmiller, line 10: [nit] There is also a [`URL`](https://nodejs.org/docs/latest-v12.x/api/url.html) module you can create to get url information. If you like not having to regex or split on the URLs yourself, that module can come in real handy since its pretty easy to use and extract URL information out of.

```js
const url = new URL(src)

if (url.hostname === 'fast.wistia.net') {
  // ...
}
```

jerelmiller (commented):
I only had `[nit]` comments. Up to you if you want to make those changes or leave what you have. Nice work on this! This is awesome 💯

jerelmiller:
I'll approve once you've had a chance to review my comments and decide whether you want to take it or leave it.

codemods/videos.js
roadlittledawn: yes! i was contemplating using a module, but wasn't sure if that would preferable to just splitting the string. i'll update to use that

scripts/utils/migrate/html-to-markdown.js
roadlittledawn: ah right. make sense.

codemods/videos.js
roadlittledawn: yep, good call.

codemods/videos.js
```diff
@@ -1,22 +1,36 @@
...
+  if (domain === 'fast.wistia.net') {
+    propsObj.type = 'wistia';
+    propsObj.id = src.match(/iframe\/([a-zA-Z0-9]+)\??/)[1];
+  }
+  if (domain === 'www.youtube.com') {
+    propsObj.type = 'youtube';
+    propsObj.id = src.match(/embed\/([a-zA-Z0-9]+)\??/)[1];
+  } else {
+    throw new Error('Video type not recognized.');
```
jerelmiller, line 19: While I think of it, you might actually consider using the [`vfile` message API](https://github.com/vfile/vfile#vfilemessagereason-position-origin) (`message`, `info`, or `fail` for this error. The nice side effect of that is that you can point to a specific location in the AST. and have that print nicely to the screen. The `file` comes through as the second prop to the plugin function below:

```
const videos = () => (tree, file) => {
  // ...
  visit(..., node => {
    file.fail(new Error(`Video type not recognized for url: ${url}`), node.position)
  })
}
```

roadlittledawn: ah i was going to ask about how to use vfile messaging API and spaced it. yeah i like that. making change now. thanks!

jerelmiller (approved):

jerelmiller:
This looks great! Nice work!

PR #112, opened 2020-10-03
Add basic layout implementation to prettify docs a bit

jerelmiller:
## Description

Adds a very basic layout so that we can render the docs content in a little prettier way. This also now renders the banner explaining that the site is in a preview state.

## Screenshots

Screen Shot 2020-10-02 at 5 00 04 PM 1


zstix (approved):
Looks good to me!

src/layouts/MainLayout.js
```diff
@@ -0,0 +1,41 @@
...
+      }
+    }
+  `);
+
+  return (
+    <>
+      <GlobalHeader />
+      <main
+        css={css`
+          margin: 0 auto;
```
caylahamann, line 26: This comes from the theme correct? Or I guess, why is it that we query for this an add it to main? wouldn't make sense to add this some global styles that applies for every instance of the theme?

package.json
```diff
@@ -15,6 +15,7 @@
     "front-matter": "^4.0.2",
     "gatsby": "^2.24.47",
     "gatsby-image": "^2.4.16",
+    "gatsby-plugin-layout": "^1.3.12",
```
caylahamann, line 18: So i looked a little bit into the documentation, and it says: 
> This plugin enables adding components which live above page components and persist across page changes.

And I was wondering if you would be able to explain this more to me?

src/layouts/MainLayout.js
jerelmiller: > why is it that we query for this

Could you clarify what `this` is in this context? I'm not sure I follow your question.

All we are doing here is wrapping the content of the page in a `main` tag and applying these styles to `main`. The `margin: 0 auto` centers the element in the middle of the page. Without this, the layout would show up on the left side of the screen. We don't currently have anything in the theme that targets the `main` tag, so nothing here is coming from the theme.

This is not expected to be permanent. I wanted to docs to look just a little bit nicer than the super raw format (that previously had no spacing on the edge of the content, no global header, etc). Once we dive more into the navigation, I would expect this to change. Definitely take the implementation of this layout with a grain of salt since its not final or permanent 😄

package.json
jerelmiller: Yes, [this section](https://www.gatsbyjs.com/plugins/gatsby-plugin-layout/?=layout#why-would-you-want-to-reimplement-the-v1-layout-behavior) of those docs help explain it a bit more in-depth, but lets illustrate with a couple examples to try and explain why this plugin helps.

Lets take the following React code:
```js
const Routes = ({ children }) => {
  const [route, setRoute] = useState('/')

  return route === '/' ? (
    <Home />
  ) : page === '/one' ? (
    <PageOne />
  ) : (
    <NotFound />
  )
}

const Home = () => {
  return (
    <>
     <Header />
     My content
    </>
  )
}

const PageOne = () => {
  return (
    <>
     <Header />
     Page 1
    </>
  )
}

const NotFound = () => {
  return (
    <>
     <Header />
     404
    </>
  )
}
```

Note here how each of the "route" components render the `<Header />` inside of their render function. Lets say the state changes from `/` to `/one`. Under the hood, React is going to fully unmount the `<Home />` component, then mount the `<PageOne />` component. Because it is fully unmounting and remounting a different component, it means that the `<Header />` inside `<Home />` is also fully unmounted, then remounted with `<PageOne />`. This presents a problem if, say, `Header` had some local state inside of it. The state gets lost during the transition between the 2 components.

So how might you fix this? The answer is to pull the `Header` up to the `Routes` component instead so that the `<Header />` component doesn't unmount between route changes:

```js
const Routes = ({ children }) => {
  const [route, setRoute] = useState('/')

  return (
    <>
      <Header />
      {route === '/' ? (
        <Home />
      ) : page === '/one' ? (
        <PageOne />
      ) : (
        <NotFound />
      )
    }
  )
}

const Home = () => {
  return 'My content'
}

const PageOne = () => {
  return 'Page 1'
}

const NotFound = () => {
  return '404'
}
```

Note here how the `Header` is no longer being rendered by the individual routes, but rather the container of these routes.

This is essentially what this plugin is doing. Gatsby manages the top-level "route" components for us and gives us the ability to specify what goes on a page. Out-of-the-box rendering pages look a little bit like the first example: If we render a `Layout` on an individual page, it means that transitioning to a new route will unmount the `Layout` and remount on the new path. This is not ideal of you want to maintain state between route changes. This plugin essentially "fixes" this and allows you to specify the layout higher up in the component hierarchy so that route changes don't force an unmount/remount of the layout.

Does that help?

src/layouts/MainLayout.js
caylahamann: Sorry yes definitely should be more specific! What I meant was would it make more sense to add this as a global style? I guess my confusion runs from the fact that we have this `MainLayout` page that just adds the styles, which seem to be stitched into the theme. I guess my question is why don't we have a `MainLayout` component for the theme if it just queries margin and width that is available in the theme?

package.json
caylahamann: Yes that explains a lot! Thank you.

src/layouts/MainLayout.js
jerelmiller: Ah that is a great question! We've held off on moving layout specific stuff over to the theme because we wanted to be sure there was actually something to share between the 3 sites. For example, the developer site has a 2-column layout with a sidebar, where the open source site has a single-column layout with top-level navigation. There's not a lot to share there yet.

I'd rather have some duplication between the developer site and docs site for now to fully understand what makes sense to abstract and what not to. We may end up with a layout structure that doesn't fully match the developer site, in which case abstracting early might cause more headaches if we have to override a bunch of the work done in the theme. Duplication is far better than the wrong abstraction, and I want to wait to understand what the abstraction might look like before we try and pull it into the theme, especially since this implementation is not the final implementation and will likely change quite a bit between now and GA.

PR #113, opened 2020-10-05
Fix missing links on index pages

jerelmiller:
## Description

Fixes the issue with missing links on some generated index pages. Heavily refactors the `createIndexPages` script to use unist trees and vfiles to parse the file structure and create markdown via mdast.


src/content/docs/agents/index.mdx
```diff
@@ -1,18 +1,58 @@
 ---
 title: Agents
 template: basicDoc
+
```
zstix, line 4: Is this extra line intentional?

zstix (commented):
This looks _so much nicer_! I had a couple of small comments / questions, but nothing critical.

scripts/utils/migrate/create-index-pages.js
```diff
@@ -1,96 +1,128 @@
-const fs = require('fs').promises;
-const fsSync = require('fs');
+/* eslint-disable no-console */
```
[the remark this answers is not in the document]
jerelmiller: We don't! Good catch. Originally there was a `console.log` if the file errored while writing, but that got moved, so we don't need this anymore. I'll remove.

scripts/utils/migrate/create-index-pages.js
```diff
@@ -1,96 +1,128 @@
...
+
+const replace = (string) =>
+  REPLACEMENTS.reduce(
+    (str, [regex, replacement]) => str.replace(regex, replacement),
+    string
+  );
+
+const capitalize = (word) => word.charAt(0).toUpperCase() + word.slice(1);
+
+const titleize = (dirname) => {
```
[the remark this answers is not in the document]
jerelmiller: Good call. I'll update the name. I was struggling with what to call this.

src/content/docs/agents/index.mdx
jerelmiller: I wanted to dig into this a bit more. I don't know if this is from the `yaml.safeDump` or if its compiling that way from the MDAST. I'll take a look further and see if I can figure out how to remove this newline.

jerelmiller: Fixed in ae1eb4c

zstix (approved):

scripts/utils/migrate/create-index-pages.js
```diff
@@ -1,96 +1,127 @@
...
+  const tree = root([
+    frontmatter({ title: sentenceCase(dir.basename), template: 'basicDoc' }),
+  ]);
+
+  visit(
+    dir,
+    (node) => node !== dir,
+    (node) => {
+      if (shouldSkipDirectory(node)) {
+        return [visit.SKIP];
```
caylahamann, line 70: How does this exactly work? `visit.SKIP`

scripts/utils/migrate/create-index-pages.js
```diff
@@ -1,96 +1,127 @@
...
+  );
+
+  return toMDX(tree);
 };
 
-const createIndexPages = async () => {
-  await createIndexPage(BASE_DIR);
+const createIndexPages = (files) => {
+  const contentDirectory = fromList(
+    files.sort((a, b) => a.path.localeCompare(b.path)),
```
caylahamann, line 98: what is the `path.localeCompare` doing? Why is this being sorted?

scripts/utils/unist-fs-builder.js
```diff
@@ -0,0 +1,23 @@
...
+
+const directory = (dirpath, children = []) => {
+  return u(
+    'directory',
+    { basename: path.basename(dirpath), path: dirpath },
+    children
+  );
+};
+
+const file = (filepath, data) => {
```
caylahamann, line 12: What is the purpose of adding these as a syntax tree?

scripts/utils/mdast.js
```diff
@@ -0,0 +1,6 @@
+const yaml = require('js-yaml');
+const u = require('unist-builder');
+
+const frontmatter = (attributes) => u('yaml', yaml.safeDump(attributes).trim());
```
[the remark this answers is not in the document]
jerelmiller: Yes. Unfortunately the `mdast-builder` module that I pulled in doesn't have a node generator for yaml frontmatter, so this adds a little helper function to create a frontmatter node (which is identifies as type `yaml`). 

Here is the specification on the MDAST: https://github.com/syntax-tree/mdast#yaml

scripts/utils/migrate/create-index-pages.js
jerelmiller: This essentially tells the utility to stop traversing the subtree if this is returned. If I had a normal `return` here, all that does is avoid processing the `node` we are on, but the visitor will continue to traverse it's subtree.

In this case, I don't want to visit any of the nodes in the subtree when we are in the directory `attribute-dictionary`, so this tells it to skip traversing it. Does that make sense? 

Here is the documentation on this particular bit: https://github.com/syntax-tree/unist-util-visit-parents#returns

scripts/utils/migrate/create-index-pages.js
jerelmiller: Here is the documentation for [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare)

Essentially I wanted the files in alphabetical order so that I could render them as such on each `index.mdx` file.

scripts/utils/migrate/create-index-pages.js
caylahamann: Yes! Thank you. I didn't really understand what it was, but it appears that it's an attribute on the object that's a string (`'skip') where when you return it, it will know to skip that node. Is that correct?

scripts/utils/unist-fs-builder.js
jerelmiller: I was able to leverage the utilities in the [`unist`](https://github.com/syntax-tree/unist#utilities) suite of utilities to traverse the directory tree, rather than having to implement the traversal myself. 

It does bend the rules just a bit since these are technically utilities for a "syntax tree", but they worked pretty well for representing the file system as well. Figured I might as well leverage a bunch of tooling that already existed to avoid having to reimplement something very similar. Does that answer your question?

scripts/utils/mdast.js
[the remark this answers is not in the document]
caylahamann: okay awesome! thank you for clarifying that

caylahamann:
slack is currently melting down on me so I can't reply to your DM, but this is good to merge for me @jerelmiller

scripts/utils/migrate/create-index-pages.js
jerelmiller: Thats correct!

PR #117, opened 2020-10-06
Use react spring to animate collapsers

jerelmiller:
## Description

Uses react-spring to animate the collapser component.


roadlittledawn (approved):
Looks good. Thanks for fixing the top border as well!

PR #123, opened 2020-10-06
Migrate icons

### LINKED ISSUE #120: [Codemod] Migrate icons to the gatsby theme icon library
## Description

Write a codemod that converts the `<i className="fa" />` tags to our `<Icon />` component

jerelmiller:
Closes #120 

# Description

Migrates the font-awesome icons to the `<Icon />` components from the Gatsby theme with its feather icon equivalent.

* Adds a mapping from font-awesome icons to feather icons
* Adds a codemod to convert `<i />` tags to `<Icon />` tags
* Adds warnings for icons that are unable to be mapped
* Adds a warning for additional class names on the icon that we don't handle

This PR maps all but 3 icons: `plug`, `area-chart`, and `file-code-o`. We will need to figure out how we want to replace these.


jerelmiller:
Added a followup issue to mention the icons we currently do not have a mapping for: https://github.com/newrelic/docs-website/issues/124

codemods/icons.js
```diff
@@ -0,0 +1,86 @@
...
+  wrench: 'tool',
+};
+
+const icons = () => (tree, file) => {
+  visit(
+    tree,
+    (node) => isMdxElement('i', node) && hasClassName('fa', node),
+    (icon) => {
+      const className = findAttribute('className', icon);
+      const [, size] = className.match(ICON_SIZE) || [];
```
zstix, line 39: In the event that there's no match (`[]`) wouldn't destructuring the second item in the array return `undefined`? Perhaps something like this might be safer:

```jsx
const size = classname.match(ICON_SIZE)?.[1] || 0;
```

codemods/icons.js
```diff
@@ -0,0 +1,86 @@
...
+
+      icon.name = 'Icon';
+      icon.children = [];
+
+      addAttribute('name', featherIcon, icon);
+      removeAttribute('className', icon);
+      removeAttribute('aria-hidden', icon);
+
+      if (size) {
+        addAttribute('size', `${size}em`, icon);
```
zstix, line 80: Nitpick: would you be willing to use `rem`? I think that's what we're using on the developer site and the theme.

zstix (approved):
Looks good to me! A couple of _super_ minor things, but nothing critical!

codemods/icons.js
jerelmiller: Yes, `size`, would be `undefined` in this case. I am [handling that below](https://github.com/newrelic/docs-website/pull/123/files#diff-26d8a9a27464d33f0b45555b8697c54fR79-R81) by checking it to see if its a truthy value before adding it as an attribute:

```js
if (size) {
  // ...
}
```

jerelmiller: Also, unfortunately, because we are in node-land, we can't use the fancy optional chaining syntax 😞

codemods/icons.js
jerelmiller: The only reason I didn't use `rem` here was because the font-awesome class `fa-2x`, `fa-3x`, etc was a _relative_ sizing based on the surrounding font size. I used `em` to maintain compatibility, but I'd be happy to use `rem` if it makes more sense to make it relative to the root instead of the surrounding font size.

zstix (approved):

jerelmiller:
@zstix I'm going to go ahead and merge with the `em` change. We can revisit later if we decide `rem` is the way to go. Might be something for us to take a look at once we've had a chance to hammer on the design a bit more.

PR #125, opened 2020-10-06
Preserves buttons in turndown and converts them to Button components

### LINKED ISSUE #122: [Codemod] Migrate buttons to use our theme's buttons
## Description

Migrate buttons from the existing site to our Gatsby theme's buttons. 

## Acceptance criteria

* [x] Ensure buttons are kept around via turndown
* [x] Create a codemod to migrate buttons to use our `<Button />` component from the theme

caylahamann:
closes #122 

This PR preserves buttons in turndown and creates a codemod to turn them into button components. If the button has an href attribute, then it converts it into an internal link button.


codemods/buttons.js
```diff
@@ -0,0 +1,53 @@
...
+          (className) =>
+            className.startsWith('btn-') && VARIANTS.includes(className)
+        )
+        .replace('btn-', '');
+
+      if (href) {
+        a.attributes.push({
+          name: 'as',
+          value: 'a',
+          type: 'mdxBlockElement',
```
jerelmiller, line 40: This type should be an `mdxAttribute` instead of an `mdxBlockElement` since this is an attribute type instead of a JSX tag in the markdown

codemods/buttons.js
```diff
@@ -0,0 +1,53 @@
...
+        .find(
+          (className) =>
+            className.startsWith('btn-') && VARIANTS.includes(className)
+        )
+        .replace('btn-', '');
+
+      if (href) {
+        a.attributes.push({
+          name: 'as',
+          value: 'a',
```
jerelmiller, line 39: By setting this value to an `a`, it means that clicking on the button will do a whole page refresh. We want to try and use Gatsby's `Link` element where possible to provide the best experience our users.

Since that value is more than just a string, the `value` type becomes a bit more complex. After some digging, it looks like this type should handle a JSX _expression_ instead of a plain string:

```js
a.attributes.push({
  type: 'mdxAttribute',
  name: 'as',
  value: {
    type: 'mdxValueExpression',
    value: 'Link',
  }
})
```

This should transform the button into the following JSX:

```js
<Button as={Link} />
```

With this change, we should likely switch the `href` to a `to` prop to be compatible with the `Link` component.

src/components/MDXContainer.js
```diff
@@ -64,6 +69,7 @@ const components = {
   code: MDXCodeBlock,
   pre: (props) => props.children,
   wrapper: Wrapper,
+  Button,
```
jerelmiller, line 72: I _think_ we may also want to add a shortcode for the `Link` element as well so that it is available for the `as` prop in MDX

jerelmiller (changes requested):
I had a few suggestions for updating! Nice work taking this on!

codemods/buttons.js
caylahamann: this is the information I was looking for! Question: where were you able to learn about this? I've been pouring over the MDX docs and they're pretty sparse.

src/components/MDXContainer.js
caylahamann: I had that before, but then I could only get it rendering if I passed it in with an a tag. With the new information I'll make that change :)

codemods/buttons.js
```diff
@@ -0,0 +1,53 @@
...
+} = require('./utils/mdxast');
+
+const VARIANTS = ['btn-primary'];
+
+const buttons = () => (tree) => {
+  visit(
+    tree,
+    (node) => {
+      return (
+        isMdxBlockElement('a', node) &&
```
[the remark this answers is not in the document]
caylahamann: I had the same question... currently there are only a tags that are buttons, but I'm not sure if there's any other ones out there. @roadlittledawn maybe you have an idea?

codemods/buttons.js
jerelmiller: Lets chat in person!

codemods/buttons.js
```diff
@@ -0,0 +1,63 @@
...
+  hasClassName,
+  findAttribute,
+} = require('./utils/mdxast');
+
+const VARIANTS = ['btn-primary'];
+
+const buttons = () => (tree, file) => {
+  if (file.path === './src/content/docs/agents/go-agent/index.mdx') {
+    console.dir(tree, { depth: 4 });
+  }
```
jerelmiller, line 14:
```suggestion
```

Oops we forgot to remove this 😅

codemods/buttons.js
```diff
@@ -0,0 +1,63 @@
...
+      }
+
+      a.name = 'Button';
+      addAttribute('variant', variant, a);
+      removeAttribute('className', a);
+      removeAttribute('href', a);
+    }
+  );
+
+  // We are assuming all pages have frontmatter
```
jerelmiller, line 53:
```suggestion
  // We are assuming all pages have frontmatter and therefore not handling
  // pages where frontmatter is missing. If the need arises, we'll have to insert
  // the import statement as the first child instead of the second child.
```

Added a little more to this comment to explain what needs to happen in case we need to handle cases with pages without frontmatter.
Adde

jerelmiller (commented):
Had a couple tiny things, then I feel good about merging!

jerelmiller (approved):

PR #127, opened 2020-10-07
added robots plugin

jpvajda:
added robots plugin to prevent crawling of the site during preview mode.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -77,7 +77,7 @@ module.exports = {
       options: {
         host: 'https://www.docs-preview.newrelic.com',
         sitemap: 'https://www.docs-preview.newrelic.com/sitemap.xml',
-        policy: [{ userAgent: '*', Disallow: '/' }],
+        policy: [{ userAgent: '*', disallow: '/' }],
       },
     },
     'gatsby-remark-images',
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -70,6 +70,16 @@ module.exports = {
         path: `${__dirname}/src/content`,
       },
     },
+    // This plugin is set to disallow to prevent crawling of the site during preview
+    // mode
+    {
+      resolve: 'gatsby-plugin-robots-txt',
+      options: {
+        host: 'https://www.docs-preview.newrelic.com',
+        sitemap: 'https://www.docs-preview.newrelic.com/sitemap.xml',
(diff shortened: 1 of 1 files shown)


gatsby-config.js
```diff
@@ -70,6 +70,16 @@ module.exports = {
         path: `${__dirname}/src/content`,
       },
     },
+    // This plugin is set to disallow to prevent crawling of the site during preview
+    // mode
+    {
+      resolve: 'gatsby-plugin-robots-txt',
```
jerelmiller, line 76: I'm not sure how Gatsby handles this being in there "twice". We have this specified in the [gatsby theme](https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/packages/gatsby-theme-newrelic/gatsby-config.js#L17-L23) with a configurable value. Could you add these options as the `robots` key in the theme options? 

As you can see from the link, we have a default value set for the robots if you don't provide any overrides.

gatsby-config.js
```diff
@@ -70,6 +70,16 @@ module.exports = {
...
+      options: {
+        host: 'https://www.docs-preview.newrelic.com',
+        sitemap: 'https://www.docs-preview.newrelic.com/sitemap.xml',
```
jerelmiller, line 79:
```suggestion
        host: 'https://docs-preview.newrelic.com',
        sitemap: 'https://docs-preview.newrelic.com/sitemap.xml',
```

Ooooo good call on this since the default is to pull from `siteMetadata.siteUrl`. Only thing I'd suggest here is to remove the `www` since we don't serve the site there (we have a redirect in place for it though).

gatsby-config.js
```diff
@@ -70,6 +70,16 @@ module.exports = {
...
+        policy: [{ userAgent: '*', Disallow: '/' }],
```
jerelmiller, line 80:
```suggestion
        policy: [{ userAgent: '*', disallow: '/' }],
```

Looking at the [docs](https://github.com/itgalaxy/generate-robotstxt#usage), it looks like this key is lowercase.

jerelmiller (changes requested):
Had a couple suggestions for you and see if you can move the config to the theme `robots` config.

gatsby-config.js
jpvajda: will do.

jpvajda:
Will move this to the gatsby theme

PR #146, opened 2020-10-08
Add additional content + bug fix to make it build

jerelmiller:
## Description

Runs the migration to pull in more of the content, which represents ~10% of the total docs content.


scripts/utils/codemod/create-processor.js
```diff
@@ -9,7 +9,6 @@ const createProcessor = ({ codemods }) => {
   const processor = unified()
     .use(stringify, {
       bullet: '*',
-      commonmark: true,
```
caylahamann, line 12: sowe aren't using commonmark? what did removing this do out of curiosity?

jerelmiller: It was wreaking havoc on some of the output and causing the build to fail, mostly because it added a comment separator in some cases, which are not recognized by MDX. For now I think this is safe to turn off.

caylahamann: okay! thank you

caylahamann (approved):

PR #150, opened 2020-10-09
Example boxes + complex titles

### LINKED ISSUE #148: Add support for example boxes
## Description

We need to also convert `dl`s with `className="example-box"` over to the collapser component

jerelmiller:
Closes #148 

## Description

Adds a codemod to convert `<dl className="example-box">` components over to a `Collapser`. This PR also handles titles on Collapsers that have additional markdown content instead of just a plain string.

### DIFF
--- a/codemods/utils/mdxast-stringify.js
+++ b/codemods/utils/mdxast-stringify.js
@@ -1,0 +1,5 @@
+const createProcessor = require('../../scripts/utils/codemod/create-processor');
+
+const stringify = (tree) => createProcessor().stringify(tree);
+
+module.exports = stringify;
--- a/scripts/utils/codemod/create-processor.js
+++ b/scripts/utils/codemod/create-processor.js
@@ -5,7 +5,7 @@ const frontmatter = require('remark-frontmatter');
 const remarkMdx = require('remark-mdx');
 const remarkMdxjs = require('remark-mdxjs');
 
-const createProcessor = ({ codemods }) => {
+const createProcessor = ({ codemods = [] } = {}) => {
   const processor = unified()
     .use(stringify, {
       bullet: '*',
--- a/codemods/clamshells.js
+++ b/codemods/clamshells.js
@@ -3,6 +3,7 @@ const toString = require('mdast-util-to-string');
 const {
   removeAttribute,
   isMdxBlockElement,
+  isPlainText,
   hasClassName,
   removeChild,
   setAttribute,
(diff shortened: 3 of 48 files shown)


codemods/utils/mdxast-stringify.js
```diff
@@ -0,0 +1,5 @@
+const createProcessor = require('../../scripts/utils/codemod/create-processor');
+
+const stringify = (tree) => createProcessor().stringify(tree);
```
caylahamann, line 3: what did adding this change? I'm a little bit confused here

scripts/utils/codemod/create-processor.js
```diff
@@ -5,7 +5,7 @@ const frontmatter = require('remark-frontmatter');
 const remarkMdx = require('remark-mdx');
 const remarkMdxjs = require('remark-mdxjs');
 
-const createProcessor = ({ codemods }) => {
+const createProcessor = ({ codemods = [] } = {}) => {
```
caylahamann, line 8: this goes along with my other question!

codemods/utils/mdxast-stringify.js
jerelmiller: Some of the collapsers have a title with embedded markdown in it. Because its not plain text, I need to be able to represent that in the prop as a React component, not a markdown string since that prop doesn't interpret markdown.

Take for example the following:

```js
// Plain text
<dt>
  This is the title of the collapser
</dt>

// Embedded markdown
<dt>
  This has some `code` in it
</dt>
```

These `dt` tags get moved over to the `title` prop on the `Collapser` component:

```js
<Collapser title="This is the title of the collapser" />

<Collapser title={<>This has some <InlineCode>code</InlineCode> in it</>} />
```
Note how the 2nd example is a JSX _expression_, which means I can't just interpret markdown like this:

```js
// This is invalid syntax and would blow up on "undefined variable 'This'" or something similar
<Collapser title={This has some `code` in it} />
```

I therefore have to convert the embedded markdown back to JSX, like you see above.

I'm doing this in the `to-jsx-expression` module you see in this PR. That module generates an MDXAST. In order to go from an MDXAST -> Markdown, I have to "stringify" it, hence this module. I wanted to use the same plugins (`remark-mdx`, `remark-mdxjs`, etc.) to stringify the MDXAST generated from the `to-jsx-expression` module, but in this case I didn't want to include the codemods on something I'm generating an AST for myself. Does that make sense?

jerelmiller: To note once more, this `stringify` module is **only** used from the `to-jsx-expression` module to generate a markdown string from an AST

caylahamann: ahhhh okay that makes alot of sense, I was confused on what the purpose of the to-jsx-expression module was doing as well. That seems all good to me :)

caylahamann (approved):

scripts/utils/codemod/create-processor.js
jerelmiller: Since I want the stringify module to _not_ use codemods, it felt awkward to call it like: `createProcessor({ codemods: [] })`, so I updated this to make the argument a default value with no plugins. This allows me to call it like `createProcessor()`

PR #161, opened 2020-10-19
Update layout to add area for navigation

### LINKED ISSUE #130: [Nav UI] Update layout to include sidebar
# Description

In the docs site, we currently only have a single column view. We would want to add a sidebar to the `MainLayout` in order to have this render on each page. Ideally we would want to create a `Sidebar` component as well.

_NOTE: this is only for the markup and styles, not populating the navigation._

## Acceptance Criteria
* [x] Change `MainLayout` in docs site
* [x] Create Sidebar component
* [x] Sidebar collapses to a mobile menu on smaller displays

### Reference 
* [ Developer Site MainLayout](https://github.com/newrelic/developer-website/blob/3f615abbf4042ab085bc224912d0dff967a25dd0/src/layouts/MainLayout.js#L121)

jerelmiller:
Closes #130 

## Description

Update the layout to include a sidebar and mobile nav. This does not populate links in the sidebar, rather provides a placeholder for them.

## Screenshots

Screen Shot 2020-10-19 at 1 55 43 PM

Screen Shot 2020-10-19 at 2 23 40 PM

### DIFF
--- a/src/layouts/MainLayout.js
+++ b/src/layouts/MainLayout.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { GlobalHeader } from '@newrelic/gatsby-theme-newrelic';
 import { graphql, useStaticQuery } from 'gatsby';
 import { css } from '@emotion/core';
+import Sidebar from '../components/Sidebar';
 
 const MainLayout = ({ children }) => {
   const {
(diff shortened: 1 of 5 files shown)


src/layouts/MainLayout.js
```diff
@@ -18,19 +22,100 @@ const MainLayout = ({ children }) => {
...
+      css={css`
+        --global-header-height: 36px;
+        --sidebar-width: 300px;
+
+        min-height: 100vh;
+      `}
+    >
       <GlobalHeader />
-      <main
+      <div
```
zstix, line 43: Out of curiosity, why did you update `main` to be `div`? Wouldn't we want to use the more semantic option (even if it means adding an extra `display: block;`)?

zstix (approved):
Looks good to me. One super minor question

src/layouts/MainLayout.js
jerelmiller: Per [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main) (emphasis mine):

> The content of a <main> element should be unique to the document. Content that is repeated across a set of documents or document sections **such as sidebars**, navigation links, copyright information, site logos, and search forms **shouldn't be included** unless the search form is the main function of the page.

Since this `div` includes the sidebar, I moved the `main` tag to be the actual page content itself. You'll see it used on [line 95](https://github.com/newrelic/docs-website/pull/161/files#diff-a80edad99cd274f77da55f52c521689a32127055d855e00ee2f0f33542069bf2R95)

PR #273, opened 2020-11-17
Sync style guide content

### LINKED ISSUE #272: migrate internal style guide docs to new site
### Summary

We need to rewrite and update our internal style guide content for its new home on an open source docs site. To start working on that, I'll need that content migrated over.

Here is all of the content that needs to be brought in: https://docs.newrelic.com/docs/new-relic-only (minus the Docs for training env and drupal).

**Import**

- All docs under Basic style guide
- All docs under Tech writer style guide

**Don't Import**
- Drupal Configuration
- Training environment

**navigation**

- The new IA calls for this to exist in the navigation under a category called New Relic Only. We need to decide if this should be called something else as this content is no longer behind a login. @paperclypse 

Once imported, I'll need a place to put that content so that it doesn't get overwritten with each new migration.

jerelmiller:
Closes #272 

## Description

Migrates the style guide content to allow our docs team to start updating the content for the new site.


src/content/docs/new-relic-only/basic-style-guide/writing-guidelines/code-formatting-guidelines-var-mark.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+The `<var>` tag is used for user-input values in code snippets that customers would be using in procedural/functional ways. For example, you would use the `<var>` tag when explaining how to do some specific task.
+
+A `<var>` tag would not be used for:
+
+* Example code: If the code is meant to just be an example, to show the format/structure of the code, and there is not an actual procedure the customer is following, var tags are not needed. Examples:
+  * [Example configuration file](/docs/integrations/host-integrations/host-integrations-list/oracledb-monitoring-integration#example-config) (or this [Java agent config template](/docs/agents/java-agent/configuration/java-agent-config-file-template)): if you are showing an example config file that isn't part of a procedure, you shouldn't use the &lt;var> tag. Several reasons for this: 1) The important part of showing an example config file is to show the overall structure of the file, 2) Usually the number of config options present in a file will vary based on whatever the customer wishes to use, so using &lt;var> tags can actually be confusing as it implies that these values must be present.
```
roadlittledawn, line 21: Looks like our codemods may be doing something weird with `<var>` and `<mark>` tags.

roadlittledawn (commented):
i wonder if we might want to move the `new-relic-only` directory outside of `docs` into `src/content/new-relic-only`? i imagine it will eventually be renamed to something like `new-relic-style-guide` and we might want to keep it distinct from docs about new relic products / features / etc.

scripts/utils/migrate/instruction-set.js
```diff
@@ -367,4 +367,13 @@ module.exports = [
     type: INSTRUCTIONS.REMOVE,
     path: ['Plugins'],
   },
+
+  // ----- New Relic only ----- \\
+  {
+    type: INSTRUCTIONS.UPDATE,
+    path: ['New Relic only'],
+    node: {
+      root: false,
```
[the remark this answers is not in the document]
jerelmiller: THANK YOU FOR SPOTTING THIS!

I could not, for the life of me, figure out a good name for this. I was hoping someone would give me some better suggestions. I like `rootNav`. Its a bit more terse, but still clear on its intent. I will update this!

src/content/docs/new-relic-only/basic-style-guide/writing-guidelines/code-formatting-guidelines-var-mark.mdx
jerelmiller: I think this is intentional. Those tags are part of the text, so by using the html entity, it prevents it from rendering as a JSX component and instead outputs the literal text to the screen.

roadlittledawn (approved):
:shipit:

PR #274, opened 2020-11-18
Migrate data dictionary

### LINKED ISSUE #250: Migrate event/attribute definitions
## Description

We want to migrate all event/attribute definitions to the docs site. These will be stored in markdown files (**NOT** MDX). We will be combining short and long descriptions into a single field, as well as keeping the related info.

This consists of the following work:

* Update migration script to migrate all event/attribute definitions to `src/data-dictionary/events` folder
  * Turn long description HTML into markdown (use `turndown`)
* Update Gatsby to source these event/attribute definitions into GraphQL

## Helpful links

* [Prototype of event/attribute definitions](https://github.com/newrelic/docs-website/tree/jerel/data-dictionary-prototype/src/data-dictionary/events)

jerelmiller:
Closes #250 

## Description

Pulls in all event/attribute definitions and saves them in the `src/data-dictionary` folder.


jerelmiller:
I wonder if it makes sense to pull in all attribute defs in a separate PR. They kind of blow out what you're looking at

scripts/utils/migrate/fetch-attribute-definitions.js
```diff
@@ -0,0 +1,14 @@
+const docsApi = require('./docs-api');
+
+const prop = (name) => (obj) => obj[name];
```
zstix, line 3: 1. Love this
2. Looks like we define this function here and in `fetch-doc-count.js`. Maybe we should make a util file for functional helpers?

jerelmiller: Thats probably a good idea!

zstix (approved):
Looks good! Only 1 super minor comment.

zstix (approved):

scripts/utils/migrate/fetch-attribute-definitions.js
jerelmiller: Addressed with d4f5187

PR #322, opened 2020-11-24
Add `content` and `site` labels to each to make UI filtering easier

austin-schaefer:


austin-schaefer:
Updated the issue templates to add a second `site` or `content` label to all issues. This is a workaround since Github Issues doesn't support OR filtering, only AND, which makes it impossible to see all the `content` or `site` issues in one place w/o some kind of shared label.

jpvajda:
I'd recommend using labels: bug, content, site, enhancement if we are taking this approach to make it more simplified, so you'd just combine the 2 labels bug + content to = a content bug, etc. If you want to make that change I'll adjust the labels @austin-schaefer

jpvajda (approved):
LGMT

PR #323, opened 2020-11-24
Mobile UI for attribute dictionary

caylahamann:
There is some weird behavior with the tables to be fixed, but I thought that could be addressed in a follow up PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,8 +39,7 @@
     "react-spring": "npm:@react-spring/web@9.0.0-rc.3",
     "react-use": "^15.3.4",
     "react-use-measure": "^2.0.2",
-    "turndown": "^6.0.0",
-    "use-media": "^1.4.0"
+    "turndown": "^6.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -39,7 +39,8 @@
     "react-spring": "npm:@react-spring/web@9.0.0-rc.3",
     "react-use": "^15.3.4",
     "react-use-measure": "^2.0.2",
-    "turndown": "^6.0.0"
+    "turndown": "^6.0.0",
+    "use-media": "^1.4.0"
```
jerelmiller, line 43: [nit] It looks like we have `react-use` installed, which also comes with its own [`useMedia`](https://github.com/streamich/react-use/blob/master/docs/useMedia.md) hook. If you want to save on an extra dependency, you could use the `react-use` version.

jerelmiller (approved):
This looks great! Thanks for doing this :)

package.json
caylahamann: That sounds good to me! I'll make that change

jerelmiller (approved):

Issue #248, opened 2020-11-09
[DDS] Sync attribute definitions from the docs site to the service


jerelmiller:
## Description

Once we have a service up and running, connected to the database, we need a way to sync the definitions from the docs site to the data dictionary service. The service will read in the event/attribute definitions and add/update the database records with the data.

zstix:
## Remaining Work
* [ ] Save attributes to the database
* [ ] Save units to the database
* [ ] Expose an endpoint that we can ping
* [ ] Set up the endpoint with a shared secret (vault via grandcentral)
* [ ] Amplify webhook (when a deploy has finished) _or_ GH Action in docs repo (on a regular interval)

PR #338, opened 2020-12-02
Handle events without event definitions

jerelmiller:
Related to #248

## Description

When creating data dictionary event nodes, some data was skipped if there was no
event definition for an event. This PR ensures that an event definition is
created for events listed from attributes that have no associated definition.
This PR also consolidates the JSON endpoint that lists data dictionary
definitions.

### DIFF
--- a/amplify.yml
+++ b/amplify.yml
@@ -3,19 +3,20 @@ frontend:
   phases:
     preBuild:
       commands:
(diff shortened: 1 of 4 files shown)


roadlittledawn (approved):
:shipit:

roadlittledawn (approved):
:shipit: again

PR #348, opened 2020-12-04
Update change-your-apdex-settings.mdx

alanwest:


jerelmiller:
Hey @alanwest, by chance is this a test PR or are you looking to get this change applied?

Unfortunately our migration process as it sits wipes out all changes between runs, which means any human edits applied to any of the content in `src/content` will get lost between runs of the migration script. We do not yet have a solution for allowing human edits for content.

If this is a change that you would like to see permanently applied to the new docs site, but not the current one, let us know and we can consider how to do this. Another option would be to apply this change to the current docs site. This would ensure that the next time we run the migration script, this change would get pulled in.

If this is a test PR to try out the new process, then disregard everything I've just said! Feel free to close this PR when you are finished with testing on this branch :)

mmfred:
This is a test PR.

PR #350, opened 2020-12-07
Add redirects for each attribute definition

### LINKED ISSUE #251: Create redirects for old attribute definition pages to new UI
## Description

As part of this migration, we are consolidating event/attribute definitions into a single page. This means that we need to route the old attribute definition URLs to their new, filtered location.

We should be able to source the data created from #250 to use as the base for these definitions.

## Helpful links

* [Sample legacy attribute definition page](https://docs.newrelic.com/attribute-dictionary/transaction/appname)
* #250  
* #243

jerelmiller:
Closes #251 

## Description

Add redirects for each attribute definition to the new filtered view.


roadlittledawn (approved):
per campfire looks good w/ follow on work to fix filtering issue on `/attribute-dictionary` page

caylahamann:
YAY!

PR #352, opened 2020-12-08
GitHub workflow to trigger sync

jerelmiller:
## Description

Add a GitHub action that will trigger the syncing process to the data dictionary service.

### DIFF
--- a/scripts/sync-to-service.js
+++ b/scripts/sync-to-service.js
@@ -6,7 +6,7 @@ const env = process.env.ENV;
 const HOSTS = {
   staging: 'https://data-dictionary.staging-service.newrelic.com',
   production: 'https://data-dictionary.service.newrelic.com',
-  'eu-production': 'https://data-dictionary.service.newrelic.com',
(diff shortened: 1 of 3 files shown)


scripts/sync-to-service.js
```diff
@@ -0,0 +1,30 @@
...
+
+const env = process.env.ENV;
+
+const HOSTS = {
+  staging: 'https://data-dictionary.staging-service.newrelic.com',
+  production: 'https://data-dictionary.service.newrelic.com',
+  'eu-production': 'https://data-dictionary.service.eu.newrelic.com',
+};
+
+const sync = async () => {
```
zstix, line 12: I'm assuming that any uncaught errors in this script will be reported as a failed action in GitHub?

jerelmiller: They should yes since it would exist with a non-zero exit code. I've run the script myself locally to test, just not through the GH action.

zstix (approved):

Issue #301, opened 2020-11-21
[Docs Site UX] Migrate What’s New posts


zstix:
## Description
Currently, we are migrating What's New content into the `src/content/whats-new` directory and all content is sitting in that directory. We would like to migrate these into a nested directory structure based on the announcement date (year > month):

```
src/
  content/
    whats-new/
      2019
        11/
        12/
      2020/
        01/
```

Each post should have the post should have all meta fields (including post date) stored in the frontmatter.

## Acceptance Criteria
* [x] What's New content is migrated into directories based on post year and month
* [x] All What's New `.md` files contain the post date in the frontmatter

zstix:
As part of this ticket, we should pull in _all_ What's New posts. As part of this, we will update the dev database.

Issue #256, opened 2020-11-10
[Misc] Migrate 63.7% of the docs site


zstix:
## Description
At the time of creating this ticket, we are migrating `10%` of the content. We would like to bring in more content and verify that we've caught any edge cases.

## Acceptance Criteria
* [ ] Migrate 25% of the content
* [ ] Verify that the site builds correctly
* [ ] Observe build-time impact of pulling in more content

jerelmiller:
@fryckbos has requested that we sync all of the integrations pages as a part of this scope so that his team can work on parsing out the current descriptions from the MDX files.

PR #358, opened 2020-12-08
Add ability to pull in paginated data

jerelmiller:
Partially addresses #301 and #256 

## Description

Paginates the docs results from the API so that we can pull in more than 1 page of data. Also specifies to pull in all whats new content. Note that this PR does not pull in additional content yet as that will be addressed in a separate PR.

### DIFF
--- a/scripts/utils/migrate/fetch-docs.js
+++ b/scripts/utils/migrate/fetch-docs.js
@@ -1,7 +1,12 @@
 const path = require('path');
 const docsApi = require('./docs-api');
 const { prop } = require('../functional');
-const { TYPES, ITEMS_PER_TYPE, DIRECT_IDS } = require('../constants');
+const {
+  TYPES,
+  ITEMS_PER_TYPE,
+  DIRECT_IDS,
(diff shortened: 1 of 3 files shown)


scripts/utils/migrate/fetch-docs.js
```diff
@@ -1,30 +1,37 @@
...
     docsApi.get(`/api/migration/content/page/${id}`)
   );
 
-  const nrOnly = await docsApi.paginate(
-    '/api/migration/content/page/nr-only/list'
-  );
-
-  const docs = await Promise.all([...requests, ...hardCodedRequests]);
+  const docs = await Promise.all(requests);
+  const hardCodedDocs = await Promise.all(hardCodedRequests);
```
caylahamann, line 29: I think I missed this previously, what are these NR hardcoded docs?

jerelmiller: We were asked to migrate a subset of integration docs so that Fred's team could take a look at them. This ensures that we get back those integration pages.

PR #365, opened 2020-12-09
Add endpoint for Whats New IDs

roadlittledawn:
Closes #365 

Adds query to config to output whats new ids in json

Fixes query for whats new json to use allMarkdownRemark instead of allMdx﻿


jerelmiller (approved):
🌮

PR #410, opened 2020-12-14
Fix edit link on What's New pages

### LINKED ISSUE #401: what's new post edit this page button links the what's new.js template not the actual markdown page
The edit this page button on the what's new post pages should link the actual MD page for editing, not the template page. 

go to a what's new post, click on edit this page, you will be taken here: 

https://github.com/newrelic/docs-website/blob/develop/src/templates/whatsNew.js

Where a user should be routed here: 

https://github.com/newrelic/docs-website/blob/develop/src/content/whats-new/2020/12/alerts-applied-intelligence-new-landing-page.md

roadlittledawn:
Closes #401 

Generates `fileRelativePath` from `fileAbsolutePath` in `markdownRemark` nodes and adds it to the context which is then used in the footer to populate the Edit this page link.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -142,14 +142,13 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
       frontmatter: { template },
       fields: { slug },
     } = node;
-    const fileRelativePath = fileAbsolutePath.match(/src\/.*/g)[0];
 
     createPage({
       path: slug,
       component: path.resolve(`${TEMPLATE_DIR}${template}.js`),
       context: {
         slug,
-        fileRelativePath,
+        fileRelativePath: getFileRelativePath(fileAbsolutePath),
       },
     });
   });
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -74,6 +74,7 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
       ) {
         edges {
           node {
+            fileAbsolutePath
             frontmatter {
               template
             }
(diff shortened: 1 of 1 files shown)


gatsby-node.js
```diff
@@ -137,15 +138,18 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
 
   allMarkdownRemark.edges.forEach(({ node }) => {
     const {
+      fileAbsolutePath,
       frontmatter: { template },
       fields: { slug },
     } = node;
+    const fileRelativePath = fileAbsolutePath.match(/src\/.*/g)[0];
```
jerelmiller, line 145: We've been using [this function](https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/packages/gatsby-theme-newrelic/gatsby-node.js#L131-L132) to get the relative path from the absolute path. Should we be consistent?

jerelmiller (commented):
Just a single question, otherwise this looks great.

gatsby-node.js
roadlittledawn: oh right. i saw that earlier, but spaced it when i got it working. but yeah, makes sense to me to use and be consistent

gatsby-node.js
```diff
@@ -146,6 +148,7 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
       component: path.resolve(`${TEMPLATE_DIR}${template}.js`),
       context: {
         slug,
+        fileRelativePath: getFileRelativePath(fileAbsolutePath),
```
zstix, line 151: I think we need to define or import `getFileRelativePath`

roadlittledawn: we have a similar function in that file as well as the theme ¯\_(ツ)_/¯

jerelmiller (approved):

gatsby-node.js
jerelmiller: @zstix looks like it was defined at the bottom of this file already.

PR #412, opened 2020-12-14
Simple Feedback in Right Rail

### LINKED ISSUE #300: Setup simple user feedback loop from Docs Site
To capture feedback from users in early stages of the release of the Docs site we'll setup a simple feedback loop using github issues and later on look to align with the global New relic approach to capture user feedback. 

### Approach 

1. add a Feedback Button component to the each page on the docs site and display in  right rail (need to design input on placement)
2. setup a new issue type in GH called: ` positive feedback` that applies a label `+1 feedback`
3. setup a new issue type in GH called: ` negative feedback` that applies a label `- 1 feedback`
4. When a user interacts feedback button push the user through the GH issue flow to allow them to quickly file a feedback issue based on the sentiment they wish to report.
5. the thumb up would link the ` positive feedback` issue type and create the issue with the correct label.
6. the thumb up would link the ` negative feedback` issue type and create the issue with the correct label.

 7.  both the OSS Site and Dev Site should have this functionality as well.

zstix:
## Description
Adds a new `<SimpleFeedback />` component to `basicDoc` pages with two buttons: one for positive feedback and one for negative feedback. When a user clicks on one of these buttons, they will be taken to the new issue screen on GitHub (in a new tab).

The title of the issue will be pre-filled explaining which page they were on: `Feedback for: {PAGE_TITLE}`. Both buttons will add the `eng` and `feedback` labels and either the `feedback-positive` or `feedback-negative` label.

Here are the links for the "Find help and use the Support portal" page:
* https://github.com/newrelic/docs-website/issues/new?labels=eng,feedback,feedback-positive&title=Feedback+for:+Find+help%20and%20use%20the%20Support%20portal
* https://github.com/newrelic/docs-website/issues/new?labels=eng,feedback,feedback-negative&title=Feedback+for:+Find+help%20and%20use%20the%20Support%20portal

## Reviewer Notes
I made some decisions around style and labels to get the conversation rolling. If you think something would work better, let me know!

* Closes #300

## Screenshot(s)
Screen Shot 2020-12-14 at 3 33 09 PM


src/@newrelic/gatsby-theme-newrelic/icons/feather.js
```diff
@@ -253,6 +253,16 @@ export default {
...
+  'thumbs-down': (
+    <>
+      <path d="M10 15v4a3 3 0 0 0 3 3l4-9V2H5.72a2 2 0 0 0-2 1.7l-1.38 9a2 2 0 0 0 2 2.3zm7-13h2.67A2.31 2.31 0 0 1 22 4v7a2.31 2.31 0 0 1-2.33 2H17" />
+    </>
+  ),
+  'thumbs-up': (
+    <>
+      <path d="M14 9V5a3 3 0 0 0-3-3l-4 9v11h11.28a2 2 0 0 0 2-1.7l1.38-9a2 2 0 0 0-2-2.3zM7 22H4a2 2 0 0 1-2-2v-7a2 2 0 0 1 2-2h3" />
+    </>
+  ),
```
jerelmiller, line 265:
```suggestion
  'thumbs-down': (
    <path d="M10 15v4a3 3 0 0 0 3 3l4-9V2H5.72a2 2 0 0 0-2 1.7l-1.38 9a2 2 0 0 0 2 2.3zm7-13h2.67A2.31 2.31 0 0 1 22 4v7a2.31 2.31 0 0 1-2.33 2H17" />
  ),
  'thumbs-up': (
    <path d="M14 9V5a3 3 0 0 0-3-3l-4 9v11h11.28a2 2 0 0 0 2-1.7l1.38-9a2 2 0 0 0-2-2.3zM7 22H4a2 2 0 0 1-2-2v-7a2 2 0 0 1 2-2h3" />
  ),
```

Since these are just single nodes, no need for the fragment syntax.

src/templates/basicDoc.js
```diff
@@ -14,14 +15,20 @@ import toString from 'mdast-util-to-string';
...
     frontmatter,
     body,
     fields: { fileRelativePath },
   } = mdx;
 
+  const issueUrl = `${site.siteMetadata.repository}/issues/new`;
+  const title = `Feedback+for:+${frontmatter.title.replace(' ', '+')}`;
+  const labels = `eng,feedback`;
+  const positiveFeedback = `${issueUrl}?labels=${labels},feedback-positive&title=${title}`;
+  const negativeFeedback = `${issueUrl}?labels=${labels},feedback-negative&title=${title}`;
```
jerelmiller, line 30: Will we ever have a case where the feedback might point to issues other than these? I wonder if maybe we should move this inside the `SimpleFeedback` component. That way if we need to include the feedback widget on other pages with right rails, we don't need to copy/paste this over. Thoughts?

src/components/SimpleFeedback.js
```diff
@@ -0,0 +1,63 @@
...
+      `}
+    >
+      Was this page helpful?
+    </div>
+    <div
+      css={css`
+        display: flex;
+        margin-top: 0.5rem;
+        justify-content: center;
+        gap: 0.5rem;
```
jerelmiller, line 21: It looks like maybe Safari [doesn't have support](https://caniuse.com/?search=gap) for `gap`. Do we care about that? Should we have a fallback for that? Thoughts?

In case we need it, we can use `@supports`:

```sass
@supports not (gap: 0.5rem) {
  > *:first-child {
    margin-right: 0.5rem;
  }
}
```

jerelmiller (commented):
Couple of questions, otherwise looks great!

src/templates/basicDoc.js
zstix: I'm down with that. We would then only need to pass in the page title (which could be optional).

src/components/SimpleFeedback.js
zstix: Ugh, this is what happens when I switch to a chromium-based browser. Thanks for catching that...I'll use this suggestion.

src/components/SimpleFeedback.js
```diff
@@ -0,0 +1,87 @@
...
+          font-size: 0.875rem;
+        `}
+      >
+        Was this page helpful?
+      </div>
+      <div
+        css={css`
+          display: flex;
+          margin-top: 0.5rem;
+          justify-content: center;
```
zstix, line 41: We should left align and stretch these buttons out (similar to how we do with the other GitHub buttons above this).

zstix:
I've updated the buttons based on some design feedback:
Screen Shot 2020-12-15 at 10 00 00 AM

src/templates/basicDoc.js
```diff
@@ -19,8 +20,9 @@ const BasicDoc = ({ data }) => {
     mdxAST,
     frontmatter,
     body,
-    fields: { fileRelativePath },
+    fields: { fileRelativePath, slug },
   } = mdx;
+  console.log(fileRelativePath);
```
jerelmiller, line 25:
```suggestion
```

jerelmiller (commented):
Noticed a rogue `console.log` otherwise looks great!

jerelmiller (approved):

PR #421, opened 2020-12-16
Auto index pages

### LINKED ISSUE #217: Automate the updates of doc site index pages
~After speaking with @djsauble, we've determined that keeping around the auto-generated index pages for each path segment in the URL doesn't make a ton of sense. We would like to keep around the "table of contents" pages for each landing page section, but only for those pages.~

~There are a few reasons we think this is a good idea:~

* ~These are auto-generated, which means we don't actually link to them anywhere~
* ~Maintaining these will be a nightmare. Any addition/removal of a page in one of the subpaths would need a change reflected in these pages. Unless we have some tooling to auto-generate this content, its going to be a nightmare to keep these up-to-date.~
* ~These pages are only there so that if someone where to start hacking with the URL and removing path segments, they'd see a page with some content instead of a 404 page. We think that if a user starts messing with the URL manually, it's expected that you'd see a 404 page for pages that don't exist.~

# MMF 9 Scope
We want the site to automatically build these pages so content contributors do not have to manually update them. 

Directory landing pages would already exist, so would not need to be accounted for here. Depending on how they're implemented, [Agent API index pages](https://github.com/newrelic/docs-website/issues/390) and [agent specific release notes](https://github.com/newrelic/docs-website/issues/386) may need to be accounted for in this work.

## Acceptance criteria
- [ ] As a content contributor, I do not have to update index files manually
- [ ] As a content contributor, when I run the site locally I can generate those pages so I can see them on front-end with minimal effort

jerelmiller:
Closes #217 

## Description

Auto generates index pages for every URL that does not currently have a page. This ensures our content contributors do not have to take on the burden of maintaining these files whenever pages are added, modified, or deleted.


plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
```diff
@@ -0,0 +1,126 @@
...
+  const { skippedDirectories } = pluginOptions;
+  const { createPage } = actions;
+
+  const { data, errors } = await graphql(`
+    query MyQuery {
+      tableOfContents: allFile(
+        filter: {
+          sourceInstanceName: { eq: "markdown-pages" }
+          base: { in: ["index.mdx", "index.md"] }
+          relativePath: { regex: "/^(?!whats-new).*$/" }
```
zstix, line 17: Should we be adding in all of the `skippedDirectories` to this regex statement?

zstix: Ah, I see...you're skipping them in code down below.

src/components/IndexContents.js
```diff
@@ -0,0 +1,54 @@
...
+    <TableOfContentsContainer>
+      {pages.map((page) => (
+        <TableOfContents key={page.title} root={page} />
+      ))}
+    </TableOfContentsContainer>
+  );
+};
+
+IndexContents.propTypes = {
+  nav: PropTypes.object,
```
zstix, line 19: [nit] we might want to add a `shape` and `isRequired` to this.

zstix (commented):
One minor comment. I _thiiink_ I reviewed all the non-content files.

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
jerelmiller: Oops, I meant to remove this regex and instead use the `skippedDirectories` only. Good catch! I will update this.

src/components/IndexContents.js
jerelmiller: You're probably right 😄 . I was being lazy 😅

jerelmiller: Addressed in 106aa6d

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
jerelmiller: Addressed in 00b6fc5

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
```diff
@@ -0,0 +1,125 @@
...
+    if (existingPaths.includes(slug)) {
+      return;
+    }
+
+    createPage({
+      path: slug,
+      component: path.resolve('src/templates/indexPage.js'),
+      context: {
+        slug,
+        html: generateHTML(dir),
```
caylahamann, line 103: why are we generating html for the table of contents? why isn't it in mdx?

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
```diff
@@ -0,0 +1,125 @@
...
+
+    if (skippedDirectories.includes(dir.path)) {
+      return [visit.SKIP];
+    }
+
+    if (existingPaths.includes(slug)) {
+      return;
+    }
+
+    createPage({
```
caylahamann, line 98: so this is creating an index page, and the path below is creating a table of contents page, what is the difference between these two? (sadly wish i could build to visualize 😢 )

caylahamann (commented):
Left a few questions that might help me better understand the PR! thank you

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
jerelmiller: In this case, we aren't actually generating a file, but rather the contents of the index page in memory (errr, rather `pageContext`). MDX is really just a text format that will compile to JSX code, which is then evaluated at runtime by the `MDXRenderer`. Since its just a list of headers and links, I don't really need to compile the body to JSX and instead can go right to HTML.

This plugin is a replacement for the index pages, like this one: https://docs-preview.newrelic.com/docs/integrations

Notice how its just a list of headers and links. By compiling to HTML, I don't need to do any fancy logic in the template since the HTML is already built for me.

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
jerelmiller: They are similar, but the table of contents _pages_ are the pages that already have a landing page associated with them. For example, here is a landing page: https://docs-preview.newrelic.com/docs/agents/java-agent. And here is its associated table of contents page (which you can navigate to by clicking the button underneath the tiles): https://docs-preview.newrelic.com/docs/agents/java-agent/table-of-contents

The table of contents page is structured based on the nav, whereas the index pages are structured based on the child URLs in any given route. They look visually similar though.

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
caylahamann: thank you!! this is helpful to know :)

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
caylahamann: gotcha! thanks for the context

zstix (approved):

PR #428, opened 2020-12-17
RSS Feed for Release Notes

### LINKED ISSUE #376: Add RSS feed for each agent's release notes
## Summary
Similar to current site, each agent's (including mobile) release notes includes an RSS feed listing the 15 most recent release notes. Essentially any folder in `docs/release-notes` that contains release note files should have an RSS feed.

The feed includes the following for each item:
* title
* link
* description
* pubDate
* content:encoded
* guid

RSS icons are displayed:
* on the agent's release note index page
* on each agent's single release note page

## Acceptance criteria
- [ ] Each folder that contains release note files inside of `docs/release-notes` has an RSS feed at a dedicated URL
- [ ] Each feed is validated via [W3 feed validator](https://validator.w3.org/feed/)

## Helpful links
[RSS feed for PHP agent release notes](https://docs.newrelic.com/docs/release-notes/agent-release-notes/php-release-notes/feed)

zstix:
## Description
This PR sets up an XML file for consumption by RSS readers. We're using [gatsby-plugin-feed](https://www.gatsbyjs.com/plugins/gatsby-plugin-feed/) to handle the generation on `yarn build`.

## Reviewer Notes
This is set up to closely match the [existing feed](https://docs.newrelic.com/docs/release-notes/agent-release-notes/php-release-notes/feed). That said, some fields are _slightly_ different. We will want to make sure this doesn't break anyone's feeds before we launch.

[https://pr-428.d1vx4w5vgyla4s.amplifyapp.com/rss.xml](https://pr-428.d1vx4w5vgyla4s.amplifyapp.com/rss.xml)

* Closes #376 

## Screenshot(s)
Screen Shot 2020-12-17 at 2 37 30 PM


zstix:
Unfortunately, the RSS is not valid because we're not using a Date**Time**: https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fpr-428.d1vx4w5vgyla4s.amplifyapp.com%2Frss.xml

zstix:
We're now valid...this is good to be reviewed!

gatsby-config.js
```diff
@@ -279,6 +281,61 @@ module.exports = {
...
+              }
+            `,
+            serialize: ({ query: { site, allMdx } }) =>
+              allMdx.nodes.map(({ frontmatter, slug, id, html }) => {
+                const date = parseISO(frontmatter.releaseDate);
+                const formattedDate = format(date, 'EE, dd LLL yyyy');
+                const time = '00:00:00 +0000'; // time necessary for RSS validity
+
+                return {
+                  guid: id,
```
jerelmiller, line 321: Does this `guid` need to be consistent across builds? I'm not sure if that `id` is stable or not. We might want to double check that just in case.

gatsby-config.js
```diff
@@ -279,6 +281,61 @@ module.exports = {
         nodesPerFeedFile: Infinity,
       },
     },
+    {
+      resolve: 'gatsby-plugin-feed',
```
jerelmiller, line 285: If I'm understanding this correctly, we only output a single RSS feed endpoint for this correct?

I see the requirements listed in the issue as this (emphasis mine):

> Similar to current site, each agent's (including mobile) release notes includes an RSS feed **listing the 15 most recent release notes**. Essentially **any folder** in docs/release-notes that contains release note files should have an RSS feed.

From my understanding, this means we should have one feed per release note type, and pull the 15 most recent releases from each. Am I understanding that correct? Or maybe the plan is to just consolidate to a single RSS feed?

jerelmiller (commented):
The changes looks great! I had 2 minor questions for you before I approve.

gatsby-config.js
zstix: I'm not actually 100% on what this guid is used for, but I noticed it in the feed on the old site.

If I recall correctly, I think that the Gatsby-generated-ID is based off the filepath. If so, this might be "stable" enough to use. It's also totally possible that nothing uses this ID, but I wanted to keep the feed format as similar to the existing setup as possible.

gatsby-config.js
zstix: Oh good call, I'll update this to generate a few different feeds.

This plugin makes that fairly straightforward (via the `feeds` array). That said, I might extract this out to a plugin since it's already adding a lot of clutter to our config file.

gatsby-config.js
zstix: @austin-schaefer (or @roadlittledawn) Do you happen to know what the `guid` is used for in the RSS feed?

jerelmiller: Ya for sure! If this value was important, I just wanted to make sure that we aren't going to screw something up if this value isn't stable.

gatsby-config.js
```diff
@@ -279,6 +281,23 @@ module.exports = {
...
+      resolve: 'gatsby-plugin-generate-rss',
+      options: {
+        feeds: [
+          {
+            title: 'Agent Release Notes',
+            path: 'agent-release-notes.xml',
+            regex: '/docs/release-notes/agent-release-notes/',
+          },
+          {
+            title: 'Mobile Release Notes',
```
jerelmiller, line 294: Just to clarify one more time on this requirement:

>  each agent's (including mobile) release notes includes an RSS feed

I'm understanding that to be an RSS feed _per agent_. While the agents are categorized under "Agent release notes", and "Mobile release notes", I'd expect to see one RSS feed per agent. For example "Ruby agent", "PHP agent", "Java agent", etc.

Am I misunderstanding that requirement? Are we just bucketing the agents together into a single RSS feed instead?

zstix: That's accurate...I think we would like a feed for each agent.

I wanted to push up my work in the current state so we could discuss whether or not we liked the plugin configuration as it is right now. If so, we will want to add a `feed` for _each_ agent in `gatsby-config.js`. (option 🅰️)

An alternative approach would be to handle all of that by the plugin (finding all the agents and making the individual feeds dynamically). This take out all the configuration requirements, but it would be a little less flexible. (option 🅱️)

I'm inclined to keep _some_ of the config (option 🅰️). Thoughts?

zstix: Just pushed up the configs for all the agents, assuming we're alright with option 🅰️

jerelmiller: Gotcha! I'm a little slow today so pardon me for missing the intent 😄 

I'd actually be curious to understand the differences between the agents. We can get a distinct set of release notes by type (`subject`) and get their associated versions. If they all conform to the same "interface", having the plugin auto detect new subjects would be pretty nice (🅱️ ). 

That being said, maybe we need the flexibility per agent. Could you walk me through what we might need that could be unique to 1 agent vs another?

gatsby-config.js
```diff
@@ -1,6 +1,8 @@
 const fs = require('fs');
 const path = require('path');
 const indentedCodeBlock = require('./codemods/indentedCodeBlock');
+const format = require('date-fns/format');
+const parseISO = require('date-fns/parseISO');
```
zstix, line 5: We no longer need these libraries in this file.

zstix:
```suggestion
```

zstix:
Well it builds...but I have no idea where the feeds are 🤔

zstix:
The build for this is still failing (locally and on Amplify) due to Node running out of memory. Not sure how to resolve this, or what the issue is.

zstix:
Good news, it looks like it's working: https://pr-428.d1vx4w5vgyla4s.amplifyapp.com/docs/release-notes/agent-release-notes/nodejs-release-notes/feed.xml

Bad news, it doesn't look like it's returning all the information. Perhaps Gatsby is limiting the results since the payload is pretty big with the inclusion of `mdAST`.

zstix:
Looks like it's also adding some extra frontmatter variables to the HTML:
```
export const _frontmatter = {"subject":"Node.js Agent","releaseDate":"2015-04-16","version":"1.18.3","downloadLink":"https://www.npmjs.com/package/newrelic"}
```

zstix:
This is now working as intended and ready to be reviewed: https://pr-428.d1vx4w5vgyla4s.amplifyapp.com/docs/release-notes/agent-release-notes/nodejs-release-notes/feed.xml

jerelmiller (approved):

PR #436, opened 2020-12-18
Update all release notes with updated frontmatter

jerelmiller:
## Description

Update the rest of the release notes with the updated frontmatter.


scripts/utils/migrate/get-frontmatter.js
```diff
@@ -77,14 +77,25 @@ const addCustomFrontmatter = {
...
+      file.message(
+        `Unable to extract subject: ${doc.title}`,
+        null,
+        'get-frontmatter'
+      );
+    }
 
     return stripNulls({
-      subject,
+      subject: match ? match[1].trim() : doc.title.trim(),
```
roadlittledawn, line 92: i like that you added the fallback to the title!

in another branch Zack and I worked on yesterday, we check for each versioning / title format with separate regexes. it seemed browser was the only one that deviated (e.g., `Browser agent v1234`). 

do you think it's worth checking for those variants or just drawing a soft line and expecting a certain format?

also, if `doc.title` contains a unique version number, would that gum up other code that expects a list of single / unique agent / subject names?

zstix: Here's the [https://github.com/newrelic/docs-website/blob/campfire/agent-api-index-pages/scripts/utils/migrate/get-frontmatter.js#L89-L92](get-frontmatter.js) file from that branch. I abstracted the regex out to a constant:

```js
const REGEX = {
  MATCH_ON_SEMANTIC_VERSION: /^(.*?)\d+\.\d+(\.\d+){0,2}/,
  MATCH_ON_V_VERSION: /^(.*?)v+\d+/,
};
```

jerelmiller: > it seemed browser was the only one that deviated

I noticed that too! I added that `file.message` warning you see on line 84 to help me find it. I updated the regex above (0b227267f37fdba87a97fe82e4212e278f6846ad) to handle the browser versioning scheme. This now properly detects the browser agent version and adds the `subject` accordingly.

> do you think it's worth checking for those variants 

Yes! I think we should maintain the same format across all agent doc types, so its important that we handle any variations. That is done as mentioned above.

> if `doc.title` contains a unique version number, would that gum up other code that expects a list of single / unique agent / subject names?

I'm not sure I follow what you're asking. Is there a particular bit of functionality that you have in mind?

jerelmiller: Here is an example of the updated frontmatter with the new regex applied: https://github.com/newrelic/docs-website/pull/436/files#diff-a390d53dd4f210fb2e26846dc98d3ac9608763b4917856d375ca87f993a92e0dR2-R4

roadlittledawn: ah gotcha. when i tested the regex via [regexr](https://regexr.com/) against one of the titles it didn't work for me, but it obviously works because the browser release notes have the correct `subject` set. and i ran that little snippet itself and it works  ¯\_(ツ)_/¯ 

re: gumming up other code. i just meant if we use title as last resort (and the title has a unique version number in it), would that mean the nav or the template buckets each of those subjects/agents? but i think we'd catch that with the check for the match and the `file.message()` so we'd presumably fix that

roadlittledawn (approved):
looks good to me!

scripts/utils/migrate/get-frontmatter.js
jerelmiller: Ah gotcha! The template is handling it at least: https://github.com/newrelic/docs-website/blob/develop/src/templates/releaseNote.js#L13

We'll have to keep watch and see if other things are affected by this or not.

PR #451, opened 2020-12-22
cache builds to speed them up

jerelmiller:
## Description

Add `public/static` and `.cache` directories to Amplify's cache on build in an attempt to speed them up.


zstix (approved):
Yeah!! Thank you!

PR #458, opened 2020-12-22
Update migration script to account for sort order

### LINKED ISSUE #349: Migration script doesn't account for navigation sort order
### Description

Currently the migration script isn't automatically setting sort order for categories and pages.

### Example: Categories

category_sort_order_issues

### Example: Pages

doc_sort_order_issues

### Expected behavior

Preserve the sort order from the current docs site as part of the migration script.

### Additional Notes 
- Expose the sort weight in the API and leverage that as part of the script.
- Scope: This is only for the migration script.
- We need to account for release notes nav that uses release date and time to sort items in the nav

### Related to
#381 

# MMF 9 Scope
When migration script is run we want arrange things (both .mdx content files AND the sub-directories) in the nav menus in the order that is set in Drupal. In the Drupal migration API, this order will be exposed via the `order` field. We don't currently have any API resources for taxonomy terms, but we could create them if needed to get the `order` value.

This applies to migration only. Post migration, content contributors can set the order of the nav items by directly editing the .yml files.

## Acceptance criteria
- [ ] As a content contributor, .mdx files and directories/categories in the nav files mirror the current site

roadlittledawn:
Partially closes #349 

## Description
This PR updates the migration script to sort Drupal nodes (authored content) by the `order` field which is read from Drupal's weight in term field. This applies to basic pages, troubleshooting, and api docs.

It also updates the migration script to update the sort of common subdirectories (get started to the top and troubleshooting to the bottom). Ordering the remaining directories is likely best handled either via instruction sets on migration OR by directly editing the nav .yml files to the proper order.

There looks to be some slight ordering issues in .mdx content links. Moving the `.sort()` for directories first fixes that, but mis-orders troubleshooting directories. ¯\_(ツ)_/¯

### DIFF
--- a/scripts/migrate.js
+++ b/scripts/migrate.js
@@ -82,16 +82,19 @@ const run = async () => {
           const bTopic = last(b.data.topics);
           const getStartedRegex = /^Get(ting)? started/i;
           const troubleshootRegex = /^Troubleshoot(ing)?/i;
-          if (aTopic === bTopic) {
-            return 0;
+
+          switch (true) {
+            case aTopic === bTopic:
+              return 0;
+            case getStartedRegex.test(aTopic):
+            case troubleshootRegex.test(bTopic):
+              return -1;
+            case getStartedRegex.test(bTopic):
+            case troubleshootRegex.test(aTopic):
+              return 1;
+            default:
+              return 0;
           }
-          if (getStartedRegex.test(aTopic) || troubleshootRegex.test(bTopic)) {
-            return -1;
-          }
-          if (getStartedRegex.test(bTopic) || troubleshootRegex.test(aTopic)) {
-            return 1;
-          }
-          return 0;
         })
     );
     await fetchDocCount(
--- a/scripts/migrate.js
+++ b/scripts/migrate.js
@@ -16,6 +16,7 @@ const createRawHTMLFiles = require('./utils/migrate/create-raw-html-files');
 const migrateNavStructure = require('./utils/migrate/migrate-nav-structure');
 const reporter = require('vfile-reporter');
 const rimraf = require('rimraf');
+const { last } = require('lodash');
 const path = require('path');
 const { prop } = require('./utils/functional');
 const {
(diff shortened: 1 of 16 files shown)


scripts/migrate.js
```diff
@@ -70,7 +71,28 @@ const run = async () => {
...
+            return 0;
+          }
+          if (getStartedRegex.test(aTopic) || troubleshootRegex.test(bTopic)) {
+            return -1;
+          }
+          if (getStartedRegex.test(bTopic) || troubleshootRegex.test(aTopic)) {
+            return 1;
+          }
+          return 0;
+        })
```
jerelmiller, line 95:
```suggestion
          const troubleshootRegex = /^Troubleshoot(ing)?/i;
          
          if (aTopic === bTopic) {
            return 0;
          }
          
          if (getStartedRegex.test(aTopic) || troubleshootRegex.test(bTopic)) {
            return -1;
          }
          
          if (getStartedRegex.test(bTopic) || troubleshootRegex.test(aTopic)) {
            return 1;
          }
          
          return 0;
        })
```

I'm a snob and I like some space 😅

jerelmiller (dismissed):
Added a minor styling suggestion, but otherwise looks great!

jerelmiller:
> There looks to be some slight ordering issues in .mdx content links

This might not be handled by the nav yaml files, but this might be due to the template itself. It looks like we don't currently order the `mdx` nodes that come back from [this query](https://github.com/newrelic/docs-website/blob/develop/src/templates/apiIndexPage.js#L43-L53), so its possible we just need to adjust to add a sort order there. We are at the mercy of the default ordering from Gatsby itself otherwise.

scripts/migrate.js
zstix: We could also use that fancy `switch (true)` pattern here as well :thinking:

roadlittledawn: i do like how much cleaner the `switch` is. will run with that. thanks!

roadlittledawn:
> > There looks to be some slight ordering issues in .mdx content links
> 
> This might not be handled by the nav yaml files, but this might be due to the template itself. It looks like we don't currently order the `mdx` nodes that come back from [this query](https://github.com/newrelic/docs-website/blob/develop/src/templates/apiIndexPage.js#L43-L53), so its possible we just need to adjust to add a sort order there. We are at the mercy of the default ordering from Gatsby itself otherwise.

FWIW: it looks like they come out of graphql in the same order as they are in the nav .yml. i just did quick test and changed the nav and sidebar updated accordingly.

jerelmiller (approved):

Issue #390, opened 2020-12-11
Build agent API method index pages


roadlittledawn:
## Summary
Some agents have an API category where all their methods are contained (1 method per page/file). There's an index page for that category that has a slightly different look and feel from regular index pages. We'd like to essentially replicate that on new site.

During migration when creating directory indexes, we want to detect if the directory contains API docs and if so, create this special kind of `index.mdx` for the directory. Also during migration, we'll want to use the `shortDescription` field to provide a method summary on the index page.

The API method index page should contain:
* intro text at top
* list (sorted by title, alphabetically) of relevant API doc titles, their short descriptions (to be added to API doc frontmatter), link to the method's page.

These should be automatically generated so a user does not have to edit this page manually if a method is added, removed, or a summary changes.

### Caveats
* The block of content above the list of methods isn't available via JSON. Will have to figure out how to get that into migration script

## Acceptance criteria
- [ ] Each agent API directory index file uses this type of index page

## Helpful links
[Example: PHP agent API index page](https://docs.newrelic.com/docs/agents/php-agent/php-agent-api)

jerelmiller:
@roadlittledawn how would we get the intro text that shows up for these pages? I assume these pages aren't available in the JSON API and won't be migrated correct?

PR #460, opened 2020-12-22
Sort api doc list alphabetically on index pages

roadlittledawn:
## Description

Small change to sort api doc list on api index pages alphabetically by title (which for api docs is their method name) so it's easier to scan.

* #390


jerelmiller (approved):
🎉

Issue #124, opened 2020-10-06
Find replacements for font-awesome icons


jerelmiller:
## Description

As a part of our migration process, we are updating icons from font awesome to feather icons. There are a few icons that we are unable to map. They include:

* [`caret-up`](https://fontawesome.com/icons?d=gallery&q=caret-up) - [source](https://docs.newrelic.com/docs/browser/new-relic-browser/browser-pro-features/browser-error-profiles-find-error-causes)
* [`quote-left`](https://fontawesome.com/icons/quote-left?style=solid) - [source](https://docs.newrelic.com/docs/browser)
* [`quote-right`](https://fontawesome.com/icons/quote-right?style=solid) - [source](https://docs.newrelic.com/docs/browser)
* `android` - [source](https://docs.newrelic.com/docs/mobile-monitoring)
* `apple` - [source](https://docs.newrelic.com/docs/mobile-monitoring)

**NOTE:** 

This will be a running list that will change over time as we discover more icons that don't have a direct equivalent. 

The links above refer to the latest version of fontawesome. The docs site is currently using fontawesome v4 which have some variance in the styles.

## Design input 

Identify the feather icons to use for ones that cannot be currently mapped.

* [`quote-left`](https://fontawesome.com/icons/quote-left?style=solid) >> Curious about what context this appears in. Perhaps [`message-square`](https://feathericons.com/?query=message-square)?
* [`quote-right`](https://fontawesome.com/icons/quote-right?style=solid) >> Why do we need separate `quote-left` and `quote-right` replacements?
* `android` >> Suggest keeping the existing icon from FontAwesome
* `apple` >> Suggest keeping the existing icon from FontAwesome

caylahamann:
Question: does this only include the icons in the current migrated docs? / will we have to expand this when we decide to migrate more docs?

jerelmiller:
@jpvajda @djsauble Would love for you to look at this and provide recommendations as we migrate the docs.

djsauble:
@jerelmiller It would be helpful to see an example of the context in which these icons appear.

paperclypse:
@djsauble Let me know if you need help tracking down examples of icon examples on the docs site.

djsauble:
@paperclypse That'd be awesome! I left some questions in the description above, basically I'm uncertain about the context of use for some of these icons, which is important in finding adequate substitutions for them.

jerelmiller:
[...]

PR #462, opened 2020-12-23
Map more missing icons

jerelmiller:
Relates to #124

## Description

Adds and maps more missing icons in the docs and runs the migration to apply it. At this stage, we are only missing 5 icons with the docs that we have pulled in.


zstix (approved):
🖼

PR #468, opened 2020-12-23
Migrate custom heading IDs

### LINKED ISSUE #111: Support custom heading IDs
## Summary

As part of the migration, we need to allow custom IDs to be set on headers. This is supported via the [extended syntax for custom heading IDs](https://www.markdownguide.org/extended-syntax/#heading-ids). With the changes to the parsing engine in MDX v2, this syntax is now treated as a JSX expression (understandably), therefore breaking the ability to use this syntax.

I've opened issues in the [`@mdx-js/mdx`](https://github.com/mdx-js/mdx/issues/1279) and [`gatsby`](https://github.com/gatsbyjs/gatsby/issues/27120) repos about this question. From the sounds of it, this is likely something that the `gatsby-remark-autolink-headers` would need to handle.

With this restriction, we won't be able to fully migrate custom IDs on headers right now. Ideally this changes between now and GA. This issue will serve as a reminder to come back to this issue before we go live with the site, otherwise we risk breaking links between pages that link to a specific section of the page.

## Alternatives

If it comes down to it, we have a couple options that we'll need to pursue ourselves to patch this issue:

1. Patch the change ourselves
  This would require us to fork the `gatsby-remark-autolink-headers` plugin and patch the change ourselves. That or we create our own plugin that mimics this one with support for MDX v2.

2. Mass-migrate all existing links to point to the automatically generated IDs 
  We already have the ability to modify the AST of the files. It might make sense to just go ahead and create a tool that would allow us to analyze all relative page links site-wide and update them to the new header. This is much more complex, but the benefit is that would could utilize this tool for contributors to have headings update if/when the text of that heading changes.

jerelmiller:
Closes #111

## Description

Migrates the IDs on headings using the custom heading syntax (#465) to ensure they remain the same after we GA.


zstix (approved):
I assume that the adjustments in this PR (other than the MDX file updates) was just the stuff in `html-to-markdown`?

jerelmiller:
Thats correct!

PR #477, opened 2020-12-24
Watermarks

### LINKED ISSUE #335: Add watermark/other way of marking an entire doc

On the current docs site, we mark beta/limited release/legacy/deprecated docs with watermarks like [the one on this custom views doc](https://docs.newrelic.com/docs/plugins/plugins-new-relic/custom-dashboards-custom-views/custom-views-deprecated):

image

The new site doesn't have an equivalent visual "tag" for an entire document.

Some way of marking the status of an entire doc. A callout doesn't work because we want it to be visible wherever you are in the doc: At the top, at the bottom, somewhere in the middle. But I'm not at all tied to the watermarking solution. A header, something in sidebar, etc could all work.

### Acceptance Criteria

1. allow a contributor to have  any easy way to mark a page as "deprecated", "beta", or "limited release" etc
2. the component should be visible at the top of a page and as a user scrolls through the page, the component should be "sticky" and always display at the top of the page.
3. This component should be part of the global theme and available to the OSS / Dev sites as well.

Allows users to easily determine the status of "special" docs to avoid confusion. (For example, a new user building on a legacy solution instead of a new one; or someone running a beta feature in a prod environment.)

jerelmiller:
Closes #335

## Description

Show watermarks for pages that include a watermark. Adds the codemod and the component. 

## Screenshots

Screen Shot 2020-12-24 at 12 05 33 AM
Screen Shot 2020-12-24 at 12 05 48 AM
Screen Shot 2020-12-24 at 12 06 04 AM

### DIFF
--- a/codemods/index.js
+++ b/codemods/index.js
@@ -14,6 +14,7 @@ const landingPageImagePlacement = require('./landingPageImagePlacement');
 const collapseLandingPageTileGrids = require('./collapseLandingPageTileGrids');
 const landingPageButtonType = require('./landingPageButtonType');
 const tocLinks = require('./tocLinks');
+const watermarks = require('./watermarks');
 
 module.exports = [
   codeBlocks,
@@ -32,4 +33,5 @@ module.exports = [
   collapseLandingPageTileGrids,
   landingPageButtonType,
   tocLinks,
+  watermarks,
 ];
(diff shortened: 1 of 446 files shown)


zstix (approved):
dat progress bar tho... 🎨  😩 👌

PR #492, opened 2020-12-29
Refresh attribute dictionary bug

### LINKED ISSUE #351: Refreshing the attribute dictionary page when filtered does not work correctly
## Description

When filtering on the attribute dictionary page, the filtering works as expected, however refreshing the page does not filter correctly. I would expect the same event/attribute to be filtered after refreshing, but this does not work correctly. After a refresh, it appears the attribute name is filtered, but it still lists all the events.

### Steps to reproduce

1. Go to the [attribute dictionary page](https://docs-preview.newrelic.com/attribute-dictionary)
2. Filter by an event and attribute
3. Refresh the page

You'll notice the attribute is filtered correctly, but the events are not.

### Expected behavior

The event/attribute is filtered according to the query params.

caylahamann:
closes #351 

can refresh the page with query parameters specified and the attribute dictionary will show the correct event and attribute! This bug only shows up on build so you might want to either build locally or wait til amplify builds the page.

### DIFF
--- a/src/pages/attribute-dictionary.js
+++ b/src/pages/attribute-dictionary.js
@@ -1,4 +1,4 @@
-import React, { memo, useMemo } from 'react';
+import React, { memo, useMemo, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/core';
 import { graphql, Link } from 'gatsby';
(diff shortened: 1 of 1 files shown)


zstix (commented):
I'm not 100% clear what was causing the issue, but this looks good. I'll wait to test it out on the amplify build before I 👍

caylahamann:
> I'm not 100% clear what was causing the issue, but this looks good. I'll wait to test it out on the amplify build before I 👍

TBH I'm still a little confused too, but I switched to a useEffect because I thought it was a good way to trigger the filtering after the site had loaded (which I assume might have been the problem, it wasn't getting updating)

roadlittledawn (approved):
fix works for me when i built it locally!

PR #496, opened 2020-12-30
Migrate short titles to nav files

### LINKED ISSUE #156: Add support for short titles
## Summary

On the docs site today, many docs have two "titles":

* A "title" that is used in the HTML `<head><title>` and the H1 at the top of the page.
* A "short title" that's used, when set, in the sidebar nav instead of the "title"

This is helpful because it lets you have longer, more descriptive titles for search engines and page heads, while using pithier titles that take into account the context in the side nav.

## Desired Behavior

* Add fields for both in the page metadata
* Pull both from Drupal for existing nodes
* Use the two fields as described above

roadlittledawn:
Closes #156 

## Description
Uses the newly added `shortTitle` field in migration API JSON to set the `title` field in nav files if applicable.

Upon running migration, also removes a rogue duplicate release note files (because it was removed from Drupal today)


caylahamann (approved):
nooooice

austin-schaefer:
I missed this one. Woohoo! Thanks @roadlittledawn

PR #498, opened 2020-12-30
Sets up JSON endpoints for in-product help xp

### LINKED ISSUE #426: Setup JSON endpoints to serve content into In Product Help XP component
When we cut the site over to the new Docs Site we need to recreate the JSON endpoints that send sends data to the in-product XP component. That component seems to render the docs site content, different then the actual site itself but it relies on the docs library service 

**Current state**
The docs library service (which is stitched into nerdgraph which the NR1 Help XP gets swiftype and docs content from) consumes JSON endpoints from Drupal.

caylahamann:
creates a JSON with the structure: 
```js
{
  body: /*html*/
}
```
for each doc at the each slug. You can see this on the preview branch by adding .json to the end of any file url.

closes #426


gatsby-config.js
```diff
@@ -190,6 +190,7 @@ module.exports = {
         path: `./src/nav/`,
       },
     },
+    'gatsby-plugin-generate-doc-json',
```
zstix, line 193: We might be able to leverage [`gatsby-plugin-json-output`](https://github.com/dominicfallows/gatsby-plugin-json-output) to generate these files. We use it below to generate a single JSON for the data dictionary (via `serializeFeed` option) but you can use it with the `serialize` option instead to generate a JSON file for each node.

Most of the logic you have would be the same, we just don't need to make a custom plugin!

caylahamann: Yeah I first tried to do that, but the problem was that there is some specific log for it (specifically having to do with the path and the slug) which is why I opted for creating my own plugin, I'm also going to do some more transformations on the data so I still push for having it's own plugin

zstix:
```js
{
  resolve: 'gatsby-plugin-json-output',
  options: {
    siteUrl,
    graphQLQuery: `
      {
        ...your query goes here
      }
    `,
    serialize: results => results.data.allMdx.nodes.map((node) => {
      // your code goes here
    });
  }
}
```

caylahamann: there's a json created for each path, rather than a json that has all the docs files (which would be generated if we use the serialize function. If we wanted to put all the docs files into one json i would advocate for using `gatsby-plugin-json-output` !

zstix: Ah alright. I _think_ this library would still work for those edge cases, but I'm not going to complain about moving logic out of `gatsby-config.js` and into specialized files.

caylahamann: OH you're suggesting using `gatsby-plugin-json-output` not the other one! Sorry I misread this

caylahamann: I'm making some changes to this PR and doing some AST codemod things so it's easier for nr1 help xp to read, so let me push up those changes and then maybe you can determine whether it would still be feasible to put it in the config file

plugins/gatsby-plugin-generate-doc-json/gatsby-node.js
```diff
@@ -0,0 +1,40 @@
...
+const htmlGenerator = unified().use(removeImports).use(removeExports).use(html);
+
+exports.onPostBuild = async ({ graphql, store }) => {
+  const { program } = store.getState();
+
+  const query = `query {
+    allMdx(filter: {fileAbsolutePath: {regex: "/src/content/docs/"}}) {
+      nodes {
+        mdxAST
+        slug
```
jerelmiller, line 18: Interesting! I didn't realize there was a `slug` field already here. It looks like we are using [`fields.slug`](https://github.com/newrelic/docs-website/blob/develop/plugins/gatsby-plugin-release-note-rss/gatsby-node.js#L30-L32) elsewhere. Are these 2 equivalent?

jerelmiller (approved):

jerelmiller:
💯

PR #500, opened 2021-01-04
Add redirects to frontmatter for basic pages

### LINKED ISSUE #364: Handle redirects from the old Docs site to the new Doc Site
With the change in site IA, some URLs will no longer route to the correct location on the new site. Also, existing redirects in the old site are in place for URLs that have changed over time. Therefore we need to support the redirect of URLs to the correct locations when the new Doc Site launches.

### Goal 

 There are currently ~7550 redirects in total.  Ideally we map all URLS to the correct location with a goal of 100% mapping, or a fall-back goal of all those that have been accessed in the last 365 Days

### TODOs
- [x] review URL data set and decide on approach
- [x] Break down work as needed.

### Considerations 

- consider mapping the old IA to new IA and how we programmatically can determine a redirect location

- perhaps we consider a `catch` approach were if there isn't a 1:1 for match we direct them to the correct category on the site.  This may allow us to feel confident we've completed the mapping but ensure if a URL was missed it would go to a location that would make sense to a user.

- finally, all broken redirects would go to 404 pages, so part of our instrumentation approach could be to capture what URL triggered a 404 for redirect analysis.

roadlittledawn:
Partially addresses #364 

## Description

This adds module to fetch redirect data for Drupal nodes (authored pieces of content). For now, it only adds redirect URLs to files with the `page` content type. It creates redirects for each basic page if available in frontmatter.

This also includes small refactor of custom frontmatter function for basic pages.


scripts/utils/migrate/fetch-redirects.js
```diff
@@ -0,0 +1,35 @@
...
+  }
+};
+
+const fetchAllRedirects = async () => {
+  const redirects = await paginate('/api/migration/redirects', {
+    perPage: 300,
+  });
+
+  return redirects;
+};
```
jerelmiller, line 33: [nit] Since `paginate` returns a promise, you can simplify this:

```suggestion
const fetchAllRedirects = () => {
  return paginate('/api/migration/redirects', {
    perPage: 300,
  });
};
```

scripts/utils/migrate/get-frontmatter.js
```diff
@@ -54,18 +54,21 @@ const addCustomFrontmatter = {
       topics,
     };
   },
-  [TYPES.BASIC_PAGE]: ({ doc, topics }, defaultFrontmatter) => {
+  [TYPES.BASIC_PAGE]: ({ doc, topics, redirects }, defaultFrontmatter) => {
+    const frontmatter = {
+      ...defaultFrontmatter,
+      topics,
+      redirect_from: redirects,
```
jerelmiller, line 61: Can we instead use `redirects` so that we are consistent with the [dev site](https://raw.githubusercontent.com/newrelic/developer-website/b575317acf8b2962f2c2fd1e82a64c4efd5b0e29/src/markdown-pages/automate-workflows/index.mdx)?

scripts/migrate.js
```diff
@@ -94,6 +95,22 @@ const run = async () => {
     rimraf.sync(NAV_DIR);
     rimraf.sync(DICTIONARY_DIR);
 
+    logger.info('Fetching redirects');
+
+    const orderedRedirects = (docs) => {
```
jerelmiller, line 100: [nit[ Can we pull this function outside of the `run` function since it doesn't depend on anything except its argument? This makes just a bit easier to read the `run` script

jerelmiller: On another note, this doesn't look like it really orders the redirects, but rather sticks them in an object. Did you mean to `sort` these? If not, maybe you could change the name to something that better reflects what this is doing like `groupRedirectsByPage` or something?

scripts/utils/migrate/fetch-redirects.js
```diff
@@ -0,0 +1,35 @@
+const fetch = require('node-fetch');
+const logger = require('../logger');
+const { paginate } = require('./docs-api');
+
+require('dotenv').config();
+
+const fetchRedirects = async (nodeId) => {
```
jerelmiller, line 7: It doesn't look like this is used. Is this needed?

scripts/migrate.js
```diff
@@ -94,6 +95,22 @@ const run = async () => {
...
+      return docs.reduce((accum, curr) => {
+        let { to, from } = curr.redirect;
+        to = to.replace(/\//g, '');
```
jerelmiller, line 103: [nit] It looks like you are looking up the [node by its ID]https://github.com/newrelic/docs-website/pull/500/files#diff-9c154e1779b4a100a8a1713152855b8554a563f6360c2393bacbb55f465747f2R38). Rather than needing to look it up via `node<id>`, could you just replace the `node` text here to make it easier to lookup?

```suggestion
        to = to.replace(/\//g, '').replace('node', '');
```

This makes it easier to lookup later:

```js
redirect[doc.docId]
```

src/content/docs/accounts-partnerships/accounts/saml-single-sign/configure-saml-multiple-accounts.mdx
```diff
@@ -6,7 +6,10 @@ topics:
   - Accounts
   - Original accounts and billing
   - SAML SSO (original users)
-japaneseVersion: ''
+redirect_from:
+  - /docs/subscriptions/configuring-saml-with-multiple-accounts/
```
jerelmiller, line 10: Can you double check that trailing slashes redirect properly? I think we are using stripped paths when creating pages, so I'm not sure if this causes issues or not. It might be best to strip the trailing slash anyways since we want to encourage content authors to use the stripped slash.

jerelmiller (changes requested):

jerelmiller:
@roadlittledawn you shouldn't need any of those redirect plugins to do the work. You can create the redirect yourself: https://github.com/newrelic/docs-website/blob/develop/gatsby-node.js#L160-L163

jerelmiller:
Here's an example of how we do it on the dev site: https://github.com/newrelic/developer-website/blob/develop/gatsby-node.js#L52-L61

scripts/migrate.js
```diff
@@ -94,6 +95,22 @@ const run = async () => {
     rimraf.sync(NAV_DIR);
     rimraf.sync(DICTIONARY_DIR);
 
+    logger.info('Fetching redirects');
+
+    const orderedRedirects = (docs) => {
+      return docs.reduce((accum, curr) => {
```
zstix, line 101: Nitpick: can we use `acc` rather than `accum`? I believe we use this convention with our other `reduce` usage.

scripts/utils/migrate/fetch-redirects.js
```diff
@@ -0,0 +1,35 @@
...
+  const url = `https://docs-dev.newrelic.com/api/migration/redirects?id=${encodeURIComponent(
+    `node/${nodeId}`
+  )}`;
+
+  try {
+    const res = await fetch(url, {
+      headers: {
+        'Content-Type': 'application/json',
+        'X-Phpshield-Key-Disable': process.env.ACQUIA_DEV_PHP_SHIELD_KEY,
+      },
```
zstix, line 17: Nit: we can probably move these headers up into a constant rather than redefining them every time.

scripts/utils/migrate/fetch-redirects.js
```diff
@@ -0,0 +1,35 @@
...
+    });
+    const json = await res.json();
+    const { redirects } = json;
+    return redirects;
```
zstix, line 21: Could be simplified to:
```suggestion
    const json = await res.json();
    return json.redirects;
```

scripts/utils/migrate/get-frontmatter.js
zstix: I _think_ one of the plugins we were talking about using expects this format?

roadlittledawn: yeah, i used `redirects_from` to make it compatible with that plugin, but since we're likely not going to use said plugin, i'll change it to `redirects`.

scripts/migrate.js
roadlittledawn: gotcha. yeah, you're right. it groups them by the Drupal ID, so will rename it.

roadlittledawn: right after I PR'ed this, i realized that i should probably move that `orderedRedirects()` code to `fetch-redirects.js`. wasn't super sure if i should i just include that as part of `fetchAllRedirects()` or as its own function that i call on `fetchAllRedirects()`. i went with the latter and will push up

scripts/utils/migrate/fetch-redirects.js
roadlittledawn: refactored that file a bit so have to now i'm awaiting the result of `paginate` to run the grouping function on it

scripts/utils/migrate/fetch-redirects.js
roadlittledawn: that was the method i used for first approach, but ditched it in favor of getting all at once. i left it in case we needed a way to get redirects by passing the docsId to it. i'll remove for now to clean up the file some

scripts/utils/migrate/fetch-redirects.js
roadlittledawn: after refactor, that method was removed because we weren't using it.

scripts/migrate.js
roadlittledawn: i wasn't sure if i should leave `node` in the key name, because the redirect data also contains taxonomy term redirects, too. so if/when we wanted to get those out, i thought it might be useful to know a key is a Drupal node or a taxonomy term.

src/content/docs/accounts-partnerships/accounts/saml-single-sign/configure-saml-multiple-accounts.mdx
roadlittledawn: yep, will do. that was a requirement of that other plugin.

scripts/utils/migrate/fetch-redirects.js
roadlittledawn: Removed this method from the module after refactor

scripts/utils/migrate/fetch-redirects.js
```diff
@@ -0,0 +1,27 @@
...
+const { paginate } = require('./docs-api');
+
+require('dotenv').config();
+
+const fetchAllRedirects = async () => {
+  const redirects = paginate('/api/migration/redirects', {
+    perPage: 300,
+  });
+
+  return groupRedirectsByDocsId(await redirects);
```
jerelmiller, line 11:
```suggestion
  const redirects = await paginate('/api/migration/redirects', {
    perPage: 300,
  });

  return groupRedirectsByDocsId(redirects);
```

minor [nit] to move the `await` to the first expression instead of the 2nd.

src/content/docs/accounts-partnerships/install-new-relic/install-agent/install-agent.mdx
```diff
@@ -6,7 +6,7 @@ topics:
   - Accounts
   - Accounts and billing
   - Account setup
-japaneseVersion: ''
+redirects: []
```
jerelmiller, line 9: Should we omit the `redirects` property if its empty?

jerelmiller (approved):
Looks great! Thanks so much for updating those comments!

jerelmiller (approved):

PR #510, opened 2021-01-05
chore: updated documentation

jpvajda:
## Description

removed an template readme file we don't need
updated readme and contributor guide


CONTRIBUTING.md
```diff
@@ -41,6 +44,27 @@ guidelines below.
...
+yarn start
+```
+
+Your site is now running at `http://localhost:8000`!
+
+### Dependencies
+
+Node v12 is used in this project as specified in [.nvmrc](https://github.com/newrelic/developer-website/blob/master/.nvmrc).
+
+## Unit tests
```
jerelmiller, line 63: It looks like the following set of content is considered a child of this `Unit tests` section (see the ToC). I'm not sure how best to update this, but could you see if you can make sure this is organized so that this unit tests section is at the same level as the "using multiple versions of node", etc.?

CONTRIBUTING.md
```diff
@@ -41,6 +44,27 @@ guidelines below.
...
+
+To run the unit tests, run `yarn test` in the terminal. If you would like to
+have the tests automatically re-run, use `yarn run test:watch
```
jerelmiller, line 66:
```suggestion
have the tests automatically re-run, use `yarn run test:watch`
```

jerelmiller (commented):
2 minor things, otherwise looks great!

CONTRIBUTING.md
jpvajda: Yeah I can fix, it's just the Markdown Header (h1, h2, h3) that sets that in the TOC.

CONTRIBUTING.md
jpvajda: Will fix in my next commit.. good catch

jerelmiller (approved):
Nice! Thanks so much for doing this!

PR #515, opened 2021-01-06
Add language toggle to docs site and build locales for src/pages

### LINKED ISSUE #483: Build index, 404 pages, attribute dictionary for Japanese site
## Summary 
We need to build our pages (under the `/pages` directory) for /jp sub-url. We can do this either by an existing plugin ([`gatsby-plugin-intl`](https://www.gatsbyjs.com/plugins/gatsby-plugin-intl/) ) or by building our own plugin to build these pages. (I'd recommend the former). This is just getting the pages to build with the jp suburl, internationalizing the content is in a different PR.

## AC 
* [ ] pages build for japanese and english url

caylahamann:
closes #483

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -48,6 +48,9 @@ module.exports = {
           contentPadding: '2rem',
           maxWidth: '1600px',
         },
+        i18n: {
+          additionalLocales: [{ name: 'Japanese', locale: 'jp' }],
+        },
         prism: {
           languages: [
             'xml',
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -49,7 +49,7 @@ module.exports = {
           maxWidth: '1600px',
         },
         i18n: {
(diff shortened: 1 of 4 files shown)


gatsby-config.js
```diff
@@ -48,6 +48,9 @@ module.exports = {
           contentPadding: '2rem',
           maxWidth: '1600px',
         },
+        i18n: {
+          additionalLocales: [{ name: 'Japanese', locale: 'jp' }],
```
jerelmiller, line 52: Now that I think of it, we should probably copy the text from the docs site to use as the name

Screen Shot 2021-01-06 at 10 22 13 AM

jerelmiller (commented):
1 minor comment, otherwise this looks great!

gatsby-config.js
caylahamann: You're very right!! good idea, I'll make that change

jerelmiller (approved):

PR #527, opened 2021-01-09
Add translations for 404, whats-new, attribute-dictionary

roadlittledawn:
Closes #527 

## Description

Implements `i18next` via the theme for 404, whats-new, and attribute-dictionary pages.

No Japanese translations currently exist for that content, so just English was added.

### DIFF
--- a/src/i18n/translations/en/translation.json
+++ b/src/i18n/translations/en/translation.json
@@ -1,12 +1,12 @@
 {
   "404": {
-    "h1": "404",
-    "text": "The URL you entered may be broken, or the page has been removed.<2>Go back to the home page</2>"
+    "statusCode": "404",
+    "errorMessage": "The URL you entered may be broken, or the page has been removed.<1> </1><2>Go back to the home page</2>"
   },
   "dataDictionary": {
     "title": "New Relic data dictionary",
-    "intro": "<0>This data dictionary lists and defines the <2>attributes</2> attached to New Relic events and other data objects (like Metric and Span data).</0>",
-    "callout": "<0>This dictionary does not contain data reported by Infrastructure integrations. To learn about that data, see the <2>integration documentation</2>.</0>"
+    "intro": "This data dictionary lists and defines the <2>attributes</2> attached to New Relic events and other data objects (like Metric and Span data).",
+    "callout": "This dictionary does not contain data reported by Infrastructure integrations. To learn about that data, see the <2>integration documentation</2>."
   },
   "whatsNew": {
     "title": "What's new in New Relic"
--- a/src/i18n/translations/en/translation.json
+++ b/src/i18n/translations/en/translation.json
@@ -1,7 +1,7 @@
 {
   "404": {
     "statusCode": "404",
-    "errorMessage": "The URL you entered may be broken, or the page has been removed.<1> </1><2>Go back to the home page</2>"
+    "errorMessage": "The URL you entered may be broken, or the page has been removed. <2>Go back to the home page</2>"
   },
   "dataDictionary": {
     "title": "New Relic data dictionary",
--- a/src/i18n/translations/en/translation.json
+++ b/src/i18n/translations/en/translation.json
@@ -1,1 +1,14 @@
-{}
+{
+  "404": {
+    "h1": "404",
+    "text": "<0>The URL you entered may be broken, or the page has been removed.</0><1> </1><2>Go back to the home page</2>"
+  },
+  "dataDictionary": {
+    "title": "New Relic data dictionary",
+    "intro": "<0>This data dictionary lists and defines the <2>attributes</2> attached to New Relic events and other data objects (like Metric and Span data).</0>",
+    "callout": "<0>This dictionary does not contain data reported by Infrastructure integrations. To learn about that data, see the <2>integration documentation</2>.</0>"
+  },
+  "whatsNew": {
+    "title": "What's new in New Relic"
+  }
+}
--- a/src/i18n/translations/en/translation.json
+++ b/src/i18n/translations/en/translation.json
@@ -1,7 +1,7 @@
 {
   "404": {
     "h1": "404",
-    "text": "<0>The URL you entered may be broken, or the page has been removed.</0><1> </1><2>Go back to the home page</2>"
+    "text": "The URL you entered may be broken, or the page has been removed.<2>Go back to the home page</2>"
   },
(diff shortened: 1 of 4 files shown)


src/i18n/translations/en/translation.json
```diff
@@ -1 +1,14 @@
-{}
\ No newline at end of file
+{
+  "404": {
+    "h1": "404",
```
jerelmiller, line 3:
```suggestion
    "statusCode": "404",
```

Could we instead name this after the status code instead of the tag? If we ever change the design of this page and this is no longer an `h1`, then this key gets kinda weird.

src/i18n/translations/en/translation.json
```diff
@@ -1 +1,14 @@
...
+    "text": "<0>The URL you entered may be broken, or the page has been removed.</0><1> </1><2>Go back to the home page</2>"
```
zstix, line 4: I _think_ this can be simplified to:
```suggestion
    "text": "The URL you entered may be broken, or the page has been removed.<2>Go back to the home page</2>"
```

jerelmiller, line 4:
```suggestion
    "errorMessage": "<0>The URL you entered may be broken, or the page has been removed.</0><1> </1><2>Go back to the home page</2>"
```

`text` feels a bit too generic to me. Could we instead make it a bit more specific so that design changes to this page allow us to better target the area on the page?

jerelmiller: On that note, I was under the impression that you didn't need the tags unless it was a component. Here I'd expect to see the value as:

```
The URL you entered may be broken, or the page has been removed. <2>Go back to the home page</2>
```

src/pages/404.js
```diff
@@ -42,12 +48,12 @@ const NotFoundPage = ({ pageContext }) => {
                 line-height: 1;
               `}
             >
-              404
+              {t('404.h1')}
             </h1>
-            <p>
+            <Trans i18nKey="404.content" parent="p">
```
jerelmiller, line 53: Looks like the key you had above was named `404.text`. Could you make sure this matches the key?

src/i18n/translations/en/translation.json
zstix: Same with the two things in the `dataDictionary` section: if the element is just a plain text, it doesn't look like we need to wrap it.

I realize that this might have been auto-generated tho, so maybe it's not worth worrying about.

zstix (commented):
Looks good to me. I'm gonna wait to check it out on the PR build before ✅

src/pages/attribute-dictionary.js
```diff
@@ -76,31 +80,34 @@ const AttributeDictionary = ({ data, pageContext }) => {
...
+            >
+              This data dictionary lists and defines the{' '}
+              <Link to="/docs/using-new-relic/welcome-new-relic/getting-started/glossary#attribute">
+                attributes
+              </Link>{' '}
+              attached to New Relic events and other data objects (like Metric
+              and Span data).
+            </p>
+          </Trans>
+          <Trans i18nKey="dataDictionary.callout">
```
jerelmiller, line 104: [nit] If you move the `Trans` inside the `Callout`, then you can avoid the need for `Trans` to interpolate the extra component. `Trans` by default doesn't render a container, so that might make this a bit nicer:

```
<Callout>
  <Trans>

  </Trans>
</Callout>
```

src/pages/attribute-dictionary.js
```diff
@@ -76,31 +80,34 @@ const AttributeDictionary = ({ data, pageContext }) => {
...
-            attached to New Relic events and other data objects (like Metric and
-            Span data).
-          </p>
-          <Callout variant={Callout.VARIANT.TIP}>
-            This dictionary does not contain data reported by Infrastructure
-            integrations. To learn about that data, see the{' '}
-            <Link to="/docs/integrations">integration documentation</Link>.
-          </Callout>
-
+          <Trans i18nKey="dataDictionary.intro" parent="div">
```
jerelmiller, line 89: It looks like we didn't have this `div` wrapping the content before, just the `p`. Could you instead merge this `Trans` component with its `p` child?

```js
<Trans parent='p' css={css`...`}>
  // ...
</Trans>
```

jerelmiller (changes requested):
Had a few suggestions for you!

src/i18n/translations/en/translation.json
roadlittledawn: Thanks! I started using the extraction tool after doing this one manually. Made the change

src/i18n/translations/en/translation.json
roadlittledawn: yep can do. i was thinking using `h1` is generic enough in case that title ever changed.

src/i18n/translations/en/translation.json
roadlittledawn: yes indeed. i used the extractor after i did this one manually. updated it. thanks for catching that! also, updated the `text` to `errorMessage` (i actually used that same name originally but changed it to be more generic 😆 )

src/pages/404.js
roadlittledawn: ✅

src/pages/attribute-dictionary.js
roadlittledawn: ah true dat. yeah i wasn't totally sure what was optimal there, but i like that. done!

src/pages/attribute-dictionary.js
roadlittledawn: oh right i can just add the emotion `css` prop on the component. sweet! done

src/i18n/translations/en/translation.json
```diff
@@ -1,12 +1,12 @@
 {
   "404": {
-    "h1": "404",
-    "text": "<0>The URL you entered may be broken, or the page has been removed.</0><1> </1><2>Go back to the home page</2>"
+    "statusCode": "404",
+    "errorMessage": "The URL you entered may be broken, or the page has been removed.<1> </1><2>Go back to the home page</2>"
```
jerelmiller, line 4:
```suggestion
    "errorMessage": "The URL you entered may be broken, or the page has been removed. <2>Go back to the home page</2>"
```

jerelmiller (dismissed):

jerelmiller (approved):

jerelmiller (approved):

PR #528, opened 2021-01-09
Transform index page content for i18n

### LINKED ISSUE #486: Transform index page to render react-intl context
## Summary 
`index.js` has a lot of content on its page, and we would like to render this content specific to the language the user is on. This means we need to transform index.js to accept the language context and render the correct content. We will need to extract out the text out to the en.json and also translate the homepage.yml file (we can do this by sending the yaml to our vendor)

## AC 
* [ ] Send homepage.yml to vendor (talk to Julien) 
* [ ] Extract text out into en.json 
* [ ] Render en.json content via react-intl context

zstix:
## Description
Moves the homepage copy out of the `pages/index.js` component and into the english translation JSON.

* Closes #486

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -7,6 +7,7 @@ import {
   Icon,
   Surface,
   useTranslation,
+  Trans,
 } from '@newrelic/gatsby-theme-newrelic';
 import { rgba } from 'polished';
 import IntegrationIcon from '../components/IntegrationIcon';
@@ -18,7 +19,7 @@ const HomePage = ({ data }) => {
     site: { layout },
   } = data;
 
-  const { t, i18n } = useTranslation();
+  const { t } = useTranslation();
 
   return (
     <>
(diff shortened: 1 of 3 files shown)


roadlittledawn (approved):
Looks good to me!

src/pages/index.js
```diff
@@ -87,68 +94,91 @@ const HomePage = ({ data }) => {
...
+            :
+          </Trans>
         </SectionDescription>
 
-        {integrations.map(({ title, tiles }, idx) => (
+        {integrations.map((integration, idx) => (
           <Fragment key={idx}>
-            <IntegrationTitle>{title}</IntegrationTitle>
+            <IntegrationTitle>
+              {t(`home.integrations.h${idx + 1}`)}
```
jerelmiller, line 178: [nit] Could we name this key `home.integrations.title${idx}`? The `h` makes me think its named after an `h1`/`h2` tag, which was a bit misleading in my brain 😅

jerelmiller (commented):
one minor suggestion, otherwise looks great!

src/pages/index.js
zstix: Yeah, I thought that could be wonky. I chose `h` because of heading, but that's certainly confusing. I'll update it with your suggestion.

caylahamann (approved):
also looks good to me :)

PR #536, opened 2021-01-11
Add translations for two docs components

### LINKED ISSUE #484: Transform docs site components to render react-intl context
## Summary 
With`gatsby-plugin-intl` set up, we will be able to create `en.json` and `jp.json` to render language specific content for our components. We want to transform our components to render the content from json. We want to this first for the two components that have hard coded content within them. These are: 
* DefaultRelatedContent.js
* Subnavigation.js 

## AC 
* [ ] Take the english content and put it into the en.json file, and then render it into the component with the correct id. For more information, consult the `gatsby-plugin-intl` docs!

roadlittledawn:
## Description

Adds English translation for two docs components. Also adds the Japanese translation for defaultRelatedContent since it was readily available.

Had to bump theme so SubNavigation could use translation context. Removed superfluous `gatsby-plugin-layout`. 

Adds `i18next-parser` and yarn command to extract new translation keys to relevant translation file.

## Related issues
Closes #484


i18next-parser.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  lexers: {
+    js: ['JsxLexer'],
+  },
+  defaultNamespace: 'translation',
+  locales: ['en', 'jp'],
+  output: './src/i18n/translations/$LOCALE/translation.json',
+  input: ['src/**/!(Trans)*.js'],
```
jerelmiller, line 8:
```suggestion
  input: ['src/**/*.js'],
```

You can tweak this slightly since we don't have a custom `Trans` component in this repo. I was encountering errors from this tool when trying to extract translations, so I ignored it as part of the input.

src/components/DefaultRelatedContent.js
```diff
@@ -12,37 +20,37 @@ const DefaultRelatedContent = () => {
...
-          <a href="/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics">
-            New Relic Diagnostics
-          </a>
-          , our troubleshooting tool for Linux, Windows, and macOS.
-        </li>
-        <li>
-          Review New Relic's <a href="/docs/security">data security</a> and{' '}
-          <a href="/docs/licenses">licenses</a> documentation.
-        </li>
+        <Trans i18nKey="defaultRelatedContent.list">
```
jerelmiller, line 27:
```suggestion
        <Trans i18nKey="defaultRelatedContent.list" parent="ul">
```

You might consider making the `Trans` the `ul` instead of nesting it, since `ul` doesn't take any props.

src/components/SubNavigation.js
```diff
@@ -2,46 +2,49 @@ import React from 'react';
...
-);
+      >
+        <Icon
+          name="fe-home"
+          size="1rem"
+          css={css`
+            margin-right: 0.5rem;
+          `}
+        />
+        {t('subNav.title')}
```
jerelmiller, line 33:
```suggestion
        {t('subNav.homeLink')}
```

This home link isn't really a title of the sidenav, but rather is just a way for the user to get back to the home page. You can see the use of `nav.title` on [line 37](https://github.com/newrelic/docs-website/pull/536/files#diff-e93720bfaaf6d5f84c5bec74dd0fdfee1b17ed461f41596406f7c83093ddd718R37) which makes this name a bit confusing.

jerelmiller (commented):
Had just a couple minor bits of feedback, otherwise looks great!

i18next-parser.config.js
roadlittledawn: oh gotcha. derp!

src/components/DefaultRelatedContent.js
roadlittledawn: ah gotcha. makes sense to me and makes things slightly cleaner

src/components/SubNavigation.js
roadlittledawn: ah yep. not sure why i reached for the word `title` TBH. good catch

jerelmiller (approved):
Looks great! Thanks for updating this :)

PR #538, opened 2021-01-12
Use localized links

jerelmiller:
## Description

Update the site to use the theme's `Link` component, which has localizes the path.


src/pages/attribute-dictionary.js
```diff
@@ -34,7 +35,7 @@ const AttributeDictionary = ({ data, pageContext }) => {
     [allDataDictionaryEvent]
   );
 
-  const filterEvents = () => {
+  useEffect(() => {
```
caylahamann, line 38: I assume this is a change that doesn't have to do with the localized links?

src/pages/attribute-dictionary.js
```diff
@@ -193,45 +196,45 @@ const EventDefinition = memo(({ event, filteredAttribute }) => {
...
+            position: relative;
+          }
+        `}
+      >
+        <div
+          css={css`
+            position: relative;
           `}
         >
+          <Link
```
caylahamann, line 223: this is adding the embedded links for the h2 elements to the event name

src/pages/attribute-dictionary.js
jerelmiller: eslint was yelling at me telling me that `filterEvents` was missing as a dependency. Because of how hooks work with regards to closures, I merge the `useEffect` and made its function the `filterEvents` function.

You're right though, doesn't have to do with the localized links.

caylahamann (approved):
This all looks good to me! Left two comments that were for my own self-edification / awareness

src/pages/attribute-dictionary.js
caylahamann: Ah okay! Just checking, wanted to make sure I wasn't missing some special property of the localized links :)

src/pages/attribute-dictionary.js
jerelmiller: They were already there, but the design on it was not consistent with other pages:

Here is a [regular doc](https://docs-preview.newrelic.com/docs/agents/ruby-agent/getting-started/introduction-new-relic-ruby):

Screen Shot 2021-01-11 at 5 09 22 PM


And here is the [attribute dictionary](https://docs-preview.newrelic.com/attribute-dictionary/): 

Screen Shot 2021-01-11 at 5 09 31 PM


Notice how the link icon is on the other side, not vertically aligned, and much bigger. I was simply trying to unify the design since I was touching links anyways.

caylahamann: ahhh that makes more sense. I was going to say that I thought we had this before so I was surprised to see it :)

PR #552, opened 2021-01-13
Add github workflow to store slugs to translate

roadlittledawn:
## Description
Adds a workflow that runs on the `pull_request` event for PRs that contain `.mdx` files. It will check for any pages that has `translate` in the frontmatter and saves them (by locale) to a queue saved in AWS DynamoDB.

## Related issues / PRs
* Closes #531

## To Do
* [x] Combine what's in the queue with what's being added
* [x] Update this PR to target `main` before merging (see `TODO` comments in the workflow file)

PR #554, opened 2021-01-13
Change cache path on amplify

jerelmiller:
## Description

The cache path for Gatsby was unfortunately pointing to the wrong location. This PR updates the cache path so that Gatsby files should be cached properly 🤞 . This will hopefully speed up builds.


.github/workflows/validate-pr.yml
```diff
@@ -19,14 +19,34 @@ jobs:
...
 
+      - name: Get yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+
       - name: Cache dependencies
         uses: actions/cache@v2
         with:
-          path: /usr/local/share/.cache/yarn/v4
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
```
zstix, line 29: Sorry, this one was my fault. I was hoping to skip a step by defining the cache path (assuming we always run this on an ubuntu container). Thanks for adding this back in.

Since we've got so much boilerplate, how would y'all feel about abstracting these setup actions (including that great gatsby stuff below) into a repo.

Either way, we will probably want to update the [other workflow PR](https://github.com/newrelic/docs-website/pull/552) to include this (cc: @roadlittledawn)

.github/workflows/validate-pr.yml
```diff
@@ -19,14 +19,34 @@ jobs:
...
           key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
           restore-keys: |
             ${{ runner.os }}-yarn-
 
+      - name: Gatsby cache folder
```
zstix, line 34: [nit] Can we rename this step (and the following) to have "Cache" as the first word? It seems like we're going with verb-first step names.

```suggestion
      - name: Cache Gatsby folder
```

zstix (dismissed):
One super minor request, but this looks great!

.github/workflows/validate-pr.yml
jerelmiller: Haha I originally had something like that, but it read as `Cache Gatsby cache folder`. The double "cache" made me change it to what it is. There really isn't a "gatsby" folder per-se. Are you ok with `Cache Gatsby cache folder`?

.github/workflows/validate-pr.yml
jerelmiller: No problem! Figured its probably best to let yarn tell us where it is rather than assume it.

> Since we've got so much boilerplate, how would y'all feel about abstracting these setup actions (including that great gatsby stuff below) into a repo?

While I agree the boilerplate is a bit annoying, I'm not sure it warrants an entire repo just to abstract these couple steps. Copy/paste works fine for me right now. Only reason I say that is because another repo means all the overhead of going through the open source checklist and putting it in the right org.

.github/workflows/validate-pr.yml
zstix: You can never have too much `cache`

jerelmiller:
Its so strange. Amplify is failing with the following error: 

```
Cannot query field "childrenDataDictionaryAttribute" on type "DataDictionaryEvent"
```

It builds locally. Not sure if this branch is in a weird state or not.

jerelmiller:
I think I might close this PR and open a PR on its own for the github actions cache. Amplify is being dumb. Seems maybe the cache in amplify is weirded out.

PR #581, opened 2021-01-15
Github action to send content to translation vendor

### LINKED ISSUE #532: [Localization Workflow] Create Github action that executes on interval to send the MDX files to vendor via slug information in artifact
## Description 
We will want to create a GitHub action that will execute on some interval (to be determined by Julien how often he wants this sent) which will send the MDX files that have been updated. It will read this information from the slugs that have been saved in the artifact. It will need to grab the content for each slug, send it to our vendor, and then save the UUID for that job to another artifact.

## AC 
* [x] Job reads files from artifact and fetches MDX content 
* [x] Send that information to our vendor 
* [x] If there were any errors with sending information to the vendor, notify us
* [ ] Save job UUID to another artifact

zstix:
## Description
Adds an action that sends content from our `to_translate` queue to the vendor for translation. After the content is sent, we store the job UUID(s) to the `being_translated` queue and clear out the `to_translate` queue.

## Reviewer Notes
I'm making this PR a draft until we can resolve the `TODO` items. We still need to transform the content into a format that the vendor can work with (another upcoming PR should handle that).

## Related Issue(s)
* Closes #532


roadlittledawn:
Just pushed up small change to ensure what we get back from dynamoDB `TranslationQueues` for `being_translated` is iterable to add batch IDs. Left the part of code that removes `to_translate` values until testing is done done.

Also when we were testing this earlier today, one file wouldn't upload. Turned out that file with that fileUri had already been uploaded as a plain text file and you can't upload same fileUri with different content type. I deleted that file and then the upload worked.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -0,0 +1,200 @@
...
+    const { response } = await resp.json();
+    const { code } = response;
+
+    if (code !== 'ACCEPTED') {
+      console.dir(response, { depth: null });
+      throw new Error(code);
+    }
+
+    console.log(`[*] Successfully uploaded ${page.file}.`);
+    return code;
```
jerelmiller, line 89: It doesn't really look like you use this `code`.  Do you need to return it here?

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -0,0 +1,200 @@
...
+  } catch (error) {
+    console.error(`[!] Unable to upload ${page.file}.`);
+    console.error(error);
+    process.exit(1);
```
jerelmiller, line 93: I wonder if we should gather all of the jobs that failed to upload and instead do this at the end of the script. By doing this here, this will halt immediately, which likely would kill any in-progress uploads. It might be best to try and get as many uploaded as we can.

Also by exiting here, it looks like you aren't able to record which of these completed successfully, which means any time we re-run this script, all of the jobs previously successful uploads will happen again. We may want to [record](https://github.com/newrelic/docs-website/pull/581/files#diff-3275c521907ff8a8e4237b2b80241b357a478e451ec141a3c21300fdfcbb7df0R186) which ones succeeded as they finish so that we avoid sending them if something fails in the middle.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -0,0 +1,200 @@
...
+  const fileRequests = batchUids.flatMap((batchUid, idx) => {
+    const [locale, localePages] = Object.entries(content)[idx];
+    return localePages.map(uploadFile(locale, batchUid, accessToken));
+  });
+
+  const fileResponses = await Promise.all(fileRequests);
+
+  console.log(`[*] Successfully uploaded ${fileResponses.length} files`);
+
+  return { jobUids, batchUids };
```
jerelmiller, line 152: It doesn't look like you [do anything](https://github.com/newrelic/docs-website/pull/581/files#diff-3275c521907ff8a8e4237b2b80241b357a478e451ec141a3c21300fdfcbb7df0R184) with `jobUids`. Do you need it?

scripts/actions/utils/vendor-request.js
```diff
@@ -0,0 +1,91 @@
...
+      console.dir(response, { depth: null });
+      throw new Error(code);
+    }
+
+    return data;
+  } catch (error) {
+    console.error(
+      `Unable to make a ${options.method} request to ${url.href}. (${error})`
+    );
+    process.exit(1);
```
jerelmiller, line 19: Rather than exiting, we should let the script calling this function deal with how to handle a failure. This may have adverse effects on the script since a lot of the requests happen concurrently.

scripts/actions/utils/vendor-request.js
```diff
@@ -0,0 +1,91 @@
...
+ * @param {Object} [body] (Optional) Data to send with the requst.
+ * @returns {Promise<Object>} The result after making the request.
+ */
+const vendorRequest = async (
+  method,
+  endpoint,
+  body = {},
+  contentType = 'application/json'
+) => {
+  const accessToken = await getAccessToken();
```
jerelmiller, line 66: How long do access tokens last? Would it be better to get a single one up front thats used for all calls?

scripts/actions/utils/vendor-request.js
```diff
@@ -0,0 +1,91 @@
...
+  };
+
+  if (method !== 'GET') {
+    options.body =
+      contentType === 'application/json' ? JSON.stringify(body) : body;
+  }
+
+  const data = await makeRequest(url, options);
+
+  return data;
```
jerelmiller, line 85:
```suggestion
  return makeRequest(url, options);
```

Rather than setting a variable just to return it, you can instead return the request itself.

jerelmiller (changes requested):

.github/workflows/send-content-to-translate.yml
```diff
@@ -0,0 +1,46 @@
...
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Add extra dependencies
+        run: yarn add form-data
```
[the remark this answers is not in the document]
zstix: Ah, this is probably a vestige of a prior solution where we were _just_ installing the dependencies we needed. I'll bring this in as a `devDependency`.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -0,0 +1,200 @@
...
+const FormData = require('form-data');
+const fetch = require('node-fetch');
+
+const loadFromDB = require('./utils/load-from-db');
+const saveToDB = require('./utils/save-to-db');
+const serializeMDX = require('./serialize-mdx');
+const { vendorRequest, getAccessToken } = require('./utils/vendor-request');
+
+/**
+ * @typedef {Object} Page
```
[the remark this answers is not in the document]
zstix: It was something I was playing around with. I'm happy to rip this stuff out if it's obnoxious.

I _did_ find it helpful to pick up this code after longer absences because I could look at the typing for hints as to what was going on.

I'd be super down for using typescript, but I know that's not a decision to be made lightly.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -0,0 +1,200 @@
...
+    body: form,
+  };
+
+  try {
+    const resp = await fetch(url.href, options);
+    const { response } = await resp.json();
+    const { code } = response;
+
+    if (code !== 'ACCEPTED') {
+      console.dir(response, { depth: null });
```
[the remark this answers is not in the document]
zstix: Woops! I'll pull this out. Good catch!

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -0,0 +1,200 @@
...
+  }, {});
+
+/**
+ * @param {string} locale The locale that this file should be translated to.
+ * @param {string} batchUid The batch that is expecting this file.
+ * @param {string} accessToken
+ * @returns {(page: Page) => Promise<string>} A function that turns a page into a request.
+ */
+const uploadFile = (locale, batchUid, accessToken) => async (page) => {
+  const filepath = path.join(process.cwd(), 'to-translate.html');
```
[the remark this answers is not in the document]
zstix: Great point, I didn't think of that. I'll have this save to a unique file based on the `page.file`path.

I was a bit bummed that we needed to write to the filesystem only to load the file in the lines directly below. I tried sending over a `Buffer` or `Stream` but couldn't make an in-memory solution work.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -0,0 +1,200 @@
...
+    const acc = await accP;
+    return {
+      ...acc,
+      [locale]: await Promise.all(
+        slugs.map(async (slug) => {
+          const mdx = fs.readFileSync(path.join(process.cwd(), slug));
+          const html = await serializeMDX(mdx);
+          return {
+            file: slug,
+            html: html ? html : false,
```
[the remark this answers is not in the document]
zstix: I was probably being overly defensive. I think we can just return the `html` and, if there happens to be no contents, then nothing will get translated.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * Take a list of filepaths (grouped by locale) and fetches the HTML content.
+ * NOTE: async inside of loops is hard.
+ * @param {Object<string, string[]>} locales The queue of slugs to be translated.
+ * @returns {Promise<Object<string, Page[]>>}
+ */
+const getContent = async (locales) =>
+  Object.entries(locales).reduce(async (accP, [locale, slugs]) => {
+    const acc = await accP;
```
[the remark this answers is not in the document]
zstix: I'd love some input on this function. We're using `reduce` with async functions. The accumulator isn't the object, but rather a _promise_. I thought we needed to await it to ensure that we could correctly spread on it (below).

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -0,0 +1,200 @@
...
+    method: 'POST',
+    headers: {
+      Authorization: `Bearer ${accessToken}`,
+    },
+    body: form,
+  };
+
+  try {
+    const resp = await fetch(url.href, options);
+    const { response } = await resp.json();
```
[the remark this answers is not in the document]
zstix: `response` is what comes back from their API. That said, I can rename it to `res` for consistency.

scripts/actions/send-and-update-translation-queue.js
zstix: Not at the moment, but we might need it for future actions. I _think_ we can just rely on the `batchUid` to check the status of the job. I'll remove this value from the return until we need it.

scripts/actions/utils/vendor-request.js
zstix: They last for 5 minutes so we can _probably_ just generate one for the job. The only time this might be an issue is if we're uploading a _ton_ of files. That said, our tests have been pretty snappy. I'll update this to just make the `accessToken` once.

scripts/actions/send-and-update-translation-queue.js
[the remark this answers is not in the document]
jerelmiller: I was half joking haha. That being said, if we find types super helpful, its definitely worth exploring.

scripts/actions/send-and-update-translation-queue.js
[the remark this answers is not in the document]
jerelmiller: > The accumulator isn't the object, but rather a _promise_. 

I see that the object _values_ are promises, but the [accumulator itself looks like an object](The accumulator isn't the object, but rather a promise. ) to me. Am I missing something?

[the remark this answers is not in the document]
jerelmiller: Oh wait, is this because the reduce function is an `async` function that resolves to an object?

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -0,0 +1,199 @@
...
+    method: 'POST',
+    headers: {
+      Authorization: `Bearer ${accessToken}`,
+    },
+    body: form,
+  };
+
+  try {
+    const resp = await fetch(url.href, options);
+    const { response: resp } = await resp.json();
```
jerelmiller, line 78: Does this work? It looks like you've defined `resp` above as a `const`, so I'd expect to see a reassignment error here., by the way, the `res` suggestion was merely for the result returned from `fetch`, since that seems to be the general name for that variable, not from this destructured result. I'm fine leaving this as `response` here.

scripts/actions/utils/vendor-request.js
```diff
@@ -0,0 +1,91 @@
...
+      'Content-Type': contentType,
+    },
+  };
+
+  if (method !== 'GET') {
+    options.body =
+      contentType === 'application/json' ? JSON.stringify(body) : body;
+  }
+
+  return await makeRequest(url, options);
```
jerelmiller, line 85:
```suggestion
  return makeRequest(url, options);
```

No need to use the `await` keyword here since you're returning a promise. Anything that `await`s this `vendorRequest` function will get the resolved value.

scripts/actions/send-and-update-translation-queue.js
zstix: Oh I see, sorry I was moving pretty quickly through those edits. Should have slowed down and thought about this. I'll put it back to `response`.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -0,0 +1,200 @@
...
+const { vendorRequest, getAccessToken } = require('./utils/vendor-request');
+
+/**
+ * @typedef {Object} Page
+ * @property {string} file The filepath for the page (from the project root).
+ * @property {string} html The HTML serialized content for the page.
+ */
+
+// NOTE: the vendor requires the locales in a different format
+// We should consider this into the Gatsby config for each locale.
```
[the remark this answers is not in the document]
roadlittledawn: Also +1 to this one

scripts/actions/send-and-update-translation-queue.js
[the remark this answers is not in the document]
zstix: Yeah...we've ran into this situation in the past. Because the reduce callback function is `async`, the return value is a promise. Which means that the first parameter (the accumulator) is a promise, not the value we want.

There's some serious code smell funk going on with this function, but I'm not sure how to break this up / synchronize it.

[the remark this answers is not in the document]
jerelmiller: Rather than `await`ing the value of the promise, you can do that later on to clean this up, which makes this just a normal reduce function.

My suggestion would be to remove the `async`/`await` here and instead amend [this code](https://github.com/newrelic/docs-website/pull/581/files#diff-3275c521907ff8a8e4237b2b80241b357a478e451ec141a3c21300fdfcbb7df0R120) to something like the following:

```js
const pages = await Promise.all(Object.values(content));
```

scripts/actions/send-and-update-translation-queue.js
zstix: I'm returning the `code` for all uploads and using it for a log:
```
[*] Successfully uploaded 10 / 12 files
```
This should allow for partial success with the file uploads.

scripts/actions/utils/vendor-request.js
zstix: I've refactored the code to only call `process.exit` from the main entrypoint function.

We still log issues, but it's up to the caller to handle and thrown errors. At the moment, we're just catching them at the `main` function, but we can add some additional error response (i.e. communicating the errors via slack).

zstix: Oh...just realized that we're still blowing away the queue, regardless if the file uploaded successfully or not. That's going to require a bit of work to implement. Gonna jump on that now.

zstix: Resolved in https://github.com/newrelic/docs-website/pull/581/commits/f32c2b71774b6e4c30a07e393f38441d04c1b710

.github/workflows/send-content-to-translate.yml
```diff
@@ -0,0 +1,39 @@
+name: Send content to be translated
+
+on:
+  schedule:
```
zstix, line 4: We should add the ability to manually trigger this.

zstix: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

jerelmiller: We also have this setup for the sync action: https://github.com/newrelic/docs-website/blob/develop/.github/workflows/sync-data-dictionary.yml#L4

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -0,0 +1,233 @@
...
+ * @param {Object<string, string[]>} locales The queue of slugs to be translated.
+ * @returns {Promise<Object<string, Page[]>>}
+ */
+const getContent = async (locales) =>
+  Object.entries(locales).reduce(async (accP, [locale, slugs]) => {
+    const acc = await accP;
+    return {
+      ...acc,
+      [locale]: await Promise.all(
+        slugs.map(async (slug) => {
```
zstix, line 37: TODO: make this synchronous.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -0,0 +1,233 @@
...
+    body: form,
+  };
+
+  try {
+    const resp = await fetch(url.href, options);
+    const { response } = await resp.json();
+    const { code } = response;
+
+    if (code !== 'ACCEPTED' || !resp.ok) {
+      throw new Error(code);
```
zstix, line 82: We don't want to throw here, since we care about this file's status (two lines below).

jerelmiller (approved):

PR #593, opened 2021-01-20
Deserialize HTML into MDX

### LINKED ISSUE #592: [Localization Workflow] Deserialize HTML from Smartling back into MDX
## Description

We serialize MDX into HTML that is sent to the Smartling for translation. Once we get the HTML file back with the updated translations, we need a way to turn that HTML back into MDX.

## Acceptance criteria

* [ ] Transforms all content back into MDX

jerelmiller:
Closes #592

## Description

Adds the ability to deserialize an HTML document back into MDX. This will be useful when getting a result from Smartling to turn it back into MDX.


scripts/actions/deserialize-html.js
```diff
@@ -0,0 +1,102 @@
...
+      td: component,
+      th: component,
+      span: component,
+      div: component,
+      h1: headingWithCustomId,
+      h2: headingWithCustomId,
+      h3: headingWithCustomId,
+      h4: headingWithCustomId,
+      h5: headingWithCustomId,
+      h6: headingWithCustomId,
```
caylahamann, line 84: do we usually have headings with custom ids on anything other than an h2 tag?

scripts/actions/utils/serialization-helpers.js
```diff
@@ -6,19 +6,28 @@ const toMDAST = require('remark-parse');
 const remarkMdx = require('remark-mdx');
 const remarkMdxjs = require('remark-mdxjs');
 const unified = require('unified');
+const visit = require('unist-util-visit');
+
+const removeParagraphs = () => (tree) => {
```
caylahamann, line 11: why was the removeParagraphs added in?

caylahamann (approved):
this looks dope! just two questions for my own edification

scripts/actions/deserialize-html.js
jerelmiller: Not sure I can say "usually", but its possible they are there. I'd rather not restrict a content contributor from being able to add a custom ID to an h6, so figured we'd cover all cases.

scripts/actions/utils/serialization-helpers.js
jerelmiller: This is only used in the case of serializing a JSX expression for a prop. You can see the effect of the change here: https://github.com/newrelic/docs-website/pull/593/files#diff-190fa2e1660b11a6496776193cc476619e6b9ae0776db130bc3d7f157e50cd82R203-R205

Notice the `p` tag wrapped around the children there. This results in this:

```js
<Collapser
  title={<><InlineCode>code</InlineCode> API</>}
>
```

getting turned into this:

```js
<Collapser
  title={<>
    <InlineCode>code</InlineCode> AP
  </>}
```

Figured its best to try and keep the formatting as consistent as possible, so by removing the paragraphs, we can get rid of the additional formatting here.

scripts/actions/deserialize-html.js
caylahamann: good point!

scripts/actions/utils/serialization-helpers.js
caylahamann: gotcha! thank you for the explanation

PR #598, opened 2021-01-20
Try a different cache key

jerelmiller:
Changes the cache key and combine paths based on this suggestion: https://dev.to/danielbayerlein/incremental-gatsby-builds-with-github-actions-2p7o

Also enables incremental builds in GitHub actions.

### DIFF
--- a/.github/workflows/validate-pr.yml
+++ b/.github/workflows/validate-pr.yml
@@ -39,7 +39,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: .cache
-          key: ${{ runner.os }}-cache-gatsby-${{ github.sha }}
+          key: ${{ runner.os }}-cache-gatsby-${{ hashFiles('.cache') }}
           restore-keys: |
             ${{ runner.os }}-cache-gatsby-
 
@@ -47,7 +47,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: public
-          key: ${{ runner.os }}-cache-public-${{ github.sha }}
+          key: ${{ runner.os }}-cache-public-${{ hashFiles('public') }}
           restore-keys: |
             ${{ runner.os }}-cache-public-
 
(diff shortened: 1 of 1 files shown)


zstix (approved):
🤞

.github/workflows/validate-pr.yml
```diff
@@ -35,27 +35,24 @@ jobs:
...
             ${{ runner.os }}-yarn-
 
-      - name: Cache Gatsby cache folder
+      - name: Cache Gatsby build
         uses: actions/cache@v2
         with:
-          path: .cache
-          key: ${{ runner.os }}-cache-gatsby-${{ github.sha }}
+          path: |
+            public
```
caylahamann, line 42: I assume this comes out to be something like this: `public/.cache`

.github/workflows/validate-pr.yml
```diff
@@ -35,27 +35,24 @@ jobs:
...
+            ${{ runner.os }}-gatsby-build-
 
       - name: Install dependencies
         run: yarn install --frozen-lockfile
 
       - name: Gatsby build
-        run: yarn build:production
+        run: yarn build:production --log-pages
+        env:
+          GATSBY_EXPERIMENTAL_PAGE_BUILD_ON_DATA_CHANGES: true
```
caylahamann, line 54: were you able to get any measurement on what the difference in build changes was 😄 ? Just out of curiosity

caylahamann (approved):

.github/workflows/validate-pr.yml
jerelmiller: Not quite. By putting these on different lines, it's caching multiple paths. You can check it out on [the documentation](https://github.com/actions/cache#whats-new)

.github/workflows/validate-pr.yml
jerelmiller: Not quite unfortunately 😢 . It deleted the cache on the last push. Figured we'd try it out on `develop` and see how it goes.

PR #603, opened 2021-01-21
Start tracking logs content

roadlittledawn:
## Description

Our `.gitignore` was ignoring all things inside of any `logs` directory. This removes that rule and adds all those untracked files / directories.


jerelmiller (approved):
I'm still laughing about this. Great find!!!

roadlittledawn:
note to self: be careful what i gitignore ;)

PR #606, opened 2021-01-21
chore: add workflow to build and cache the site on pushes to develop branch

jerelmiller:
## Description

I came to the revelation that I think our cache isn't populating because we have no cache built for our develop branch. Currently, we only build and populate the cache when a PR is opened. According to [GitHub's documentation](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#using-the-cache-action) on the cache action:

> A workflow can access and restore a cache created in the current branch, the base branch (including base branches of forked repositories), or the default branch (usually `main`). For example, a cache created on the default branch would be accessible from any pull request. Also, if the branch `feature-b` has the base branch `feature-a`, a workflow triggered on `feature-b` would have access to caches created in the default branch (`main`), `feature-a`, and `feature-b`.

> Access restrictions provide cache isolation and security by creating a logical boundary between different workflows and branches. For example, a cache created for the branch `feature-a` (with the base `main`) would not be accessible to a pull request for the branch `feature-b` (with the base `main`).

This PR attempts to build and cache the site on our default branch (`develop`) so that PRs can restore their cache from this one.

### DIFF
--- a/.github/workflows/build-cache.yml
+++ b/.github/workflows/build-cache.yml
@@ -1,0 +1,84 @@
+name: Build and cache site
+
+on:
+  push:
+    branches:
+      - develop
+
+env:
+  NODE_OPTIONS: '--max-old-space-size=4096'
+  CHOKIDAR_USEPOLLING: 1
+
+jobs:
+  build:
+    name: Gatsby Build
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


zstix (approved):
🤞

PR #625, opened 2021-01-22
chore: added Austin's colon

jpvajda:
This PR adds a colon after Data Source on the attributes dictionary page.

Screen Shot 2021-01-21 at 6 35 04 PM

### DIFF
--- a/src/pages/attribute-dictionary.js
+++ b/src/pages/attribute-dictionary.js
@@ -246,7 +246,7 @@ const EventDefinition = memo(({ location, event, filteredAttribute }) => {
             margin-right: 0.5rem;
           `}
         >
-          Data {(pluralize('source', event.dataSources.length), ':')}
+          Data {pluralize('source', event.dataSources.length)}:
         </span>
         <TagList>
           {event.dataSources.map((dataSource) => (
--- a/src/pages/attribute-dictionary.js
+++ b/src/pages/attribute-dictionary.js
@@ -246,7 +246,7 @@ const EventDefinition = memo(({ location, event, filteredAttribute }) => {
             margin-right: 0.5rem;
           `}
         >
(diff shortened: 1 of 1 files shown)


src/pages/attribute-dictionary.js
```diff
@@ -246,7 +246,7 @@ const EventDefinition = memo(({ location, event, filteredAttribute }) => {
             margin-right: 0.5rem;
           `}
         >
-          Data {pluralize('source', event.dataSources.length)}
+          Data {pluralize('source:', event.dataSources.length)}
```
jerelmiller, line 249: Can you make sure that in the plural case it doesn't render as `source:s`? You may need to add the colon after the closing `}`

jpvajda: oh I didn't realize there was a plural case. I think I fixed it,  but didn't see a way to select multiple sources in the navigation.

src/pages/attribute-dictionary.js
```diff
@@ -246,7 +246,7 @@ const EventDefinition = memo(({ location, event, filteredAttribute }) => {
             margin-right: 0.5rem;
           `}
         >
-          Data {pluralize('source', event.dataSources.length)}
+          Data {(pluralize('source', event.dataSources.length), ':')}
```
jerelmiller, line 249:
```suggestion
          Data {(pluralize('source', event.dataSources.length))}:
```

Does what you have work? I can't say I've tried it with parens like that 😄 . In case it doesn't, the above suggestion should do it.

jpvajda: I did work, but I trust your code more than mine. lol

austin-schaefer:
Thank you @jpvajda . The site was literally unusable.

PR #627, opened 2021-01-22
Check for content, fetch content, deserialize content, and make a branch with content

### LINKED ISSUE #533: [Localization Workflow] Create GitHub action that executes on interval to read for jobs from vendor and check for completion
## Description 
On another interval, we will want to have a GitHub action that checks for the completion of the jobs. 

## AC 
* [ ] Reads job artifact 
* [ ] Checks each job number for completion 
* [ ] Transform content back into MDX 
* [ ] Check in new translated content in correct folder via opening a new PR

caylahamann:
## Summary 
This PR has a github action that will run a node script to check for the status of each batch uid. If the batch is completed, will run the deserialization script to get the fileURIs from the batch, fetch a zip file of the translated files with an array of the fileURIs and then will run the deserialization script on those files. It will then save those files into the correct folder (based on the fileUri) and then in the github actions, commit and add the files, and then lastly open up a PR. 

## TO DO: 
* [x] Get the batch UIDs from the AWS DynamoDB queue 
* [x] Decide on who opens the PR, commit message, ect. 

## Related Issue(s)
 * Closes #533


scripts/actions/check-job-progress.js
```diff
@@ -0,0 +1,35 @@
...
+const main = async () => {
+  //get list of current job batch uids from dyanmoDB, for each one execute checkJobStatus
+  try {
+    const isCompleted = checkJobStatus(batchUid);
+
+    if (isCompleted) {
+      await fetchAndDeserialize(batchUid);
+      process.exit(0);
+    } else {
+      process.exit(1);
```
caylahamann, line 26: If these are not completed, I want to stop the job and I'm not sure what the best way to do this is.

package.json
```diff
@@ -53,9 +55,12 @@
     "remark-html": "^13.0.1",
     "remark-rehype": "^8.0.0",
     "rss": "^1.2.2",
+    "stream": "^0.0.2",
```
caylahamann, line 58: gotta remove some of these that I added in for testing out different libraries.

scripts/actions/check-job-progress.js
zstix: I see `process.exit(1);` as a way to say "something went wrong with this script. Perhaps we simply want to log out that the job is still being processed and exit with a `0` code (success).

I also worry that failed actions might blow up our email.

package.json
zstix: We might want to move these action-related dependencies to `devDependencies` so that our bundle doesn't get bloated with dev-related stuff.

scripts/actions/fetch-and-deserialize.js
```diff
@@ -0,0 +1,85 @@
...
+const fetchTranslatedFilesZip = async (fileUris, locale) => {
+  const fileUriStr = fileUris.reduce((str, uri) => {
+    return str.concat(`&fileUris[]=${uri}`);
+  }, '');
+
+  const localeIdStr = `localeIds[]=${locale}`;
+
+  return vendorGetRequest(
+    `/files-api/v2/projects/${projectId}/files/zip?${localeIdStr}${fileUriStr}`
+  );
```
zstix, line 35: I think we need the HTTP method as the first parameter.

```suggestion
  return vendorGetRequest(
    'GET'
    `/files-api/v2/projects/${projectId}/files/zip?${localeIdStr}${fileUriStr}`
  );
```

scripts/actions/fetch-and-deserialize.js
```diff
@@ -0,0 +1,85 @@
...
+    const zipEntries = zip.getEntries();
+
+    const translatedHtml = zipEntries.map((entry) => {
+      return {
+        path: entry.entryName.split('/').pop().split('.').slice(0, -1), //get the last element from the folder
+        html: zip.readAsText(entry, (encoding = 'utf8')),
+      };
+    });
+
+    const deserializedMdx = Promise.all(
```
zstix, line 56: If you `await` this here, you wont need to on line 65 below.
```suggestion
    const deserializedMdx = await Promise.all(
```

scripts/actions/fetch-and-deserialize.js
```diff
@@ -0,0 +1,85 @@
...
+      translatedHtml.map(async ({ path, html }) => {
```
zstix, line 57: This might be a good place to add a log so we can see what the action did, maybe something like:
```js
console.log(`[*] Deserializing ${path}`);
```

scripts/actions/fetch-and-deserialize.js
```diff
@@ -0,0 +1,85 @@
...
+          extname: '.mdx',
+        }),
+      'utf-8'
+    );
+
+    createDirectories(files);
+
+    files.forEach((file) => write(file, 'utf-8'));
+  } catch (error) {
+    console.error(error);
```
zstix, line 79: In addition to this error, we'll probably want to `system.exit(1);` so that we know that the job failed.

zstix (commented):
I know it's still a work-in-progress, but I left a few comments.

package.json
caylahamann: good idea! I'll move those other to `devDependencies` 😄

scripts/actions/fetch-and-deserialize.js
caylahamann: So I needed to create a separate `vendorRequest` function (I should probably rename it something more specific) because I didn't want the function to return `response.json()`

scripts/actions/__tests__/kitchen-sink.mdx
```diff
@@ -173,5 +180,6 @@ def handler(event, context):
         </tr>
       </tbody>
     </Table>
+
```
jerelmiller, line 183: I think the tests are failing due to the changes in this file. We should revert them.

roadlittledawn:
latest commit gets deserialization working. but it now errors at the `Commit changes` part of the workflow. here's the error: 

```
Run Commit changes
| /github/workflow/commit-changes: line 1: git: command not found
```

caylahamann:
> latest commit gets deserialization working. but it now errors at the `Commit changes` part of the workflow. here's the error:
> 
> ```
> Run Commit changes
> | /github/workflow/commit-changes: line 1: git: command not found
> ```

Is this from using `act`?

package.json
```diff
@@ -58,6 +58,11 @@
     "unist-util-remove": "^2.0.1"
   },
   "devDependencies": {
+    "file-type": "^16.2.0",
+    "adm-zip": "^0.5.1",
+    "stream": "^0.0.2",
+    "stream-buffers": "^3.0.2",
+    "unzipper": "^0.10.11",
```
caylahamann, line 65: `stream-buffers`, `unzipper` can be removed

caylahamann: actually all of these except `adm-zip` can be removed :)

roadlittledawn:
@caylahamann . re: that error. that was my first thought that it had something to do with testing via `act`. didn't get a chance to look into it before leaving for the night. looks like we can't yet trigger the workflow manually now that we added `workflow_dispatch` (maybe because the PR is in draft now?)

caylahamann:
> @caylahamann . re: that error. that was my first thought that it had something to do with testing via `act`. didn't get a chance to look into it before leaving for the night. looks like we can't yet trigger the workflow manually now that we added `workflow_dispatch` (maybe because the PR is in draft now?)

If you got that error using act, act actually doesn't have git installed in its container. You have to source another container in order to get it to work. The command that you can run in order to get a container with more utilities that mimic the github workflow is `act -P ubuntu-18.04=nektos/act-environments-ubuntu:18.04`

caylahamann:
You have to pull that image in with `docker pull nektos/act-environments-ubuntu:18.04` before you run that. I was running into this issue last Friday night and it was driving me insane. 😆

PR #629, opened 2021-01-23
Speed up installation of node modules in GH actions

jerelmiller:
## Description

Rather than caching the yarn cache directory, this PR will now cache the node modules directory and only install depdencies if there is a cache miss. This should speed up the relevant GitHub actions that rely on node modules.

Here is a great article that talks about it: https://dev.to/mpocock1/how-to-cache-nodemodules-in-github-actions-with-yarn-24eh


zstix (approved):
🤞

PR #648, opened 2021-01-26
added page title and path

jpvajda:
this PR attempts to fix the issue with the `https://docs-preview.newrelic.com/attribute-dictionary`  not loading when a user clicks on Attributes Dictionary from the left nav by added a trailing slash.

Also the PR attempts to add a single navigation element for Events.

Both of these things aren't working at the moment.

### DIFF
--- a/src/nav/attribute-dictionary.yml
+++ b/src/nav/attribute-dictionary.yml
@@ -1,5 +1,5 @@
 title: Attribute Dictionary
-path: /attribute-dictionary/
+path: /attribute-dictionary
 pages:
   - title: Events
-    path: /attribute-dictionary/
+    path: /attribute-dictionary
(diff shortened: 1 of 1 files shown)


jpvajda:
@jerelmiller suggestions welcome. :)

src/nav/attribute-dictionary.yml
```diff
@@ -1,2 +1,5 @@
 title: Attribute Dictionary
-path: /attribute-dictionary
+path: /attribute-dictionary/
+pages:
+  - title: Events
+    path: /attribute-dictionary/
```
jerelmiller, line 5:
```suggestion
path: /attribute-dictionary
pages:
  - title: Events
    path: /attribute-dictionary
```

All of our other paths are not using a trailing slash. While the browser will show it in the URL, we should be consistent. Our nav elements should be matching regardless, so if it isn't, we have a bug.

jerelmiller (approved):

PR #674, opened 2021-01-28
Linting checks for PRs - eslint and prettier

caylahamann:
Adds linting checks for eslint and prettier


.github/workflows/validate-pr.yml
```diff
@@ -75,3 +75,55 @@ jobs:
...
+          with:
+            path: '**/node_modules'
+            key: ${{ runner.os }}-node-modules-${{ hashFiles('**/yarn.lock') }}
+
+        - name: Install dependencies
+          if: steps.yarn-cache.outputs.cache-hit != 'true'
+          run: yarn install --frozen-lockfile
+
+        - name: Run Prettier
+          run: prettier --check
```
jerelmiller, line 129: Do we have eslint setup with prettier rules? I _think_ the eslint step above may already handle this. Would be good to double check. Thoughts?

.github/workflows/validate-pr.yml
```diff
@@ -75,3 +75,55 @@ jobs:
 
       - name: Jest test
         run: yarn test --passWithNoTests
+
+    lint:
```
jerelmiller, line 79: I _think_ this may need to be dedented one level. If I'm looking at this right, this looks like it falls under the `test` field, which we don't want.

.github/workflows/validate-pr.yml
caylahamann: Let me double check!

jerelmiller (approved):

PR #680, opened 2021-01-29
Fix double click issue for table of contents in right rail

### LINKED ISSUE #443: Doc page not scrolling to correctly location until second click in right rail
### Description

After the first section click on the sidebar, the page no longer scrolls when you click another section unless you click twice. The URL does update.

### Steps to reproduce

1. Go to https://docs-preview.newrelic.com/docs/agents/net-agent/custom-instrumentation/add-detail-transactions-xml-net
2. Click on `Example Myinstrumentation.xml`
3. Click `Name transactions`. Notice that the doc doesn't scroll but the URL updates
4. Click `Name transactions` again. The doc scrolls. 
5. Try it with other headers

### Expected behavior

Doc scrolls to appropriate header with each click.

### Environment

* Chrome
* Edge 87.0.664.66

jerelmiller:
Closes #443

# Description

Fixes the issue of the table of contents links not scrolling the page to the proper section until the user clicked twice.

### DIFF
--- a/src/components/TableOfContents.js
+++ b/src/components/TableOfContents.js
@@ -112,7 +112,7 @@ const TableOfContents = ({ page }) => {
                   `}
                 >
                   <Icon
-                    name={Icon.TYPE.ARROW_LEFT}
+                    name="fe-arrow-left"
                     css={css`
                       display: ${isActive && !isMobileScreen
                         ? 'inline-block'
--- a/src/components/TableOfContents.js
+++ b/src/components/TableOfContents.js
@@ -30,46 +30,14 @@ const TableOfContents = ({ page }) => {
       });
   }, [mdxAST]);
 
-  const raf = useRef();
-  const navRef = useRef();
-  const activeRef = useRef();
   const headingIds = useMemo(() => headings.map(prop('id')), [headings]);
   const activeHash = useActiveHash(headingIds);
-  const previousActiveHash = usePrevious(activeHash);
-  const changedActiveHash = activeHash !== previousActiveHash;
-
   const isMobileScreen = useMedia('(max-width: 1240px)');
-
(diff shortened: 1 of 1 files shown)


src/components/TableOfContents.js
```diff
@@ -30,46 +30,14 @@ const TableOfContents = ({ page }) => {
...
-  const changedActiveHash = activeHash !== previousActiveHash;
-
   const isMobileScreen = useMedia('(max-width: 1240px)');
 
-  useEffect(() => {
-    if (!activeRef.current) {
-      return;
-    }
-
-    const navRect = navRef.current.getBoundingClientRect();
```
caylahamann, line 48: How is it scrolling now? Where does that behavior come from?

caylahamann (commented):
Just confused on how removing this solved the problem :) but maybe there was another PR that I missed

src/components/TableOfContents.js
jerelmiller: So this scroll behavior you see deleted was actually controlling the scroll of the table of contents. This was more of a nice feature than not. Basically this just analyzed if the current "active hash" changed, and whether or not it was visible in the right rail. If it wasn't visible, it would scroll the right rail to show that link.

Unfortunately it seems like this scroll behavior caused it to capture the event awkwardly or something, which prevented the page itself from scrolling. Not entirely sure why, but removing this functionality seemed to fix it. 

Since this was more of a nice feature than a must have, I went ahead and removed it. Its a tricky problem to solve correctly. For now, this just means the user would have to scroll the right rail if they scroll to a header that is not visible.

caylahamann: Fair enough! thank you for the explanation

caylahamann (approved):

PR #683, opened 2021-01-29
Fixes the linting errors

caylahamann:
imagine: a lint error free docs site


jerelmiller (approved):

caylahamann:
nevermind, there appears to be more that I missed! will update with those as well

PR #686, opened 2021-01-29
Fix parsing issue with indented code blocks

### LINKED ISSUE #347: Migration scripts don't handle fenced code blocks within MDX well
### Description

The NRQL syntax doc has some complicated syntax within collapsers, and the migration script is getting confused and pulling in all the content as code.

### Steps to reproduce

1. Go to https://docs-preview.newrelic.com/docs/query-your-data/nrql-new-relic-query-language/get-started/nrql-syntax-clauses-functions
2. Compare with https://docs.newrelic.com/docs/query-your-data/nrql-new-relic-query-language/get-started/nrql-syntax-clauses-functions

### Expected behavior

New doc should reproduce formatting/hierarchy of old doc.

### Considerations

- This functionality is now available in MDX  https://github.com/mdx-js/mdx/pull/1367

### Useful links

* [MDXv2 Umbrella Issue](https://github.com/mdx-js/mdx/issues/1041)

jerelmiller:
Closes #347

# Description

Fixes the issue where indented fenced code blocks, generally inside collapsers,
were capturing much more text content than what was inside the code block. This
PR attempts to fix that by slicing the text up to the fence, then re-parsing the
rest of the text through MDX and adding that to the AST.

## Screenshot(s)

**Before**

Screen Shot 2021-01-29 at 1 18 41 AM
Screen Shot 2021-01-29 at 1 18 54 AM

**After**
Screen Shot 2021-01-29 at 1 19 00 AM
Screen Shot 2021-01-29 at 1 19 25 AM 1


plugins/gatsby-remark-mdx-v2-fenced-code-blocks/index.js
```diff
@@ -0,0 +1,25 @@
+const { createCompiler } = require('@mdx-js/mdx');
+const visit = require('unist-util-visit');
+
+const INVALID_CODE_BLOCK = /`{3,}\n/;
```
zstix, line 4: It seems like we're relying a lot on regex for all these transform operations we've been doing (which is great, this is not a complaint).

Perhaps we should consider putting all these regex statements in one location (or maybe even wrap them in a helper class with some common functionality). Either way, that's out of scope for this PR.

zstix (approved):
LGTM 👍

plugins/gatsby-remark-mdx-v2-fenced-code-blocks/index.js
jerelmiller: > putting all these regex statements in one location 

This is definitely an idea, though I'd rather put them in a spot due to _function_, not just because they are regexes. These are just a tool like anything else (function, variable, etc.) so I don't necessarily think we should treat them as unique. For example, we don't have a file just called "functions". That being said, if there are regexes that are copied across several locations, there is definitely an opportunity to consolidate those.

As for the helper function, it didn't seem like wrapping this in a function provided much value, hence why I tried to make the variable name descriptive. This didn't seem to really give me much:

```
const isInvalidCodeBlock = (string) => INVALID_CODE_BLOCK.test(string)
```

I'd rather just use `INVALID_CODE_BLOCK.test(string)` directly.

PR #697, opened 2021-01-29
Saves the localized file to correct filepath

caylahamann:
## Summary 

Before the file path was saved to `src/i18n/content/jp/docs/some-file.mdx` when we want to save it to `src/i18n/content/jp/docs/full/filepath/some-file.mdx`. This solves this!

### DIFF
--- a/scripts/actions/fetch-and-deserialize.js
+++ b/scripts/actions/fetch-and-deserialize.js
@@ -44,10 +44,10 @@ const fetchAndDeserialize = (accessToken) => async ({ locale, fileUris }) => {
     return {
       path: entry.entryName
         .split('/')
-        .slice(1)
+        .slice(2) //get rid of locale and src
         .join('/')
         .split('.')
-        .slice(0, -1),
+        .slice(0, -1), //get rid of .mdx
       html: zip.readAsText(entry, 'utf8'),
     };
   });
--- a/scripts/actions/fetch-and-deserialize.js
+++ b/scripts/actions/fetch-and-deserialize.js
@@ -54,7 +54,6 @@ const fetchAndDeserialize = (accessToken) => async ({ locale, fileUris }) => {
 
   const deserializedMdx = Promise.all(
     translatedHtml.map(async ({ path, html }) => {
-      console.log(`[*] Deserializing ${path}`);
       return {
         path: `src/i18n/content/${localesMap[locale]}/docs/${path}`,
         mdx: await deserializedHtml(html),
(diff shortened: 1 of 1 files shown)


scripts/actions/fetch-and-deserialize.js
```diff
@@ -40,15 +40,20 @@ const fetchAndDeserialize = (accessToken) => async ({ locale, fileUris }) => {
   const zipEntries = zip.getEntries();
 
   const translatedHtml = zipEntries.map((entry) => {
+    console.log(entry.entryName);
```
zstix, line 43: We can probably remove this `console.log`.
```suggestion
```

zstix (commented):
One minor suggestion, but this looks great!

scripts/actions/fetch-and-deserialize.js
```diff
@@ -40,15 +40,20 @@ const fetchAndDeserialize = (accessToken) => async ({ locale, fileUris }) => {
...
     return {
-      path: entry.entryName.split('/').pop().split('.').slice(0, -1), // get the last element from the folder
+      path: entry.entryName
```
jerelmiller, line 45: Its a bit hard to follow what this is doing. This also feels like it could be a touch brittle. I would suggest a couple changes.

1) Since you're trying to get rid of `src/jp` as part of the name, I would suggest just using a `replace`. Its a bit clearer the intention and harder to break:

```js
entry.entryName.replace(`src/${locale}`, '')
```

2) Use `path.basename` + `path.extname` to get rid of the extension. This makes it a bit clearer and ensures edge cases are handled:

```js
path.basename(entry.entryName, path.extname(path.entryName)) // => index.mdx -> index
```

Putting that altogether makes it easier to understand the purpose:

```js
const path = require('path')

const filepath = entry.entryName.replace(`src/${locale}`)

return {
  path: path.basename(filepath, path.extname(filepath))
}
```

scripts/actions/fetch-and-deserialize.js
jerelmiller, line 43: Was this meant to be logged? If so, should this include our little `[*]` denotation?

scripts/actions/fetch-and-deserialize.js
caylahamann: yes! I like this suggestion :)

scripts/actions/fetch-and-deserialize.js
caylahamann: No it was just for testing purposes! It's removed now :)

jerelmiller (approved):
Shibby!

PR #703, opened 2021-01-29
Remove batchIds as last step in workflow

### LINKED ISSUE #701: Review batchUid saving / clearing

roadlittledawn:
Closes #701 

## Description

Adds one last step to workflow to run a script where we run the code to remove batchUids from the translation queue after we've fetched, deserialized, and successfully PR'ed the repo.


jerelmiller (approved):
Nicceeee! Thanks so much for doing this! I know it will save us some future headaches

roadlittledawn:
@jerelmiller far as i can tell, this should be working, but the workflow step can't seem to find the `remove-completed-batch.js` file when the command is run. https://github.com/newrelic/docs-website/runs/1794336937?check_suite_focus=true#step:9:18

caylahamann:
@roadlittledawn I think it might be because that file isn't available in `develop` yet

jerelmiller:
Did you run that job against this branch as the base?

roadlittledawn:
> Did you run that job against this branch as the base?

yep, when i ran it and chose this branch in the dropdown. but now that i look at the wording, looks like it may just use the workflow file and not the code from that branch (to @caylahamann 's point), even though it looks like it's checking out the repo from that branch in earlier steps (but maybe i'm also misinterpreting what it's doing there).

caylahamann:
@roadlittledawn I figured out why this was happening! I put a test PR for you to check it out

https://github.com/newrelic/docs-website/pull/706

So when we do the create-pull-request action, we checkout develop, so we have to recheck out the base branch and install the dependencies for it to get it work

caylahamann:
@roadlittledawn this is what needs to get added into this PR: 
https://github.com/newrelic/docs-website/blob/bb1baef431c4d8d735cd14779c373dcd3c2e117f/.github/workflows/check-translations-and-deserialize.yml#L72-L86

roadlittledawn:
@caylahamann ooooh, interesting. i see i see. thanks a heap!

roadlittledawn:
That did the trick! Will merge once checks pass

PR #706, opened 2021-01-29
Test remove batches

caylahamann:
Testing a theory that the base branch switches on the `create-pull-request` action, meaning that we will need to checkout the base again.


caylahamann:
this was just a demo

Issue #116, opened 2020-10-05
Support <var>, <mark>, and <a> tags inside code blocks


jerelmiller:
## Summary

Some of the pages in the docs site include code blocks with embedded `<var>`, `<mark>` `<a>` and `<pre>` tags. Unfortunately our `CodeBlock` component, nor the fenced code blocks support the additional HTML tags, since these render as actual text inside the code block. We need a way to allow these kinds of nested elements so we can style them appropriately. 

## Useful links

* [Code block with a var, and a tag](https://docs.newrelic.com/docs/apis/nerdgraph/examples/nerdgraph-nrql-tutorial)
* [MDXv2 Umbrella Issue](https://github.com/mdx-js/mdx/issues/1041)

paperclypse:
Are we going to support <pre> tags as well?

paperclypse:
Does the new site currently support <var>, <mark>, and <pre> tags or not? I realize on rereading that this issue might be addressing a slightly different thing (whether you can use these tags in codeblocks or not). Thanks!

austin-schaefer:
Example of a doc where the lack of formatting is very visible. Compare https://docs-preview.newrelic.com/docs/agents/java-agent/api-guides/java-agent-api-instrument-using-annotation with https://docs.newrelic.com/docs/agents/java-agent/api-guides/java-agent-api-instrument-using-annotation .

![image](https://user-images.githubusercontent.com/55203603/101090030-e33a8200-356a-11eb-98db-43334af2e0ae.png)

![image](https://user-images.githubusercontent.com/55203603/101090038-e6ce0900-356a-11eb-95b3-49a316d3fe55.png)

PR #712, opened 2021-01-30
Keep var/mark tags in documents

### LINKED ISSUE #670: Old What's New URLs not mapped on new site
### Description

The old WN posts use a different URL structure than the new site. Compare:

* https://docs.newrelic.com/whats-new/new-relic-lambda-extension
* https://docs-preview.newrelic.com/whats-new/2020/11/new-relic-lambda-extension

The old URLs, however, are broken when you map them to the new site:

* https://docs-preview.newrelic.com/whats-new/new-relic-lambda-extension

### Expected behavior

Create redirects for existing WN URLs

jerelmiller:
Closes #670 
Partially addresses #116

# Description

Migrates all `<var>` and `<mark>` tags into the docs. Note this does **NOT** migrated `<var>` and `<mark>` tags inside of code blocks. This will be done inside a followup PR. 

This PR also fixes issues with pages that redirect to themself. During our migraiton, we were pulling in redirects and attaching them to a matching page. This confused Gatsby because we had both a page and redirect created for the same location.

## Screenshots

**var**

Screen Shot 2021-01-29 at 5 57 55 PM

**mark**

Screen Shot 2021-01-29 at 5 57 59 PM


scripts/utils/migrate/html-to-markdown.js
```diff
@@ -193,6 +193,14 @@ module.exports = (file) => {
       filter: (node) => node.nodeName === 'I' && node.classList.contains('fa'),
       replacement: (content, node) =>
         htmlToJSXConverter.convert(node.outerHTML),
+    })
+    .addRule('var/mark', {
+      filter: ['var', 'mark'],
+      replacement: (content, node) => {
+        const tag = node.nodeName.toLowerCase();
+
+        return `<${tag}>${content}</${tag}>`;
```
caylahamann, line 202: how is this different than just returning the node ( `<var>` or `<mark>` ))?

gatsby-node.js
```diff
@@ -90,6 +90,16 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
         }
       }
 
+      whatsNewPosts: allMarkdownRemark(
+        filter: { fileAbsolutePath: { regex: "/src/content/whats-new/" } }
```
caylahamann, line 94: I'm also gathering this has some redirect work! would you be able to explain what is going on here? or maybe a separate PR for the redirect stuff?

caylahamann (commented):
Just two comments! overall it looks good to me (for the var/mark work, I'm not sure about the redirect stuff because i'm not sure what problem is being solved)

gatsby-node.js
caylahamann: Nevermind! I just read the full PR! Sorry about that

caylahamann (approved):

scripts/utils/migrate/html-to-markdown.js
jerelmiller: In this case, I'm filtering this rule by both `var` and `mark`. The `replacement` function doesn't know which is which when it is called, so this just makes sure that it takes the `nodeName` and uses that as the tag. The alternative was this (which didn't feel any better)

```js
return node.nodeName.toLowerCase() === 'var' ? `<var>${content}</var>` : `<mark>${content}</mark>`
```

gatsby-node.js
jerelmiller: Ah right, I kinda snuck in a separate ticket in this PR. This fixes #670 to make sure the old whats new URLs map to the new ones. I did this work in this PR because this PR [was failing](https://github.com/newrelic/docs-website/actions/runs/522497102) due to some weird cache issues from deleting all of the data dictionary stuff. By changing `gatsby-node`, this force-cleared the cache and allowed this PR to build successfully. Sorry about the confusion here!

PR #713, opened 2021-01-30
Sends visual context to vendor in send content to be translated action, deletes on deserialization

caylahamann:
this has been quite the adventure with apis and form-data 

## Summary 
Sends the visual context and runs text matching against our uploaded files! This allows translators to see our beautiful docs site. This PR 
* fetches the html from `docs-preview.newrelic.com` for the specific `fileUri` after the file has been successfully uploaded (note: this will have to be changed to docs.newrelic.com once we GA, once merged I will create a follow up ticket) 
* uploads html to smartling and runs async matching


caylahamann:
There will be a follow up PR to delete the context after the file is done being translated (in the fetch and deserialize action) ensuring translators always have the latest visual context

scripts/actions/remove-completed-batch.js
```diff
@@ -17,4 +24,63 @@ const saveRemainingBatches = async () => {
...
+/**
+ * Gets the fileUris from the deserializedBatchUids, takes the fileUris and fetches
+ * the contextUids, and then the contextUids are deleted for deserialized batches
+ * @param {Array} deserializedBatchUids
+ * @returns {Promise}
+ */
+const removePageContext = async (deserializedBatchUids) => {
+  const accessToken = await getAccessToken();
+
+  const fileUris = await deserializedBatchUids.map(async (batchUid) => {
```
zstix, line 36: I _think_ we might want to `await Promise.all(deserializedBatchUids.map(async ...` here since the `map` will return an array of promises

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -77,13 +77,60 @@ const uploadFile = (locale, batchUid, accessToken) => async (page) => {
...
+    await sendPageContext(page.file, accessToken);
   } else {
     console.error(`[!] Unable to upload ${page.file}.`);
   }
 
   return { code, locale, slug: page.file };
 };
 
+/**
+ * sends the html file as a visual context for each uploaded file
```
zstix, line 89:
```suggestion
 * Sends the html file as a visual context for each uploaded file
```

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -77,13 +77,60 @@ const uploadFile = (locale, batchUid, accessToken) => async (page) => {
...
+ * @param {string} fileUri
+ * @param {string} accessToken
+ * @returns {Promise}
+ */
+const sendPageContext = async (fileUri, accessToken) => {
+  const filepath = fileUri.replace(`src/content/`, '');
+  const slug = filepath.replace(`.mdx`, '');
+  const contextUrl = new URL(slug, 'https://docs-preview.newrelic.com'); //need to change this once we migrate to docs-newrelic-com
```
zstix, line 97: Perhaps w should move this into a constant at the opt of the file like we did with `LOCALE_IDS` and `PROJECT_ID`

zstix (commented):
This looks great! I had a couple of small comments.

scripts/actions/send-and-update-translation-queue.js
caylahamann: I like that idea! that's what I was also thinking, because then we just have to change the var instead of finding where it is in this file :)

scripts/actions/remove-completed-batch.js
caylahamann: yes very true! thank you

jerelmiller:
@caylahamann I just merged a PR that should fix the build for you. If you merge `develop` once more, the build should succeed.

scripts/actions/__tests__/kitchen-sink.mdx
```diff
@@ -32,7 +32,7 @@ For an index of documentation available during the private Beta, see the [C SDK
 To instrument errors in transactions:
 
 1. [Start a transaction](/docs/instrument-transactions-c-agent#start-transaction).
-2. Record an error with [`newrelic_notice_error()`](https://source.datanerd.us/c-agent/c-agent/blob/master/examples/ex_notice_error.c), supplying the required parameters.
+2. Record an with [`newrelic_notice_error()`](https://source.datanerd.us/c-agent/c-agent/blob/master/examples/ex_notice_error.c), supplying the required parameters.
```
caylahamann, line 35: this was removed during a test on `mark` and `var`! if it's really important I can add it back in but the snapshot has already been updated :)

scripts/actions/remove-completed-batch.js
```diff
@@ -17,4 +29,65 @@ const saveRemainingBatches = async () => {
   );
 };
 
+/**
+ * Gets the fileUris from the deserializedBatchUids, takes the fileUris and fetches
+ * the contextUids, and then the contextUids are deleted for deserialized batches
+ * @param {Array} deserializedBatchUids
```
zstix, line 35: It looks like the function signature has changed (it now takes `fileUri`s.

scripts/actions/remove-completed-batch.js
```diff
@@ -17,4 +29,65 @@ const saveRemainingBatches = async () => {
...
+    .map((item) => {
+      return {
+        fileUri: item.name,
+        contextUid: item.contextUid,
+      };
+    });
+
+  let returnCode = 'SUCCESS';
+
+  for (const { contextUid, fileUri } of contextUids) {
```
zstix, line 65: As much as I love `for...of`, we might want to use `reduce` here snce we're looping over something (`contextUids`) and hoping to have one value at the end (`returnCode`).

zstix (approved):
Couple of small comments, but this looks good to me!

scripts/actions/remove-completed-batch.js
caylahamann: yeah I agree! I at first was not returning anything (which is why I was doing this) but I was running into some promise issues and returning a New Promise was causing a bunch of other errors.

PR #722, opened 2021-02-02
Taxonomy Redirects

### LINKED ISSUE #364: Handle redirects from the old Docs site to the new Doc Site

zstix:
## Description
Sets up redirects for _Taxonomy_ (index pages in Gatsby).

## To Do
* [x] Filter out taxonomy IDs that no longer exist in Drupal (because it doesn't prune this stuff)
* [x] Ensure that the redirects are being created at build time

## Related Issue(s)
* Closes #364

### DIFF
--- a/plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
+++ b/plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
@@ -3,10 +3,11 @@ const fromList = require('./utils/unist-fs-util-from-list');
 const visit = require('unist-util-visit');
 const generateHTML = require('./utils/generate-html');
 const { sentenceCase } = require('./utils/string');
+const taxonomyRedirects = require('../../src/data/taxonomy-redirects.json');
 
 exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
   const { skippedDirectories } = pluginOptions;
-  const { createPage } = actions;
+  const { createPage, createRedirect } = actions;
 
   const { data, errors } = await graphql(`
     query MyQuery {
@@ -129,6 +130,20 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
         fileRelativePath: null,
       },
     });
+
+    console.log(`Setup taxonomy redirects for ${slug}`);
+    Object.entries(taxonomyRedirects)
+      .filter(([to]) => to === slug)
+      .map(([_to, from]) => from)
+      .forEach((from) => {
+        console.log(`Redirect from ${from}`);
+        createRedirect({
+          fromPath: from,
+          toPath: slug,
+          isPermanent: true,
+          redirectInBrowser: true,
+        });
+      });
 
     locales.forEach(({ localizedPath }) => {
       const localizedSlug = path.join(`/${localizedPath}`, slug);
--- a/plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
+++ b/plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
@@ -134,10 +134,10 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
     const redirectsFrom = taxonomyRedirects.find(({ url }) => url === slug);
 
     if (redirectsFrom) {
-      console.log(`Setup taxonomy redirects for ${slug}`);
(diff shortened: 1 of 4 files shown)


plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
```diff
@@ -130,6 +131,20 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
       },
     });
 
+    console.log(`Setup taxonomy redirects for ${slug}`);
+    Object.entries(taxonomyRedirects)
+      .filter(([to]) => to === slug)
+      .map(([_to, from]) => from)
+      .forEach((from) => {
+        console.log(`Redirect from ${from}`);
```
zstix, line 139:
```suggestion
```

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
```diff
@@ -130,6 +131,20 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
       },
     });
 
+    console.log(`Setup taxonomy redirects for ${slug}`);
```
zstix, line 134:
```suggestion
```

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
```diff
@@ -130,6 +131,22 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
       },
     });
 
+    const redirectsFrom = taxonomyRedirects.find(({ url }) => url === slug);
+
+    if (redirectsFrom) {
+      console.log(`Setup taxonomy redirects for ${slug}`);
+
+      redirectsFrom.paths.forEach((from) => {
+        console.log(`\tRedirect from ${from}`);
```
jerelmiller, line 140: This will likely get pretty noisy, pretty fast. If we value the logging, could we instead use [`reporter.verbose`](https://www.gatsbyjs.com/docs/reference/config-files/node-api-helpers/#reporter) so that these only show up if you pass a `--verbose` flag when running `gatsby develop` or `gatsby build`?

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
```diff
@@ -130,6 +131,22 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
       },
     });
 
+    const redirectsFrom = taxonomyRedirects.find(({ url }) => url === slug);
+
+    if (redirectsFrom) {
+      console.log(`Setup taxonomy redirects for ${slug}`);
```
jerelmiller, line 137: Same here, could we instead use `reporter.verbose`?

jerelmiller (commented):
I only had 2 minor comments. This is really nice work!

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
zstix: Oh good call. I had this in for debugging and was on the fence about taking it out.

I might just strip it out, since this will almost always be the same results every time.

zstix: Actually, after looking at `reporter.verbose` I changed my mind. That looks pretty handy.

jerelmiller (approved):

Issue #587, opened 2021-01-19
Review nav instruction set


jerelmiller:
## Description

Now that we have all content migrated (#577), we need to review the instruction set for our nav and ensure it matches what we expect. We still show quite a few warnings during a migration about nodes that are not found. Let's go over the instruction set with a fine tooth comb and update/remove any instructions to get rid of those warnings.

### Specific Nav issues 

- Remove licenses  from left nav
- Create integration should be under integrations
- NR1 > use NR1 > Get started (should be first, a la Anna)
- OpenTelemetry first in docs/integrations

jpvajda:
@austin-schaefer if you have other navigation corrections you can add to this issue.

PR #729, opened 2021-02-02
Sort nav subdirs on migration

### LINKED ISSUE #622: Order the sorting of navigation programmatically so it's closer to the old docs site
Throughout the Docs site, the navigation sort order differs from the old docs site, is there a way to programatically adjust the order to align closer to the old docs site? vs having this corrected manually?

## Scope of remaining work
On migration (in `migrate.js`) we apply two `.sort()` to order docs based on the `order` field AND to try and order subdirs Get Started and Troubleshooting in their parent directory. More notes on some caveats in PR linked below.

The remaining work is to ensure docs are ordered correctly AND to order subdirs correctly (either via instruction sets or if needed get the order for subdirs from Drupal)

## Acceptance criteria
- [ ] Links to content pages are in the expected order (same as current Drupal site)
- [ ] Subdirs are ordered similar to the current Drupal counterparts in their parent dir

## Related issues / PRs
* https://github.com/newrelic/docs-website/pull/458
* https://github.com/newrelic/docs-website/issues/349

## Helpful links
- [Example: PHP page with ordered nav items](https://docs.newrelic.com/docs/agents/php-agent/getting-started/introduction-new-relic-php)

roadlittledawn:
## Description
This applies two more sorts during migration on newly added fields in docs JSON that specify the sort weight of topics 2 and 3 in their respective parent. It worked for many of the subdirs, but not on some. Where there were obvious ordering issues, I manually added reorder instructions (`Agents > Manage APM agents`).

I also moved links to latest release note for each agent into the agent's getting started subdir alongside the link to all release notes for the agent

Looking at rest of the site, there may be tweaks to the nav here or there, especially where we added instructions to update IA. Those can be addressed in #587. Some I noticed include:
* [Alerts and Applied Intelligence](http://localhost:8000/docs/alerts-applied-intelligence)
* [FSO](https://docs-preview.newrelic.com/docs/full-stack-observability)
* [TDP](https://docs-preview.newrelic.com/docs/telemetry-data-platform)
* [Security](https://docs-preview.newrelic.com/docs/security)

## Related issues / PRs
* Closes #622 
* Related to #587


scripts/utils/migrate/migrate-nav-structure.js
```diff
@@ -401,4 +404,4 @@ const write = (file, contents) => {
 const shouldRemoveNode = (node) =>
   !node.path && (!node.pages || node.pages.length === 0);
 
-module.exports = migrateNavStructure;
+module.exports = { migrateNavStructure };
```
jerelmiller, line 407: It looks like we are still importing this as [the default](https://github.com/newrelic/docs-website/pull/729/files#diff-fb209cfbca6f9a5efb2360c1518d59e68c6974106513282a3317e1ab4d158edeR16). Can you revert this, or update the import to destructure this?

src/nav/release-notes.yml
```diff
@@ -3,2958 +3,1968 @@ path: /docs/release-notes
 pages:
```
jerelmiller, line 3: This nav is automatically generated now, so it might be best for us to delete this and filter it out in our migration script.

jerelmiller (changes requested):
The only reason I'm blocking this is due to the way the migrate nav structure function is exported. I assume the changes to the nav are correct 😄

scripts/utils/migrate/migrate-nav-structure.js
roadlittledawn: oh yep. thanks for catching that!

src/nav/release-notes.yml
roadlittledawn: you mean remove all child pages from `nav/release-notes.yml` right? that way we still have a link to the release notes in the root nav i assume?

roadlittledawn: updated migration script so it creates the `release-notes.yml` with no children pages so link still appears in root nav

plugins/gatsby-source-nav/gatsby-node.js
```diff
@@ -7,7 +7,7 @@ exports.createSchemaCustomization = ({ actions }) => {
     type Nav {
       id: ID!
       title(locale: String = "en"): String
-      pages: [NavItem!]!
+      pages: [NavItem!]
```
jerelmiller, line 10: It looks like we have a resolver here that ensures this will always return an empty array, even if there aren't child pages: https://github.com/newrelic/docs-website/pull/729/files#diff-c610a67ff8be52857e50c81f69210028602f97bfcf381d404585d458dfb22de4L76

Can you revert this?

jerelmiller:
@roadlittledawn would it be better to just delete that release notes file completely and instead hardcode the nav entry in the [RootNav](https://github.com/newrelic/docs-website/blob/develop/src/components/RootNavigation.js#L27) (like we do with whats new)?

The biggest reason I would push for that is because the nav is automated for release notes pages. By having a `yml` file, it implies that adding to that file would mean changes to the release notes nav. This might confuse others. Thoughts?

roadlittledawn:
> @roadlittledawn would it be better to just delete that release notes file completely and instead hardcode the nav entry in the [RootNav](https://github.com/newrelic/docs-website/blob/develop/src/components/RootNavigation.js#L27) (like we do with whats new)?
> 
> The biggest reason I would push for that is because the nav is automated for release notes pages. By having a `yml` file, it implies that adding to that file would mean changes to the release notes nav. This might confuse others. Thoughts?

i suppose my reasoning for trying to keep it was if a content contributor wants to add a link to the root nav (which i imagine will happen time to time), they may look here first? it also puts it in alpha order instead of the bottom (but i suppose that could always be changed via the `RootNavigation` component). think we should ensure that use case is supported via .yml?

but yeah, i do agree since release notes are different, we could just include them via `RootNavigation`

plugins/gatsby-source-nav/gatsby-node.js
roadlittledawn: ah gotcha. seemed like build was failing because of this. going to include release notes link via RootNavigation a la https://github.com/newrelic/docs-website/pull/729#issuecomment-773526770

jerelmiller:
>  if a content contributor wants to add a link to the root nav (which i imagine will happen time to time), they may look here first

That is a very valid point. I would say 95% of the time, this will be the case. We almost always want an associated yaml file with a set of navigation. The whats new and release notes nav just happen to be special cases, and we really want a hands off approach as much as possible. 

I guess the question is: Is it more confusing to have a `.yml` file associated with a nav that is fully automated (and therefore its `pages` property ignored), or more confusing to have that file completely gone, but still show up in the root nav?

> it also puts it in alpha order 

This is very true. This shouldn't be a problem (yet), but could be something we look at if we add a future nav that starts with a letter after "R" and we want that release notes nav item sorted in with it.

jerelmiller (approved):
Doppeeee. Thanks so much!!!

PR #743, opened 2021-02-03
Fix for removing context from vendor

caylahamann:


scripts/actions/check-job-progress.js
```diff
@@ -74,9 +74,11 @@ const main = async () => {
       .filter((batch) => batch && !batch.done)
       .map((batch) => batch.batchUid);
 
-    console.log(
-      `::set-output name=batchUids::${JSON.stringify(remainingBatches)}`
-    );
+    const outputBatchUids = remainingBatches.length
+      ? remainingBatches.join('[]')
+      : JSON.stringify(remainingBatches);
```
jerelmiller, line 79: Do you need this `JSON.stringify` if the `remainingBatches` are empty? Would just `[]` be sufficient here?

And out of curiosity, why use the `[]` delimiter here?

jerelmiller (commented):
Just 1 question before I approve, but otherwise looks great!

scripts/actions/check-job-progress.js
caylahamann: There's some wonkiness with the `JSON.stringify` that I couldn't figure out. It wasn't able to parse the array when there was something in the array (got a Unknown JSON token error) but if I didn't have the JSON.stringify when there was no files in the array, then remainingbatches wouldn't get the correct number of arguments

caylahamann: Thus why there's this ternary logic. It ensures that there is an empty array ('[]') or a delimiter for the remaining batches `batchuid1[]batchuid2`

scripts/actions/remove-completed-batch.js
```diff
@@ -60,32 +60,35 @@ const removePageContext = async (fileUris) => {
...
-
-    if (code === 'SUCCESS' && resp.ok) {
-      console.log(`[*] Successfully deleted ${fileUri} context.`);
-    } else {
-      console.error(`[!] Unable to delete ${fileUri} context.`);
-      return code;
-    }
-    return returnCode;
-  }, 'SUCESSS');
+  return await contextUids.reduce(
```
jerelmiller, line 63: Since the value returned here is not a promise, but an array of promises, this `await` might return too early. You'll want to wrap this in a `Promise.all` so that you ensure all promises are resolved/rejected before the array of values is fully resolved.

Since you're returning this value directly from this function, you shouldn't need the `await` keyword here either. By returning a `Promise.all`, you'll want the thing that calls this function to `await` its result.

scripts/actions/check-job-progress.js
jerelmiller: I guess what I mean is, if its empty, why even use `JSON.stringify` since you know its an empty array. Couldn't you just write the string `[]` itself? That being said, you may not even need this at all.

As a best practice, you'll want your values in both cases to be the same "type". Since your non-empty case is now a string delimited by a `[]`, you'll want your empty case to resemble this. Your empty case should no longer be an array, but rather a "string delimited by a []". In this case, the simplest from of an empty set of uids would just be an empty string.

This likely helps your parsing a bit too. Take the following 2 outputs:

```js
'[]'.split('[]') // => ["", ""]
''.split('[]') // => [""]
```

All of this being said, the `[]` delimiter is a bit strange. Instead, why not just use a comma delimiter? It looks like the array of values will be alphanumeric, so this should work great and won't feel quite as wonky. When doing so, `join`ing on an empty array gets you an empty string.

tl;dr; I would recommend simplifying this to the following:

```
const outputBatchUids = remainingBatches.join(',')

console.log(`::set-output name=batchUids::${outputBatchUids}`)
```

scripts/actions/remove-completed-batch.js
caylahamann: When I wrap this in `Promise.all` I don't get back a single value, I get back an array which is something like: `['S', 'U', 'C', 'C', 'E', 'S', 'S']` and I haven't found any errors in my testing while doing the await on the reduce. Do you know how to get around this?

jerelmiller: Ok I finally see your conundrum 😄 

In this case, it might be best to split this into 2 steps:

1) Go do all of the async fetches in a loop
2) Reduce on the returned value

You're right in that you want to reduce to a single value, not the array of values. Because you're executing the `fetch` in the loop, it gets pretty difficult to manage both the async callback and the single value.

Without the `Promise.all`, what you're left with is an array of promises. Take the following example:

```js
const timeout = (duration) => new Promise((resolve, reject) => {
  setTimeout(() => resolve(duration), duration)
})

const test = async () => {
  const value = await [timeout(10), timeout(5), timeout(100)]
  
  console.log(value)
}

test()
```

You'll notice this `await`s the array of promises (which mimic a bit of what you're doing here). In this case, the value that is logged is `Promise`, not the actual resolved value.

By splitting this into 2 steps, it will make it much easier to reduce the single value from the array of fetch results:

```js
const results = await Promise.all(
  contextUids.map(uid => fetch('/url', { method: 'DELETE', ... }).then(res => res.json())
)

return results.reduce((returnCode, { response }) => {
 // ... you get the idea
}, 'SUCCESS')
```

Thoughts?

caylahamann: I like that idea! I will implement that

caylahamann: Implemented that! let me know if it looks all good :)

scripts/actions/remove-completed-batch.js
```diff
@@ -60,32 +60,38 @@ const removePageContext = async (fileUris) => {
...
 
-    if (code === 'SUCCESS' && resp.ok) {
+  return results.reduce(({ code, fileUri }) => {
+    if (code === 'SUCCESS') {
       console.log(`[*] Successfully deleted ${fileUri} context.`);
     } else {
       console.error(`[!] Unable to delete ${fileUri} context.`);
       return code;
     }
     return returnCode;
```
jerelmiller, line 93:
```suggestion
    if (code === 'SUCCESS') {
      console.log(`[*] Successfully deleted ${fileUri} context.`);
      return returnCode;
    } else {
      console.error(`[!] Unable to delete ${fileUri} context.`);
      return code;
    }
```

[nit] Since you're using an `if`/`else` here, it might be best to return the `returnCode` in the `if` block, since thats the only time line 93 will be run:

jerelmiller (approved):
Thanks for making those changes! That loop looks much nicer 🙂

scripts/actions/remove-completed-batch.js
```diff
@@ -60,32 +60,38 @@ const removePageContext = async (fileUris) => {
...
+      const { code } = response;
 
-    const { response } = await resp.json();
-    const { code } = response;
+      return { code, fileUri };
+    })
+  );
 
-    if (code === 'SUCCESS' && resp.ok) {
+  return results.reduce(({ code, fileUri }) => {
```
jerelmiller, line 86: Ah shoot, looks like this `reduce` is missing its reduced value in the callback (i.e. the `memo` argument).

I assuming you meant this?

```suggestion
  return results.reduce((returnValue, { code, fileUri }) => {
```

jerelmiller (approved):

PR #748, opened 2021-02-03
Filter out more dummy docs

### LINKED ISSUE #682: Please remove intermediate, empty docs between menu links in main body and landing pages.
I discovered that we have intermediate (mostly empty) docs that appear between the main menu links in the body of a page and the landing pages. Here's what you need to do to reproduce this:

1. Go to https://docs-preview.newrelic.com/docs/full-stack-observability
2. In the main body of the page under "Observability solutions," click any of the menu items, such as "Application monitoring (APM)." 
3. On the intermediate page (https://docs-preview.newrelic.com/docs/full-stack-observability/monitor-everything/observability-solutions/application-monitoring-apm), click "APM."

As a user, I'm not expecting these intermediate pages.

This menu behavior in the main body of the page is different from the behavior of the same navigation in the left pane. The left page takes me directly to the landing pages and does not go to an intermediate page.

jerelmiller:
Closes #682

# Description

Filters out more dummy docs by checking the content and removing docs that contains < 100 characters.


jerelmiller:
@roadlittledawn it seems some of the redirects are missing. I've filtered out some additional content that contains less < 100 characters (for example, to treat this as a dummy doc: https://docs-preview.newrelic.com/docs/full-stack-observability/monitor-everything/observability-solutions/application-monitoring-apm), but it seems those pages did not get added as a redirect to any other page. Clicking these links now result in a 404. Do you know why that might be the case?

zstix (approved):

austin-schaefer:
@jerelmiller @zstix Are we removing the dummy docs, or handling them? The dummy docs are there for a speciic reason and just removing all of them is going to create some brokene xperiences.

If we have to remove them, can we get a list of all of the ones removed so we can manually re-add those docs to the necessary taxonomies?

jerelmiller:
@austin-schaefer in the new world, dummy docs are no longer a thing. What replaces them is an entry in the `redirects` frontmatter of the doc that a nav item links to. 

What we are doing with dummy docs is removing the `.mdx` file and adding both its path + its redirects to the doc that it points to. You can see an example of that on this set of docs:

Here are the redirects: https://github.com/newrelic/docs-website/pull/748/files#diff-96edeefd105f37be1f2b7cba9e47ea16ef73e292cc61280f43e95b6300d1b096R34-R48
And here is the dummy doc that was deleted: https://github.com/newrelic/docs-website/pull/748/files#diff-90c7181bf8ddeec94d303d605c147695f2fd648d3215600ae97f40de65ca1e4eL1-L20

You'll see that both the path to that deleted file + its redirects are in the doc it redirects to. This ensures the functionality isn't lost for the migration.

jerelmiller:
@austin-schaefer One other thing to note is that all the nav items remain the same. If a nav item links to a dummy doc, that now just means it links to a path that gets redirected to the doc it points to. That functionality remains.

My question above is more-so that we have some dummy docs getting deleted, but we weren't able to find a redirect for it, so the redirect path didn't get added to the doc it points to. I'm hoping to figure out why that was the case.

austin-schaefer:
AWESOME, thanks @jerelmiller ! Very clever.

This came up because @urbiz-nr stumbled across this old category that was mostly dummy docs that's acting a little funny https://docs-preview.newrelic.com/docs/integrations/new-relic-integrations

jerelmiller:
Funny you mention that. Those docs are fixed as a part of this PR. If you look at the file diff, you'll see the docs in that category have been deleted.

That being said, it looks like my question to @roadlittledawn pertains to these as well. We've deleted those, but weren't able to detect a redirect for those docs.

roadlittledawn:
> @roadlittledawn it seems some of the redirects are missing. I've filtered out some additional content that contains less < 100 characters (for example, to treat this as a dummy doc: https://docs-preview.newrelic.com/docs/full-stack-observability/monitor-everything/observability-solutions/application-monitoring-apm), but it seems those pages did not get added as a redirect to any other page. Clicking these links now result in a 404. Do you know why that might be the case?

looks like this falls under a redirect to a taxonomy term, which Zack has a PR up for #722 . i'll look and see if that one is covered there. 

looking at the redirects db table,  this url: `https://docs-dev.newrelic.com/docs/full-stack-observability/monitor-everything/observability-solutions/application-monitoring-apm` redirects to `taxonomy/term/2526` (which is `/docs/apm` page))

roadlittledawn:
> Funny you mention that. Those docs are fixed as a part of this PR. If you look at the file diff, you'll see the docs in that category have been deleted.
> 
> That being said, it looks like my question to @roadlittledawn pertains to these as well. We've deleted those, but weren't able to detect a redirect for those docs.

and i think this also falls under taxonomy term redirects since those docs link to a menu / index page. if redirects were added from those dummies to a taxonomy term page, it should be available in the redirects as such and handled via Zack's PR.

roadlittledawn:
Confirmed that in #722:
* `/docs/full-stack-observability/monitor-everything/observability-solutions/application-monitoring-apm` redirects to `/docs/apm`
* the dummy doc links in `docs/integrations/new-relic-integrations` redirect accordingly as well.

jerelmiller:
Oh wonderful! That is fantastic. I'll merge develop into this branch and check it out. Thanks for looking into that @roadlittledawn!

jerelmiller:
Ok, after merging this in, it looks like its still not quite right. It looks like the redirects are only added in the auto index pages plugin. These landing pages are definitely not auto generated so I think they are still being missed. I do see that they are listed in that file though as you said, so I'll do a bit of work on the migration to make sure the taxonomy redirects are added to the proper landing pages. Thanks for the help @roadlittledawn!

jerelmiller:
Moving to "Blocked" until #798 is done and merged. To fix, I'll need to run additional migrations, but I'd rather not pull in all of the data from the other PR.

zstix (approved):

PR #755, opened 2021-02-03
Add path fix

caylahamann:
Adds path fix so that the deserialized values save to the correct spot


caylahamann:
this works now: https://github.com/newrelic/docs-website/pull/765

zstix (approved):

PR #772, opened 2021-02-04
Serializes fenced codeblocks correctly

caylahamann:
Pulls in the logic from the gatsby-plugin jerel created in this PR: https://github.com/newrelic/docs-website/pull/686/files to correctly parse fenced code blocks


caylahamann:
This is the reason that this PR: https://github.com/newrelic/docs-website/pull/765 is not correctly building. When it was serialized, large segments of the page were turned into codeblocks

zstix (approved):

PR #774, opened 2021-02-04
Update get-know-telemetry-data-platform.mdx

bradleycamacho:
I did not think the intro was that good or friendly! I rewrote it in a friendlier tone and style.


jpvajda:
this doesn't seem like a legit PR so closing.

PR #779, opened 2021-02-04
Revert amplify yaml file 

caylahamann:
Reverting the amplify yaml file because our docs preview site has not updated since this was merged in previously.


zstix (approved):
We're going to investigate the catalyst to our builds not running, but I'm down to revert this while we're looking into it!

PR #784, opened 2021-02-04
agent doc hack

jpvajda:
# Description


jpvajda:
Just testing with @austin-schaefer  an example agent release not modification that eventually we'd want to automate some checks around to prevent this from occurring.

jerelmiller:
> we'd want to automate some checks around to prevent this from occurring.

Agreed. This would be good to do. For now, a PR review should be good. We'll have to watch out for that.

PR #799, opened 2021-02-05
Release

jerelmiller:


amplify.yml
```diff
@@ -1,7 +1,4 @@
-version: 1
```
jerelmiller, line 1: Hold up: It looks like these got reverted for some reason. We'll want to revert this change before we merge

Issue #497, opened 2020-12-30
Improve Product Category Landing Page experience


jpvajda:
When accessing a product catagory landing page, like FSO, TDP or Alerts and Intel, a user should be presented with a more visually appealing experience then just a index page of links. 

Also some intro descriptive content would be valuable to explain each product category and add SEO capabilities to these pages.

### Category Landing Pages 

https://docs-preview.newrelic.com/docs/telemetry-data-platform

https://docs-preview.newrelic.com/docs/full-stack-observability

https://docs-preview.newrelic.com/docs/alerts-and-applied-intelligence

### Acceptance Criteria

- [x]  a design and approach is provided by @djsauble 
- [x]  the design should try to use existing components so we can build these page templates quickly 
- [ ] content is provided by the docs team

jpvajda:
@djsauble / @austin-schaefer Let me know your thoughts on this need, I recall us discussing these pages and wanted to do something more visual, but I don't remember where we ended up with it. I would like to try to get something more visual on these pages for launch as the index page approach is not ideal for these particular pages.

djsauble:
In terms of using existing components, the approach taken to APM, etc seems quite reasonable.

https://docs-preview.newrelic.com/docs/apm

To make them a bit more visually appealing, we could put a video or screenshot to the right of the description text (as is done on the existing APM page, come to think of it).

https://docs.newrelic.com/docs/apm

jpvajda:
@djsauble great suggestions! would you have some time to just provide a higher level mock fore these pages? It always help to have a visual to go along with these discussions. 😄

austin-schaefer:
@jpvajda : Are you hoping this content will be ready for launch? If so, can you file a DOC jira for writing these? Those pages like the APM one take a while to write and research.

jpvajda:
@austin-schaefer for sure, once we have some mock design I can create that issue in the DOCS backlog. cc / @djsauble

jpvajda:
product videos are available here: https://www.youtube.com/c/NewRelicInc

jpvajda:
@djsauble how are things coming along for the final mock up for this page?

djsauble:
@jpvajda I didn't get any strong pushback to the sketch I shared in Slack, so that's the plan of record as far as I'm concerned. Happy to continue iterating if there's anything else you'd like considered.

![Uploading IMG_0471.jpeg…]()

jpvajda:
[...]

PR #813, opened 2021-02-09
Keep captions and fix landing page layout

### LINKED ISSUE #586: Update layout on landing pages
## Description

Now that we have images rendered in the site (#582), the landing pages need a format similar to the current docs site. Right now, the docs preview site renders the image in the same column as the introductory paragraph.

Screen Shot 2021-01-19 at 10 05 28 AM

This should be rendered as a 2-column layout instead.

Screen Shot 2021-01-19 at 10 05 42 AM

### LINKED ISSUE #786: Image Captions are hard to distiguinsh from page text
### Description

On the current docs site, image captions have their own styling. This ensures they don't blend in with the surrounding text and confuse readers.

### Screenshots

The preview site makes it hard to see the relationship between the caption and the image. Compare:

_Drupal site_
image

_Preview site_
image

### Additional context

Labeling as a launch-blocker because if we don't fix this by launch, we'll lose the formatting that tells us which text is supposed to be image captions and which is running text.

jerelmiller:
Closes #586
Closes #786
Partially addresses #497 

# Description

* Update migration script to wrap image captions in a `figcaption` element
* Update landing page layout to use 2-column look like in the current docs site
* Add custom remark plugin to remove paragraph from button
* Minor mobile style fixes on landing pages
* Adds fso and tdp landing pages with lorem ipsum

## Screenshots

localhost_8000_docs_apm_size=n_10_n


codemods/landingPageHero.js
```diff
@@ -0,0 +1,44 @@
...
+};
+
+// This codemod assumes it is run after the landing page tile grid has been run
+const landingPageHero = () => (tree, file) => {
+  if (file.data.doc.type !== 'landing_page') {
+    return;
+  }
+
+  visit(tree, 'root', (root) => {
+    const frontmatterIdx = 0; // frontmatter always comes first
```
zstix, line 19: If frontmatter always comes first, is there any advantage to just writing `.slice(1, gridIdx)` below? I'm not against it...just curious if we gain anything by attaching some context to `0`.

scripts/utils/migrate/taxonomy-redirects.js
```diff
@@ -25,8 +25,6 @@ const getTaxonomyPath = async (uri) => {
 
     return get(json, 'terms[0].term.urlPath');
   } catch (error) {
-    console.log(`Taxonomy ID ${id} does not exist, skipping`);
```
zstix, line 28: 😅 Ooops! Thanks for removing this.

zstix (approved):
One minor question, but it's more of a curiosity.

codemods/landingPageHero.js
jerelmiller: Not sure. `1` felt like a magic number to me, hence why I opted for the variable. That being said, this was probably more of a long-term strategy than it needs to be. This codemod really only needs to survive until GA. I'll go ahead and switch to use `1`. If this were something that stuck around for the lifetime of the application, I'd probably opt for this so that updating this 6 months from now was obvious was that `1` was for 😄

zstix (approved):

PR #820, opened 2021-02-09
Use integration tiles for release notes landing page

### LINKED ISSUE #613: Release notes landing page design modifications
in Our stakeholder feedback review the [release notes page](https://docs-preview.newrelic.com/docs/release-notes) was identified as a page that could use a more subtle design to reduce the size of the release notes CTA components.

### General Feedback 
- reduce the size of the CTA boxes
- remove the need for content within the boxes, and instead just show a logo relating to the agent language type
- Perhaps make it look and feel more like what we have for our integrations section on the home page 

### Current View

Screen Shot 2021-01-21 at 4 48 24 PM

### Design Idea

Screen Shot 2021-01-21 at 4 48 17 PM

jerelmiller:
Closes #613

# Description

Updates the release notes landing page to use the integration tiles on the home page. Refactors the components to ensure these can be shared.

Screen Shot 2021-02-09 at 2 29 45 PM


caylahamann:
It all looks goo to me! I just have a question about how you structured this. Is there a reason why there's a folder for gatsby theme newrelic and you put these changes in them, rather than putting it in the theme?

jerelmiller:
> Is there a reason why there's a folder for gatsby theme newrelic and you put these changes in them

In general, I'm using [shadowing](https://www.gatsbyjs.com/docs/how-to/plugins-and-themes/shadowing/) to add these icons, which is why they are in the gatsby newrelic folder. Essentially this allows me to add those icons and have the `Icon` component consume them, without actually needing to touch `node_modules` or release new versions.

> rather than putting it in the theme?

No good reason, other than they were already in this project and I didn't want to wait for a review/merge/release/update cycle before I could do this work. We definitely should consider moving all of these over to the theme, especially since now its a much easier copy/paste.

caylahamann (approved):

PR #821, opened 2021-02-09
Remove all translation context when job is complete

### LINKED ISSUE #796: Delete all page contexts when fetching content from vendor
## Summary 
When we run the `Send content to be translated` job, we send two contexts. One comes from the serialized html that automatically gets uploaded when we upload the file. The other one comes from a request to the context API that contains the full page html. When we get the content back, we only delete the full html page, not the serialized html file from the contexts. We want to delete both. This is important because our vendor runs automatic matching, and we don't want them to match against an outdated version of the file. 

## AC 
* [ ] serialized html is deleted from contexts when getting translated content back

zstix:
## Description
When a translation job is complete, we remove the "context" files from our translation vendor's platform. However, when we upload files, the vendor will make a _second_ "context" from the uploaded file. This PR ensures that we're removing _all_ context associated with a particular page.

## Related Issue(s)
* Closes #796

### DIFF
--- a/scripts/actions/remove-completed-batch.js
+++ b/scripts/actions/remove-completed-batch.js
@@ -57,7 +57,7 @@ const removePageContext = async (fileUris) => {
     .filter(({ name }) => [...fileNames, ...fileUris].includes(name))
     .map(({ name, contextUid }) => ({
       fileUri: name,
-      contextUid: contextUid,
+      contextUid,
     }));
 
   const results = await Promise.all(
--- a/scripts/actions/remove-completed-batch.js
+++ b/scripts/actions/remove-completed-batch.js
@@ -51,12 +51,14 @@ const removePageContext = async (fileUris) => {
     accessToken,
   });
 
+  // Find the smartling context to be removed. This includes context manually
+  // uploaded (fileNames) as well as the "context" created by smartling (fileUris).
   const contextUids = items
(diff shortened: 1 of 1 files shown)


scripts/actions/remove-completed-batch.js
```diff
@@ -51,14 +51,14 @@ const removePageContext = async (fileUris) => {
...
-    .map((item) => {
-      return {
-        fileUri: item.name,
-        contextUid: item.contextUid,
-      };
-    });
+    .filter(({ name }) => [...fileNames, ...fileUris].includes(name))
+    .map(({ name, contextUid }) => ({
+      fileUri: name,
+      contextUid: contextUid,
```
jerelmiller, line 60:
```suggestion
      contextUid,
```
[nit] can use shorthand here

jerelmiller (approved):
I assume this is working correctly. Nice find!

scripts/actions/remove-completed-batch.js
zstix: Oh duh...I had turned this into an implicit return for the purpose of removing that 🤦‍♂️

jerelmiller (approved):

scripts/actions/remove-completed-batch.js
jerelmiller: Haha no problem 😄

PR #822, opened 2021-02-09
Review nav instructions / order

### LINKED ISSUE #587: Review nav instruction set

roadlittledawn:
Closes #587 

## Description

This work updates instructions to the migration nav instruction set to address remaining nav issues.


zstix (approved):
Technically, this looks good to me. I'll let the content folks weigh in on the ordering.

Issue #728, opened 2021-02-02
Fix the OVERWHELMING docker image + consider transparent background images for these pages


mmfred:
### Description

On a couple of pages, the icons have gone a little haywire in migration. Specifically, the docker icon is huge compared to the others, and then I'm also unsure if the plan is for the icons and labels to be in one big list. 

as a bonus if we can also use transparent background images here, that would be ideal, preferably svgs. 

### Steps to reproduce

1. Go to https://docs-preview.newrelic.com/docs/logs/enable-log-management-new-relic/enable-log-monitoring-new-relic/forward-your-logs-using-infrastructure-agent#install 
2. OR https://docs-preview.newrelic.com/docs/infrastructure/install-infrastructure-agent/linux-installation/install-infrastructure-agent-linux#install-agent
3. See error and compare to this nice layout: https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/linux-installation/install-infrastructure-monitoring-agent-linux#install-agent

![Screen Shot 2021-02-02 at 5 41 27 PM](https://user-images.githubusercontent.com/1395158/106681653-28386000-657e-11eb-935b-3dd69ba7da17.png)

jpvajda:
this is definitely a bug.. lol. I'd also like to look into to using images with transparent backgrounds here so we don't have this effect in dark mode.

bradleycamacho:
Hey @jpvajda,

I just took a look at some of these icons and realized one of two things is happening for each:

1. The icons HAVE a white background, and are "correctly" rendering with the white background. The docs team should go through and update these occurrences going forward.
2. The icons have a transparent background, but render with a white background on the site. Not sure why this is happening, it seems like a bug. An example of this is the Docker icon. 
![docker-logo-crop](https://user-images.githubusercontent.com/42678939/107067511-39f35080-6794-11eb-8823-bdac9b36413d.png)

jpvajda:
@bradleycamacho yep yep. we'll fix it!

bradleycamacho:
Nice, thanks you!

mmfred:
This is another page that has giant docker icon: 
https://docs-preview.newrelic.com/docs/infrastructure/install-infrastructure-agent/get-started/install-infrastructure-agent

bradleycamacho:
The only solution:

![whale](https://user-images.githubusercontent.com/42678939/107077399-a45ebd80-67a1-11eb-8b9b-8a32f7fecc86.gif)

[...]

PR #826, opened 2021-02-10
Fix large icons for tile grid on some docs pages

jerelmiller:
Partially addresses #728

# Description

Addresses one of the main concerns on a few of the docs pages that use an icon
inside a tile, similar to our integration tile. This PR adds a codemod to
capture these during migration, and updates the `TechTile` to handle images as
icons.

## Screenshots

Screen Shot 2021-02-09 at 6 21 25 PM


codemods/techTile.js
```diff
@@ -0,0 +1,113 @@
...
+                    image.alt && mdxAttribute('alt', image.alt),
+                  ].filter(Boolean)
+                ),
+              ])
+            ).trim()
+          )
+        ),
+      ].filter(Boolean);
+
+      node.children = [];
```
caylahamann, line 71: Since my mind is on localization, we should note that we also need to add a handler for the techTile component. I wonder if there's a way to easily always add a new component to handled with the serialize, deserialize function (in case we forget to add it the handlers for localization).

codemods/techTile.js
```diff
@@ -0,0 +1,113 @@
...
+      const tiles = parent.children.slice(idx, idx + count);
+      const tileGrid = mdxBlockElement('TechTileGrid', [], tiles);
+
+      parent.children.splice(idx, count, tileGrid);
+
+      return idx + 1;
+    }
+  );
+
+  imageImports.reverse().forEach(({ name, path }) => {
```
caylahamann, line 105: why is `reverse()` used here?

gatsby-config.js
```diff
@@ -181,12 +183,20 @@ module.exports = {
...
+          ],
+        ],
         gatsbyRemarkPlugins: [
           {
             resolve: 'gatsby-remark-images',
             options: {
               maxWidth: 1200,
               linkImagesToOriginal: false,
+              backgroundColor: 'transparent',
+              disableBgImageOnAlpha: true,
```
caylahamann, line 199: what does this option do? `disableBgImageOnAlpha`?

src/components/TechTile.js
```diff
@@ -1,15 +1,43 @@
...
-    return cloneElement(icon, { className });
+  if (isValidElement(icon) && icon.props?.mdxType === 'img') {
+    return (
+      <span
+        className={className}
+        css={css`
+          flex: 1;
+        `}
+      >
+        <ClassNames>
```
caylahamann, line 28: Why use `ClassNames` instead of applying the `css` class directly?

caylahamann (commented):
just some questions :)

gatsby-config.js
jerelmiller: According to the [docs](https://www.gatsbyjs.com/plugins/gatsby-remark-images/):

> Images containing transparent pixels around the edges results in images with blurry edges. As a result, these images do not work well with the “blur up” technique used in this plugin. As a workaround to disable background images with blurry edges on images containing transparent pixels, enable this setting.

There were a few of these transparent pngs that looked rather weird when they loaded when using that background image, so I opted to turn that off.

codemods/techTile.js
jerelmiller: No great reason other than I'm adding these to the AST bottom to top, so the first import actually started at the bottom.  This just reverses the order to start with so that as its inserted in the AST, it remains the same order. No great reason other than I liked the way it looked 😄

src/components/TechTile.js
jerelmiller: The reason is because of the `cloneElement` below. The `css` prop actually has a compile-time step to it where emotion does some work behind the scenes to turn this into a classname prop at build time. 

In this case, because this is not JSX, its difficult to use that `css` prop on `cloneElement` and have it work nicely with that build time step. This `ClassNames` utility gives you an "workaround" that allows you to attach to `className` directly. 

You can see more about this here: https://emotion.sh/docs/class-names

(though now that I re-read that, I wonder if it would work with `css`? I've had troubles in the past, but maybe thats been resolved with newer versions 🤔 )

codemods/techTile.js
jerelmiller: Good catch! I'll update this to include a deserializer.

> I wonder if there's a way to easily always add a new component to handled with the serialize, deserialize function

We talked about this early on and opted for it to be loud when it fails. The reason is because components may need to do some special logic other than the default, and we didn't want to miss something in the localization if we introduced a new component and forgot to add a handler for it. In this case, its better to fail than to be partially wrong.

For example, lets say we introduced the following component:

```
<Dropdown title="Pick something">
  <Dropdown.Item label="Option 1" />
  <Dropdown.Item label="Option 1" />
</Dropdown>
```

If we decided to use default serialize/deserializers, this would mean that none of these props would get serialized in a way that translators would be able to use (because we base64 encode all props). In this case, we'd get back a translated doc with untranslated dropdown items.

By failing, we are able to catch this much easier than if we had decided to use a default serialize/deserialize.

jerelmiller:
@caylahamann this PR has been updated to ensure `TechTile` and `TechTileGrid` are properly serialized/deserialized for translation.

caylahamann (approved):

PR #851, opened 2021-02-12
Create redirects for more dummy docs to tax term / index paths

### LINKED ISSUE #758: Links to agent release notes are returning 404s
### Description

Links to the Agent release notes from the APM docs left nav are returning 404s. I was able to replicate this issue in three sets of APM docs and assumed that it would continue that way.

### Steps to reproduce

1. Go to https://docs-preview.newrelic.com/docs/agents/go-agent
2. In the left nav, click Get started and then click C SDK release notes (https://docs-preview.newrelic.com/docs/agents/c-sdk/get-started/c-sdk-release-notes)
3. See error

I replicated this in Go and Java as well.

### Expected behavior

I think we should see the release notes page for each agent.

### LINKED ISSUE #804: The PHP agent API "View all methods" in the left nav returns a 404
### Description

The PHP agent API "View all methods" left nav link is returning a 404. 

### Steps to reproduce

1. Go to https://docs-preview.newrelic.com/docs/agents/php-agent/php-agent-api
2. In the left nav, click on **View all methods**
4. See 404 error

### Expected behavior

What's strange is that the link in step 1 has the content I expect to see after clicking **View all methods**.

roadlittledawn:
Closes #758, closes #804

## Description
This supports redirecting dummy docs (like those in agent docs that link to that agent's release note index page AND view all agent API methods links) that link to an index page.

Uses a new migration API resource to get a Drupal node's URL alias given the node URI in the redirect data.

### DIFF
--- a/scripts/utils/migrate/taxonomy-redirects.js
+++ b/scripts/utils/migrate/taxonomy-redirects.js
@@ -28,6 +28,13 @@ const getTaxonomyPath = async (uri) => {
   }
 };
 
+/**
+ * Gets the URL path for a Drupal node.
+ * If unable to find the path, null is returned.
+ * @param {string} uri The node URI in Drupal
+ * @returns {Promise<string>}
+ */
+
 const getNodePath = async (uri) => {
   const id = uri.replace(/.*\/node\//, '');
   const url = new URL(`/api/migration/content/node/${id}/urlPath`, BASE_URL);
@@ -53,19 +60,20 @@ const fetchTaxonomyRedirects = async (redirects) => {
       .map(async ([url, paths]) => [await getTaxonomyPath(url), paths])
   );
 
-  return await taxonomy.reduce(async (memo, [url, paths]) => {
-    const filteredPaths = await Promise.all(
-      paths
-        .filter((path) => path !== url && !path.startsWith('/taxonomy'))
-        .map(async (path) =>
-          path.startsWith('/node') ? await getNodePath(path) : path
-        )
-    );
-    const finalPaths = filteredPaths.filter((path) => Boolean(path));
+  const filteredPaths = await Promise.all(
+    taxonomy.map(async ([url, paths]) => [
+      url,
+      await Promise.all(
+        paths
+          .filter((path) => path !== url && !path.startsWith('/taxonomy'))
+          .map((path) => (path.startsWith('/node') ? getNodePath(path) : path))
+      ),
+    ])
+  );
 
-    return Boolean(url) && finalPaths.length > 0
-      ? { ...(await memo), [url]: finalPaths }
-      : await memo;
+  return filteredPaths.reduce((memo, [url, paths]) => {
+    paths = paths.filter(Boolean);
+    return Boolean(url) && paths.length > 0 ? { ...memo, [url]: paths } : memo;
   }, {});
 };
 
--- a/scripts/utils/migrate/taxonomy-redirects.js
+++ b/scripts/utils/migrate/taxonomy-redirects.js
@@ -29,6 +29,23 @@ const getTaxonomyPath = async (uri) => {
   }
 };
 
+const getNodePath = async (uri) => {
+  const id = uri.replace(/.*\/node\//, '');
+  const url = new URL(`/api/migration/content/node/${id}/urlPath`, BASE_URL);
+  const headers = {
+    'Content-Type': 'application/json',
+    'X-Phpshield-Key-Disable': process.env.ACQUIA_DEV_PHP_SHIELD_KEY,
+  };
+  try {
+    const res = await fetch(url, { headers });
+    const json = await res.json();
+    return get(json, 'docs[0].doc.path');
+  } catch (error) {
+    fs.writeFileSync('src/data/nodes.txt', url);
+    return null;
+  }
+};
+
 const fetchTaxonomyRedirects = async (redirects) => {
   const taxonomy = await Promise.all(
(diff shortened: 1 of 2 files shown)


src/data/taxonomy-redirects.json
```diff
@@ -10,6 +10,7 @@
       "/docs/servers/new-relic-servers-linux/maintenance/server-monitor-known-issues",
       "/docs/servers/new-relic-servers-linux/maintenance/servers-linux-known-issues",
       "/docs/servers/new-relic-servers-linux/troubleshooting/servers-linux-known-issues",
+      null,
```
jerelmiller, line 13: Can you make sure these `null`s are filtered out?

scripts/utils/migrate/taxonomy-redirects.js
```diff
@@ -29,24 +29,42 @@ const getTaxonomyPath = async (uri) => {
...
-        !path.startsWith('/node') &&
-        !path.startsWith('/taxonomy')
+  return await taxonomy.reduce(async (memo, [url, paths]) => {
+    const filteredPaths = await Promise.all(
+      paths
+        .filter((path) => path !== url && !path.startsWith('/taxonomy'))
+        .map(async (path) =>
+          path.startsWith('/node') ? await getNodePath(path) : path
+        )
+        .filter((path) => Boolean(path))
```
jerelmiller, line 62: Since you're returning a list of Promises from the `map` right above this, this `filter` is going to return `true` 100% of the time (since `Boolean(Promise<any>) === true`). You'll need to run this `filter` after you've `await`ed the results from the `Promise.all`

scripts/utils/migrate/taxonomy-redirects.js
```diff
@@ -29,24 +29,42 @@ const getTaxonomyPath = async (uri) => {
...
       .map(async ([url, paths]) => [await getTaxonomyPath(url), paths])
   );
 
-  return taxonomy.reduce((memo, [url, paths]) => {
-    const filteredPaths = paths.filter(
-      (path) =>
-        path !== url &&
-        !path.startsWith('/node') &&
-        !path.startsWith('/taxonomy')
+  return await taxonomy.reduce(async (memo, [url, paths]) => {
```
jerelmiller, line 55: Let's try and avoid reducing on an `async` function. They are a bit hard to manage. That `await taxonomy.reduce` isn't doing what you think its doing 😄 

This `taxonomy.reduce` is actually going to return an array of promises. Unfortunately reducing on an array of promises doesn't resolve when all of the items have been resolved. As a test, try this:

```js
const reducePromises = async () => {
  const nums = [1, 2, 3]
  return await nums.reduce(async (memo, num) => {
     return memo + num
  }, 0)
}

const test = async () => {
  const result = await reducePromises()  
  
  console.log(result)
}

test()
```

What you'll get is `[object Object]3`, not quite what you'd expect. Instead, I find it best to break this up into 2 steps:

1) do all the async work in a `Promise.all`
2) Reduce on that value

Since you're trying to fetch data in here, I'd break it up that way:

```
const filteredPaths = await Promise.all(
  taxonomy
    .filter((path => ...)
    .map(async ([url, paths]) => ...)
)

return filteredPaths.reduce(...)
```

I think you'll find this to be much more reliable

scripts/utils/migrate/taxonomy-redirects.js
```diff
@@ -29,24 +29,42 @@ const getTaxonomyPath = async (uri) => {
...
+        .map(async (path) =>
+          path.startsWith('/node') ? await getNodePath(path) : path
+        )
+        .filter((path) => Boolean(path))
     );
 
-    return Boolean(url) && filteredPaths.length > 0
-      ? { ...memo, [url]: filteredPaths }
-      : memo;
+    return Boolean(url) && (await filteredPaths.length) > 0
```
jerelmiller, line 65: `filteredPaths.length` should never return a promise, so this `await` is unnecessary here.

scripts/utils/migrate/taxonomy-redirects.js
```diff
@@ -29,24 +29,42 @@ const getTaxonomyPath = async (uri) => {
...
+      ? { ...(await memo), [url]: await filteredPaths }
+      : await memo;
```
jerelmiller, line 67: See above. `reduce` gets really goofy with `async`, so I'd recommend just doing the async work in step 1, then reduce, which will get rid of all these weird `await`s

jerelmiller (changes requested):

jerelmiller (approved):

roadlittledawn:
hmmm... checking the Amplify build for this PR, some taxonomy redirects don't seem to work? digging further

This [node agent link](https://pr-851.d1vx4w5vgyla4s.amplifyapp.com/docs/agents/nodejs-agent/getting-started/nodejs-release-notes) should redirect to [node release notes index page](https://pr-851.d1vx4w5vgyla4s.amplifyapp.com/docs/release-notes/agent-release-notes/nodejs-release-notes).

roadlittledawn:
i think i see why. for release notes, it's because we exclude that directory via `gatsby-config.js` when creating auto index pages (because we do special handling for for those index pages).

    options: {
      skippedDirectories: ['', 'whats-new', 'docs/release-notes'],
    },

and for "view all api methods" it's because we use an index.mdx file to add authored content to top of page, so the auto index plugin ignores those (so it doesn't overwrite the content in index.mdx, which is mainly landing pages)

PR #853, opened 2021-02-12
chore: testing mark tags (Don't Merge)

jpvajda:
This PR tests using `mark` tags in content sections outside of code blocks. It also has markdown linter corrections. 😉


jpvajda:
We don't need to merge this, as it doesn't change the visualization of the marked text outside the code block, but it's more for reference for how it could work.

jerelmiller:
@jpvajda do we need this PR anymore? Could we close it if not?

PR #855, opened 2021-02-12
Remove paragraphs from table components

caylahamann:
Remove paragraphs from table pieces, `html-react-parser` on the nr1-help-xp side is having trouble parsing the table with the `<p>` tags in between

### DIFF
--- a/plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
+++ b/plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
@@ -144,8 +144,8 @@ module.exports = {
   thead: (h, node) => h(node, 'thead', {}, all(h, node)),
   tbody: (h, node) => h(node, 'tbody', {}, all(h, node)),
   tr: (h, node) => h(node, 'tr', {}, all(h, node)),
-  th: (h, node) => h(node, 'tr', {}, all(h, node)),
-  td: (h, node) => h(node, 'tr', {}, all(h, node)),
-  var: (h, node) => h(node, 'tr', {}, [u('text', toString(node))]),
-  mark: (h, node) => h(node, 'tr', {}, [u('text', toString(node))]),
+  th: (h, node) => h(node, 'th', {}, all(h, node)),
+  td: (h, node) => h(node, 'td', {}, all(h, node)),
+  var: (h, node) => h(node, 'var', {}, [u('text', toString(node))]),
+  mark: (h, node) => h(node, 'mark', {}, [u('text', toString(node))]),
 };


zstix (approved):

plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
```diff
@@ -144,8 +144,8 @@ module.exports = {
   thead: (h, node) => h(node, 'thead', {}, all(h, node)),
   tbody: (h, node) => h(node, 'tbody', {}, all(h, node)),
   tr: (h, node) => h(node, 'tr', {}, all(h, node)),
-  th: (h, node) => h(node, 'tr', {}, all(h, node)),
-  td: (h, node) => h(node, 'tr', {}, all(h, node)),
```
jerelmiller, line 148: Careful with this change. Let's say you had the following MDX:

```
<td>
  Call the `newrelic.sendCustomEvent` using your <var>ACCOUNT_ID</var> provided.
</td>
```

By using `toString` here, you lose all of the formatting inside this node.

jerelmiller (changes requested):

plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
caylahamann: I need to get rid of the paragraphs from that node, do you have any idea of how I could accomplish that?

jerelmiller: Out of curiosity, why is that?

Our docs site renders the paragraphs:

Screen Shot 2021-02-12 at 1 01 38 PM

Is this more a styling thing or is there a technical reason why these need to be removed?

caylahamann: Nevermind I think I found a way 🤓

caylahamann: Not a styling thing, the parser that we use to turn the dom into react is having trouble with the paragraphs inside the `td` and `th` blocks.

caylahamann: The documentation also kind of sucks but basically what happens is that the it will change this format: 
```html
<table>
   <thead>
      <tr>
      <tr>
         <p><strong>NRQL condition</strong></p>
      </tr>
      <tr>
         <p>Details</p>
      </tr>
      </tr>
   </thead>
   <tbody>
      <tr>
      <tr>
         <p>Required values</p>
      </tr>
      <tr>
         <p>The <a href="#state-select"><code>SELECT</code></a> statement and <a href="#sel-from"><code>FROM</code></a> clause are required. All other clauses are optional. You can start your query with either <code>SELECT</code> or <code>FROM</code>.</p>
      </tr>
....
```
into this: 
```html
 <p><strong>NRQL condition</strong></p>
<p>Details</p>

<table>
   <thead>
      <tr>
      <tr>
         <p><strong>NRQL condition</strong></p>
      </tr>
      <tr>
...
```

jerelmiller: Thats very strange 🤔 

I hestitate to remove all paragraphs because it also means we lose formatting on multiple paragraphs used in a table column (of which the docs site has several of these)

```html
<td>
  <p>This is a paragraph</p>
  <p>This is also a paragraph that is meant to visually separated
</td>
```

In fact, the screenshot above shows a great example of multiple paragraphs in a table cell. Are we sure its the parser and not something else?

Sorry to push back on this, but making this change has the potential to fundamentally alter the text structure of the table cell, which I'm not sure is acceptable.

PR #857, opened 2021-02-12
Related resources

### LINKED ISSUE #427: Add related content links to the Docs Site in the right rail
Adding dynamic content links using our related content component in the right rail under the page header links would allow us to drive users to OSS and Dev Sites and connect users to more discoverable content. 

### Considerations

- **display logic** what about really long right rail scenarios [like this page](https://docs-preview.newrelic.com/docs/agents/java-agent/configuration/java-agent-configuration-config-file). Would we want to NOT show the related content component if a page has a very long set of navigational headers in the right rail? 

- **cost** What costs would we incur with swift type as costs would be driven up by all the docs pages needing to refresh with dynamic content links. (WIP)

- **event tracking** How can we measure if this is valuable to the end user? Perhaps track click events on the dynamic links to see if people use it? Or look at bounce rates on pages that have the dynamic links, It will likely increase bounce rates on those pages but increase traffic to our other sites. Measuring when users click on the related links would be very useful to understand when users feel they need to go to another site for content needs.

- **Localization Impacts** What about searching through a translated site like Japan? It's currently done through the proxy, so what is our approach now?  - once the related content is available via a query, it shouldn't really impact swifttype cost as we'll have the related content already from the initial query.

### Swift type impacts and approach

We refetch results per page once a day on a cron job. The data is available as part of the build: https://github.com/newrelic/developer-website/blob/develop/src/data/related-pages.json

That data gets rebuilt in this action once a day: https://github.com/newrelic/developer-website/blob/develop/.github/workflows/fetch-related-content.yml 

We can specify whatever interval we want if we go this route. When we execute a query per page each time we fetch related content for a page. We have a limit on the number of queries we can run before we are charged more.

If we fetch once a day, and we have 3000 queries we have to run, thats 3k queries thats ~90k queries per month. We could decrease the interval and instead do once a week, or once a month.

Doing so would put costs down, but means that auto related content could become stale (_maybe thats a fair tradeoff_)

### Right Rail Mockup

Untitled Diagram

### LINKED ISSUE #845: reduce padding in search box in the global header
Yeah, reducing the left and right padding from 1rem to 0.5rem seems about right.

jerelmiller:
Closes #845 
Closes #427 

# Description

Add related resources to the docs site.

NOTE: The screenshots below are using sample data. We do not actually have related resource data yet.

## Screenshots

Screen Shot 2021-02-12 at 3 05 38 PM

Screen Shot 2021-02-12 at 1 46 29 AM


gatsby-config.js
```diff
@@ -64,6 +65,37 @@ module.exports = {
...
           ],
         },
+        relatedResources: {
+          swiftype: {
+            resultsPath: `${__dirname}/src/data/swiftype-resources.json`,
+            engineKey: 'Ad9HfGjDw4GRkcmJjUut',
+            refetch: Boolean(process.env.BUILD_RELATED_CONTENT),
+            filter: ({ slug }) =>
+              slug ===
+              '/docs/integrations/kubernetes-integration/understand-use-data/kubernetes-cluster-explorer',
```
jerelmiller, line 75: Oops, will update this. I did this for testing purposes so that I wouldn't trigger a full rebuild

gatsby-config.js
```diff
@@ -64,6 +65,37 @@ module.exports = {
             'python',
           ],
         },
+        relatedResources: {
+          swiftype: {
+            resultsPath: `${__dirname}/src/data/swiftype-resources.json`,
+            engineKey: 'Ad9HfGjDw4GRkcmJjUut',
```
zstix, line 71: Just to double-check, this is public and okay to commit, right?

jerelmiller: Yep! Its a client token specifically used for frontends to tell it which search engine to use.

gatsby-config.js
jerelmiller: Actually, I'm going to leave this in for now so that we can test the build end-to-end once or twice with related resources to make sure its all hooked up correctly. Once we confirm, then I'll update this to be the real logic.

zstix (approved):

PR #861, opened 2021-02-13
Switch query to prevent sharp warnings on build

jerelmiller:
# Description

Currently PR builds emit a bunch of warnings related to querying `childImageSharp` on files that will return null. This PR attempts to silence those warnings by querying for image through allImageSharp.

### DIFF
--- a/plugins/gatsby-plugin-generate-doc-json/gatsby-node.js
+++ b/plugins/gatsby-plugin-generate-doc-json/gatsby-node.js
@@ -29,39 +29,40 @@ const htmlGenerator = unified()
 exports.onPostBuild = async ({ graphql, store }) => {
   const { program } = store.getState();
 
-  const query = `query {
-    allMdx(filter: {fileAbsolutePath: {regex: "/src/content/docs/"}}) {
-      nodes {
-        mdxAST
-        slug
-        fields {
-          fileRelativePath
-        }
-      }
-    }
-    allFile(filter: {childImageSharp: {id: {ne: null}}}) {
-      nodes {
-        childImageSharp {
-          fluid {
-            src
+  try {
+    const { data } = await graphql(`
+      query {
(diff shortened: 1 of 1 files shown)


plugins/gatsby-plugin-generate-doc-json/gatsby-node.js
```diff
@@ -29,39 +29,40 @@ const htmlGenerator = unified()
...
 
-    const imageHashMap = allFile.nodes.reduce(
-      (obj, { childImageSharp, relativePath }) => {
-        return {
-          ...obj,
-          [relativePath]: childImageSharp.fluid.src,
-        };
-      },
+    const imageHashMap = allImageSharp.nodes.reduce(
+      (memo, { fluid, parent }) => ({
```
caylahamann, line 62: is there a reason for call this memo? It's not really a memoized function right?

caylahamann (approved):
all loooks goood to me, just curious about the naming for `memo`

plugins/gatsby-plugin-generate-doc-json/gatsby-node.js
jerelmiller: Ah this harkens back to my Ruby days where this variable was called `memo` as a generic term for the accumulator (https://ruby-doc.org/core-2.4.0/Enumerable.html#method-i-reduce). I forget that in JS its usually called "accumulator" as the shorthand. Thats the only reason. This wasn't meant to mean "memoized value", my brain just went to ruby-land generic term.

caylahamann: ah that makes sense! would it be possible to change it to `acc`? Just because in my brain, memo always means some sort of memoized value or function (like `useMemo`)

jerelmiller: Sure!

caylahamann (approved):
Thank you!

PR #864, opened 2021-02-16
Remove json plugin to test amplify build

caylahamann:


jerelmiller (approved):
Approving so we can get builds to succeed

PR #875, opened 2021-02-16
Add Tessen pageView tracking

### LINKED ISSUE #789: [Tessen] Add Tessen to our docs site
## Summary 
In order to reach parity with the current docs site, we need to add tessen to our docs site. We currently have it on our developer site: 
* https://github.com/newrelic/developer-website/blob/d9961a32caaa315546fbf1e5fa98ef87f1c8bccf/src/components/Seo.js

We will probably want to move it out of an SEO component and into some global head component for the site.

zstix:
## Description
Adds the necessary configuration for Tessen to track `pageView` actions.

## Reviewer Notes
The configuration for this was taken from our existing documentation site. Most of the work to get this working was done in https://github.com/newrelic/gatsby-theme-newrelic/pull/257

## Related Issues
* Closes #789

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -125,6 +125,9 @@ module.exports = {
           pageView: {
             name: 'pageView',
             category: 'DocPageView',
+            getProperties: ({ env }) => ({
+              env: env === 'production' ? 'prod' : env,
+            }),
           },
         },
       },
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -126,7 +126,7 @@ module.exports = {
             name: 'pageView',
             category: 'DocPageView',
             getProperties: ({ location, env }) => ({
-              location,
+              path: location.path,
               env: env === 'production' ? 'prod' : env,
             }),
           },
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -117,6 +117,16 @@ module.exports = {
             },
           },
         },
+        tessen: {
+          product: 'DOC',
+          subproduct: 'TDOC',
+          segmentWriteKey: 'AEfP8c1VSuFxhMdk3jYFQrYQV9sHbUXx',
+          trackPageViews: true,
+          pageView: {
+            name: 'pageView',
+            category: 'DocPageView',
+          },
+        },
       },
     },
     {
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -126,7 +126,7 @@ module.exports = {
             name: 'pageView',
             category: 'DocPageView',
             getProperties: ({ location, env }) => ({
-              path: location.path,
+              path: location.pathname,
               env: env === 'production' ? 'prod' : env,
             }),
           },
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -125,7 +125,8 @@ module.exports = {
           pageView: {
             name: 'pageView',
             category: 'DocPageView',
-            getProperties: ({ env }) => ({
+            getProperties: ({ location, env }) => ({
+              location,
               env: env === 'production' ? 'prod' : env,
             }),
           },


gatsby-config.js
```diff
@@ -117,6 +117,16 @@ module.exports = {
             },
           },
         },
+        tessen: {
+          product: 'DOC',
+          subproduct: 'TDOC',
+          segmentWriteKey: 'AEfP8c1VSuFxhMdk3jYFQrYQV9sHbUXx',
```
jerelmiller, line 123: I assume this is safe to commit?

gatsby-config.js
```diff
@@ -117,6 +117,16 @@ module.exports = {
...
+          trackPageViews: true,
+          pageView: {
+            name: 'pageView',
+            category: 'DocPageView',
+          },
```
jerelmiller, line 128: I noticed we have a couple other properties on the docs site, namely `env` and `content_type`. Is that something we should also instrument using the [`getProperties`](https://github.com/newrelic/gatsby-theme-newrelic/tree/develop/packages/gatsby-theme-newrelic#tessen) option?

gatsby-config.js
zstix: Yep! It's used as a way to let segment know where to put data ([docs](https://segment.com/docs/connections/find-writekey/)). This was taken straight from the current docs site.

gatsby-config.js
zstix: I was unsure about this. We don't have any other environments and we've sorta done away with the "content type" concept. I'm happy to add additional metadata to these events, but there wasn't any additional acceptance criteria on the ticket 🤔

gatsby-config.js
```diff
@@ -117,6 +117,16 @@ module.exports = {
             },
           },
         },
+        tessen: {
+          product: 'DOC',
+          subproduct: 'TDOC',
+          segmentWriteKey: 'AEfP8c1VSuFxhMdk3jYFQrYQV9sHbUXx',
+          trackPageViews: true,
+          pageView: {
```
roadlittledawn, line 125: i can't quite recall, but we explicitly send the URL path along with the Tessen.page() call

jerelmiller: I do see it there in the screenshot above https://github.com/newrelic/docs-website/pull/875#discussion_r577186771

Definitely something we should add as well

gatsby-config.js
roadlittledawn: yep. i checked way back when if it was okay that this was exposed, and security signed off on it

gatsby-config.js
zstix: I've added the environment and location to the event properties.

gatsby-config.js
zstix: I believe I've covered this in https://github.com/newrelic/docs-website/pull/875/commits/4df3d72dfcd5a8b13f34e4c7e920cbd191d5750c

gatsby-config.js
```diff
@@ -117,6 +117,20 @@ module.exports = {
...
+            name: 'pageView',
+            category: 'DocPageView',
+            getProperties: ({ location, env }) => ({
+              location,
```
jerelmiller, line 129: 2 things with this one:

* It looks like we call this `path` in the current docs site. Assuming we want to maintain consistency, we should probably do the same
* This is the `location` object for the page. We likely just want the `pathname` property off the location. Could you update to use `pathname`?

jerelmiller:
```suggestion
              path: location.pathname,
```

zstix: Oh gotcha, I was unclear what `location` represented. Good to know...I'll update.

gatsby-config.js
```diff
@@ -117,6 +117,20 @@ module.exports = {
...
+        tessen: {
+          product: 'DOC',
+          subproduct: 'TDOC',
+          segmentWriteKey: 'AEfP8c1VSuFxhMdk3jYFQrYQV9sHbUXx',
+          trackPageViews: true,
+          pageView: {
+            name: 'pageView',
+            category: 'DocPageView',
+            getProperties: ({ location, env }) => ({
+              path: location.path,
```
jerelmiller, line 129:
```suggestion
              path: location.pathname,
```

Since this is a location object, `path` will be undefined

jerelmiller (approved):
Thanks for getting that updated!

PR #891, opened 2021-02-17
Add classnames for images in help-xp

caylahamann:
Adds classnames back to images for rendering in the in product xp as well styling for width, height and vertical align

### DIFF
--- a/plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
+++ b/plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
@@ -8,6 +8,7 @@ const toString = require('mdast-util-to-string');
 const u = require('unist-builder');
 const { compileStyleObject } = require('../../../rehype-plugins/utils/styles');
 const { set, get } = require('lodash');
+const path = require('path');
 
 const stripNulls = (obj) =>
   Object.fromEntries(Object.entries(obj).filter(([, value]) => value != null));
@@ -20,13 +21,10 @@ const getAllAttributes = (node) =>
     : {};
 
 const getSrcUrl = (fileRelativePath, url) =>
-  fileRelativePath
-    .replace('.mdx', '')
-    .replace('src/content/', '')
-    .split('/')
-    .slice(0, -1)
-    .join('/')
-    .concat(url.replace('./', '/'));
+  path.join(
+    path.dirname(fileRelativePath.replace('src/content', '')),
+    url.replace('./', '')
+  );
 
 module.exports = {
   image: (h, node, imageHashMap, fileRelativePath) => {
--- a/plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
+++ b/plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
@@ -40,7 +40,7 @@ module.exports = {
       node,
       isBlockImage ? 'div' : 'span',
       {
-        className: isBlockImage ? 'block-image' : 'inline-image',
+        className: [isBlockImage ? 'block-image' : 'inline-image', 'image'],
       },
       [
         h(
--- a/plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
+++ b/plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
@@ -167,7 +167,7 @@ module.exports = {
     const [image] = node.children;
     set(image, 'data.style', compileStyleObject(style));
 
-    return one(h, image, node);
+    return one(h, image, node.parent);
   },
   InlineCode: (h, node) => h(node, 'code', {}, [u('text', toString(node))]),
   table: (h, node) => h(node, 'table', {}, all(h, node)),
--- a/plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
+++ b/plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
@@ -2,6 +2,7 @@ const all = require('mdast-util-to-hast/lib/all');
 const { findAttribute } = require('../../../codemods/utils/mdxast');
 const toString = require('mdast-util-to-string');
 const u = require('unist-builder');
+const { compileStyleObject } = require('../../../rehype-plugins/utils/styles');
 
 const stripNulls = (obj) =>
   Object.fromEntries(Object.entries(obj).filter(([, value]) => value != null));
@@ -22,14 +23,31 @@ module.exports = {
       .slice(0, -1)
       .join('/')
       .concat(node.url.replace('./', '/'));
-    return h(
-      node,
-      'img',
-      {
+
+    const isBlockElement =
+      isMdxElement('paragraph', node.parent) &&
+      node.parent.children.length === 1;
+
+    if (isBlockElement) {
+      return h(node, 'img', {
         src: imageHashMap[srcUrl] || node.url,
         alt: node.alt,
+        className: 'article-image',
+      });
+    }
+
+    return h(
+      node,
+      'div',
+      {
+        className: 'type-image',
       },
-      all(h, node)
+      [
+        h(node, 'img', {
+          src: imageHashMap[srcUrl] || node.url,
+          alt: node.alt,
+        }),
+      ]
     );
   },
   CodeBlock: (h, node) => {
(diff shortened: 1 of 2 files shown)


zstix (approved):

plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
```diff
@@ -13,23 +17,50 @@ const getAllAttributes = (node) =>
...
     : {};
 
+const getSrcUrl = (fileRelativePath, url) =>
+  fileRelativePath
+    .replace('.mdx', '')
+    .replace('src/content/', '')
+    .split('/')
+    .slice(0, -1)
+    .join('/')
+    .concat(url.replace('./', '/'));
```
jerelmiller, line 27: Is this trying to get the full relative path to the image given the MDX file its in? You can simplify this with `path` helpers:

```js
path.join(path.dirname(fileRelativePath.replace('src/content', '')), url)
```

plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
```diff
@@ -13,23 +17,50 @@ const getAllAttributes = (node) =>
...
-      .replace('src/content/', '')
-      .split('/')
-      .slice(0, -1)
-      .join('/')
-      .concat(node.url.replace('./', '/'));
+    const srcUrl = getSrcUrl(fileRelativePath, node.url);
     return h(
       node,
-      'img',
+      'div',
```
jerelmiller, line 54: Since this is outputting to a `div`, how does this work when the image is used inline with text? 

Here are a couple pages that show this:

* https://docs-preview.newrelic.com/docs/infrastructure/install-infrastructure-agent/update-or-uninstall/uninstall-infrastructure-agent
* https://docs-preview.newrelic.com/docs/infrastructure/install-infrastructure-agent/get-started/install-infrastructure-agent

jerelmiller (commented):
Can we do a quick video chat? I'd love to work through this with you a bit and see if we can eliminate the `imageHashMap` and `fileRelativePath` passed to the `mdx*Element` handlers. Ideally we'd use a single `image` handler for rendering images, regardless of the parent. I've got some ideas on how to do it, but its a bit complicated to lay out in a PR review.

plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
```diff
@@ -13,23 +20,37 @@ const getAllAttributes = (node) =>
       }, {})
     : {};
 
+const getSrcUrl = (fileRelativePath, url) =>
+  path.join(
+    path.dirname(fileRelativePath.replace('src/content', '')),
+    url.replace('./', '')
```
jerelmiller, line 26:
```suggestion
    url
```

You can simplify this to simply `url`. Setting `url` here, even if it has the `./` still works correctly. Try the following in a node console:

```js
path.join('src/content', './images/one.png') // => 'src/content/images/one.png'
path.join('src/content', 'images/one.png') // => 'src/content/images/one.png'
```

plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
```diff
@@ -13,23 +20,37 @@ const getAllAttributes = (node) =>
...
+      node.parent.children.length === 1;
+
     return h(
       node,
-      'img',
+      isBlockImage ? 'div' : 'span',
       {
-        src: imageHashMap[srcUrl] || node.url,
-        alt: node.alt,
+        className: [isBlockImage ? 'block-image' : 'inline-image', 'image'],
```
jerelmiller, line 41: I see that you're setting `data.style` in the `ImageSizing` handler (which is 💯 ), but you'll also need to access it here and set it as a property, otherwise the data just gets discarded.

One other question: Does it makes sense to do the style string compilation here or from the parent component that wants to add inline styles to the image? That would affect where you call `compileStyleString`.

One advantage of doing it here vs a parent is that the parent would then be able to use object literal syntax to set styles on `data.style` rather than having to convert to a string. Thoughts? Totally fine the way it is, but figured I'd spur some discussion 🙂

jerelmiller (commented):
These changes are looking FANTASTIC! The only reason I'm not approving yet is because I think the `style` gets lost on the image node when compiled.

plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
caylahamann: ah I set it on the image, but I'm going to want to pass it in as 
```js 
stripNulls({
        className: [isBlockImage ? 'block-image' : 'inline-image', 'image'],
        style: get(node, 'data.style', null),
```

correct?

Also: when you say object literal syntax, what do you exactly mean?

jerelmiller: I just mean this:

```js
const style = {
  display: 'inline-block',
  margin: '0 auto'
}
```

vs this:

```js
const style = 'display: inline-block; margin: 0 auto;
```

Typically "object literal" refers the the bare object as a variable (i.e. `{ }`). Since everything is technically an "Object" in JavaScript, the term "object literal" was meant to describe those kinds of variables. Does that help?

jerelmiller: And yes! That change looks great 😄

jerelmiller (dismissed):

plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
```diff
@@ -148,4 +178,6 @@ module.exports = {
   td: (h, node) => h(node, 'td', {}, all(h, node)),
   var: (h, node) => h(node, 'var', {}, [u('text', toString(node))]),
   mark: (h, node) => h(node, 'mark', {}, [u('text', toString(node))]),
+  figcaption: (h, node) =>
+    h(node, 'div', { className: ['meta'] }, all(h, node)),
```
jerelmiller, line 182: Is there any reason this couldn't be a [`figcaption`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption) element instead of a `div`?

caylahamann: I guess there is no reason (except for that how's the nr1-help-xp currently handles it). I'm also not super familiar with `figcaption` but I thought it would have to be wrapped by a `figure` element

plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
```diff
@@ -139,6 +157,18 @@ module.exports = {
...
+    const style = stripNulls({
+      height: findAttribute('height', node),
+      width: findAttribute('width', node),
+      verticalAlign: findAttribute('verticalAlign', node),
+    });
+
+    const [image] = node.children;
+    set(image, 'data.style', compileStyleObject(style));
+
+    return one(h, image, node);
```
jerelmiller, line 170: tl;dr; you might need to make the `parent` the `node.parent` instead of the `ImageSizing` node itself.

**Why?**

Now that I look at this one more time, I'm realizing that those `rehype` plugins that handle image sizing and inline images have a bit of an order dependency. If you look at the [`image-sizing` plugin](https://github.com/newrelic/docs-website/blob/develop/rehype-plugins/image-sizing/index.js), you'll notice that the `ImageSizing` removes itself from the tree and instead [replaces itself with the image node](https://github.com/newrelic/docs-website/blob/develop/rehype-plugins/image-sizing/index.js#L34). This means there will be no additional wrapper. Because of this change, the image node's parent now becomes the `ImageSizing` node's parent.

This is important, because the [`gatsby-inline-images`](https://github.com/newrelic/docs-website/blob/develop/rehype-plugins/gatsby-inline-images/index.js) plugin [runs after](https://github.com/newrelic/docs-website/blob/develop/gatsby-config.js#L231-L237) the `image-sizing` plugin runs. This is important because the `image-sizing` plugin checks to see if its parent is a `paragraph` and has the image as its only child (much like you're doing here).

If the image handler checks to see if its parent is a `paragraph`, but the `ImageSizing` node is not "removed", then you may not be able to detect block level images that use `ImageSizing` (of which there are definitely pages that do).

In other words, you won't be able to properly handle this MDX:

```mdx
Some paragraph

<ImageSizing width="20px">
  ![icon.png](./images/icon.png)
</ImageSizing>

More stuff
```

The space in between those paragraphs are what cause the AST to wrap a paragraph around the `ImageSizing` and `image` markdown. To ensure you can detect the above as a block image, you'll want the image's parent to be the `ImageSizing` parent. Does that make sense?

jerelmiller (approved):

PR #893, opened 2021-02-17
Add Google Search Console verification

### LINKED ISSUE #791: [GA] Add our site to the google search console
## Summary 
Add the docs site to the google search console (similar to how we did with the OSS site and dev site). This will require adding a meta tag with the verification code. This how we implemented it in the dev site: 
https://github.com/newrelic/developer-website/blob/d9961a32caaa315546fbf1e5fa98ef87f1c8bccf/src/components/Seo.js#L29-L30

## AC 
* [ ] Docs site is added to the google search console (verify with @jpvajda because he has access)

zstix:
## Description
Creates a `SEO` component that extends the theme with docs-site-specific meta content (so that we don't need to constantly update a ton of files).

Also adds the site verification code for docs.newrelic.com.

## Related Issue(s)
* Closes #791

### DIFF
--- a/src/layouts/MainLayout.js
+++ b/src/layouts/MainLayout.js
@@ -47,9 +47,7 @@ const MainLayout = ({ data = {}, children, pageContext }) => {
 
   return (
     <>
-      <SEO location={location}>
-        <meta name="google-site-verification" content="foobar" />
-      </SEO>
+      <SEO location={location} />
       <GlobalHeader />
       {isSmallScreen && (
         <MobileHeader
--- a/src/layouts/MainLayout.js
+++ b/src/layouts/MainLayout.js
@@ -5,12 +5,12 @@ import {
   Layout,
   Link,
   Logo,
-  SEO,
   useLayout,
 } from '@newrelic/gatsby-theme-newrelic';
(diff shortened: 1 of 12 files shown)


src/layouts/MainLayout.js
```diff
@@ -47,7 +47,9 @@ const MainLayout = ({ data = {}, children, pageContext }) => {
 
   return (
     <>
-      <SEO location={location} />
+      <SEO location={location}>
+        <meta name="google-site-verification" content="foobar" />
```
jerelmiller, line 51: It doesn't look like the `SEO` component accepts a `children` prop. I also noticed the `foobar` set as `content`. I assume this was likely for testing purposes. Could you remove this?

jerelmiller (changes requested):
Just one minor change, but otherwise looks great! This will be really be a really nice foundation when we get into meta tags for swiftype as well

src/layouts/MainLayout.js
zstix: Sorry...I had that for testing and forgot to remove it. This should be resolved now.

jerelmiller (approved):

PR #895, opened 2021-02-18
Meta description for basic doc pages

### LINKED ISSUE #551: [SEO] Include meta description tag

The meta header for docs doesn't include the meta description. This is essential for SEO and a GA blocker.

* Include the SEO description in frontmatter of file
* Include description in the header in the correct format to be picked up by search engines
* Automatically import existing meta tags from Drupal site
* Not a requirement but super handy: Enforce length limitation of 160 characters. (Descriptions longer than this commonly get cut off.)

Search engines and other tools rely heavily on this field. For example, Slack uses it in link previews. Google often shows it in search results when there's no precise keyword match. Swiftype uses it to build search results.

### Example

Compare source of this doc on prod and preview site:

* https://docs.newrelic.com/docs/agents/php-agent/installation/install-php-agent-shared-hosting-service
* https://docs-preview.newrelic.com/docs/agents/php-agent/installation/install-php-agent-shared-hosting-service

The preview site lacks the `description` field ("Info on PHP agent compatibility with shared hosting providers and how to determine if your host supports installing the agent.")

zstix:
## Description
Adds a meta descrtipion in the header that will either use the `shortDescription` or `title` (from frontmatter) for a basic page.

Includes an update to the docs-site-specific `<SEO />` component to support adding a `descrtipion` prop.

## Related Issue(s)
* Closes #551


src/components/SEO.js
```diff
@@ -10,17 +10,31 @@ const METADATA = [
...
-    {METADATA.map((data) => (
-      <meta key={data.name} {...data} />
-    ))}
-  </SEO>
-);
+const DocsSiteSeo = ({ location, title, description }) => {
+  const metadata =
+    description || title
+      ? [
+          ...METADATA,
```
jerelmiller, line 17: Rather than sticking this one the array, I think the dynamic `meta` tags might be a little cleaner if you put it in React itself:

```js
<SEO>
  {metadata.map(...)}
  {(description || title) && <meta name='description' content={description || title} />}
</SEO>
```

src/templates/docPage.js
```diff
@@ -130,6 +130,7 @@ export const pageQuery = graphql`
       body
       frontmatter {
         title
+        shortDescription
```
jerelmiller, line 133: Should this get passed into the `description` field on the `SEO` component on [line 67](https://github.com/newrelic/docs-website/pull/895/files#diff-1d1e91cf1df9133860992dc5d7fb0566a86cbae77ff1c92aea195586dc730f4dL67)?

jerelmiller (commented):
One minor nit and one bit of data that might have been lost

src/templates/docPage.js
zstix: Yep

src/components/SEO.js
```diff
@@ -10,17 +10,21 @@ const METADATA = [
...
 
-const DocsSiteSeo = ({ location, title }) => (
+const DocsSiteSeo = ({ location, title, description }) => (
   <SEO location={location} title={title}>
     {METADATA.map((data) => (
       <meta key={data.name} {...data} />
     ))}
+
+    {description ||
+      (title && <meta name="description" content={description || title} />)}
```
jerelmiller, line 20:
```suggestion
    {(description || title) && <meta name="description" content={description || title} />}
```

I think prettier got ya 😄. The parens should wrap `description || title` to ensure the boolean logic runs correctly, otherwise you'll only get this meta tag when `title` is present

jerelmiller: You may want to make that change locally as I'm not sure the above suggestion is prettier friendly

zstix: Ugh yeah that's prettier.

Sorry, I'm moving much too quickly and making a lot of stupid mistakes.

roadlittledawn:
i imagine we'll want to update this once we add the author written meta descriptions to frontmatter right? unless we call that field `shortDescription` though that name doesn't feel quite right.

zstix:
@roadlittledawn Do we have a different frontmatter field that authors are using to describe the page? That was the only field that I could find that was relevant (and it's only used in a few places).

roadlittledawn:
> @roadlittledawn Do we have a different frontmatter field that authors are using to describe the page? That was the only field that I could find that was relevant (and it's only used in a few places).

yeah it's in the migration JSON but we just haven't updated the migration script to add that to frontmatter yet

jerelmiller (approved):

PR #921, opened 2021-02-22
Use translated titles on locale auto index pages

### LINKED ISSUE #565: Dynamic TOC not showing Japanese menus
The dynamic [menu for Key transactions](https://docs-preview.newrelic.com/jp/docs/apm/transactions/key-transactions) does not show an item that has been translated though it shows on the [page's left nav](https://docs-preview.newrelic.com/jp/docs/apm/transactions/key-transactions/introduction-key-transactions).
Screenshot attached.
Screen Shot 2021-01-14 at 1 00 15 PM

roadlittledawn:
Closes #565

## Description
Adds necessary graphql query and swaps out a title of a page if a translated version exists for each locale when creating auto index pages via `gatsby-plugin-auto-index-pages`. It also adds the locale slug for each link.

### DIFF
--- a/plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
+++ b/plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
@@ -178,30 +178,47 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
         });
       });
     }
+  });
 
-    locales.forEach(({ localizedPath }) => {
+  locales.forEach(({ localizedPath }) => {
+    visit(list, 'directory', (dir) => {
+      const slug = `/${dir.path}`;
       const localizedSlug = path.join(`/${localizedPath}`, slug);
 
+      if (skippedDirectories.includes(dir.path)) {
+        return [visit.SKIP];
+      }
+
+      if (existingPaths.includes(slug)) {
+        return;
+      }
+
       dir.children
-        .filter((child) => child.type === 'file')
-        .forEach((child) => {
-          const localizedFileSlug = path.join(
-            '/',
-            localizedPath,
-            child.data.fields.slug
-          );
-          const matchedNode = translatedFileNodes.find(
-            ({
-              childMdx: {
-                fields: { slug },
-              },
-            }) => slug === localizedFileSlug
-          );
-          if (matchedNode) {
-            child.data.frontmatter.title =
-              matchedNode.childMdx.frontmatter.title;
+        .filter((child) => child.type === 'directory')
+        .forEach((subDir) => {
+          if (subDir.children) {
+            subDir.children
+              .filter((child) => child.type === 'file')
+              .forEach((child) => {
+                const localizedFileSlug = path.join(
+                  '/',
+                  localizedPath,
+                  child.data.fields.slug
+                );
+                const matchedNode = translatedFileNodes.find(
+                  ({
+                    childMdx: {
+                      fields: { slug },
+                    },
+                  }) => slug === localizedFileSlug
+                );
+                if (matchedNode) {
+                  child.data.frontmatter.title =
+                    matchedNode.childMdx.frontmatter.title;
+                }
+                child.data.fields.slug = localizedFileSlug;
+              });
           }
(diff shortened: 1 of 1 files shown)


plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
```diff
@@ -150,13 +182,31 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
     locales.forEach(({ localizedPath }) => {
       const localizedSlug = path.join(`/${localizedPath}`, slug);
 
+      const localeDir = dir;
```
zstix, line 185: Do we need to create a new variable for this?

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
```diff
@@ -150,13 +182,31 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
...
+
+      localeDir.children.forEach((child) => {
+        if (child.type === 'file') {
```
zstix, line 188: You might want to `filter` the list before looping over it. That way you don't need an `if` statement.

```js
dir.children
  .filter((child) => child.type === 'file')
  .forEach((child) => {
    // do stuff here
  });

**Review by zstix — COMMENTED** [review]
Couple of small syntax comments

File: plugins/gatsby-plugin-auto-index-pages/gatsby-node.js (hunk 1 above)
Comment (roadlittledawn) [reply plugins/gatsby-plugin-auto-index-pages/gatsby-node.js:185]: good question. i was getting some funky behavior when i was using a nested `visit()` buuut since i changed that this may not be necessary. will test it out

**Comment by roadlittledawn** [comment]
i noticed that this only worked for the leaf directory. so, on the page: `/docs/apm/transactions/key-transactions` it worked as expected. but on `/docs/apm/transactions` it did not. i tried updating the code to traverse a directory's children, but it would then only work for `/docs/apm/transactions` and not the leaf directory (would always use japanese titles and slugs).

File: plugins/gatsby-plugin-auto-index-pages/gatsby-node.js [hunk 3]
```diff
@@ -146,10 +178,49 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
...
+      }
+
+      if (existingPaths.includes(slug)) {
+        return;
+      }
+
+      dir.children
+        .filter((child) => child.type === 'directory')
+        .forEach((subDir) => {
+          if (subDir.children) {
```
zstix, line 199: We might be able to get rid of this `if` statement to help with some of the indentation:

```js
dir.children
  .filter((child) => child.type === 'directory)
  .filter((subDir) => subDir.children && subDir.children.length)
  .forEach(({ children }) => {
    // do your filtering and whatnot
  });
```

You might be able to reduce nesting even further to remove that nested loop:

```js
dir.children
  .filter((child) => child.type === 'directory)
  .filter((subDir) => subDir.children && subDir.children.length)
  .flatMap(({children}) => children)
  .filter((child) => child.type === 'file')
  .forEach((child) => {
    // do your child-specific stuff here
  });
```

zstix: And if you wanna go _super_ functional:

```js
const isDirectory = ({ type }) => type === 'directory';
const isFile = ({ type }) => type === 'file';
const hasChildren = (dir) => dir.children && dir.children.length;

dir.children
  .filter(isDirectory)
  .filter(hasChildren)
  .flatMap(prop('children'))
  .filter(isFile)
  .forEach((child) => {
    // do your child-specific stuff here
  });

caylahamann (commented):
This all looks good to me, but I do like Zack's suggestions, so if you wanna implement that (or not) I'll give a re-review :)

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
roadlittledawn: much neater! just pushed that up

PR #941, opened 2021-02-24
Add New Relic deployment marker action

### LINKED ISSUE #644: Add `newrelic/deployment-marker-action` to track deployments
## Description
We should consider adding https://github.com/newrelic/deployment-marker-action to a workflow that's triggered whenever we deploy the site (merge into `main`). This will help us troubleshoot errors and gain deeper insight from within New Relic.

zstix:
## Description
Adds [`newrelic/deployment-marker-action@v1`](https://github.com/newrelic/deployment-marker-action) to a new workflow that is triggered whenever a PR into main is merged (or a release is published). This relies on credentials which are already added to the repository secrets.

The `applicationId` is for our [browser application](https://staging.one.nr/0LkjnevlRo1).

We may want to consider adding a slack message that gets triggered on release as well, but it's outside the scope of this PR.

## Related Issue(s)
* Closes #644

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,7 +3,7 @@ name: Release
 on:
   pull_request:
     types:
-      - main
+      - closed
     branches:
       - main
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,27 @@
+name: Release
+
+on:
+  release:
+    types:
+      - created
+  pull_request:
+    types:
+      - main
+    branches:
+      - main
+
+jobs:
+  newrelic:
+    name: New Relic deployment marker
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set release version from tag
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
+name: Release
+
+on:
+  release:
```
jerelmiller, line 4: Will this create 2 deployment markers if we both create a GitHub release and merge the PR to `main`?

zstix: Yeah. I wasn't sure if there was a scenario where we would do one without the other (such as publish a release from an _older_ commit). It's a bit of a weird use case, I could just remove the `release` part since we have no tooling set up for that right now anyway.

jerelmiller: Awesome. Definitely something we can add later if need be

jerelmiller (approved):

PR #942, opened 2021-02-24
Instrument hash requests

### LINKED ISSUE #788: [NR Browser] Instrument hash requests
## Summary 
When a user goes to one of our pages with a hash in the url, we want to record that hash (on page load). Not when the hash changes while the user is on the page. 

## AC 
* [ ] Instrument `hash_request` with NR Browser

jerelmiller:
Closes #788

## Description

Instrument requests that contain a hash in the URL. Will only instrument hash
requests when the page first loads, or navigating to a different page with a
hash.

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -1,1 +1,2 @@
+export { default as onRouteUpdate } from './gatsby/on-route-update';
 export { default as wrapPageElement } from './gatsby/wrap-page-element';
--- a/gatsby/on-route-update.js
+++ b/gatsby/on-route-update.js
@@ -1,0 +1,11 @@
+const onRouteUpdate = ({ location, prevLocation }) => {
+  if (
+    window.newrelic &&
+    location.hash &&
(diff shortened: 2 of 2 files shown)


zstix (approved):
Nice! For this to work, all we need is to export `onRouteUpdate`?

roadlittledawn:
i accessed a this [PR preview URL w/ a hash](https://pr-942.d1vx4w5vgyla4s.amplifyapp.com/docs/alerts-applied-intelligence/new-relic-alerts/get-started/introduction-alerts#security), but i'm not seeing the data reported in [staging NR1 DevEx acct](https://staging.one.nr/08dQeYorQez). that's where it should be reporting right?

jerelmiller:
@roadlittledawn it looks like we [don't have a `staging` config](https://github.com/newrelic/docs-website/blob/develop/gatsby-config.js#L118) for New Relic, which is set for PR previews, so you won't see anything reported there.

PR #946, opened 2021-02-24
Add feedback to right rail on attribute dictionary page

### LINKED ISSUE #882: Attribute dictionary page is missing the feedback section in the right rail
The [attribute dictionary](https://docs-preview.newrelic.com/attribute-dictionary) page is missing the feedback section in the right rail. For consistency, we should include it here as well.

Screen Shot 2021-02-16 at 4 34 50 PM

jerelmiller:
Closes #882

### Tell us why

The feedback widget was missing on the right rail for the attribute dictionary
page. This adds it back.

### DIFF
--- a/src/pages/attribute-dictionary.js
+++ b/src/pages/attribute-dictionary.js
@@ -13,6 +13,7 @@ import {
   useQueryParams,
   Icon,
   useTranslation,
+  SimpleFeedback,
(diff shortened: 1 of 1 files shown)


zstix (approved):
It's awesome how simple it is to make some of these adjustments.

PR #949, opened 2021-02-24
Update NR in Practice nav path

### LINKED ISSUE #940: [Bug Bash] Missing a page for "New Relic in practice"
## Description
In the top-level navigation, we have a link for `New Relic in practice` that goes to our `404` page. We should either find a path for this page or remove it from the navigation.

roadlittledawn:
Closes #940

## Description

Update nav migration instructions to set NR in Practice to link to https://docs-preview.newrelic.com/docs/using-new-relic . The previous path led to a 404 page. 

Also included are redirects to existing index pages in `manual-edits` that I must've missed. And on migration a single quote came through un-encoded which looked legit to me so committed it.

### DIFF
--- a/scripts/utils/migrate/instruction-set.js
+++ b/scripts/utils/migrate/instruction-set.js
@@ -431,6 +431,13 @@ module.exports = [
     to: ['New Relic in practice'],
   },
   {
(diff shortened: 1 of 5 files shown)


jerelmiller (approved):
Nice!

Issue #892, opened 2021-02-17
Add three new release notes category


austin-schaefer:
### Is your feature request related to a problem? Please describe

As part of the platform release notes content type EoL, Austin manually migrated three categories of former platform release notes over to standard release notes. These new categories need to be included in the migration script.

### Affected categories

* https://docs.newrelic.com/docs/release-notes/infrastructure-release-notes/kubernetes-integration-release-notes/
* https://docs.newrelic.com/docs/release-notes/infrastructure-release-notes/cloud-integration-release-notes/
* https://docs.newrelic.com/docs/release-notes/diagnostics-release-notes/diagnostics-cli-release-notes/

### Describe the solution you'd like

All of these categories need the following implemented and/or included in the migration script:

* Migrate the actual content
* Ensure a version number is optional in `releaseNote.js` template, graphql, etc
* Migrate the taxonomies
* Migrate redirects (for both content and taxonomy)
* Migrate existing RSS feeds
* Add these new categories to https://docs-preview.newrelic.com/docs/release-notes
* Handle the following dummy docs:
    * https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/diagnostics-cli-nrdiag/diagnostics-release-notes
    * https://docs.newrelic.com/docs/integrations/kubernetes-integration/get-started/release-notes
    * https://docs.newrelic.com/docs/integrations/infrastructure-integrations/cloud-integrations/cloud-integrations-release-notes

### Complexifiers and potential gotchas

* Some of this content is Markdown like other release notes, some is HTML
* Cloud integrations release notes *do not have version numbers*

### Additional context

Lots and lots of context and detail in Jira DOC-6550

jpvajda:
I'm evaluating if this is actually a launch-blocker, would like to get a rough estimate on it to get a sense if we should wait to do this post GA, there are some pretty bad trade offs to waiting: 

. It will mean the content disappears and the links are broken, so I'd like to avoid if at all possible.

austin-schaefer:
There are 108 total release notes in these categories.

jerelmiller:
Confirmed that these pages were pulled in from the final migration!

PR #952, opened 2021-02-24
Account for new release notes, optional version number

roadlittledawn:
## Description

Sets up support for #892. Updates nav and `releaseNote.js` template so `version` isn't required for a release note (cloud integrations do not have a version number).

Adds dummy doc and original platform release note index link to redirects in the index.mdx files for each directory. Adds those index.mdx files on migration via `manual-edits`.

## Screenshot(s)

Example of a cloud integration release note will look like (the subject / name will actually be `New Relic infrastructure cloud integration`)


jerelmiller (approved):
Nice work! 

Do we need to add tiles to the release note landing page for these as well?

roadlittledawn:
@jerelmiller d'oh! i was going to do that but got ahead of myself. i'll do that now. thanks for catching that!

jerelmiller (approved):

PR #953, opened 2021-02-25
Doc 6554 navigation panes

rhetoric101:
### Tell us why

This PR accidentally includes two Jiras: doc-6554 and doc-6508:

The first part was to move the left-navigation pane material from the CONTRIBUTING.MD file to the main docs site. I made a few tweaks to the material.

I inserted a link in CONTRIBUTING.MD that points to the new topic in the docs site. I crafted what I thought the new URL might be and just inserted that into the redirect of "update-left-navigation-pane.mdx"

I stood this up in my local preview to look at it, and to do so, I inserted a copy of it in the current directory (not protected from the migration). This should get overwritten by the next migration, right? If not, I can remove it.

The second part is for some additional, unrelated work in Jira doc-6508 to move the "What's New?" material over to the site (from CONTRIBUTING.MD). After talking with Shawn, we decided to move the "What's New?" where it's going to be seen easily by tech writers. I simply removed it from CONTRIBUTING.MD and created a new topic. Be sure to check my new topic because I did some reformatting (table and callouts).


src/content/docs/new-relic-only/tech-writer-style-guide/processes-procedures/update-left-navigation-pane.mdx
```diff
@@ -0,0 +1,46 @@
...
+            path: /docs/insights/use-insights-ui/manage-dashboards/filter-insights-dashboards
+          - title: View, organize, share Insights dashboards
+            path: /docs/insights/use-insights-ui/manage-dashboards/view-organize-share-insights-dashboards
+```
+
+## Configuration file options
+
+When you're changing the navigation, keep in mind that each node in the `.yml` configuration file can have the following properties:
+
+| Key        | Required? | Description                               |
```
bradleycamacho, line 36: This is fine, but keep in mind you can use html tables instead of markdown if you want!

bradleycamacho (approved):
Looks good!

src/content/docs/new-relic-only/tech-writer-style-guide/processes-procedures/update-left-navigation-pane.mdx
rhetoric101: @bradleycamacho, I apologize! I hadn't pushed up the file! I only noticed that I hadn't pushed it up after your comment about the Markdown table, which I had replaced with the <Table> component. I'm must not have saved this, which is why this didn't show up in GitHub Destop as a commit to make. Anyway, could you re-read this? I made a variety of changes to it--not just the the table.

bradleycamacho: @rhetoric101 Have you pushed up the change?

bradleycamacho (approved):
Looks good to me!

PR #956, opened 2021-02-25
generate-docs-json checks for node parent 

caylahamann:
## Summary 
the docs json was not building because of not checking the node.parent existance when checking for whether the element is a block element. This PR now checks for that, as well as errors when the generate-doc-json fails to build


jerelmiller (approved):

caylahamann:
@jerelmiller I'm going to wait for it to build gain before merging in! It built locally but I wanna make sure bc you never know 👀

jerelmiller:
Agreed! Always good to make sure it builds nicely on these systems :)

PR #972, opened 2021-02-26
Remove leading slash

caylahamann:
removes leading slash from `srcUrl` so that the hashmap works :)

### DIFF
--- a/plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
+++ b/plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
@@ -44,7 +44,7 @@ module.exports = {
           node,
           'img',
           stripNulls({
-            src: imageHashMap[srcUrl] || node.url,
+            src: imageHashMap[srcUrl.substr(1)] || node.url,
             alt: node.alt,
           })
         ),


plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
```diff
@@ -44,7 +44,7 @@ module.exports = {
           node,
           'img',
           stripNulls({
-            src: imageHashMap[srcUrl] || node.url,
+            src: imageHashMap[srcUrl.substr(1)] || node.url,
```
jerelmiller, line 47: So after [looking at the JSON](https://pr-972.d1vx4w5vgyla4s.amplifyapp.com/docs/new-relic-solutions/best-practices-guides/full-stack-observability/monitor-apps-hosts-health-maps.json) for a doc with an image, there will still be a problem, and that it looks like these images are relative paths. That presents a problem with a different site, since these images aren't available at one.newrelic.com or staging-one.newrelic.com.

We'll need to make sure these images have the site URL prefix so that they show up correctly.

See that JSON doc above. You'll see an image with `/static/health-map-entity-box-f79da9aace992f8491d13d149114fff4.png` as the `src`, but that should be `https://docs-preview.newrelic.com//static/health-map-entity-box-f79da9aace992f8491d13d149114fff4.png`

You can test this by going to the doc in the product here: https://staging.one.nr/0BoQDDM1Qyv and manually replacing the broken image with the URL above.

jerelmiller (changes requested):

plugins/gatsby-plugin-generate-doc-json/utils/handlers.js
caylahamann: That actually gets handled in the in help xp, https://source.datanerd.us/nr1-dev-experience/nr1-help-xp/blob/394876aa74d7fbc699f333d018c2f311dda7f5e7/shared/components/ParsedDocsArticle/ParsedDocsArticle.js#L78

jerelmiller: Oh duh haha. I didn't realize it did it on that end. Thanks!

jerelmiller (approved):

PR #975, opened 2021-02-26
Fix: Release notes tiles and titles

austin-schaefer:
### Tell us why

* We're currently mis-categorizing hundreds of release notes, creating redundant categories.
* Some release notes categories also had naming that's inconsistent with our curent brand guidelines


austin-schaefer:
Nominating @jerelmiller as reviewer since I think he's spent the most time with these pages, but could be anyone.

jerelmiller (approved):
THANK YOU!!! This is SO much better

PR #988, opened 2021-02-26
Sort release notes nav as case insensitive

### LINKED ISSUE #974: Sort order bug in release note sidebar
### Description

The sort order in release notes sidebar has two issues:

* Sorts lower-case letters after upper-case letters, resulting in iOS agent and tvOS agent appearing erroneously at the bottom of the list.
    * 
image
* Does not ignore `.` which puts .NET at the top of the list
    * 
image

### Steps to reproduce

1. Go to http://localhost:8000/docs/release-notes
2. Look at the sidebar

jerelmiller:
Closes #974

### Tell us why

This PR updates the sorting of the release notes nav by case insensitive alpha. Also compares against word-only characters to ensure release notes such as `.NET` show up with the `n`s.

## Screenshots

Screen Shot 2021-02-26 at 12 14 21 AM


roadlittledawn (approved):
makes sense to me. :shipit:

PR #992, opened 2021-02-26
Updated the Learn about AI tile to point to a better place

paperclypse:
We found this incorrect link during the Docs team test.


jerelmiller (approved):
Thanks for fixing!

PR #996, opened 2021-02-26
Replacing an ugly tile icon with a better one

dbarnesbrownNR:
Wanted to replace icon for realsies but also just making sure I am getting the process right with some tiny changes.


bradleycamacho (approved):
Looks good!

PR #998, opened 2021-02-26
New guided install overview

paperclypse:
### Tell us why

The guided install overview doc.

I'm not super happy with how the button styles work. I'm thinking we may want to go with links in the short term, at least.

It's important this doc is in place before we go live on Monday, March 1. 

No

### DIFF
--- a/src/content/docs/full-stack-observability/observe-everything/get-started/new-relic-guided-install-overview.mdx
+++ b/src/content/docs/full-stack-observability/observe-everything/get-started/new-relic-guided-install-overview.mdx
@@ -4,8 +4,6 @@ tags:
   - Full-Stack Observability
   - Observe everything
   - Get started
-contentType: page
-template: basicDoc
 metaDescription: Our guided installation is a single command that you can use to discover and start monitoring integrations on your system. 
 ---
 
--- a/src/content/docs/full-stack-observability/observe-everything/get-started/new-relic-guided-install-overview.mdx
+++ b/src/content/docs/full-stack-observability/observe-everything/get-started/new-relic-guided-install-overview.mdx
@@ -1,0 +1,148 @@
+---
+title: New Relic guided install overview
+tags:
+  - Full-Stack Observability
+  - Observe everything
+  - Get started
+contentType: page
+template: basicDoc
+metaDescription: Our guided installation is a single command that you can use to discover and start monitoring integrations on your system. 
+---
+
+Instrument your systems and send telemetry data to New Relic with guided install.
+
+<ButtonLink
+  role="button"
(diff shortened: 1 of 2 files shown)


jerelmiller:
@paperclypse I'd love to get an idea of what you'd like to see with the buttons. I would agree, they are a sore spot when used in docs right now, so it would be great to do something about it.

paperclypse:
@jerelmiller I think if the buttons just had some slightly better padding (some on the left and right, and more consistent padding on the top and bottom) that would go a long way.

If they just matched what we've got on the old Docs site, I'd be happy with that: https://docs.newrelic.com/docs/full-stack-observability/observe-everything/get-started/new-relic-guided-install-overview

src/content/docs/full-stack-observability/observe-everything/get-started/new-relic-guided-install-overview.mdx
```diff
@@ -0,0 +1,150 @@
+---
+title: New Relic guided install overview
+tags:
+  - Full-Stack Observability
+  - Observe everything
+  - Get started
+contentType: page
+template: basicDoc
```
bradleycamacho, line 8: These aren't necessary.

src/content/docs/full-stack-observability/observe-everything/get-started/new-relic-guided-install-overview.mdx
```diff
@@ -0,0 +1,150 @@
...
+<ButtonLink
+  role="button"
+  to="https://one.newrelic.com/launcher/nr1-core.home?pane=eyJuZXJkbGV0SWQiOiJucjEtY29yZS5ob21lLXNjcmVlbiJ9&cards[0]=eyJuZXJkbGV0SWQiOiJucjEtaW5zdGFsbC1uZXdyZWxpYy5ucjEtaW5zdGFsbC1uZXdyZWxpYyIsImFjdGlvbkluZGV4IjoxfQ=="
+  variant="primary"
+>
+  Guided install
+</ButtonLink>
+
+If you're hosted in the EU:
+
```
bradleycamacho, line 23: I'm not a fan of how these render on the site. If you're going to have the "If you're hosted in the eu" line, I'd do a similar line for the other button.

src/content/docs/full-stack-observability/observe-everything/get-started/new-relic-guided-install-overview.mdx
bradleycamacho: Removed these in my commit below.

bradleycamacho:
@paperclypse This looks good! Two notes:

I removed the extra front matter from the doc. Basic docs don't need any front matter designating them as such, since the default is already `basicDoc`. Also, `contentType` and `template` are outdated, the correct front matter field is `type`

The buttons are a bit wonky. I would either remove the EU line, or add a line above the first button. If you do the second option, I would make sure it goes something like:

```
If you're hosted in the Us:
Button
//New line
If you're hosted in the EU:
Button
```

The blank line between the text and button throws me off a bit.

paperclypse:
@bradleycamacho Good call on the buttons. Thanks!

paperclypse:
@bradleycamacho I replaced the buttons with a tip callout and a couple of links.

PR #1015, opened 2021-02-26
Removed unnecessary pages

zstix:
## Description
Removes the unnecessary `new-relic-only` pages and related navigation.


jerelmiller (approved):

jerelmiller:
We may want to get sign off from a doc writer just to make sure we aren't losing anything here. I'm going to revoke my review for now.

PR #1021, opened 2021-02-27
Create GA announcement, delete beta banner (don't merge until GA)!

austin-schaefer:
### Tell us why

The site is GA now!

Can we set this PR to merge just before the new site goes live?


jerelmiller (approved):

zstix:
@austin-schaefer announcement banners only display when the current date is within the supplied date range. We should be alright to merge this whenever, but I'm happy to hold off if you would like.

src/announcements/docs-ga-banner.mdx
```diff
@@ -0,0 +1,10 @@
+---
+startDate: 2020-03-01
```
jerelmiller, line 2:
```suggestion
startDate: 2021-03-01
```

Just realized, this has the wrong year!

jerelmiller (changes requested):

jerelmiller (approved):
As @zstix said, we can merge this whenever. It won't go live until the `startDate`, so we are safe to merge it now.

austin-schaefer:
@jerelmiller @zstix : Well won't it go live at 12:01 am on March 1, hours ahead of scheduled GA?

jerelmiller:
@austin-schaefer You're right, that is true. This does use an ISO 8601 date format, so we _could_ in theory use a time as well here. Meh, let's just wait until Monday.

PR #1026, opened 2021-02-27
Add workloads tile home page

mmfred:
These are the changes to create the workloads tile on the home page.


mmfred:
My local build never showed this tile. But Bradley's did. So we'll see.

jerelmiller (approved):
This looks right to me. I'll checking it locally to make sure.

jerelmiller:
Looks good!

PR #1030, opened 2021-02-27
Move landing page template guide to style-guides

jerelmiller:
### Tell us why

We are in the process of deprecating the `new-relic-only` folder. A recent edit
has been made to the landing page template guide. This PR moves this to the new
`style-guide` folder so that we do not lose those changes.

This should fix the conflict in https://github.com/newrelic/docs-website/pull/1015


jerelmiller:
@paperclypse If you could take a look and make sure I put this in the right place, that would be great!

jerelmiller:
Closing as the page has already been added to the correct location

PR #1033, opened 2021-02-27
Plugin to check links

caylahamann:
Simple console log for checking links in our mdx files, not extensive but can be combined with checking our nav yaml and other file types.


plugins/gatsby-plugin-check-links/utils/check-nav-links.js
```diff
@@ -0,0 +1,61 @@
...
+    const file = fs.readFileSync(filepath, 'utf8');
+    const data = yaml.load(file);
+
+    return { links: [...new Set(extractLinks(data))], filepath };
+  } catch (e) {
+    console.error(`Unable to fetch ${filepath}:\n${e}`);
+  }
+};
+
+const codeRequests = async (nav) => {
```
zstix, line 28: If we find that this takes a while to run, we might want to consider running these requests in batches in parallel.

zstix (approved):
Looks good to me.

Is the intention that this is something we would run every so often on one of our machines?

caylahamann:
The intention is to have it as a utility to check links! We can always comment back in the plugin in the gatsby file to check it 

OR alternatively we could have it run every single time, since there's several changes a day (in IA, in quality checks) we may want to run it whenever we release. 

Could turn this into a github action that we could run manually whenever. I'm not sure, but I'll comment out it in the gatsby-config once it's done building so it doesn't slow down our builds even more so, and we'll have a quality check on the latest files :)

PR #1035, opened 2021-02-27
Force trailing slashes

jerelmiller:
### Tell us why

Force trailing slashes when creating pages.


jerelmiller:
Closing this as I believe the theme will end up doing the heavy lifting

PR #1039, opened 2021-03-01
Landing page and dummy doc

rhetoric101:
@jerelmiller, please not merge this because I have some questions for the docs reviewer. Thanks!

This PR includes a variety of work that isn't all related to the name of the branch:

* I updated the landing page template. It now has a link that send users to "Embed images" for details about adding icons.
* I removed "For more help" at the bottom of "html-template.mdx" since this is created automatically.
* I corrected some entries in "style-guide.mdx" that I stumbled across.
* I moved "update-left-navigation-pane.mdx" from the "manual-edits" directory to the main style-guide directory.
* I added icon information to "embed-images.mdx." Austin, if you look at this, I was not able to figure out where the product logos were listed (for example, logo-aws). I have locations for both the New Relic and Feather icons.
* I added the "dummy-docs.mdx" document for the style guide.
* I added navigation entries to "style-guide.yml" for my new style guide pages. It also looked like there were some broken links in various style guide sections, so I changed those. If those were corrected in another PR, we may have to merge these.


austin-schaefer:
Looking really good @rhetoric101 ! Pulled this down locally and it's all rendering properly. I really like the conversational style you're using, nice contrast to the slightly stiffer tone of older style guide docs. Also really nice, clear procedures. 

A few minor suggetions.

## Landing page template

* Should we refer to "feather" icons, or should this just refer to icons more generally (since we also have other icons like AWS, Azure, etc etc)
* Can we use <var> tags around the INSERT_HERE values in the pre blocks? 
* I didn't quite follow this bullet: "Insert a value for href that links to the target landing page. If the target landing page is index.html, you can just include the directory path with no filename."

## Update left-navigation pane

* I think the `src/nav` link in the first ¶ probably shouldn't be a relative link. Should instead link to https://github.com/newrelic/docs-website/tree/develop/src/nav ?
* Ditto the link to Insights in the second ¶. Also might want to use a different product since Insights is on its way out in favor of NR1 dashboards?

## Embed images

* Screenshots and inline images appear to be broken
* The links for "Docs site icons" didn't quite make sense to me (it's pointing at a js file. Should this be https://github.com/newrelic/docs-website/tree/develop/src/%40newrelic/gatsby-theme-newrelic/icons ?

## "Dummy docs"

I'm not actually sure if "dummy doc" is a standard industry term, or just something our team came up with at some point. I wonder if you can cut the preface and re-title this to something like, "Include a doc in multiple taxonomies" or "Include a doc in multiple menus" or similar?

PR #1043, opened 2021-03-01
Basic user restrictions enforced for original user model users

zuluecho9:
For users on original user model, we now are enforcing basic user restrictions. https://newrelic.atlassian.net/browse/DOC-6598


zuluecho9:
Closing to resubmit, hopefully without conflicts.

PR #1044, opened 2021-03-01
Basic user restrictions for original user model users

zuluecho9:
Basic user restrictions going in place for users on original user model. https://newrelic.atlassian.net/browse/DOC-6598


zuluecho9:
Closing cause I keep getting conflicts. Will resubmit later.

PR #1081, opened 2021-03-02
Renders inline jsx as children of component for collapsers and tech tiles

caylahamann:
## Summary 
This allows for titles for Collapsers and icons Tech Tiles to be rendered in the in-product help xp. Before this gets merged in, there will need to be some work on the side of the in-help-xp to handle the collapser titles as children and support for tech tiles in general. 

In the future (when react 17 is up and running in the product) we should move away from this and try to serve the mdx and use mdx in the product to render the components! but that's a thing for the future :)

### DIFF
--- a/codemods/utils/mdxast.js
+++ b/codemods/utils/mdxast.js
@@ -125,6 +125,7 @@ const containsImport = (tree, node) => {
 module.exports = {
   addAttribute,
   containsImport,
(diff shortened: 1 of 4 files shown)


jpvajda:
Zack is currently working on this..

LizBaker (approved):

PR #1106, opened 2021-03-03
Hero edit: Corrected two links in "Manage synthetic monitors via REST API"

rhetoric101:
Two links in this doc had problems: they were pointing to the same doc. I confirmed with Brian Peck that these two links should point to a lower section in the same doc called “Script API for scripted browser and API test monitors.”

Also, this commit included a bonus change that I didn't create for `src > data > whats-new-ids.json`.   @jerelmiller said I should commit this.


urbiz-nr:
Checked in my localhost, looks fine.

PR #1107, opened 2021-03-03
Generate JSON for release notes

### LINKED ISSUE #961: Provide a JSON endpoint for release notes data
a JSON endpoint for release notes will provide teams the ability to reference release note data in NerdGraph. In the old Docs world, a scrapper was pulling in agent release note data and making that data available in the Nerdgraph customer schema. So we'll want to provide a publically facing JSON endpoint that will provide similar data.

### References

- @erabug inquired about this as she wrote the code for the [scrapper](https://source.datanerd.us/agents/agent-releases)
- We currently offer [release notes data in XML format](https://docs-preview.newrelic.com/docs/release-notes/agent-release-notes/nodejs-release-notes/feed.xml)for our RSS feeds 💯 so the same data would be provided in JSON format.
- The new endpoint should be added to the WIKI and explaination of how it can be used should also be documented. 

### Use case(s)

[Agent grounds keeper pulls in this data](https://discuss.newrelic.com/t/groundskeeper-nerdpack/84168) seems to be using this[query](https://github.com/newrelic/nr1-groundskeeper/blob/main/nerdlets/groundskeeper-nerdlet/queries.js)

```
in the last week it looks like agent groundskeeper had 32 active users. so it's definitely an app some users find useful
but a very small % of our 1, 270 weekly app users over the same week
```

Support Innovation has a  Release Notes Parser.  Cade Conklin developed that against the docs site, and now needs to migrate it to work with the new Open Source docs  >  @cade-conklin

zstix:
## Description
~This can be used by [services](https://source.datanerd.us/agents/agent-releases) to fetch all release notes and [save them](https://s3.amazonaws.com/agent-releases/agent-releases.json) as they see fit.~

This exposes a JSON "endpoint" for all of the release notes for use in NerdGraph (no extra service and S3 bucket required!).

## Reviewer Notes
The JSON format is the same as the [previous](https://s3.amazonaws.com/agent-releases/agent-releases.json) JSON file:

```json
[
  {
    "agent": "java",
    "version": "1.0.0",
    "date": "2018-10-04"
  }
]
```

https://pr-1107.d1vx4w5vgyla4s.amplifyapp.com/api/agent-release-notes.json

This will require an update to NerdGraph (to point to this new JSON file).

## Related Issue(s)
* Closes #961


zstix:
This can be previewed here: https://pr-1107.d1vx4w5vgyla4s.amplifyapp.com/release-notes-1.json

gatsby-config.js
```diff
@@ -438,6 +438,39 @@ module.exports = {
...
+          }
+        }
+        `,
+        serializeFeed: ({ data }) =>
+          data.allMdx.nodes
+            .map(({ frontmatter, fields }) => ({
+              agent:
+                fields.slug !== '/docs/release-notes'
+                  ? fields.slug.split('release-notes/')[1].replace('-', '')
+                  : null,
```
zstix, line 466: This looks like it's `agent` for every result. We'll need to do a better job at identifying the agent type.

jerelmiller: That data should be available as the “subject” in frontmatter. Can you use that instead?

gatsby-config.js
```diff
@@ -438,6 +438,39 @@ module.exports = {
         nodesPerFeedFile: Infinity,
       },
     },
+    {
+      resolve: `gatsby-plugin-json-output`,
```
jerelmiller, line 442: Out of curiosity, why use the `json-output` plugin here instead of our `generate-json` plugin?

zstix: The previous script expects pagination, which we get for free from the `json-output` plugin. It looks like our plugin creates a single JSON file, which would be far too big for _all_ the release notes.

gatsby-config.js
zstix: Oh duh 🤦  Good call!

gatsby-config.js
jerelmiller: Got it. How are we planning to handle pagination? Because we can't build pages with query params, are we going to do some redirect magic in Amplify? Or modify the service that handles this?

zstix: This creates `release-notes-1.json`, `release-notes-2.json`, etc and python script handles crawling the pages. Ideally we would upload these files directly to s3 and remove the need for a python script to fetch these, but we don't currently have an IAM role with access to the necessary bucket.

jerelmiller: > This creates release-notes-1.json, release-notes-2.json

Totally understand that. Sorry, I should have been more clear what I'm asking. Let me try again 😄 

> Ideally we would upload these files directly to s3

This is really more of what I'm poking at. Is this "ideal" scenario going to happen right away?

If we release this change now, will the python script get updated to work with the new format, or will it continue to expect to paginate between pages using query params? In my mind, one of 3 things need to happen in order to get this working again.

1. Get the solution in place to upload this data directly to s3
2. Get the python script updated to use the new endpoint format `-1.json`, `-2.json` for pagination
3. Add redirect magic to amplify to rewrite `?page=2` requests to `-2.json` endpoints

Without any of these changes, we won't be able to source the data correctly. I was more curious on what we plan to do **now** once this change is merged.

zstix: My aim was to take the fastest, least disruptive, approach to solving the problem. Since we already had a python script in place, we had an engineer willing to update it, and we had a plugin set up to created paginated results in this format, I thought this was the best direction.

Since this PR isn't building at the moment, and this is turning out to _not_ be a quick fix, I'll start looking into what it would take to set this up with S3.

zstix: After speaking with Erika and Vince, we've decided to just supply one JSON file and have NerdGraph reach out to us directly. I've updated the code and the PR description to reflect this.

jerelmiller (approved):

Issue #945, opened 2021-02-24
Too much space between nested list items


jerelmiller:
### Description

Visit the following page: https://docs-preview.newrelic.com/docs/integrations/open-source-telemetry-integrations/opentelemetry/view-your-opentelemetry-data-new-relic#explorer

You'll notice the space between the list items is too large. Here is a screenshot:

<img width="903" alt="Screen Shot 2021-02-23 at 10 13 30 PM" src="https://user-images.githubusercontent.com/565661/108956416-80fe9400-7624-11eb-98ba-c0915f061b9c.png">

AubreyHewes:
@jerelmiller the docs-preview url quoted in this and other issues are a 404 ; 

e.g. (this issue) https://docs-preview.newrelic.com/docs/integrations/open-source-telemetry-integrations/opentelemetry/view-your-opentelemetry-data-new-relic#explorer

Removing the `-preview` gets the current doc...

jerelmiller:
Hey @AubreyHewes good catch! We were hosting our site at docs-preview up until yesterday when we launched our docs site public! As you mentioned, removing the `-preview` should get you to the right place. Thanks for mentioning that! We should probably do a quick audit on some of our open issues to make sure they link to the right place.

AubreyHewes:
@jerelmiller or a redirect from docs-preview; maybe easier that rewriting X issues ? ;-)

jerelmiller:
@AubreyHewes Haha good point! Normally I'd say yes, but we are repurposing docs-preview for some other things in the future, so the 404 will actually be the correct behavior. Growing pains 😬

jerelmiller:
Opening back up for now. After more discussion in #1110, we might opt for an update to our styles to account for this.

PR #1110, opened 2021-03-03
incorrect markdown caused p-tags (and expected style margins).. this removes it

AubreyHewes:
### Tell us why

Incorrect markdown in list caused a p tag, so also the p tag style (margin et al)

fixes https://github.com/newrelic/docs-website/issues/945 

[#gatsbyconf](https://github.com/newrelic/docs-website/issues?q=is%3Aissue+is%3Aopen+label%3A%23gatsbyconf)

### DIFF
--- a/src/content/docs/integrations/open-source-telemetry-integrations/opentelemetry/view-your-opentelemetry-data-new-relic.mdx
+++ b/src/content/docs/integrations/open-source-telemetry-integrations/opentelemetry/view-your-opentelemetry-data-new-relic.mdx
@@ -20,10 +20,8 @@ To get started with the explorer:
 1. Open [https://one.newrelic.com](https://one.newrelic.com), and go to **explorer**.
 2. Using the filter bar at the top, narrow your searches:
    1. Limit results to OpenTelemetry entities with the following:
-
       ![Screen capture showing how to filter for OpenTelemetry.](./images/explorer_filter_1.png "explorer_filter_1.png")
    2. After the OpenTelemetry filter, you can add another filter for your service based on the `service.name` attributes you've used. For example, to find `CartService`:
-
       ![Screen capture showing how to add a secondary filter for OpenTelemetry in explorer.](./images/explorer_filter_2.png "explorer_filter_2.png")
 3. Click the service you want to know more about, and if you need help understanding the data, see the explanations below.
 


jerelmiller (approved):
This is a really great solution to this problem! I'll make sure to communicate this with the teams that contribute content daily to make it easy to determine when best to apply the newline or not. Appreciate the contribution!

AubreyHewes:
@jerelmiller Hm, actually I dont think this actually fixes the problem... I was to eager.. It will have to be a styling fix for p in lists.. 
As the spacing in the screenshot is the margin of the p from the parent list item; this removes the spacing from the child item and the image....

:sheep:

jerelmiller:
Thats interesting! I actually tried this locally and it _seemed_ to do what I expected. Perhaps we should open up the other issue again?

That being said, I do see what you're saying and I think it might be good to try and remove that margin if the `p` tag is the "last child of type" so that nested lists don't end up with the gap as well.

Our global styles are controlled by [our theme](https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/packages/gatsby-theme-newrelic) so that we can share it among all of our sites. Perhaps [this CSS rule](https://github.com/newrelic/gatsby-theme-newrelic/blob/develop/packages/gatsby-theme-newrelic/src/components/GlobalStyles/GlobalStyles.js#L55-L57) needs to handle `:last-of-type` as well 🤔 (which I believe might also solve this issue).

AubreyHewes:
Yeah currently the theme uses p:last-child... last-of-type could be better

AubreyHewes:
The theme has no source map so I cant say where... but if you inspect the "Using the filter bar at the top, narrow your searches" you will see it has a margin-bottom... if you inspect the first item (without a sub list) it has no margin bottom.

jerelmiller:
That sounds accurate!

jpvajda:
@AubreyHewes thanks so much for the contribution!! Do you mind completing this form so we can send you a gift from New Relic. 🎁  https://forms.gle/EQNmAV9X3UsMxHJy7

AubreyHewes:
@jpvajda Yeah but the fix was not the fix... and should be reverted is still ongoing (@ theme)

jpvajda:
@AubreyHewes sure thing. Just when the merge is approved in our theme and deploy just send us the info so we can get you the 🎁 .

AubreyHewes:
@jpvajda form does not work for international postcodes ;-) (error: must be a number)

jpvajda:
@AubreyHewes sorry about that. It's fixed now, will accept strings. 😄

AubreyHewes:
@jpvajda while you are about it... :sheep: email is required twice (at the start and nearly the end) x

jpvajda:
@AubreyHewes thanks for testing our form. I removed the duplicate field. 🎉

PR #1137, opened 2021-03-03
Bring in final style guide docs (I hope)

paperclypse:
### Tell us why

I'm not 100% sure, but I think this is the final batch of style guide docs we need to bring over.

I definitely need a second pair of eyes on this, because it's a lot of changes: moving files around and a lot of updates to the navigation file.


austin-schaefer:
@paperclypse : Not grabbing for a full review, but do we need redirects for the moved/removed files?

bradleycamacho:
Looks good just scanning. Building locally now.

bradleycamacho (approved):
Looks good to me, good job @paperclypse! Let me know if you decide to add redirects as @austin-schaefer recommends above - I'll take another look if so,

paperclypse:
@austin-schaefer We shouldn't need redirects... but I should probably add them, just to be on the safe side.

bradleycamacho:
@paperclypse Seems like the redirects work for me! I think this is ready to merge.

PR #1146, opened 2021-03-04
fix: compatibility link in win installation page

Noly:
Compatibility and requirements link was broken.


urbiz-nr:
Checked locally and it works.

PR #1151, opened 2021-03-04
feat:  add additional permissions to Cloud Integrations

morecar:
### Tell us why

Updating the Cloud Integrations documentation with some missing permissions. For more information #Beyond


bradleycamacho:
@morecar Thanks for adding these missing permissions! This looks good to me, so I'm merging now! This should be up with our 9am build :)

PR #1155, opened 2021-03-04
Adding missing agent RN redirects

austin-schaefer:
### Tell us why

Someone reported in #documentation that the old root/index pages for agent release notes are missing redirects. 

I found the new URLs by playing around in the old Drupal-based docs-dev site.

Opening as draft because I still need to test the redirects locally.


bradleycamacho:
@austin-schaefer: Note that redirects on a local build will throw a 404 error. To test them, you have to sit on that 404 error page for ~1 minute until it redirects you.

austin-schaefer:
Checked all new redirects locally and all working.

PR #1157, opened 2021-03-04
Sentence case attribute dictionary

austin-schaefer:
Please test this locally! I tested locally but it's a big enough change I think a second set of eyes would be goodl.

### Tell us why

Taxonomy names should be lower case.


bradleycamacho (approved):
Looks good to me @austin-schaefer! We might want to scan through the entire taxonomy at some point to ensure all titles are lowercase.

Feel free to merge once the checks complete!

PR #1170, opened 2021-03-04
Provides wrapper around buttons in mdx docs

### LINKED ISSUE #802: add a little bit more padding around buttons

I think it would be great if we had a little bit of padding around the bottom of our buttons. As it is, the text and button are pretty jammed up together there.

button padding screenshot

Let's add a little padding there under our buttons.

It's a minor thing that nags at me a little every time I see it. :)

caylahamann:
## Summary 
We want to make all buttons center on the docs site! If there's more than one button, we will want to have them both centered in a row. This PR creates two components: `MDXButton` which is a wrapper around the `Button` component and `MDXButtonGroup` which creates a `ButtonGroup` component to be used to wrap two buttons. This PR also changes all the buttons to match this pattern! The details of this was specified by austin in [this slack thread](https://newrelic.slack.com/archives/C017ZNASGUX/p1614889222478200) to handle these two cases. 

### Before
Screen Shot 2021-03-04 at 9 33 57 PM

### After
Screen Shot 2021-03-04 at 9 33 42 PM

### Before 
Screen Shot 2021-03-04 at 9 30 55 PM

### After (requiring a `ButtonGroup` to be wrapped around the two)
Screen Shot 2021-03-04 at 9 35 43 PM

### DIFF
--- a/src/components/MDXButton.js
+++ b/src/components/MDXButton.js
@@ -1,0 +1,48 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css } from '@emotion/core';
+import { Button } from '@newrelic/gatsby-theme-newrelic';
+
+const ALIGNMENTS = {
+  LEFT: 'left',
+  CENTER: 'center',
+  INLINE: 'inline',
+};
+
+const styles = {
+  alignment: {
+    [ALIGNMENTS.CENTER]: css`
+      justify-content: center;
+    `,
+    [ALIGNMENTS.LEFT]: css`
+      justify-content: flex-start;
+    `,
+  },
+};
+
+const MDXButton = ({ alignment = 'center', ...props }) => {
+  if (alignment === ALIGNMENTS.INLINE) {
+    return <Button {...props} />;
+  }
+
+  return (
+    <div
+      css={css`
+        display: flex;
+        &:not(:last-child) {
+          margin-bottom: 1rem;
+        }
+        ${styles.alignment[alignment]}
+      `}
+    >
+      <Button {...props} />
+    </div>
+  );
+};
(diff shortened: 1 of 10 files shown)


jerelmiller:
This is a really solid approach to this problem. I'd be curious, with the default being `center`, how would this look on a page like this? https://docs.newrelic.com/docs/agents/python-agent/getting-started/introduction-new-relic-python

We may consider defaulting to `inline`. Thoughts?

zstix (approved):

caylahamann:
@jerelmiller I totally agree - but I think right now there's a few more instances where `center` might apply. I'm mostly referring to the requests from the ticket and specifically this comment from @austin-schaefer: 
https://github.com/newrelic/docs-website/issues/802#issuecomment-786443024
> The buttons should also be centered in the column—right now they float to the left awkwardly. This is something we should prioritize soon—it's an increasingl common pattern across the site.

@austin-schaefer - do you have a preference for what the default should be?

jerelmiller:
@caylahamann definitely understand you there, but I know this change will be applied to all buttons/button links on the whole site, which would include pages with multiple buttons inline. Because this new behavior uses `center` as its default, and because that wraps a div around a single button, this means it will move pages with multiple inline buttons to their own lines. That may or may not be desirable. 

I just want to make sure that by solving one problem, we aren't creating another.

caylahamann:
Yeah I totally agree!! I definitely don't want this to cause a bunch of pages to look wonky - I just wanna figure out which one is more preferable (to automatically center buttons and specify inline, or automatically keep them as is and specify center). From my understanding most of the buttons should be centered, with a few being inline (like the ones you pointed out above). Let me slack austin to make sure this is visible so he can give his input!

zstix (approved):

PR #1174, opened 2021-03-04
Sorts release notes by version number when there's multiple on the same day

### LINKED ISSUE #910: when more than 1 release note is posted in a single day, we aren't sorting them correctly
Check the timestamp on the python release note example where they are out of order. We want to be sure that 2 or more releases post in the correct order based on their version number

Screen Shot 2021-02-18 at 5 21 28 PM

caylahamann:
Evaluates the version number to a number value, and then sorts based the version number
closes #910

### DIFF
--- a/src/templates/releaseNoteLandingPage.js
+++ b/src/templates/releaseNoteLandingPage.js
@@ -81,43 +81,50 @@ const ReleaseNoteLandingPage = ({ data, pageContext, location }) => {
 
             return (
               <Timeline.Item label={date} key={date}>
-                {posts.map((post) => {
-                  const excerpt = getBestGuessExcerpt(post.mdxAST);
+                {posts
+                  .sort((postA, postB) => {
+                    return (
+                      Number(postB.frontmatter.version.replaceAll('.', '')) -
+                      Number(postA.frontmatter.version.replaceAll('.', ''))
+                    );
+                  })
+                  .map((post) => {
+                    const excerpt = getBestGuessExcerpt(post.mdxAST);
 
-                  return (
-                    <div
-                      key={post.version}
-                      css={css`
-                        margin-bottom: 2rem;
-
-                        &:last-child {
-                          margin-bottom: ${isLast ? 0 : '2rem'};
-                        }
-                      `}
-                    >
-                      <Link
-                        to={post.fields.slug}
+                    return (
+                      <div
+                        key={post.version}
                         css={css`
-                          display: inline-block;
-                          font-size: 1.25rem;
-                          margin-bottom: 0.5rem;
+                          margin-bottom: 2rem;
+
+                          &:last-child {
+                            margin-bottom: ${isLast ? 0 : '2rem'};
+                          }
                         `}
                       >
-                        {post.frontmatter.title
-                          ? post.frontmatter.title
-                          : `${subject} v${post.frontmatter.version}`}
-                      </Link>
-                      <p
-                        css={css`
-                          margin-bottom: 0;
-                        `}
-                      >
-                        {excerpt.slice(0, EXCERPT_LENGTH)}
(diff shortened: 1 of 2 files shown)


zstix (approved):

src/templates/releaseNoteLandingPage.js
```diff
@@ -81,43 +81,50 @@ const ReleaseNoteLandingPage = ({ data, pageContext, location }) => {
...
-                          margin-bottom: ${isLast ? 0 : '4rem'};
-                        }
-                      `}
-                    >
-                      <Link
-                        to={post.fields.slug}
+                {posts
+                  .sort((postA, postB) => {
+                    return (
+                      Number(postB.frontmatter.version.replaceAll('.', '')) -
```
jerelmiller, line 87: There are a few release notes that don't have versions attached to them (see [this doc](https://docs.newrelic.com/docs/release-notes/infrastructure-release-notes/kubernetes-integration-release-notes/deprecation-notice-kubernetes-v17-or-lower) or any of the "Cloud integrations" release notes). Does this blow up for those pages?

jerelmiller (commented):
Spotted a potential for pages to blow up. Could you verify?

src/templates/releaseNoteLandingPage.js
jerelmiller: Indeed the build fails for this reason 😄

caylahamann: ah I did not realize that, let me add a conditional to this

src/templates/releaseNoteLandingPage.js
```diff
@@ -81,43 +81,60 @@ const ReleaseNoteLandingPage = ({ data, pageContext, location }) => {
...
+                    );
+                    const versionB = parseInt(
+                      postB.frontmatter.version
+                        ?.replace('-beta', '')
+                        ?.replace(/\./g, '')
+                    );
+                    if (!(isNaN(versionA) && isNaN(versionB))) {
+                      return versionB - versionA;
+                    }
+                    return 0;
```
jerelmiller, line 99: Couple things here that I think could help you clean this up just a little bit. I'll go through them piece by piece.

If we are comparing 2 posts without a version, we should avoid sorting them (return value `0`). We can catch that early on and return early. By doing this, you can avoid doing the whole `isNan` dance on line 96.

```js
if (!postA.frontmatter.version || !postB.frontmatter.version) {
  return 0
}
```

2nd of all, I see you're doing `replace` twice here to try and get rid of any unwanted characters in the version name. While what you've got works for now, if somebody came along and added an `-alpha`, this would break. Since what you care about here is comparing 2 integers, I'd just remove any non-digits from the string, which you can do in a single regex:

```js
frontmatter.version.replace(/\D/g, '')
```

Lastly, whenever you use `parseInt`, its a good practice to provide the radix argument so that you don't accidentally parse the string as a base of anything other than 10.

```js
parseInt(..., 10) // <-- forces it to parse in base 10
```

With this, you can simplify to returning what you have on line 97:

```js
return versionB - versionA
```

jerelmiller (changes requested):
I'm only blocking because I'd like to see the last 2 points of my comment changed before approving to avoid any future bugs.

src/templates/releaseNoteLandingPage.js
caylahamann: Sweet! thank you, these are all great suggestions :) and yes I knew there was someway to accomplish what I wanted with regex, but somehow regex just doesn't jive with my brain. Thank you and I'll make these changes after I get the swiftype PR up :)

jerelmiller (approved):
This is a GREAT looking PR! Thanks for making that change 😄

PR #1196, opened 2021-03-05
Update rules-limits-alerts.mdx

wrekone:
Update condition name max length to 128 characters

# Description

The condition name character limit is supposed to be 128. This brings the documentation in line with the limit enforced by the UI and the backend API.


mmfred:
Hi @wrekone! Thanks for your fix. Can you help us verify this change? Steps to check this change or a screenshot would be great. Thank you!

wrekone:
@mmfred You can verify this limit by starting the process of creating an NRQL Alerts condition, entering as many characters as the name will allow, then counting those characters using `echo -n "{the string you entered in the name field}" | wc -m` at the command line.

See AINTER-7716 for additional details.

mmfred:
@wrekone - Thanks for letting us know - really helpful!

PR #1198, opened 2021-03-05
Terraform configuration for alerts and synthetics

zstix:
## Description
We need to adjust which account owns the browser agent, sytnethetics monitor, and related alert conditions. Unfortunately, entities and policies can not be moved between accounts - they need to be recreated. I'd like to take another shot at configuring this stuff with _Terraform_ and automating the application of it using [atlantis](https://www.runatlantis.io/).

## Reviewer Notes
This is not a high-priority item, but it would be nice to get it set up sometime in the next two weeks.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -30,11 +30,24 @@ data "newrelic_entity" "browser" {
   }
 }
 
+resource "newrelic_synthetics_monitor" "docs_site_monitor" {
+  name      = "docs.newrelic.com"
+  type      = "SIMPLE"
+  frequency = 5
+  status    = "ENABLED"
+
+  # Portland, London, Hong Kong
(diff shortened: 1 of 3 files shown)


terraform/main.tf
```diff
@@ -0,0 +1,61 @@
+terraform {
+  required_version = "~> 0.13.0"
+
+  required_providers {
+    newrelic = {
+      source  = "newrelic/newrelic"
+      version = "~> 2.12"
```
jpvajda, line 7:
```suggestion
      version = "~> 2.21"
```

jpvajda: bumped this to the most current TF version so we can leverage dashboard apis... 🥳 https://github.com/newrelic/terraform-provider-newrelic/releases/tag/v2.21.0

jpvajda (commented):
just suggested a bump in TF provider version

zstix:
We've set this up manually for now. I'm going to close the PR, but keep the branch open.

PR #1199, opened 2021-03-05
Add meta tag to disable swiftype robots on auto-generated localized pages

### LINKED ISSUE #1191: English homepage search generates Japanese search results
### Description

I just noticed today some new behavior: If I search starting from an English language page, the results include Japanese listings. The search result links take me to the Japanese site.

### Steps to reproduce

1. Go to https://docs.newrelic.com/ (English site)
2. In the search bar, enter "Kubernetes" (this also worked for "Prometheus")
  NOTE: The results include Japanese characters.
3. Click any of the results, which take you to the Japanese site.

### Expected behavior

I would expect to get English search results if I start from the English site.

### Related Issue

#1204 

Due to a few issues with language tags on pages and a site map problem, we were having issues rendering pages in the product help, to fix it, we black listed the jp site in Swiftype, until we fixed those other issues, in doing this the current problem is persisting search terms in Japanese still need to be enter in English on the JP site

Once we resolve the issues mention above we can remove the JP black list entry from swiftype

## Eng work
- We need to differentiate jp namespaced docs pages from english namespaced docs pages via the swiftype meta tag `type`. We'll do that via the theme, because we can apply that globally and it already has the logic in place to work with locales.
- Once that is in the theme, we need to bump the theme version for each site to take advantage of that, docs being first because it's the only site with translations available right now.
- Once the relevant meta tags are in place for jp namespaced pages, we need to remove the crawl rule in swiftype to allow indexing pages with `/jp/docs/` in the url. 
- Once swiftype indexes all translated pages under their new `type`, we can update the theme search to account for searching with japanese characters against `docs-jp`. See caveat below.

### Caveats
If a locale namspaced page only has English, we tell swiftype to not index it. However, a few pages have some translations but are mostly in English (ex: `jp/attribute-dictionary`, `jp/whats-new`). So instead of just adding `docs-jp` to the list of site `type` values to query against, we may need to add some affordance to the search UI to account for that and avoid what will look like duplicate records when a user searches using English.

The work to bump the theme is straight forward on the docs site. The develop site requires some more refactor because it wasn't using the theme's SEO component. And I assume opensource will require similar if not more effort to bump the theme, since it's on a fairly old version.

### LINKED ISSUE #1195: Search engine results are returned for languages other than english

caylahamann:
closes #1191  after it is deployed and we trigger a recrawl. 

### Summary 
Disables swiftype robots by a meta-tag when the page generated automatically rather than from a localized mdx page. According to the [documentation](https://swiftype.com/documentation/site-search/crawler-configuration/meta-tags#swiftbot) when added to the page this will override all other swiftype meta tags on the page (which is what we want, so it doesn't really matter that if in the SEO component there's any ones being added or elsewhere). This will make sure that when a user is searching in Japanese they will get a page that has actual japanese content on it.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -271,6 +271,7 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
       createPageFromNode(i18nNode || node, {
         prefix: i18nNode ? '' : locale,
         createPage,
+        disableSEO: !i18nNode,
       });
     });
   });
@@ -329,7 +330,7 @@ exports.createResolvers = ({ createResolvers }) => {
 
 exports.onCreatePage = ({ page, actions }) => {
   const { createPage, deletePage } = actions;
-  const oldPage = Object.assign({}, page);
+  const oldPage = { ...page };
 
   if (page.path.match(/404/)) {
     page.context.layout = 'basic';
@@ -372,7 +373,10 @@ const createLocalizedRedirect = ({
   });
 };
 
-const createPageFromNode = (node, { createPage, prefix = '' }) => {
+const createPageFromNode = (
+  node,
+  { createPage, prefix = '', disableSEO = false }
+) => {
   const {
     fields: { fileRelativePath, slug },
   } = node;
@@ -398,6 +402,7 @@ const createPageFromNode = (node, { createPage, prefix = '' }) => {
         fileRelativePath,
         slug,
         slugRegex: `${slug}/.+/`,
+        disableSEO,
       },
     });
   }
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -330,7 +330,7 @@ exports.createResolvers = ({ createResolvers }) => {
 
 exports.onCreatePage = ({ page, actions }) => {
   const { createPage, deletePage } = actions;
-  const oldPage = { ...page };
+  const oldPage = Object.assign({}, page);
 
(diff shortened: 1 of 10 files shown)


gatsby-node.js
```diff
@@ -329,7 +330,7 @@ exports.createResolvers = ({ createResolvers }) => {
 
 exports.onCreatePage = ({ page, actions }) => {
   const { createPage, deletePage } = actions;
-  const oldPage = Object.assign({}, page);
+  const oldPage = { ...page };
```
caylahamann, line 333: my linter automatically made this change ?? I will revert it haha

gatsby-node.js
```diff
@@ -271,6 +271,7 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
       createPageFromNode(i18nNode || node, {
         prefix: i18nNode ? '' : locale,
         createPage,
+        disableSEO: !i18nNode,
```
caylahamann, line 274: originally I had named this because I thought we might wanna disable all SEO for these pages, but after thinking about it I realized that might not be the best option (Google has a way better way for organizing these pages based on locale, so I would want to interrupt that). Maybe, `disableSwiftype` or `disableSwiftypeIndex`?

caylahamann: changed it :)

zstix (approved):

PR #1201, opened 2021-03-05
Added guided install callout to two infrastructure docs

paperclypse:
### Tell us why

We want to make sure Infrastructure users get the word about our new guided install feature.


barbnewrelic (approved):
Excellent!

PR #1202, opened 2021-03-05
DOC-6317 Added guided install quick start section to supported OHIs (on-host integration)

paperclypse:
### Tell us why

We want to get the word out about our new guided install. This PR adds a tip callout with links to the guided install on all of the supported OHIs.


barbnewrelic (approved):
Thx for adding this important callout to all these docs!

PR #1226, opened 2021-03-08
Adding the quick start guided install section and tip

paperclypse:
### Tell us why

The new guided install is a great way to install the infrastructure agent. Seems like we should let people know about it!


bradleycamacho (approved):
This looks great!

Feel free to merge as it is, but I _think_ buttons are styled to be decent looking now. Maybe look into that?

paperclypse:
@bradleycamacho Yeah, I'm definitely going to roll out the button option sometime this week :) Thanks for reviewing this so quickly!

PR #1238, opened 2021-03-08
Update New Relic One link

lixiong-intelex:
### Tell us why

Current New Relic One link is a relative path, leading to a 404 page. So change it to full path.

None

No


jpvajda (approved):
LGTM

PR #1256, opened 2021-03-09
Ensure existing redirects are preserved

austin-schaefer:
### Tell us why

We should ensure we're not losing old redirects when we delete a doc.


urbiz-nr:
Looks good and ready to merge.

PR #1263, opened 2021-03-10
Fixing broken embedded youtube videos

paperclypse:
### Tell us why

The embedded YouTube videos were broken when I used the standard YouTube markdown.


jerelmiller:
@paperclypse we forgot to mention, those "whats new" posts are regular markdown, not MDX, so you aren't able to use the `Video` component like you can on other pages.

We opted for regular markdown here since these posts are shared with the platform. We wanted to make it easy to generate basic HTML without needing to add room for special components in the platform. The fix you have in this PR is perfect for videos. 

Apologies for not communicating this better/sooner!

PR #1277, opened 2021-03-10
Remove unneeded escape characters from NRQL function doc collapsers

jerelmiller:
### Tell us why

As I was looking at [this NRQL doc](https://docs.newrelic.com/docs/query-your-data/nrql-new-relic-query-language/get-started/nrql-syntax-clauses-functions/), I noticed some of the collapsers included a `\` before the optional arguments where `[` was used. I believe this was due to our migration process and something that we forgot to remove when moving it to the `title` attribute. This PR removes those stray `\` characters to provide better clarity so someone doesn't assume this is part of the NRQL itself.


bradleycamacho:
Looks good to me! Feel free to merge whenever @jerelmiller

bradleycamacho (approved):

jerelmiller:
Thanks @bradleycamacho!

dbarnesbrownNR:
@jerelmiller @bradleycamacho I was just this afternoon talking with Austin about how there is some weird stuff goin on in that doc, so thank you for this! I have a few more checks to do but this is one sweep item I don't need to scan for now. :)

PR #57, opened 2020-09-10
Prototype Script To Fetch Docs & Create MDX Files

zstix:
## Description
A small script that gets a JSON representation of different content types from the docs site and creates `mdx` files for each. The implementation has a lot of gaps, but it should serve as a starting point and can help us think through upcoming hurdles.

The general flow for this script is:
1. Get the docs. We're using `node-fetch` library to help make the request.
2. Create the directory structure. Turn the URL path into a directory structure (excluding the slug).
3. Create a file for each doc.
4. Create frontmatter based on content type. We have some _very_ basic fields based on the proposed MDX files.
5. Add the content to the file.
6. 🎉 

## Related Issue(s)
* Closes #52

PR #1291, opened 2021-03-11
fix: set nri-jmx defaults for user and pass to empty string

brushknight:
### Tell us why

Defaults for user and pass for nri-jmx were rolled back to `""` in this PR [#57](https://github.com/newrelic/nri-jmx/pull/57)

Add any context that will help us review your changes such as testing notes,


zuluecho9:
@brushknight Hi. Wanted to make sure you thought it added value to give the default of "" versus just leaving it out entirely. I wondered if most people would assume the default was empty if we just left off anything about the default entirely. What do you think?

PR #1295, opened 2021-03-11
Only run build in PR validation for merges to develop, not main

### LINKED ISSUE #1209: Prevent double build on release PRs
### Description

When opening release PRs from `develop` -> `main`, both the action to build and cache the site + the PR validation workflow is run. This results in 2 builds for any release PR. Seeing as we are already building the site when caching the merges, I'm not sure we need to also build the site for the PR. This can slow down release PRs because depending on when the validate PR workflow is run, it might add quite a bit of time before a PR can be merged.

Let's try and consolidate to a single build to allow us to merge the release PRs a bit quicker.

LizBaker:
### Tell us why

Currently we run a build on PRs merging into develop and again when releasing develop to main. The build from develop -> main is unnecessary and adds a lot of time to releases. This PR adjusts the action so the build job is only run on PRs to develop

Closes #1209

### DIFF
--- a/.github/workflows/validate-pr.yml
+++ b/.github/workflows/validate-pr.yml
@@ -13,7 +13,7 @@ env:
 jobs:
   build:
     name: Gatsby Build
-    if: github.ref == 'refs/heads/develop'
+    if: github.base_ref == 'develop'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


zstix (approved):
This looks good to me!

I was on the fence between preferring this approach vs running it on all PRs _except_ into `main`, but I think this is probably the right path.

LizBaker:
@zstix ahh yeah I was wondering about that, I do feel like this catches the most important case. if we have a different branch flow for content later we can always tweak things

PR #1303, opened 2021-03-11
Fixing broken link for AI decisions doc

zuluecho9:


urbiz-nr:
Checked locally, it works.

Issue #1204, opened 2021-03-05
search terms in Japanese need to be enter in English on the JP site.


jpvajda:
Due to a few issues with language tags on pages and a site map problem, we were having issues rendering pages in the product help, to fix it, we black listed the jp site in Swiftype, until we fixed those other issues, in doing this the current problem is persisting 

`search terms in Japanese still need to be enter in English on the JP site` 

Once we resolve the issues mention above we can remove the JP black list entry from swiftype

<img width="342" alt="Screen Shot 2021-03-05 at 4 31 27 PM" src="https://user-images.githubusercontent.com/1395158/110185736-653b8080-7dd0-11eb-8fd1-3b4bb64ef812.png">

jpvajda:
We should also check if  the languages specific search will rank higher in the results  based on the language site they are on.

jpvajda:
Will be resolved by #1191

PR #1311, opened 2021-03-12
Set swiftype type via theme's updated SEO component

roadlittledawn:
## Description
Bumps theme version to use updated `<SEO>` theme component that sets swiftype meta tag based on `siteMetadata.siteUrl`. Supports setting separate `type` value for localized pages.

## Related issues
- https://github.com/newrelic/docs-website/issues/1204


zstix (approved):
:tada: Looks good to me!

PR #1323, opened 2021-03-12
DOC-6627 Creating a variety of guided install links/buttons

paperclypse:
### Tell us why

The goal of this work is to create a variety of buttons and links styles so that we can collect some data on which one is the most effective at sending readers to the guided install UI.

I'd love to narrow the six options I've presented down to three, so please be opinionated. :) 

Also, if you have a great idea about an alternative to what I've done, please feel free to suggest it.


bradleycamacho:
@paperclypse Looks good to me!

PR #1345, opened 2021-03-16
Removed broken RSS feed link

dbarnesbrownNR:
Just a very small edit to remove a bullet linking to RSS feed, which is not currently supported. No peer or tech review needed.


dbarnesbrownNR:
This one was low hanging fruit, merging now.

Issue #1342, opened 2021-03-16
Issue: Introduction to AWS integrations


jsbnr:
The first link on page to "AWS Integrations list" takes you to a random FSO overview page.

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/integrations/amazon-integrations/get-started/introduction-aws-integrations/
* **OS**: Unknown
* **Browser**: Unknown
* **Device**: Unknown

urbiz-nr:
Fixed in #1365 .

PR #1365, opened 2021-03-17
Fixed link from #1342

urbiz-nr:
Fixes AWS integrations link in intro doc.


urbiz-nr:
Tested locally, looks good.

Issue #1367, opened 2021-03-17
Issue: Introduction to the Log API


jsbnr:
### Tell us what you need
The link to Payload limits in the first set of bullet points takes you to the Event API limits, but the limits are actually detailed lower down on the same page (and are different)

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/logs/log-management/log-api/introduction-log-api/
* **OS**: Mac OS 10.16
* **Browser**: Firefox 86.0
* **Device**: Unknown

urbiz-nr:
Fixed in #1368.

PR #1368, opened 2021-03-17
Updated link to payload limits section

urbiz-nr:
Fixing #1367.


urbiz-nr:
Tested locally, works fine.

PR #1377, opened 2021-03-17
Update ssl-or-connection-errors-java.mdx

XiXiaPdx:
call out specifically this hard to reason about exception


dbarnesbrownNR (commented):
Looks like the h2s are appearing as repetitions of Problem and Solution headers in the right nav, so I want to try to fix that up before merging. Placing this on hold until I or another TW can tinker a bit.

dbarnesbrownNR:
Moving this back to to do column as I didn't have time to get to it when I was heroing. This one needs to either have the repetitious problem/solution headers fixed or be split into a separate troubleshooting page, at the discretion of the writer who takes it.

urbiz-nr:
@dbarnesbrownNR - what do you think of this version?

urbiz-nr:
Thanks @dbarnesbrownNR! Merging.

PR #1395, opened 2021-03-17
Fix indentation for Java yaml infinite_tracing config

jasonjkeller:
Indentation was wrong for the `infinite_tracing` config yaml example which made it look like it was supposed to be nested under the `distributed_tracing` config stanza.


urbiz-nr:
Tried locally, looks good.

PR #1399, opened 2021-03-17
DUE 3/18 BY 9AM: Added root cause analysis what's new

bradleycamacho:
Creates What's New post that needs to be live by 9am tomorrow morning.


dbarnesbrownNR (approved):
Looks good, made one tiny edit but this is good to go now!

Issue #1392, opened 2021-03-17
Issue: Upload source maps via API


CherylFrankenfield:
### Tell us what you need

* Can we edit the New Relic-owned NPM package's documentation to clarify that it doesn't support changing the path for the API to the EU region? Link to: https://www.npmjs.com/package/@newrelic/publish-sourcemap

### Anything else you'd like to share?

We had a customer reach out about this request. Thank you again! 

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/browser/new-relic-browser/browser-pro-features/upload-source-maps-api/
* **OS**: Mac OS 10.14.6
* **Browser**: Chrome 88.0.4324.192
* **Device**: Unknown

CherylFrankenfield:
Link to ticket : https://newrelic.zendesk.com/agent/tickets/450083

urbiz-nr:
Info needs to go to README file, closing this and deleting branch.

PR #1405, opened 2021-03-18
Added path info in callout

urbiz-nr:
See #1392.


urbiz-nr:
Info needs to go in the README file.

PR #1406, opened 2021-03-18
DOC-6619 Admin portal and user mgmt changes

zuluecho9:
This is set to go out the morning of March 18th. I may be away from keyboard when it does but I've told them to ping docs hero. It's fairly important so if we can make an attempt to merge to main when it goes out, that'd be awesome.


paperclypse (approved):
Looks good.

paperclypse:
@zuluecho9 asked me to merge this in.

PR #1423, opened 2021-03-18
Add 3.0.40 CPM release notes

cleon-nr:
### Tell us why

Releasing a new CPM version

N/A

No

### DIFF
--- a/src/content/docs/release-notes/synthetics-release-notes/containerized-private-minions-release-notes/containerized-private-minion-cpm-3040.mdx
+++ b/src/content/docs/release-notes/synthetics-release-notes/containerized-private-minions-release-notes/containerized-private-minion-cpm-3040.mdx
@@ -6,5 +6,5 @@ version: 3.0.40
 
 ### Fixes
 
-* Sticky bit is now correctly assigned for /tmp mounted volume on Kubernetes deployment
-* Allows setting Selenium timeouts to a value of 0 for ScriptTimeout, ImplicitTimeout, and PageLoadTimeout
+* Sticky bit is now correctly assigned for `/tmp` mounted volume on Kubernetes deployment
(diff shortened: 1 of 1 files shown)


src/content/docs/release-notes/synthetics-release-notes/containerized-private-minions-release-notes/containerized-private-minion-cpm-3040.mdx
```diff
@@ -0,0 +1,10 @@
+---
+subject: Containerized private minion (CPM)
+releaseDate: '2021-03-18'
+version: 3.0.40
+---
+
+### Fixes
+
+* Sticky bit is now correctly assigned for /tmp mounted volume on Kubernetes deployment
```
urbiz-nr, line 9: I'm not sure I understand this sentence, @cleon-nr. Could you clarify it?

cleon-nr: This is in reference to a Linux permission option that's being set on a mounted `/tmp` directory when the CPM is deployed on Kubernetes. We have an initContainer that recursively changes some permissions on these mounted subpaths and the "sticky bit" wasn't being applied. So going from `777` to `1777` is mainly what this is referring to. The `1` referring to the "sticky bit" as described below.

> The restricted deletion flag or sticky bit is a single bit, whose interpretation depends on the file type. For directories, it
> prevents unprivileged users from removing or renaming a file in the directory unless they own the file or the directory; this is called the restricted deletion flag for the directory, and is commonly found on world-writable directories like /tmp. For regular files on some older systems, the bit saves the program's text image on the swap device so it will load more quickly when run; this is called the sticky bit.

dbarnesbrownNR (approved):
Updated to fix the clarity issue @urbiz-nr asked about. All good!

PR #1427, opened 2021-03-18
Update create-nrql-alert-conditions.mdx

jaymcgrath:
### Tell us why

Add histogram to the list of things from which you can't make a NRQL alert. We just added validation for this to the NRQL alert condition builder and are directing folks to this doc in the error message.

I changed the column header to element from "clause" since this covers both aggregation functions and clauses.


dbarnesbrownNR (approved):
I fixed an extra space and committed that change, otherwise all good!

PR #1429, opened 2021-03-18
Set Insights to not display in the main nav

paperclypse:
It doesn't deserve such top billing. :)


dbarnesbrownNR:
Trying to preview this on local site but not sure if I'm previewing the right thing. Let's discuss tomorrow when we sync. :)

dbarnesbrownNR (approved):
Looks good!

PR #1450, opened 2021-03-19
chore: add requirements section, add log format callout

mnevilleoneill:
Added a requirements section, as well as a note explaining that Heroku doesn't support customizing logs via Logplex.

### Tell us why

Customer feedback indicated that our docs could be clearer around requirements/formatting.

N/A

No.


urbiz-nr:
Works locally.

PR #1452, opened 2021-03-19
Hero DT troubleshoot images

rhetoric101:
### Tell us why

Four images were not included in the Drupal-to-Gatsby migration, so I replaced them. Each of these sections should now have two images:

* Problems with spans > Missing span not getting exported
* Problems with trace details > Middleware doesn't recognize proprietary trace header


mmfred:
All look good @rhetoric101

PR #1470, opened 2021-03-22
Replaced link for reserved words

urbiz-nr:
As mentioned in #documentation, the link https://docs.newrelic.com/docs/insights/nrql-new-relic-query-language/nrql-resources/nrql-reserved-words/ no longer works.

I've started a sweep to replace it for https://docs.newrelic.com/docs/insights/event-data-sources/custom-events/data-requirements-limits-custom-event-data/#reserved-words. I want to double check with @barbnewrelic this is the right doc before I finish the sweep, I've changed 3 instances so far.

### DIFF
--- a/src/content/docs/agents/java-agent/attributes/java-agent-attributes.mdx
+++ b/src/content/docs/agents/java-agent/attributes/java-agent-attributes.mdx
@@ -108,7 +108,7 @@ In addition to the [default APM attributes](/docs/insights/new-relic-insights/de
     * Browser events: Disabled
 
     <Callout variant="important">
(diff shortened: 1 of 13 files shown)


barbnewrelic:
Yes this is exactly right, thanks for checking! Years ago we had lists of reserved words directly in Insights docs. Finally linked over to the standalone NRQL doc that had a section about reserved words. But anchors seem to have problems post-GitHub migration. Thanks so much for fixing this!

dbarnesbrownNR (approved):
Spot checked on local build and all looks good!

urbiz-nr:
Thanks @dbarnesbrownNR !

PR #1472, opened 2021-03-22
Update Security Docs to add new browser agent security bulletin

robertprast:
### Tell us why

The browser agent recently released a new update to fix a security issue. To alert customers , the security team will post a bulletin that describes the issue and how to patch the bug. 

Thanks for the help and super easy way to update the site!

Only content update :)


mmfred:
Looks good. I'll merge when the checks pass.

PR #1489, opened 2021-03-22
Fix taxonomy redirects

### LINKED ISSUE #1460: Numerous 404s in taxonomy-redirects.json
Many, many paths in `taxonomy-redirects.json` lead to 404s. Out of 1131 lines, 208 return 404s:

* [List of all URLs that result in 404](https://gist.github.com/austin-schaefer/fe0848fe12ba13091d749757c6a3c317)
* [Spreadsheet used to create this list](https://docs.google.com/spreadsheets/d/10_Sxrg6NKOvUdF1xZn0LMMFU-zE7YDGqCih5y01ytoA/edit#gid=0) (NR ONLY)

### Acceptance criteria

For this issue to be resolved, we need:

* We know why these 404s weren't caught by the original 404-catching script, and we've updated the script to catch them in future
* We've also used the script to test all links in nav .yml files
* IDEALLY: We know what went wrong in the redirects migration script so we can see if it happened elsewhere.
* WRITER MAY BE ABLE TO DO THIS: All 404s in taxonomy-redirects.json now resolve to the correct destination 
    * For a writer to do this, we’ll need to know with confidence how to edit/work with this file and have it work reliably on local. 

### Example issue

I was not able to find much of a pattern. For example, consider lines 1155 thru 1167:

```
{
    "url": "/docs/integrations/amazon-integrations/",
    "paths": [
      "/docs/integrations/amazon-integrations/aws-integrations-list",
      "/docs/infrastructure/amazon-integrations/aws-integrations-list",
      "/docs/kubernetes-integration/amazon-integrations/aws-integrations-list",
      "/docs/integrations/new-relic-integrations/cloud-integrations/aws-integrations",
      "/docs/infrastructure-integrations/amazon-integrations/aws-integrations-list",
      "/docs/infrastructure/amazon-integrations",
      "/docs/kubernetes-integration/amazon-integrations",
      "/docs/infrastructure-integrations/amazon-integrations"
    ]
  },
```

Of these, the original `url` works, the first one under paths works but goes to the wrong place, and the others just 404:

```WORKS: https://docs.newrelic.com/docs/integrations/amazon-integrations/
RESOLVES BUT WRONG PLACE: https://docs.newrelic.com/docs/integrations/amazon-integrations/aws-integrations-list
404: https://docs.newrelic.com/docs/infrastructure/amazon-integrations/aws-integrations-list
404: https://docs.newrelic.com/docs/kubernetes-integration/amazon-integrations/aws-integrations-list
404: https://docs.newrelic.com/docs/integrations/new-relic-integrations/cloud-integrations/aws-integrations
404: https://docs.newrelic.com/docs/infrastructure-integrations/amazon-integrations/aws-integrations-list
404: https://docs.newrelic.com/docs/infrastructure/amazon-integrations
404: https://docs.newrelic.com/docs/kubernetes-integration/amazon-integrations
404: https://docs.newrelic.com/docs/infrastructure-integrations/amazon-integrations```

### PR DESCRIPTION
**Pull request by roadlittledawn**
## Description

Originally it seemed like redirects were not working via `gatsby-plugin-auto-index-pages`, but the issue was a gotcha in the way we attach nav menus (loaded into gatsby via `nav/*.yml` files). This updates the logic when resolving which nav to add to a given page at a slug to not add one if the page is a programmatically generated page (not authored .mdx driving it)

### Cause
Every page has a `nav` object in its data via `MainLayout` query, which passes it a slug and either gets back a nav menu object based on one of the nav .yml files OR null. That's set via `gatsby-source-nav` via a resolver that looks for the slug in all the `NavYaml` data in graphql (which is loaded via nav .yml files). We do that to know what left nav menu to show so a user stays in context when clicking on links in a particular menu. The gotcha is if there are links in a nav .yml that are links to directory index pages, the script doesn't know that and adds the nav object. If that link is in more than one nav .yml, the first one it sees (which i think is alpha order) is the nav menu it attaches to the index.

### Fixes re: to the issue
This fixes some of the 404s listed in the issue this relates to. But there are many other links that 404. At quick glance some are caused by:
* links that do not have anything to resolve to (so there's no .mdx file, directory, page in Gatsby)
* paths in `taxonomy-redirects.json` with `docs/release-notes` and `whats-new` are intentionally excluded via the auto index plugin and thus redirects aren't created there (they can be created via frontmatter on index.mdx files in release notes subdirs)

The original scope of this PR was to not look at file nodes in addition to landing page nodes. Still unsure why we look at all file nodes and compare that to directory paths, but reverted that for now.

## Related issues / PRs
Closes #1460

### DIFF
--- a/plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
+++ b/plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
@@ -135,9 +135,7 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
 
   const { nodes: locales } = allLocale;
 
-  const existingPaths = tableOfContentsNodes
-    .map(getSlug)
-    .concat(fileNodes.map(getSlug));
+  const existingPaths = tableOfContentsNodes.map(getSlug);
 
   const files = fileNodes.map((node) => ({
     basename: node.base,
@@ -153,7 +151,7 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
     if (skippedDirectories.includes(dir.path)) {
       return [visit.SKIP];
     }
-
+    // I think we want to just look at tableofcontents, not all filenodes
     if (existingPaths.includes(slug)) {
       return;
     }
--- a/plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
+++ b/plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
@@ -135,7 +135,9 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
 
   const { nodes: locales } = allLocale;
 
-  const existingPaths = tableOfContentsNodes.map(getSlug);
+  const existingPaths = tableOfContentsNodes
+    .map(getSlug)
+    .concat(fileNodes.map(getSlug));
 
   const files = fileNodes.map((node) => ({
     basename: node.base,
--- a/plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
+++ b/plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
@@ -151,7 +151,7 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
     if (skippedDirectories.includes(dir.path)) {
       return [visit.SKIP];
     }
-    // I think we want to just look at tableofcontents, not all filenodes
+
     if (existingPaths.includes(slug)) {
       return;
     }
--- a/plugins/gatsby-source-nav/gatsby-node.js
+++ b/plugins/gatsby-source-nav/gatsby-node.js
@@ -254,6 +254,29 @@ const groupBy = (arr, fn) =>
 const createNav = async ({ args, createNodeId, nodeModel, locales }) => {
   const { slug } = args;
 
+  const fileNode = await nodeModel.runQuery({
+    type: 'File',
+    query: {
+      filter: {
+        sourceInstanceName: { eq: 'markdown-pages' },
+        relativePath: { regex: '/docs/' },
+        childMdx: {
+          fields: {
+            slug: {
+              eq: slug
+                .replace(/\/table-of-contents$/, '')
+                .replace(new RegExp(`^\\/(${locales.join('|')})(?=\\/)`), ''),
+            },
+          },
+        },
+      },
+    },
+  });
+
+  if (fileNode === null) {
+    return null;
+  }
+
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: plugins/gatsby-plugin-auto-index-pages/gatsby-node.js [hunk 1]
```diff
@@ -153,7 +151,7 @@ exports.createPages = async ({ actions, graphql, reporter }, pluginOptions) => {
     if (skippedDirectories.includes(dir.path)) {
       return [visit.SKIP];
     }
-
+    // I think we want to just look at tableofcontents, not all filenodes
```
zstix, line 154: Nitpick: can we either (a) remove this comment or (b) add more context? It feels like this might have been a note while debugging that snuck into the PR.

zstix (approved):
One small comment, but the code otherwise looks good to me!

plugins/gatsby-plugin-auto-index-pages/gatsby-node.js
roadlittledawn: thanks for catching that. was a superfluous comment

roadlittledawn:
looking at this more closely, not sure what i removed should have mattered. still can't recall the reason we want to check all file nodes. BUT i noticed that the `indexPage` template has a nav in its data somehow. dug in last night to see why that might be, and can't find any reason why that it is

roadlittledawn:
yep, i think the problem is caused by how we recursively look through a nav yml file to find the slug of the current page to see which one it may belong to to know which nav to display. https://github.com/newrelic/docs-website/blob/c46814ff2489fdf2c40f365b8f01e18b261eca82/plugins/gatsby-source-nav/gatsby-node.js#L257

paperclypse:
The sooner we can merge this in, the better, I'd say @roadlittledawn. We're getting pinged about this issue once or twice a day from internal folks.

roadlittledawn:
To fix the root cause, we could try to detect if the page at a particular slug is an .mdx file (instead of an auto index page) and return null so we don't attach the nav object to index pages. I have something locally now but it's proving hard to test.

https://github.com/newrelic/docs-website/blob/b6c09a8f804ab358e7acf15cd516967f71c3ff3e/plugins/gatsby-source-nav/gatsby-node.js#L254

nr-kkenney (approved):
lgtm. was also over your shoulder to see the thought process & that it works.

not too knowledgeable in js, but nothing stands out.

roadlittledawn:
fyi. i canceled the Run tests portion of the Validate pull request workflow because it looked like it was stuck. i ran tests locally and it passed (tests don't really apply to the work in this PR)

PR #1508, opened 2021-03-23
chore: add missing permission in EMR

HiDAl:
### Tell us why

Users of Cloud integrations have to add the permission `emr:ListInstanceFleets` to work as expected.


mmfred:
Thanks again for your fix!

PR #1512, opened 2021-03-23
Update usage-queries-alerts.mdx

nedl86:
Updated callout about Evaluation Offset to caution level and explained that not doing this may cause conditions not to trigger.

### Tell us why

The previous language around the Evaluation Offset (and level of the callout being set to Tip) did not strongly enough suggest that *not* doing this will cause alert conditions to not trigger. Added additional language stating this, and changed callout level from tip to caution.

Based on my testing, leaving the evaluation offset at the default causes usage alerts to not trigger.

No.


urbiz-nr:
Thanks for updating this, @nedl86. Merging!

PR #1535, opened 2021-03-25
Restore "Single page apps" to left navigation panel

rhetoric101:
### Tell us why

In the migration from Drupal to Gatsby, the docs related to single-page apps were not included in the left navigation. Evidently, the migration just moved navigation sub-category item " Browser monitoring." 

To fix this, I moved the navigation category "Browser monitoring" down one level and created a sibling category of "Single page app monitoring." I then created the main navigation node for these related topics as "Browser" and is a sibling to other main topics, such as APM and Logs.

The one thing that doesn't quite match Drupal is that when you go to "View all Browser docs," you'll see that I had to rename the landing page from "Browser monitoring" to just "Browser" to avoid two stacked, identical headers on that TOC page. So, the new browser landing page now just says "Browser"; I hope that is OK!

You can compare this with Drupal at https://docs-dev.newrelic.com/docs/browser

### DIFF
--- a/src/content/docs/browser/index.mdx
+++ b/src/content/docs/browser/index.mdx
@@ -1,5 +1,5 @@
 ---
-title: Browser monitoring
+title: Browser
 type: landingPage
 tags:
   - Browser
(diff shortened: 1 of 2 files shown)


urbiz-nr:
Tried locally, works fine. Good catch!

PR #1539, opened 2021-03-25
Guided Install now enabled process monitoring

josemore:
Highlighting that enable_process_metrics is now enabled by default when using the Guided Install

### Tell us why

The new Guided Install flow is setting enable_process_metric enabled by default. This helps new New Relic customers experience all NR1 features around process monitoring.

Some customers asked if Guided Install is enabling process monitoring by default in the infrastructure agent. 


mmfred:
@josemore - thanks for the fixes!

PR #1554, opened 2021-03-26
Add rename logic to getAgentName

LizBaker:
Currently https://docs.newrelic.com/api/agent-release-notes.json outputs entries for both `nodejs` and `node` as well as `.net` and `dotnet`. This PR simplifies these entries and renames `.net` and `node` to `dotnet` and `nodejs`. This is in line with the naming used in a previous service

### DIFF
--- a/src/utils/getAgentName.js
+++ b/src/utils/getAgentName.js
@@ -16,16 +16,28 @@ const AGENTS = [
   'sdk',
 ];
 
+// keeping the naming streamlined and in matching with previous data
+const RENAMES = { '.net': 'dotnet', node: 'nodejs' };
+
 /**
  * @param {string} subject The release note "subject" in frontmatter
  * @returns {string}
  */
-const getAgentName = (subject) =>
-  subject &&
-  AGENTS.find(
-    (agent) =>
-      subject.toLowerCase().includes(agent) &&
-      !subject.toLowerCase().includes('insights')
-  );
+
+const renameKeys = Object.keys(RENAMES);
+
+const getAgentName = (subject) => {
+  if (subject) {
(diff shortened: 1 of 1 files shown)


src/utils/getAgentName.js
```diff
@@ -16,16 +16,28 @@ const AGENTS = [
...
+  if (subject) {
+    const subjLowercase = subject.toLowerCase();
+    const agentName = AGENTS.find(
+      (agent) =>
+        subjLowercase.includes(agent) && !subjLowercase.includes('insights')
+    );
+
+    return agentName && renameKeys.includes(agentName)
+      ? RENAMES[agentName]
+      : agentName;
```
zstix, line 39: You might be able to simplify this with the [in operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in):

```js
return agentName in RENAMES ? RENAME[agentName] : agentName;
```

Or perhaps even shorter by using the [short-circuit OR evaluation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR#short-circuit_evaluation):

```js
return RENAMES[agentName] || agentName;
```

I'm not 100% that the second example works, might require a little tweaking.

zstix (commented):
Looks good! I think we can probably avoid making a `renameKeys` variable, but it's your call on if how we do that.

src/utils/getAgentName.js
LizBaker: ooh yeah! that `in` is nice, and i think the or should totally work

zstix (approved):

PR #1556, opened 2021-03-26
Redundant image docs

austin-schaefer:
### Tell us why

I found there were still a number of redundant style guide docs lurking in both the main `style`guide` directory and in the `manual-edits` directory.

The `embed-images.mdx` doc I deleted had some info about snagit usage that's not covered in the `edit-images.mdx` doc I kept, but I don't think that SnagIt-specific info is (a) up to date or (b) particularly valuable.


paperclypse (approved):
Yeah, I think this change makes sense. It occurs to me that we may want to bump the Embed videos doc underneath Embed images. Thematically they sort of make sense being closer together.

austin-schaefer:
Done @paperclypse . I also found a duplicate entry in that nav file that I deleted.

The "processes and procedures" category as a whole is pretty bloated, perhaps we should break it up? I'm a little leery of messing with the taxonomy that much in a quick hit, though, so I'll leave that one to your discretion.

austin-schaefer:
Tested locally and all links seem to be behaving nicely.

PR #1569, opened 2021-03-29
Add in iOS app, handled exception release note

jvanmetre:
### Tell us why

New version of the New Relic iOS app has been released with support for handled exceptions.


bradleycamacho:
@jvanmetre Looks good to me!

PR #1583, opened 2021-03-29
Fix: Auto triage jobs failing for issues

aswanson-nr:
### Tell us why
The auto triage jobs currently work for pull requests, but they error out with a 404 for issues. This PR pulls that workflow out into separate workflows, allowing the branching to happen at the event level. This way the code itself does not have to determine what type of event it is running against.


aswanson-nr:
I tested this on a personal repository, https://github.com/aswanson-nr/action-test-repo, if you'd like to take a look. I can confirm that it works!

.github/workflows/issue-triage.yml
```diff
@@ -0,0 +1,43 @@
...
+        gh api -H "$HEADER" -X POST projects/columns/$COLUMN/cards -f content_type='Issue' -F content_id=$ISSUE_ID
+
+  localization_project:
+    runs-on: ubuntu-latest
+    env:
+      PROJECT_ID: 12004783
+      PROJECT_NAME: Docs Site Localization PRs and Issues
+    name: Triage to Localization Project
+    steps:
+    - name: Triages NEW issues and NEW pull requests to the Localization Project
```
nr-kkenney, line 34: i think you can update the name for this now

.github/workflows/pr-triage.yml
```diff
@@ -36,10 +34,10 @@ jobs:
     - name: Triages NEW issues and NEW pull requests to the Localization Project
```
nr-kkenney, line 34: same as above, i think you can update the name for this now.s

.github/workflows/issue-triage.yml
aswanson-nr: Good catch! I have updated it

.github/workflows/pr-triage.yml
aswanson-nr: Updated!

nr-kkenney (approved):
lgtm. might want another reviewer though, since i might be too biased / involved with this.

PR #1603, opened 2021-03-30
Fixes 'whats new' link on release note template page.

michaelgoin:
Fixes the broken "What's new style guidelines" link at the top of the 'create-relase-notes' page.

Current link 404s.


urbiz-nr:
Thanks for fixing this @michaelgoin! Tried locally.

PR #1604, opened 2021-03-30
Renames 6.x and 7.x Node.js release notes to new naming convention.

michaelgoin:
Renames 6.x and 7.x release notes for the Node.js agent to the naming convention discussed here: https://github.com/newrelic/docs-website/issues/1255 which we've started using going forward.

This is a quality-of-life change for our team to make clear to stick to the new convention and have all recent notes grouped nearby. This should not have any functional impacts other than changing existing URLs. 

There is a small risk of breaking someone's existing link to a release note but seems worth this one-time risk to avoid future mistakes.


urbiz-nr:
Thanks for updating these files @michaelgoin!

PR #1617, opened 2021-03-31
Update store-secure-credentials-scripted-browsers-api-tests.mdx

rudawg:
### Tell us why

The Synthetics Engineers have confirmed that the character limit has been increased from `3000` to `10000` so reflecting that change in the Docs.
Ref [GTSE-7743](https://newrelic.atlassian.net/browse/GTSE-7743)


urbiz-nr:
Thanks for the update, @rudawg!

PR #1620, opened 2021-03-31
fix: newrelic-fluent-bit-output released version

Noly:
### Tell us why
* `newrelic-fluent-bit-output` included in the last released agent was with wrong version.


urbiz-nr:
Thanks for the update, @Noly !

PR #1625, opened 2021-03-31
Fix issue with collapser ID

nr-ahemsath:
### Tell us why

The `completeTransactionsOnThread` service config section could not be linked to because the collapser ID was incorrect.


bradleycamacho:
Looks good to me, thanks @nr-ahemsath !

PR #1644, opened 2021-04-01
Document how to associate entities / infra with logs

RyanCopley:
### Tell us why

It took way too much effort to figure out how to associate logs with infra resulting in "Enable Logs" button in my Infrastructure UI.

I work at New Relic


urbiz-nr:
Thanks for helping with docs, @RyanCopley!

PR #1677, opened 2021-04-02
Added table of contents example to the end of the path for the final …

rhetoric101:
### Give us some context

The final button at the bottom of our landing page template has a link to all the docs, which displays a table of contents page. To load this page, you need to have "/table-of-contents" following the path for the button, or else the page won't load (it just takes you to the node you were on).


paperclypse (approved):
Looks good!

PR #1683, opened 2021-04-05
Disable deployment marker action until we've got versioning

LizBaker:
### Give us some context

Comments out the release workflow which depends on tags. We can uncomment this once we have some kind of versioning in place 

closes https://github.com/newrelic/docs-website/issues/1682


zstix (approved):

dudss37:
- [ ]

PR #1694, opened 2021-04-05
Update Filter Function Description

ahunger603:
PR from internal QUEX Team.

### Give us some context

Improves clarity of use-case description for the `filter()` function.


urbiz-nr:
Thanks for the update, @ahunger603!

dbarnesbrownNR (approved):
Looks good!

dbarnesbrownNR:
(And you can ignore that weird other commit; it's due to a specific endpoint that gets updated automatically on local builds, which is not needed but also harmless.)

PR #1713, opened 2021-04-06
fix: make rss guids deterministic

### LINKED ISSUE #1271: Release notes RSS feed guid's are changing/regenerated, breaking some RSS readers
### Description

The docs website Java Agent release notes RSS feed includes a guid field. This guid field is used by some RSS feed readers to uniquely identify new posts and prevent re-downloading already seen posts. 

In this specific case we're using Huginn and it's built in RSS agent to monitor the Java Agent Release Notes RSS feed. 

The guid field in the XML feed is changing/being regenerated on a regular basis. My suspicion would be when some kind of build event occurs, but that's just a guess. 

Because the guid is changing, the RSS reader thinks there are new posts available for download. Because every single guid for every single post is changing, it re-downloads hundreds of posts each time. 

### Steps to reproduce

1. Make note of a guid for a post on the Java Agent Release Notes RSS feed
2. Wait 24-48 hours for whatever event is occurring (automated build?)
3. Check the guid for the post you made note of, it will be different

I do not know what is triggering the guid to change, but whatever it is occurring daily at some time in the evening UTC (approx 1800-2000 UTC).

### Expected behavior

Expected that guid do not change so that posts can be uniquely identified and prevent re-downloading. 

### Additional context

Link to specific feed: 
https://docs.newrelic.com/docs/release-notes/agent-release-notes/java-release-notes/feed.xml

### Environment

N/A

aswanson-nr:
### Give us some context
The current GUIDs for our release notes RSS feeds can change, causing downstream issues for consumers of these feeds.
This PR makes the GUID deterministic and based on the subject of the release note and the version.
Closes #1271

### DIFF
--- a/plugins/gatsby-plugin-release-note-rss/gatsby-node.js
+++ b/plugins/gatsby-plugin-release-note-rss/gatsby-node.js
@@ -66,7 +66,7 @@ const releaseNotesQuery = async (graphql) => {
 
 // converts graphQL response (node) into a consumable format for RSS
 const getFeedItem = (node, siteMetadata) => {
-  const { frontmatter, slug, id, mdxAST } = node;
+  const { frontmatter, slug, mdxAST } = node;
   const { releaseDate, subject, version } = frontmatter;
 
   const transformedAST = htmlGenerator.runSync(mdxAST);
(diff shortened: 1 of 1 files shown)


aswanson-nr:
I confirmed that two separate cache-less builds completed with the same RSS Guids.

roadlittledawn (approved):
Using `subject` and `version` make sense to me :shipit:

PR #1717, opened 2021-04-06
Enabling related resources

### LINKED ISSUE #1353: Add related resources component to Docs Site
As a follow up on #427 we want to now enable the right rail component to return related resource content links. In order to this we need to clarify some questions around cost and how we'll need to implement the refresh rate of content per page.

### Considerations

- We refetch results per page once a day on a cron job. The data is available as [part of the build: ](https://github.com/newrelic/developer-website/blob/develop/src/data/related-pages.json)
- That data gets rebuilt [in this action once]( https://github.com/newrelic/developer-website/blob/develop/.github/workflows/fetch-related-content.yml) a day:
- We can specify whatever interval we want if we go this route. When we execute a query per page each time we fetch related content for a page. We have a limit on the number of queries we can run before we are charged more. We could decrease the interval and instead do once a week, or once a month.  Doing so would put costs down, but means that auto related content could become stale (maybe thats a fair tradeoff)

### Acceptance Criteria

- [x] confirm who will be paying for the additional costs to run the swiftype queries. - `APPROVED`  it's all in Slack. in #swiftype-devs 👍 

### Cost
```
estimated cost: about 60k searches/month (assuming 2k searches * 30 days), just to populate the 
links for Related Content between $400 - $600 a month in swiftype 
cost. This is based on our current pricing plan

```

### Interval
```
To cap it at 60k queries a month, that means we can do 2K a day max...since there's a query per page and we have 5kish pages,  we would want to run this job once every 3 days (72 hrs) as the most frequent interval we could do 
```

- [x] Finish the Implementation of the right rail component on the docs site and generate related content links into pages. We are approved to do `60k searches/month (assuming 2k searches * 30 days)

> NOTE: JV tested this out in #1356 so it seems to be working with static resource links in frontmatter.

- [ ] [Update the Style Guide](https://docs.newrelic.com/docs/style-guide/) to ensure we define how this feature works. we can pass this off to a Docs Writer, so let's create a content ticket and have @paperclypse review it.

- [x]  There is a small bug when running the build for related resources which is why there are no pages with it currently. Right now you can run the build for related resources on demand, which will query Swiftype for all pages, but last I [ran this](https://github.com/newrelic/docs-website/actions/runs/604770713) it failed due to an error from the Swiftype API that we did not catch properly.

### Mock Up

Screen Shot 2021-03-16 at 3 15 21 PM

zstix:
## Description
Enables the "Build with swiftype" workflow to run twice a week and ensures that we're only showing related resources for the same language.

[Here](https://github.com/newrelic/docs-website/runs/2282202029?check_suite_focus=true) is an example of a successuful run of this workflow. I've also run this a few times on my machine (with one page, some pages, and all pages). I haven't run into any issues from _Swiftype_ so I think this work is...ready?

## Related Issue(s)
* Closes #1353

## Screenshot(s)
Screen Shot 2021-04-06 at 4 42 36 PM

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -8,6 +8,7 @@ const getAgentName = require('./src/utils/getAgentName');
 
 const siteUrl = 'https://docs.newrelic.com';
 const dataDictionaryPath = `${__dirname}/src/data-dictionary`;
+const additionalLocales = ['jp'];
 const quote = (str) => `"${str}"`;
 
 const autoLinkHeaders = {
@@ -56,7 +57,7 @@ module.exports = {
         },
         i18n: {
           translationsPath: `${__dirname}/src/i18n/translations`,
-          additionalLocales: ['jp'],
+          additionalLocales,
         },
         prism: {
           languages: [
@@ -76,6 +77,7 @@ module.exports = {
             resultsPath: `${__dirname}/src/data/swiftype-resources.json`,
             engineKey: 'Ad9HfGjDw4GRkcmJjUut',
             refetch: Boolean(process.env.BUILD_RELATED_CONTENT),
+            limit: 3,
             filter: ({ node }) => {
               if (node.internal.type !== 'Mdx') {
                 return false;
@@ -106,6 +108,11 @@ module.exports = {
             getParams: ({ node }) => {
               const { tags, title } = node.frontmatter;
 
+              const locale = slug && slug.split('/')[0];
+              const postfix = additionalLocales.includes(locale)
+                ? `-${locale}`
+                : '';
+
               return {
                 q: tags ? tags.map(quote).join(' OR ') : title,
                 search_fields: {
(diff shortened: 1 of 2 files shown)


gatsby-config.js
```diff
@@ -8,6 +8,7 @@ const getAgentName = require('./src/utils/getAgentName');
 
 const siteUrl = 'https://docs.newrelic.com';
 const dataDictionaryPath = `${__dirname}/src/data-dictionary`;
+const additionalLocales = ['jp'];
```
aswanson-nr, line 11: This probably doesn't need to be done now, but would it be worth adding a config variable to hold all the locales? That way when we add another we don't need to go through a bunch of plugins.

aswanson-nr (approved):

gatsby-config.js
zstix: Yes! We had discussed moving this into some sort of configuration file that also included locale codes in different formats (`ja-JP` for example). Out of scope of this PR, but certainly something we should consider.

PR #1718, opened 2021-04-07
chore: updated css

jpvajda:
This PR updates the CSS based on feedback from @bradleycamacho


jpvajda:
Closing this as we are going to take a different approach which would be to create a new component for these docs pages that would be a banner element that would be sticky and clearly articulate the the state of the doc for BETA docs.

PR #1731, opened 2021-04-07
Move JSON schema into its own section

paperclypse:
I got a request from a member of the Applied Intelligence team to move this JSON schema code example out into its own section.


urbiz-nr:
Checked locally, looks good. You may merge, @paperclypse!

PR #1740, opened 2021-04-08
add mutation sample for link a Metric streams account

HiDAl:
### Give us some context

Add instructions of how to link an AWS Account sending data through Cloudwatch Metric Stream (PUSH mode)


urbiz-nr:
@HiDAl, thanks for updating this doc. I've tried it locally and it looks good. Merging!

PR #1742, opened 2021-04-08
Slight improvements/clarifications to capabilities and roles info based on SME feedback 

zuluecho9:


zuluecho9:
@dbarnesbrownNR thanks much! good catch

PR #1747, opened 2021-04-08
Clean up Insights mobile app docs, as those apps are EOL

jvanmetre:
* Add in New Relic tvOS release note for updated version 
* Remove Insight iOS / Android documentation as they are EOL on April 11th.


dbarnesbrownNR (approved):
Made one tiny edit to fix a punctuation issue. Otherwise this CHONKY PR is good to go. :)

PR #1776, opened 2021-04-09
Removed guided install from Java docs

paperclypse:
Java on guided install needs a bit more time in the oven.


bradleycamacho:
Looks good to me! Not sure why the first check is failing.

bradleycamacho (approved):

PR #1778, opened 2021-04-09
Fixed broken link in trace observer step.

rhetoric101:
### Give us some context

We had a broken link to a table of contents page in the trace observer instructions (step 6) in the following page:

https://docs.newrelic.com/docs/distributed-tracing/infinite-tracing/set-trace-observer/


paperclypse (approved):
Looks good!

PR #1780, opened 2021-04-12
Clarifying agent versions vs. subscription roles

urbiz-nr:
As discussed with the #gts-browser team.


urbiz-nr:
Reviewed by @zuluecho9.

PR #1787, opened 2021-04-12
DOC-6653 Update attribute-dictionary.js

zuluecho9:
Update to intro of data dictionary.


barbnewrelic:
@zuluecho9  Thanks for making these updates! I'm not clear what to do with them next, so I've temporarily moved this over to Blocked so you and I can chat. Thanks!

zuluecho9:
@barbnewrelic I think we can just merge them at any time, so if you think they look good, I'll do that.

PR #1804, opened 2021-04-13
Update required permissions for GCP

mkrichevsky:
For monitoring GCP Cloud Integrations we now require new permission that is required to monitor all services


barbnewrelic (approved):
Thanks for making this change!

barbnewrelic:
Thanks for updating this!

PR #1807, opened 2021-04-13
Update find-unusual-logs-log-patterns.mdx

barbnewrelic:
redirect URL was missing, reported in documentation slack channel 04-13-21


rhetoric101:
@barbnewrelic, your redirect entry in the front matter looks correct, although my local build won't redirect to it. I think you can still merge this (The file renders fine--just not the redirect).

PR #1851, opened 2021-04-15
Doc 6682 - Some more edits to user migration workflow

zuluecho9:
This is another round of peer edits for user migration doc. Just need a light PR review for merging.


src/content/docs/accounts/original-accounts-billing/original-users-roles/user-migration.mdx
```diff
@@ -52,13 +52,14 @@ For more about these tools, see the [new user management docs](/docs/accounts/ac
...
 
-    ![login-multiple-accounts-view.png](./images/login-multiple-accounts-view.png "login-multiple-accounts-view.png")
+
+    ![lA screenshot of what is shown when you have an email address associated with multiple New Relic logins](./images/login-multiple-accounts-view.png "Login UI for email addresses with multiple user records")
```
barbnewrelic, line 62: I can't tell if that's supposed to be "1A" or that's an "l" from previous text

barbnewrelic (approved):
I have one question in my review, but otherwise I think you're good to go. Thanks for continuing to work on these important account docs!

mmfred (approved):

PR #1859, opened 2021-04-15
Use tag name for release name

zstix:
## Description
Updates our release workflow so that the title of the release is the same as the version name. Without this, the release title was the last commit.

## Reviewer Notes
The last time the release workflow ran, [it failed](https://github.com/newrelic/docs-website/runs/2355397196?check_suite_focus=true). It doesn't look like this was an issue with our workflow, but rather an issue with Github actions (it never made it to our step). I'll monitor it :eyes:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ jobs:
           echo "::set-output name=tag::release-$MERGES"
 
       - name: Create release
-        run: gh release create ${{ steps.get-tag.outputs.tag }}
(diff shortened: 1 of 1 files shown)


nr-kkenney (approved):
lgtm

paperclypse:
Thanks for doing this! 

What if, and I know this seems trivial and nonsensical, what if in addition to a number, we appended a randomly generated two or three word phrase. It would be fun and help make the releases a bit more memorable. :)

I mean, who wouldn't want `yellow-babbage-monkey` or `dessert-dancemaster-flue` releases? 

Just a funny idea I had.

PR #1865, opened 2021-04-16
chore: Adjust gh cli command

LizBaker:
### Give us some context

This PR swaps a flag used in our github release action which was previously causing an issue (using `wc` `-L` counts the length of the longest line while `-l` counts the number of lines ). also updates the release logic to only count merged PRs (not including every closed PR) and starts the count from march 1st


nr-kkenney (approved):
lgtm. ran it locally and got the right number (i think).

PR #1867, opened 2021-04-16
chore: update .NET Agent MongoDB support

jaffinito:
### Give us some context

Updates the compatibility page for MongoDB with the .NET Agent to better describe what we support.  Breaks things out into package and version which is more accurate.

Nope.


barbnewrelic (approved):
Thanks for updating this info, very helpful!

PR #1872, opened 2021-04-16
Doc 6677 - Improvements to NerdGraph intro and nav

zuluecho9:
Asking @rhetoric101 if he wants to do review on this. See jira for details. Basically overhauled the intro doc, making it easier to understand what NerdGraph is, and how it works, and using simpler, more direct language, and including a helpful video. I also added all the assorted NerdGraph tutorials in one place in the intro doc, because previously we didn't have the full list there (as they are a bit scattered throughout docs site). And I retitled all nerdgraph-related docs and changed their short titles to make them all more aligned.


rhetoric101 (approved):
@zuluecho9, these changes look good. I only added one missing quotation mark in the front matter for that dashboard export page.

PR #1882, opened 2021-04-17
Update create-metrics-other-data-types.mdx

nr-jbiggley:
Added an example of using Events to Metrics on a non-numeric field by using summary(1) instead of the summary(non-numeric_field) as the latter will return zeros and NULLs

### Give us some context

Added a new example for E2M rules to demonstrate translating a **count(string)** function into a **summary()** metric by using **summary(1)** vs **summary(string)**. The latter will not return the desired results.

No site code changes


urbiz-nr:
Thanks for updating this doc, @nr-jbiggley!

urbiz-nr (approved):

PR #1895, opened 2021-04-19
Fix PageView and PageViewTiming links

joshwhatk:
### Context

When accessing the links on the [Events reported by Browser Monitoring page](https://docs.newrelic.com/docs/insights/event-data-sources/default-events/events-reported-browser-monitoring/), several of the links (`PageView`, `PageViewTiming`, etc.) only go to the `AjaxRequest` event and not the correct event.

### Resolution

I could not find the proper `event_tids` to use, so I used the resolved URL for the specific event pages. I have only fixed the `PageView` and `PageViewTiming` events. 

This may not work for you all, but I thought I'd throw this out there.

### DIFF
--- a/src/content/docs/insights/event-data-sources/default-events/events-reported-browser-monitoring.mdx
+++ b/src/content/docs/insights/event-data-sources/default-events/events-reported-browser-monitoring.mdx
@@ -38,7 +38,7 @@ redirects:
   <tbody>
     <tr>
       <td>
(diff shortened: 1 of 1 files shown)


zuluecho9:
Closing, created another PR because all those data dictionary links in that category were broken so fixed them all. https://github.com/newrelic/docs-website/pull/1901

zuluecho9:
thanks @joshwhatk. this turned out to be a bigger issue with a bunch of broken data dictionary links so we're working on that. Thanks much.

PR #1899, opened 2021-04-19
Section on X-Ray and Step Functions

MattWhelan:
### Give us some context

Customers sometimes want to use X-Ray to monitor step functions and view the traces in NR.

I don't think we need deep coverage of how to use step functions with X-Ray.
Neither of those things is a New Relic product. But links to AWS docs should
help users get the most out of their integration.


zuluecho9 (approved):

zuluecho9:
thanks much @MattWhelan .

PR #1903, opened 2021-04-19
Just messing around on a high-wire

paperclypse:
No big deal :)

Hey @bradleycamacho, does this get us closer to the docs you love?


bradleycamacho (dismissed):

bradleycamacho:
Look out Gatsby docs, we're coming for you! Approved.

jpvajda:
Is this a legit PR? seems "maybe" 🤔

mmfred:
For a short term improvement, if we can revise this intro to define NR just by adding a few phrases, that's a quick win. So you could move that initial line back up, and then use the chunkier middle paragraph to introduce New Relic with more definition.
I'd love to avoid using the terms TDP and FSO in that intro to what New Relic is paragraph, and focus on the things that are mentioned in the subtitles for those products. Like, use some of the descriptions for TDP: Ingest, visualize, and alert on all your telemetry data in one place. and similarly, a phrase or two from the descriptions from FSO and Agile etc. on the landing page.

paperclypse:
@jpvajda It's not _not_ legit :)

paperclypse:
What do you think about my last commit, @mmfred? Closer to what you're thinking?

mmfred:
yes! so much better! I believe that @Broseberg said we need to have this at the top, next to the three clickable boxes: If you're new, follow these three steps to create an account and get going. (It's free!) 

And if so, this can just go right below it. So the order, using the current numbers, would be:
p4
p1
p2
p3

paperclypse:
As fun as it is to make changes to the landing page, we probably need to think through them a little more :)

PR #1934, opened 2021-04-20
Main changes to fix too many notes and callouts.

dbarnesbrownNR:
This is just a delayed PR relating to DOC-6717. There were some scream-y callouts that detracted from clarity so I took those away and integrated the content that was still important into the pages in appropriate places.


paperclypse (approved):
They all seem like good changes to me.

PR #1946, opened 2021-04-21
Update to gatsby v3

nr-kkenney:
Changes for upgrading to Gatsby V3


package.json
```diff
@@ -4,34 +4,34 @@
...
     "date-fns": "^2.17.0",
     "front-matter": "^4.0.2",
-    "gatsby": "^2.32.4",
-    "gatsby-image": "^2.11.0",
-    "gatsby-plugin-manifest": "^2.12.0",
+    "gatsby": "^3.3.1",
+    "gatsby-image": "^3.3.0",
+    "gatsby-plugin-manifest": "^3.3.0",
     "gatsby-plugin-json-output": "^1.2.0",
     "gatsby-plugin-mdx": "^1.10.1",
```
aswanson-nr, line 22: This has a newer version, `2.3.0`

package.json
```diff
@@ -4,34 +4,34 @@
...
     "gatsby-plugin-meta-redirect": "^1.1.1",
     "gatsby-plugin-offline": "^3.10.1",
```
aswanson-nr, line 24: This has a newer version as well, `4.3.0`

roadlittledawn:
one weird thing i just noticed is on the data dictionary page. when i go to filter for a specific `Data Type` i get an error in the console and the page doesn't seem to apply the filter. https://pr-1946.d1vx4w5vgyla4s.amplifyapp.com/attribute-dictionary/

aswanson-nr (commented):
Looks great! I just noticed two plugins that can be updated

package.json
nr-kkenney: updated.

package.json
nr-kkenney: updated.

nr-kkenney:
resolved the issue @roadlittledawn mentioned. switched `useNavigate` to `navigate` in our import from reach/router.

zstix (approved):

PR #1968, opened 2021-04-22
DOC6581

barbnewrelic:
Didn't make sense to have Span listed with the WorkloadStatus attributes. Returning this to 2 separate items until we have a better plan for attributes that apply to multiple events.


zuluecho9:
Closed due to doing some data dictionary work in another PR.

PR #1974, opened 2021-04-23
Add "go forward" callout

mnevilleoneill:
Data partitions is a "go forward" feature which does not affect logs that have already been written to NRDB. We've received a few customer inquiries on this point, updating the docs for clarity.


mmfred (approved):
looks good! thanks @mnevilleoneill!

PR #1907, opened 2021-04-20
Update references to New Relic's language-specific exporters for OpenTelemetry

jgals:
See DOC-6641 for details.

This PR updates references to New Relic's language-specific exporters for OpenTelemetry in favor of the collector and OTLP.

PR #1975, opened 2021-04-23
Clarify use of pre-release OTLP endpoint

jgals:
A follow up from #1907.  For more details see DOC-6641. @rvanderwal-newrelic, please review!

This changes language around the pre-release to reference the OTLP endpoint, not the OTLP exporter, which is slightly ambiguous. In other places, I've tried to be more clear that we mean [the OTLP exporter in the OpenTelemetry collector](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/otlpexporter).


mmfred (approved):
Thanks @jgals!

RichVanderwal (commented):
Yeah, I think this is pretty good. Thanks for dealing with my suggestions!

PR #2002, opened 2021-04-26
DOC 6811 - FedRAMP compliance for Trace API

zuluecho9:
For details, see jira. Hold off on publishing but we can do review; it'll be ready to publish soon. Will ask @rhetoric101 if he wants to review.

### DIFF
--- a/src/content/docs/security/security-privacy/compliance/fedramp-compliant-endpoints.mdx
+++ b/src/content/docs/security/security-privacy/compliance/fedramp-compliant-endpoints.mdx
@@ -525,4 +525,8 @@ To use the Log API directly, you'd edit the [Log API endpoint configuration](/do
 
 ### Trace API [#trace-api]
 
-FedRAMP is not yet available for the Trace API itself, but note that some of our agents report trace data and you can [enable FedRAMP for some of those agents](#agents). 
+To ensure FedRAMP compliance for data sent via the [Trace API](/docs/distributed-tracing/trace-api/introduction-trace-api/) or [telemetry integrations](/docs/integrations/open-source-telemetry-integrations/) that use the Trace API, replace the `https://trace-api.newrelic.com/trace/v1` endpoint with `https://gov-trace-api.newrelic.com/trace/v1`. 
+
+<Callout variant="tip">
+Note that distributed trace data is also reported by some agents, like our APM agents, browser agent, and mobile agent. To enable FedRAMP compliancy for that data, you would [enable FedRAMP for the applicable agent](#agents).
+</Callout>
--- a/src/content/docs/security/security-privacy/compliance/fedramp-compliant-endpoints.mdx
+++ b/src/content/docs/security/security-privacy/compliance/fedramp-compliant-endpoints.mdx
@@ -526,7 +526,7 @@ To use the Log API directly, you'd edit the [Log API endpoint configuration](/do
 ### Trace API [#trace-api]
 
 Notes about FedRAMP compliancy for trace data: 
-* Trace data is also reported by some of our agents, like our APM agents, browser agent, and mobile agent. To enable FedRAMP compliancy for that data, you would [enable FedRAMP for the applicable agent](#agents).
+* Trace data is reported by some of our agents, like our APM agents, browser agent, and mobile agent. To enable FedRAMP compliancy for that data, you would [enable FedRAMP for the applicable agent](#agents).
 * Currently [Infinite Tracing](/docs/distributed-tracing/infinite-tracing/introduction-infinite-tracing/) is not FedRAMP compliant. 
 
 To ensure FedRAMP compliance for data sent via the [Trace API](/docs/distributed-tracing/trace-api/introduction-trace-api/) (including [telemetry integrations](/docs/integrations/open-source-telemetry-integrations/) that use this API), replace the `https://trace-api.newrelic.com/trace/v1` endpoint with `https://gov-trace-api.newrelic.com/trace/v1`. 
--- a/src/content/docs/security/security-privacy/compliance/fedramp-compliant-endpoints.mdx
+++ b/src/content/docs/security/security-privacy/compliance/fedramp-compliant-endpoints.mdx
@@ -525,9 +525,8 @@ To use the Log API directly, you'd edit the [Log API endpoint configuration](/do
 
 ### Trace API [#trace-api]
 
-Notes about FedRAMP compliancy for trace data: 
-* Trace data is reported by some of our agents, like our APM agents, browser agent, and mobile agent. To enable FedRAMP compliancy for that data, you would [enable FedRAMP for the applicable agent](#agents).
+To ensure FedRAMP compliance for data sent via the [Trace API](/docs/distributed-tracing/trace-api/introduction-trace-api/) (including [telemetry integrations](/docs/integrations/open-source-telemetry-integrations/get-started/introduction-new-relics-open-source-telemetry-integrations) that use this API), replace the `https://trace-api.newrelic.com/trace/v1` endpoint with `https://gov-trace-api.newrelic.com/trace/v1`. 
+
+Notes about FedRAMP compliance for other trace data: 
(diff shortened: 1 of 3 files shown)


src/content/docs/security/security-privacy/compliance/fedramp-compliant-endpoints.mdx
```diff
@@ -525,4 +525,10 @@ To use the Log API directly, you'd edit the [Log API endpoint configuration](/do
 
 ### Trace API [#trace-api]
 
-FedRAMP is not yet available for the Trace API itself, but note that some of our agents report trace data and you can [enable FedRAMP for some of those agents](#agents). 
\ No newline at end of file
+Notes about FedRAMP compliancy for trace data: 
+* Trace data is also reported by some of our agents, like our APM agents, browser agent, and mobile agent. To enable FedRAMP compliancy for that data, you would [enable FedRAMP for the applicable agent](#agents).
+* Currently [Infinite Tracing](/docs/distributed-tracing/infinite-tracing/introduction-infinite-tracing/) data is not FedRAMP compliant.
```
jonathanpdx, line 530: I think `data` might be confusing to our customers.

```suggestion
* Currently [Infinite Tracing](/docs/distributed-tracing/infinite-tracing/introduction-infinite-tracing/) is not FedRAMP compliant. 
```

src/content/docs/security/security-privacy/compliance/fedramp-compliant-endpoints.mdx
```diff
@@ -525,4 +525,9 @@ To use the Log API directly, you'd edit the [Log API endpoint configuration](/do
 
 ### Trace API [#trace-api]
 
-FedRAMP is not yet available for the Trace API itself, but note that some of our agents report trace data and you can [enable FedRAMP for some of those agents](#agents). 
\ No newline at end of file
+Notes about FedRAMP compliancy for trace data: 
+* Trace data is reported by some of our agents, like our APM agents, browser agent, and mobile agent. To enable FedRAMP compliancy for that data, you would [enable FedRAMP for the applicable agent](#agents).
```
rhetoric101, line 529: I think "compliance" might be a better form here, unless "compliancy" is used elsewhere in FedRamp docs.

src/content/docs/security/security-privacy/compliance/regulatory-audits-new-relic-services.mdx
```diff
@@ -81,7 +81,7 @@ New Relic's time frames for supported regulatory frameworks and annual audits in
 
     <tr>
       <td>
-        APM (plus language agents)
+        [APM](/docs/security/security-privacy/compliance/fedramp-compliant-endpoints/#apm-endpoints) (plus language agents)
```
rhetoric101, line 84: Why not call this "APM language agents"? The "plus language agents" seems confusing to me because that is all you have in APM, right?

src/content/docs/security/security-privacy/compliance/fedramp-compliant-endpoints.mdx
```diff
@@ -525,4 +525,9 @@ To use the Log API directly, you'd edit the [Log API endpoint configuration](/do
...
+* Currently [Infinite Tracing](/docs/distributed-tracing/infinite-tracing/introduction-infinite-tracing/) is not FedRAMP compliant. 
+
+To ensure FedRAMP compliance for data sent via the [Trace API](/docs/distributed-tracing/trace-api/introduction-trace-api/) (including [telemetry integrations](/docs/integrations/open-source-telemetry-integrations/) that use this API), replace the `https://trace-api.newrelic.com/trace/v1` endpoint with `https://gov-trace-api.newrelic.com/trace/v1`.
```
rhetoric101, line 532: I wonder if this final sentence should be made into a third bullet, since the agents also send trace data to the Trace API. Maybe you need to clarify that this is for non-agent apps. How about something like this?

* To ensure FedRAMP compliance if you have custom tracing apps or telemetry integrations that send data to the Trace API, replace the endpoint `https://trace-api.newrelic.com/trace/v1` with `https://gov-trace-api.newrelic.com/trace/v1`.

Also, that table of contents page for the telemetry integrations isn't very user-friendly. Although our main page telemetry page isn't a lot better, it's at least a bit more conventional: /docs/integrations/open-source-telemetry-integrations/get-started/introduction-new-relics-open-source-telemetry-integrations. I could go either way on that.

src/content/docs/security/security-privacy/compliance/regulatory-audits-new-relic-services.mdx
```diff
@@ -241,7 +241,7 @@ New Relic's time frames for supported regulatory frameworks and annual audits in
 
     <tr>
       <td>
-        Mobile monitoring
+        [Mobile monitoring](/docs/security/security-privacy/compliance/fedramp-compliant-endpoints/#mobile-endpoints)
```
rhetoric101, line 244: We're not consistent here because we use "Mobile monitoring" in this document but "Mobile agent" over in the endpoints doc. Maybe change to "Mobile monitoring agents"? Does this matter?

src/content/docs/security/security-privacy/compliance/regulatory-audits-new-relic-services.mdx
```diff
@@ -361,7 +361,7 @@ New Relic's time frames for supported regulatory frameworks and annual audits in
 
     <tr>
       <td>
-        Trace API
+        [Trace API](/docs/security/security-privacy/compliance/fedramp-compliant-endpoints/#trace-api)
```
rhetoric101, line 364: On my local instance, I couldn't load the URL with the anchor; it would only go to the top of the page: /docs/security/security-privacy/compliance/fedramp-compliant-endpoints/#trace-api

I eventually would get this error: Unhandled Runtime Error. You might try this out on your box.

rhetoric101 (commented):
I wonder if we need to have at least one mention of FedRAMP compliance over in the distributed tracing docs. How about a reference over in "How it works"? We could send users over to the compliance stuff.

src/content/docs/security/security-privacy/compliance/regulatory-audits-new-relic-services.mdx
zuluecho9: yeah that's weird I don't get why that was like that.

src/content/docs/security/security-privacy/compliance/fedramp-compliant-endpoints.mdx
zuluecho9: actually the agents don't use the API path, in my understanding. But I"m not exactly clear on that. My understanding is that they have their own path, which is why traces sent via the agents were previously fedramp compliant before the Trace API (and things that use it) were fedramp compliant.

zuluecho9: but maybe that's what you were saying, actually.

src/content/docs/security/security-privacy/compliance/regulatory-audits-new-relic-services.mdx
zuluecho9: that's weird. I can't imagine why it wouldn't work. it's working on public docs.

src/content/docs/security/security-privacy/compliance/regulatory-audits-new-relic-services.mdx
zuluecho9: it probably doesn't matter but I changed to align to 'mobile agents'

zuluecho9:
> I wonder if we need to have at least one mention of FedRAMP compliance over in the distributed tracing docs. How about a reference over in "How it works"? We could send users over to the compliance stuff.

I'd say maybe but the current truth is that we don't mention fedramp in our 'enable' docs for basically any other solutions that I know of. And I think the reason is that, if a customer needs fedramp, they will find the fedramp doc, and see what data they can use/enable. So in that sense, I think it makes sense as is. Otherwise, if we added it to this one, we'd have to add it to all the fedramp compliant tools and the fact that I've never heard a complaint about this makes me feel that it's working as described above. In short, the 'fedramp' would come first for customers who need it and they'd find that doc.

Issue #1561, opened 2021-03-26
whats-new-ids.json changes needs local build to update prior to submitting a PR


austin-schaefer:
### Description

Every other part of the site gets updated when we merge from develop to main as part of the build process. But there's a totally bizarre thing where whats-new-ids.json doesn't get updated unless you build _locally_, first.

This has bit us a couple times, most notably with a fix @mmfred did to fix a duplicate post earlier today as she didn't build locally before merging to develop.

### Steps to reproduce

1. Create or delete a WN post
2. Build locally
3. Notice that this file changes: [what's-new-ids.json](https://github.com/newrelic/docs-website/blob/develop/src/data/whats-new-ids.json) 
4. merge to develop 
5. push to main

1. Create or delete a WN post
2. DO NOT Build locally
3. merge to develop 
4. push to main
3. Notice that this file doesn't change: [what's-new-ids.json](https://github.com/newrelic/docs-website/blob/develop/src/data/whats-new-ids.json) 

### Expected behavior

local build shouldn't be needed to update [what's-new-ids.json](https://github.com/newrelic/docs-website/blob/develop/src/data/whats-new-ids.json)  and that build process that runs when the site deploys, should detect these changes and update [what's-new-ids.json](https://github.com/newrelic/docs-website/blob/develop/src/data/whats-new-ids.json)

austin-schaefer:
@jpvajda @zstix : Odd bug we might want to look at. Not a dealbreaker but it's led to some confusing issues with WN being out of sync. Perhaps more importantly, it's a "gotcha" that's (AFAIK) the only exception to the rule that the process of building locally shouldn't affect the actual build/publish process.

jpvajda:
@austin-schaefer thanks for reporting this, we'll consider it for Sprint 1 as one of our primary focuses in the next  few months is reducing contributor toil.

jpvajda:
@austin-schaefer It's not clear exactly the issue here can you clarify the following?


    1. Create or delete a WN post
    2. Build locally
    3. Notice that this file changes for some reason 
> what file? do you mean the [what's-new-ids.json](https://github.com/newrelic/docs-website/blob/develop/src/data/whats-new-ids.json) Also, is this what you'd expect as you just modified a file that relates to this JSON file.
    Discard that change, open a PR, and try merging
    Notice that the json file doesn't get updated

> ok, but that's because you discarded the change.

So what is your are expecting to happen here?

jpvajda:
[...]

PR #2008, opened 2021-04-26
Create whats new ID workflow

LizBaker:
### Give us some context

This PR adds a workflow to generate whats new IDs used for the `https://docs.newrelic.com/api/nr1/content/nr1-announcements.json` endpoint. Previously we generated these during the build, which meant new IDs lived in memory until someone ran a local build and committed those changes. This will get rid of the process on build and run it in a workflow that occurs on PRs to main. This way new IDs should get committed as necessary on each release

Closes https://github.com/newrelic/docs-website/issues/1561

### DIFF
--- a/.github/workflows/build-with-swiftype-content.yml
+++ b/.github/workflows/build-with-swiftype-content.yml
@@ -107,7 +107,7 @@ jobs:
               },
               restrictions: {
                 users: [],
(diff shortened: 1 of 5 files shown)


aswanson-nr (approved):
LGTM

PR #2019, opened 2021-04-27
chore: removed deployment marker

jpvajda:
This PR removes the new relic deployment marker as it doesn't work for non-APM applications.


aswanson-nr (approved):
:shipit:

PR #2027, opened 2021-04-27
Update react-spring and theme versions

LizBaker:
### Give us some context

Updates the react spring version and the theme (which also got a react spring version update) which should fix issues we were running into during local development with any components that had animation (nav, collapers, etc)


aswanson-nr (approved):
LGTM :shipit:

PR #2045, opened 2021-04-28
Reverting back to the opensource bot token

zstix:
## Description
We recently attempted to update the Github tokens used in our workflows (#2037 and #2030). It turns out that the original token we had in place may have been accurate, but the bot user didn't have the right _permission_ in the repository. Since neither of the token updates we did yesterday worked, I'm reverting those updates and will explore giving the bot the correct permissions.


nr-kkenney (approved):
lgtm

PR #2048, opened 2021-04-28
Revert token for swiftype workflow

zstix:
## Description
Like we did in #2045, we'd like to keep using the `OPENSOURCE_BOT_TOKEN` for our workflows that require altering branch protection.


nr-kkenney (approved):
lgtm

Issue #2063, opened 2021-04-29
Store files for the Docs site in the github repo


jpvajda:
### Issue

the docs site had a few cases where actual files were being stored on the old drupal site for download, so for this case we'd want to provide an S3 bucket (or similar) where a file could be uploaded and referenced in a doc page.  I'm of the opinion that we shouldn't be storing files in our github repo, and that file storage should be left to a service like S3. 

**Temporary approach:** Store the file in the repo
**Longer term approach**: TBD

### Acceptance Criteria 

- [x] Decide where to store the files 
- [x] Setup File storage directory
- [x] Provide documentation on how to upload and reference a file in a doc

jpvajda:
Temporarily this will file will be stored in the site repository if we find we are in need of adding more files to the docs site, then we'll want to look at a longer term solution like S3.

jpvajda:
@austin-schaefer 👋  to clarify if other cases exist for file storage longer term.

jpvajda:
@LizBaker could you add some points to this so we can account for it?

PR #2069, opened 2021-04-29
Add FedRAMP pdf

LizBaker:
### Give us some context

This adds a PDF we need to link to from a doc. We're still figuring out how we want to deal with adding PDFs but for now it's going in the repo (https://github.com/newrelic/docs-website/issues/2063)
By adding the file to the directory that the doc is in we can simply link to it with `[link](filename.pdf)` in that doc, so in this case `[FedRAMP PDF](New-Relic-FedRAMP-Customer-Responsibility-Matrix.pdf)` from within regulatory-audits-new-relic-services.mdx


mmfred:
@barbnewrelic - I believe you're working on this. I'm going to leave this for you to address.

LizBaker:
also @barbnewrelic feel free to use this same branch to add the link where you need it. There is one caveat and it's that when you run it locally you'll get a 404 upon clicking the link. if you refresh that 404 page it will then load the pdf. 

this _wont_ happen once its live (and the PR preview will work correctly as well), it's just something that doesn't quite load with a local build

nr-kkenney (approved):
lgtm

jpvajda:
Confirming with the Docs Team this is OK to merge.

PR #2075, opened 2021-04-30
Fix wrong definitions displayed in data dictionary

roadlittledawn:
## Description

Adds filtered attributes via query params into state as well to fully render the attribute and its data. Also makes keys unique.

## Screenshot

TBD


aswanson-nr (approved):
:shipit:

PR #2078, opened 2021-04-30
Fix anchor in Go attributes documentation

RichVanderwal:
This anchor was a duplicate of the one above it. I've adjusted the name so it'll link to the correct section.


mmfred:
thanks for the fix @RichVanderwal

PR #2079, opened 2021-04-30
PR for testing: DO NOT MERGE

rhetoric101:
I ran my redirect script on agent files, moved the agents.yml tree in the FSO.yml file.


src/nav/full-stack-observability.yml
```diff
@@ -177,6 +177,1191 @@ pages:
             path: /docs/apm/new-relic-apm/getting-started/view-app-alert-information-apm
           - title: APM data security
             path: /docs/apm/new-relic-apm/getting-started/apm-agent-data-security
+      - title: Manage APM agents
+        pages:
+          - title: Agent data
+            pages:
+              - title: Real time streaming
+                path: /docs/agents/manage-apm-agents/agent-data/real-time-streaming
```
LizBaker, line 185: each of these paths under the agents nav will need to be updated from `/docs/agents` to `/docs/apm/agents` and that should solve the missing left nav issue

rhetoric101: @LizBaker, so are you saying that these navigation links don't invoke the redirects the same way that redirects are invoked from within content files?

src/content/docs/apm/agents/c-sdk/index.mdx
```diff
@@ -0,0 +1,80 @@
+---
+title: C SDK
+type: landingPage
+tags:
+  - Agents
+  - C SDK
+redirects:
+  - /docs/agents/c-sdk/index
```
LizBaker, line 8: this needs to be `/docs/agents/c-sdk` so just deleting the `index` on the end should fix the 404s. this needs to happen in each of the agent index.mdx files

src/nav/full-stack-observability.yml
roadlittledawn: @rhetoric101  the nav and redirects are kinda of separate things. when we attach nav menus to each node (aka, .mdx file in this context), we take the node's URL path (which is generated based on where it is in the file system) and we look for that URL path in each nav yaml file in `src/nav`. if we find a `path` value in a yaml file that matches the node's URL path, we attach that nav yaml menu to the node to display in the left nav. 

so since the agents nav items were brought into `full-stack-observability.yml` but the paths reflect all the nodes' old URLs / file paths, it's not going to attach a nav menu to the new nodes' paths.

so like Liz said, if you update all those `path` values to reflect the current URL / file paths for the agent `.mdx` files, that should attach the FSO nav to them all.

happy to hop on a call too if it's easier to walk through an example.

rhetoric101: @roadlittledawn, thanks for the clarification! I switched out all those old navigation prefixes with the new one and this worked!!

src/content/docs/apm/agents/c-sdk/index.mdx
rhetoric101: @LizBaker, just as you said, when I removed these "index" references, I was able to build and see the new changes!

jpvajda:
@aswanson-nr just a reminder to add some points to this once you finish up so we can account for the effort in sprint 3. 🎫

jpvajda:
This PR is being closed in favor of #2119  I'll keep it in our sprint as we did 3pts of work on it, and it's good to keep track of that effort. cc @zstix

PR #2112, opened 2021-05-04
fix: RightRail rendering list items with the same href

rudouglas:
The `Mentioning users...` section had the same `href` as `Create description...NRQL` , so just edited the reference in the md page


zstix (approved):

dbarnesbrownNR (approved):
Looks good!

PR #2119, opened 2021-05-04
DRAFT (DO NOT MERGE) Second attempt to move agents to apm.

rhetoric101:
In this test, I ran the latest version of the script that now removes index files, too. Also, I kept the Agent node when I moved it to APM this time.


rhetoric101:
Here's an update: I received feedback that maybe we should keep the "APM agents" quick-link in the left navigation. 

Three issues came up with my solution:

* Now that the agent navigation tree is in the FSO tree, I don't want to keep a separate tree for agents, because that takes users out of the flow of the FSO doc navigation. To get around this, I restored the `agents.yml` file and removed all the lines except the first two. This seems to work: it creates a link to a my new page:  `/docs/apm/agents/intro-to-apm-agents.mdx`.  Is it acceptable for me to gut the `agents.yml` file just to create a quick link?
* It looks like my creation of a deep directory structure in the FSO navigation file breaks the table-of-contents pages invoked when you click the button to view all docs for a section (in this case APM). The error looks like it only wants two levels in the navigation (see the **View all** button at the bottom of this page: `http://localhost:8000/docs/apm/` ):
```
Error in function eval in ./src/components/IndexContents.js:26
Cannot read property 'length' of undefined

./src/components/IndexContents.js:26
Open in Editor
 24 | const TableOfContents = ({ root, depth = 2 }) => {
  25 |   const { title, url, pages } = root;
> 26 |   const idx = pages.findIndex((page) => page.pages.length);
     |                                                   ^
  27 |   const Heading = `h${depth}`;
  28 | 
  29 |   const links = idx === -1 ? pages : pages.slice(0, idx);
```
* Also, my new page at `/docs/apm/agents/intro-to-apm-agents.mdx` does not trigger the left navigation list, even though it is in the navigation file for FSO. Any ideas why?

src/nav/agents.yml
```diff
@@ -1,1188 +1,2 @@
...
-            path: /docs/agents/ruby-agent/troubleshooting/passenger-troubleshooting
-          - title: No data with Unicorn
-            path: /docs/agents/ruby-agent/troubleshooting/no-data-unicorn
-          - title: Update deprecated API calls
-            path: /docs/agents/ruby-agent/troubleshooting/update-deprecated-api-calls
-          - title: Update private API calls to public Tracer API
-            path: /docs/agents/ruby-agent/troubleshooting/update-private-api-calls-public-tracer-api
-          - title: Not installing Grape
-            path: /docs/agents/ruby-agent/troubleshooting/not-installing-new-relic-supported-grape
+path: /docs/apm/agents/intro-to-apm-agents
```
aswanson-nr, line 2: I'm not 100% sure, since the builds take forever, but I think this is the line that's causing the build to break. You probably want to delete this file completely!

aswanson-nr: Just realized you added it back in! My understanding of the nav pages is that they also need the `pages` field.

jpvajda:
@rhetoric101 well there goes all your Github Stats 😉 📈

PR #2123, opened 2021-05-05
Add RSS feed for security bulletins

roadlittledawn:
## Description

Adds local plugin to generate RSS feed endpoint for security bulletin content. The query gets all `.mdx` files (excluding any index.mdx file) in a directory named `security-bulletins` anywhere inside of `src/content/docs/security`. The feed path is inferred by looking at the first node (all security bulletin .mdx files should be in same directory) and automatically setting the feed to correspond with the directory all the files are in.

so if that directory and its contents are moved around anywhere inside of `docs/security` the plugin shouldn't need updating to work.

per Austin's feedback, i just added the RSS icon/link to two .mdx files, because there's nothing in Gatsby right now to distinguish security bulletin content from any other basic page.

if all bulletins are in:

```
docs/security/new-relic-security/security-bulletins
```

then the feed path will be:

```
docs/security/new-relic-security/security-bulletins/feed.xml
```

## Related issues
Closes #2123 

## Screenshots
Screen Shot 2021-05-05 at 15 09 49

Screen Shot 2021-05-05 at 15 09 40


plugins/gatsby-plugin-security-bulletins-rss/gatsby-node.js
```diff
@@ -0,0 +1,130 @@
...
+  const { title, metaDescription } = frontmatter;
+
+  const transformedAST = htmlGenerator.runSync(mdxAST);
+  const html = htmlGenerator.stringify(toHast(transformedAST));
+
+  // time is necessary for RSS validity
+  const date = new Date();
+  const pubDate = `${format(date, 'EE, dd LLL yyyy')} 00:00:00 +0000`;
+  const link = new URL(slug, siteMetadata.siteUrl).href;
+  const id = Buffer.from(`${title}`).toString('base64');
```
zstix, line 58: We might want to add the date to help insure uniqueness:
```suggestion
  const id = Buffer.from(`${date}-${title}`).toString('base64');
```

zstix (commented):
One small comment, but this otherwise looks great!

plugins/gatsby-plugin-security-bulletins-rss/gatsby-node.js
roadlittledawn: unfortunately this date is going to reflect the current date/time when it builds, so it'll change each time which seemed bad. additionally, the `title` for most bulletins contains a unique ID for a finding. for example: `Security Bulletin NR20-01` which is two digit year and the number of the finding for that year.

roadlittledawn: i originally had it using the file `birthTime`, but since the builds copy files over those dates will also always reflect the date/time of the build. i couldn't find any other way to reference the date/time the file was created, short of adding it for frontmatter.

zstix (approved):

roadlittledawn:
**Note to self**
If we see any issues re: the publish date in feed readers, we _could_ add the original creation/publication date for each of the bulletins since it’s in the body of all but couple files. for the others i could look at Drupal. that would be a little extra onus on content contributors to add that date to the frontmatter.

PR #2126, opened 2021-05-05
feat(nri-prometheus): updated docs with new config option

paologallinaharbur:
Following this change https://github.com/newrelic/nri-prometheus/pull/148 and in https://github.com/newrelic/helm-charts/pull/359 we need to update the documentation


dbarnesbrownNR (approved):
Looks good, thank you!

PR #2131, opened 2021-05-05
Fixed headings

dbarnesbrownNR:
H2 still refers specifically to "aggregator functions" but not all of them are aggregator functions. Should split list into aggregator and non aggregator at some point also.


rhetoric101 (approved):
@dbarnesbrownNR, these changes look good!

PR #2149, opened 2021-05-06
Fix a spelling error

djzevenbergen:
Infrastrucute, while a fun word, is probably not the correct terminology here


dbarnesbrownNR (approved):
Thanks, approved!

PR #2150, opened 2021-05-06
Update fluentd doc layout and eu endpoint examples

Rilsta:
Setting the EU logging endpoint was listed as an optional configuration using the base_uri property. EU customers must manually set this to the EU endpoint to get logs from our Fluentd plugin. I used the collapsable examples in the previous config steps but with the base_uri value added to the <match **> tags. I figured adding the whole config would be better since EU customers will likely just copy and paste the whole code snippet.

We also mentioned the github examples twice, but the github examples do not show how or where to set the base_uri property in the fluentd config. I removed the second link to the example and put the first one in a "Tip" box prior to config to make it more noticeable.

### Give us some context


barbnewrelic (approved):
Thanks so much for updating the FluentD doc. Looks great!

PR #2151, opened 2021-05-06
Add Heroku Log Streaming and Syslog

Rilsta:
Heroku log streaming and the syslog TCP endpoint were missing from the list of available log forwarding solutions.


dbarnesbrownNR:
Awaiting another team member's review on the logs component of the edit.

PR #2175, opened 2021-05-06
Small fixes to user mgmt docs

zuluecho9:
Some quick fixes based on feedback, per liaison work: 

* Addition of warning to user mgmt docs section about avoiding multiple admins being in UI at same time. 
* Addition of pertinent billing factor to billing section


dbarnesbrownNR (approved):
Good to merge when you're ready. If you want to you could change "access grants" to "granting access" but it works either way I think.

Issue #2193, opened 2021-05-08
Feedback: addPageAction (Browser agent API)


LeslieMurphy:
### Tell us more about your feedback

Sorry to hear this doc wasn't helpful, but we'd love to make it better! We've 
recorded your feedback. Help us improve this content by providing more info:

* How can we improve this doc?
TRY ME LINK IS BROKEN.

Example does not show what JS to include such that var newrelic is defined so I can use newrelic.addPageAction

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/browser/new-relic-browser/browser-agent-spa-api/add-page-action/
* **OS**: Windows 10
* **Browser**: Edge 90.0.818.51
* **Device**: Unknown

urbiz-nr:
@LeslieMurphy, thanks for reporting this! We'll be fixing it soon, you can check the details in #2196.

PR #2196, opened 2021-05-10
Fix code block example to display a tag and not render html link

roadlittledawn:
## Description

Adds spaces to `<a>` tag in code block to prevent parsing and rendering as HTML link. This is supposed to be part of the example and not rendered as a link, like we do in other code blocks.

## Related issues
Closes https://github.com/newrelic/docs-website/issues/2193

## Example
The first example in this doc has a link that should just show the `<a>` code as is: http://localhost:8000/docs/browser/new-relic-browser/browser-agent-spa-api/add-page-action/#examples

## Screenshots
Screen Shot 2021-05-10 at 09 00 33


bradleycamacho (approved):
Thanks @roadlittledawn, merging in now!

PR #2201, opened 2021-05-10
DOC-6812: Standardizing on a single guided install messaging pattern

paperclypse:
This PR updates a bunch of docs and standardizes on a single pattern for promoting and linking to the guided install UI.


bradleycamacho (approved):
These all look good to me! Ready to merge @paperclypse

PR #2210, opened 2021-05-10
Notice that building site locally for the first time will take a while

bradleycamacho:
A SME suggested we mention that building the site locally can take upwards of 20-30 minutes the first time.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,6 +52,8 @@ yarn start
 
 Your site is now running at [`http://localhost:8000`](http://localhost:8000)!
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -52,6 +52,8 @@ yarn start
 
 Your site is now running at [`http://localhost:8000`](http://localhost:8000)!
 
+The first time you build the site locally, it will take upwards of 20 minutes. This is normal, and will take signifigantly less time afterwards.
```
paperclypse, line 55: You could just do "it often takes 20+ minutes". Also I think you can lose the comma from the second sentence.

paperclypse (approved):
Looks good! You can take or leave my comment of course :D

PR #2214, opened 2021-05-10
Update 'APM metric alert conditions' to reflect improved features

founddrama:
### Give us some context

* Documentation outlined steps for adding NRQL conditions for External Service metrics with some copy-and-paste action from APM
* We can now create NRQL conditions from those charts directly with the "Create alert condition" chart menu option
* ...so that streamlines the process and these changes address that.


urbiz-nr (approved):
Thanks for updating this doc, @founddrama - looks really good!

urbiz-nr:
Tried locally, merging.

Issue #2217, opened 2021-05-11
Issue: NerdGraph tutorial: View and configure an entity's golden metrics


jsbnr:
The last query for reseting golden tags is invalid. the 'key' attribute needs to be removed from the outer 'tags' group.

(at least for me it thre an error with this)

urbiz-nr:
Thanks for getting in touch, @jsbnr - which document are you referring to?

urbiz-nr:
Thanks, we'll check it out!

PR #2219, opened 2021-05-11
Fix the GM reset query

naxhh:
### Give us some context

Solves #2217

There's an extra `key` in the query


urbiz-nr (approved):
Thanks for the fix @naxhh!

PR #2227, opened 2021-05-11
Added missing redirect

bradleycamacho:
An internal contributor pointed out a google search result pointed to a 404. This PR adds a redirect for the page to this doc.


paperclypse (approved):
Got the redirect to work in my local build... for the first time? Anyway, looks good!

PR #2237, opened 2021-05-11
chore: update muting rules docs to reflect preferred nomenclature

founddrama:
### Give us some context

* When muting rules was first launched, we did use `tag.` as the prefix
* ...but that's been updated to accept `tags.` _and_ `tag.` (with the former being the preferred naming)
* This update reflect that.


barbnewrelic (approved):
I corrected the anchor on line 484, but everything else looks great. Thanks!

PR #2243, opened 2021-05-12
Bb doc6671 fed ramp pdf

barbnewrelic:
TW: When you peer review this, please let me know whether you can see the PDF. The slack channel thread I copied into DOC-6900 says it should load but I can't still can't see it. Thanks!


src/content/docs/logs/log-management/troubleshooting/no-log-data-appears-ui.mdx
```diff
@@ -84,11 +84,11 @@ If no data appears after you send some log payloads and wait about five minutes,
 
     <tr>
       <td>
-        Checking for Data
+        Checking for data
       </td>
 
       <td>
-        Try querying the `Log` data type using Insights. For example, you can use a [NRQL](/docs/query-data/nrql-new-relic-query-language/getting-started/introduction-nrql) query to look at everything in the Log eventType. If no data appears in Insights, then no data will appear in the Logs UI.
+        Try querying the `Log` data type. For example, you can use the [NRQL query builder](https://docs.newrelic.com/docs/query-your-data/explore-query-data/query-builder/introduction-query-builder/) to look at everything in the `Log` eventType. If no data appears in the query builder, then no data will appear in the Logs UI.
```
urbiz-nr, line 91: Use relative link for docs.

urbiz-nr (dismissed):
Check minor comment, other than that looks good to merge!

barbnewrelic:
I made some additional edits based on previous peer review. It looks like GH wants one more official review. I'd feel better if that was another TW not myself. Thanks again1

LizBaker:
looks like this branch doesn't have the latest from develop so it's missing the PDF, i merged it locally and was able to run it. if you'd like you can merge develop into this branch and get a fresh PR preview to see it working, or if you're feeling spicy you can assume it will work once merged in since it will have access to the file then. i feel 99% confident it will 😝

barbnewrelic:
If at first you don't succeed... :-O anyway I have added an images folder, added the PDF, and corrected the hyperlink. It still doesn't work in my localbuild but I hope that someone else can verify this is correct. Thanks again!

bradleycamacho (approved):
Pushed up a small commit, but approved otherwise!

LizBaker:
@barbnewrelic no need to add an images folder. PDF files go directly in the same folder that the mdx file is in. so in this case that compliance folder. however, the PDF is already in develop, your branch just doesn't have the latest changes so you don't see the PDF. i would delete the images folder so we don't have two PDF files getting checked in

bradleycamacho (approved):
Looks good!

PR #2251, opened 2021-05-12
Add pie chart troubleshooting to pie chart section

zuluecho9:
When doing DOC-6819, Insights cat deprecation, realized I'd forgot to add back this 'pie chart percentage' doc into the pie chart section.


paperclypse (approved):
Looks good!

Issue #2007, opened 2021-04-26
 Follow up work to improve the navigation sort order for index pages.


jpvajda:
### Related

#1844 
#1585

###  User Story 

As a user  if I go to this[ billing page]( https://docs.newrelic.com/docs/accounts/accounts-billing/new-relic-one-pricing-billing/) I should see this navigation, which is essentially the nav structure found in accounts.yml but I see the navigation “expanded” in the section I am actually in. More importantly the Index view of the page reflects the exact URLS that exist in this nav structure. In this approach, no matter where  I land on any index page, I always be shown the the primary navigation of where that page is found, where I am actually at in the navigation and the URLS for the index section.

### Mock Up

![image (1)](https://user-images.githubusercontent.com/1395158/117517852-9f550a00-af5a-11eb-87df-9bae2c65de6a.png)

### Problem 

We always want to determine the correct navigation to show on the left side and then we want to determine the correct part of the navigation to show as the main page content for that particular index page so the user has clear indicators of where they are in the navigation hierarchy. 

### Acceptance Criteria 

- [x] When a user traverses navigation, regardless of the level of navigation they are at, the navigation should always reflect what doc they are viewing so it's clear to them where they are in the Nav Hierarchy
- [x] Sort order on the index pages matches that in navigation file
- [x] Always show the correct docs in the navigation file.
- [x] Always have a 1:1 mapping between the contents of the left nav and the contents of the index pages
- [x] if a new page is added to that index directory and there is no corresponding nav item added to a yml file, then this page would NOT have a corresponding nav item. note: This is a manual process the Docs Team enforces, so we don't need to try to automate this in this ticket).
- [x] please work with @zuluecho9 for acceptance of this change as he has deep knowledge of the expected experience here
- [x] ensure documentation is updated in the style guide to explain how this works. @zuluecho9

jpvajda:
This continues to generate a lot of issues and complaints of user experience for navigation so let's look to try to continue to improve this in Sprint 4, if we have bandwidth.

jpvajda:
Updated the AC based on our chat with @zuluecho9 and @austin-schaefer .. cc @zstix

jpvajda:
We need to clarify exactly how this work for a case such as: https://docs.newrelic.com/docs/integrations/amazon-integrations/aws-integrations-list/

jpvajda:
I'm working through this with a little internal doc and will update the AC when we get consensus on how to proceed.

[...]

PR #2260, opened 2021-05-12
Use nav yaml to generate all index page content

roadlittledawn:
## Description

Uses nav yaml file data to generate content on all auto generated index pages (similar to `/table-of-contents` pages).

If we attached the correct nav yaml data for an index page, but we only want to show just the section of the nav that matches where we are in the URL path, we look through the nav object and pull out the object where the `url` matches the slug of the page we're on.

If an index page path is used in a nav yaml file where the top level path differs from the beginning of its path, we try to find the nav yaml file that matches it and use that.

If an `.mdx` file or a file directory are not referenced in any nav yaml file, we cannot attach a nav menu to them, so the left nav is empty when viewing them. Examples include:
- https://docs.newerlic.com/docs/plugins
- https://docs.newrelic.com/docs/plugins/plugin-developer-resources/develop-plugins/checklist-developing-plugins/

If a file directory with content exists, but is not referenced in any nav YAML file, we fall back to the old behavior of creating auto index page and populating that with the file system data via `generateHtml()`.

## Docs
Style guide doc updated for this: https://github.com/newrelic/docs-website/pull/2260/files#diff-43de305ddfb201cf81b0cee4a20f4e700d6f154847bbbd6eddb91be29d6ef51e

## Related issues / PRs
Closes https://github.com/newrelic/docs-website/issues/2007

Follow up issue to create auto index pages for all nav paths: https://github.com/newrelic/docs-website/issues/2340

## Screenshots

[AWS integration list index page](https://docs.newrelic.com/docs/integrations/amazon-integrations/aws-integrations-list/)

Screen Shot 2021-05-18 at 12 28 25

[NR1 pricing and billing page](https://docs.newrelic.com/docs/accounts/accounts-billing/new-relic-one-pricing-billing/new-relic-one-pricing-billing/)

Screen Shot 2021-05-18 at 13 12 11

## Reviewer notes
- We may be able to refactor to combine `indexPage` and `tableOfContents` templates, but a lot of the old functionality is in place as a fallback in case a file directory with content exists, but is not present anywhere in the nav.
- If  a path exists in a nav YAML file that `gatsby-plugin-auto-index-pages` did not create a page for, it currently cannot get an index page. Upon digging in a bit, I'm a [little leery of supporting that](https://github.com/newrelic/docs-website/pull/2260#issuecomment-843640205).


roadlittledawn:
build failed due to `/docs/mobile-apps` being an index page used in `new-relic-in-practice.yml` with a top level path that uses `/docs/using-new-relic`. there isn't a nav yaml that uses a top level path of `/docs/mobile-apps`, there's only a file directory for that. and i imagine there will likely be several instances like this where the top level URL path of a given index page references a file directory that doesn't match 1:1 with a nav yaml file.

not sure i currently see a way to distinguish between:
- an index page menu item that actually belongs to a nav yaml menu that doesn't share the same top level path that when viewed shows the nav menu that's referencing it 
VS
- an index page menu item that actually belongs to another nav yaml menu that when viewed, shows  its "true" nav menu

roadlittledawn:
re: trying to avoid the need to add paths for all sections in nav yaml files. it's ideal to have those path values in the nav yaml because:
- i found many instances where within a given section the URL paths don't match the top level nav path (ex. linking to an index page in another top level nav), so nav menus would be incomplete in many cases if all we can do is try to match with a menu item's `url` value including the `location.pathname`.
- nav item URL paths must point to an actual folder path. the `gatsby-plugin-auto-index-pages` looks at the file directories when determining if it should create a page and if so sets the URL path based on the file directory path.

roadlittledawn:
turns out the work i had locally to address creating pages for nav paths not found in `.mdx` or `.md` file paths worked, but didn't quite work while running local dev server.

looking at [how many such paths there](https://gist.github.com/roadlittledawn/ff98fbc36632ee18ed740a23f48c429e) are in this branch's nav YAML files, there's 156 paths. but on a quick inspection, many paths already have pages created via other `gatsby-node.js` code. examples:
- /attribute-dictionary
- /nru
- /docs/release-notes/agent-release-notes/java-release-notes/current

not yet quite sure how we check to ensure a page already exists at a given path. and also not quite sure we really want any ol' path added to a YAML file to drive page creation.

zuluecho9 (approved):
I check this out locally and also Austin Schaefer reviewed it, and we say go for it.

aswanson-nr (approved):
:shipit:

zstix (approved):
This looks great, nice work!

roadlittledawn:
merging without review this time. only change was to resolve merge conflict on one line in a yaml file.

PR #2278, opened 2021-05-13
DOC-6870 - Sorted NRQL functions into aggregate and non-aggregate

x8a:


zuluecho9:
I'll take peer edit on this one.

dbarnesbrownNR (approved):
Looks great! Might need to move histogram but I am waiting on info so let's publish and I can always move that one later if needed.

PR #2279, opened 2021-05-13
chore: changed Trace API link

rudouglas:
This link was leading to a `404`


x8a (approved):
It looks great! Thanks for spotting it 🙌🏻

PR #2281, opened 2021-05-13
Fix broken link to the Ingest APIs overview page

wulczer:


x8a (approved):
Thanks for spotting it!

PR #2308, opened 2021-05-14
Fix redirect loop on /docs/telemetry-data-platform/manage-data/manage…

austin-schaefer:
Fixes redirect loop reported in Documentation channel


austin-schaefer:
@x8a @urbiz-nr : Are either of you able to review this small PR? I'd like to get this live with the Monday deploy if possible.

bradleycamacho:
@austin-schaefer Looking at this now. I can do a small release since redirect loops are very annoying.

PR #2309, opened 2021-05-16
chore: fixed incorrect link

rudouglas:
The link for Trace API was leading to a 404


urbiz-nr (approved):
Thanks for the fix @rudouglas!

PR #2329, opened 2021-05-18
Add mapping tables plus real examples in insights api migration guide

hredondo:
### What

* Adding some mapping tables in the Insights API -> NR1 dashboards API migration doc
  * Also updating the new dashboards API examples so people can copy/paste and start playing around
* Adding some callouts to drive people to the migration doc in case they are coming from Insights

### Why?

* In the context of the Insights dashboards API being EOLed soon, help Insights API users migrate to the new dashboards API

image

image

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # docs.newrelic.com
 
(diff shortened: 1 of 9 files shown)


zuluecho9:
@paperclypse I checked this one out and made some edits (including moving general features to the main migration doc so the API doc was focused on API, and making better cross linking between docs). I think it looks good but maybe could use a quick check.

paperclypse (approved):
Looks good!

PR #2332, opened 2021-05-18
Keyboard shortcut that no longer works

austin-schaefer:
The `h` shortcut doesn't work on the new site (to my chagrin)


paperclypse:
To my chagrin also, @austin-schaefer

paperclypse (approved):
Not sure we need quotation marks around "anchor" but otherwise I'm good with this.

PR #2333, opened 2021-05-18
Fixing some missing TDP nav file categories

zuluecho9:
Just reinstituting what should have been there to begin with.


paperclypse (approved):
Looks good!

PR #2336, opened 2021-05-18
Update JavaScriptError Attributes Definitions

krede-nr:
### Give us some context
These changes are to improve documentation around the JavaScriptError event and its attributes.


paperclypse (approved):
Looks good!

PR #2342, opened 2021-05-18
Added a tip related to policy condition deletion

paperclypse:
New Relic account admins can remove the ability to delete a policy condition. I thought this collapser section could use a tip to that effect.


urbiz-nr (approved):
Looks good!

PR #2346, opened 2021-05-19
Added Fedramp exception

urbiz-nr:
Added a note about gov-bam.nr-data.net. See DOC-6867 for more details.


obarney:
@urbiz-nr depending on how we change this I think the text can be shortened to:

> * if you are using FedRAMP-compliant endpoints

obarney:
I was thinking a box on the left side but open to your ideas!

Screen Shot 2021-05-20 at 7 47 29 AM

zuluecho9 (approved):

PR #2373, opened 2021-05-20
Removing logging dependecies that are not needed anymore from docs

danybmx:
Hi!

This PR removes some dependencies from the docs that are not needed anymore on the infra-agent for enable logging:

- Microsoft Visual C++ Redistributable is not needed anymore since version 1.12.4 for Windows.


paperclypse (approved):
Looks good!

Issue #2388, opened 2021-05-21
Number of integrations on home page does not match


jerelmiller:
It appears a nav item was added to the root nav with the text `See our 360+ integrations`. Further down on the page in the `New Relic Integrations` section, it says `370+ integrations`. These numbers should probably match since they link to the same page.

## Screenshots

<img width="1610" alt="Screen Shot 2021-05-20 at 9 06 23 PM" src="https://user-images.githubusercontent.com/565661/119080764-89f6cb80-b9af-11eb-9287-f9cd8e3ed3c6.png">

### Doc information (don't delete this section)

* https://docs.newrelic.com/
* **OS**: Unknown
* **Browser**: Unknown
* **Device**: Unknown

urbiz-nr:
We say [369 integrations](https://newrelic.com/integrations) in our site, so I'd go for +360.

urbiz-nr:
Fixed in #2391 .

PR #2391, opened 2021-05-21
Corrected amount of integrations

urbiz-nr:
The amount of available integrations didn't match in the UI, as reported in #2388. This PR corrects this.


x8a (approved):
Looks great!

PR #2401, opened 2021-05-21
DOC-6847 New issue entity map feature

paperclypse:
From what I've been told, this feature has already been released. I'm keeping this very high level and concise at this point, because there's still a lot I don't know about it.


rhetoric101 (approved):
@paperclypse, these changes look good!

PR #2408, opened 2021-05-21
Adjust Amplify build settings

zstix:
## Description
Our builds are not getting deployed on AWS Amplify. Their engineering team suspects that this is due to our builds taking longer than an hour (presumably, we are given an authorization token that expires after an hour). This change was suggested as a possible way to help with the problem while they continue to look into the problem.

### DIFF
--- a/amplify.yml
+++ b/amplify.yml
@@ -9,9 +9,9 @@ frontend:
       commands:
         - |
           if [ "${AWS_BRANCH}" = "main" ]; then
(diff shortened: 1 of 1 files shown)


roadlittledawn (approved):
:shipit:

LizBaker (approved):

nr-kkenney (approved):
looks reasonable

PR #2410, opened 2021-05-21
Pulling back Fargate beta for doc-6858

rhetoric101:
Based on PM comments, we need to pull back this Fargate beta documentation.


paperclypse (approved):
Looks good.

PR #2448, opened 2021-05-25
DOC-6891: EOL Synthetics Label API 

bradleycamacho:
This PR deletes a deprecated doc and adds redirects from the old doc to a new labels doc.


urbiz-nr (approved):
Go for it!

PR #2450, opened 2021-05-25
update supportability for tomcat and jersey

tbradellis:
Double checking TomEE. May make an additional change here.

### Give us some context
Discovered changes that will impact coming versions of tomcat, tomEE and jersey.


rhetoric101 (approved):
@tbradellis, these changes look good!

PR #2452, opened 2021-05-25
Updating requirements in 'Migrate users' doc 

zuluecho9:
Small follow up to DOC-6860, as liaison. Some additions to requirements and guidelines for using the user migration wizard.


urbiz-nr (approved):
All good!

Issue #2461, opened 2021-05-26
Issue: Monitor services running on Amazon ECS


djaffinito:
### Tell us what you need

* If you've found something inaccurate, what is it?

The suggested userdata snippets on Step 1, subsection 6 (both snippets) don't install the agent. 

Specifically this section:

```
runcmd:
  - [ systemctl, daemon-reload ]
  - [ systemctl, enable, newrelic-infra ]
  - [ systemctl, start, --no-block, newrelic-infra ]
```

Should be:

```
runcmd:
  - [ yum, install, newrelic-infra, -y ]
  - [ systemctl, daemon-reload ]
  - [ systemctl, enable, newrelic-infra.service ]
  - [ systemctl, start, --no-block, newrelic-infra.service ]
```

For both "CentOS 7, RHEL 7, Amazon Linux 2" and "For CentOS 6, RHEL 6, Amazon Linux 1" snippets

Then in Step 2, we never mention that you need to add a `runcmd` to install the desired integration. Using the nginx example that the doc is currently using, it should say to add the following to the `runcmd` section:

```
  - [ yum, install, nri-nginx, -y ]
```

So the full `runcmd` section would be:

```
runcmd:
  - [ yum, install, newrelic-infra, -y ]
  - [ yum, install, nri-nginx, -y ]
  - [ systemctl, daemon-reload ]
  - [ systemctl, enable, newrelic-infra.service ]
  - [ systemctl, start, --no-block, newrelic-infra.service ]
```

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/integrations/host-integrations/host-integrations-list/monitor-services-running-amazon-ecs/
* **OS**: Unknown
* **Browser**: Unknown
* **Device**: Unknown

djaffinito:
Also note that in the doc, our examples use `newrelic-infra` as the service name, but that changed to `newrelic-infra.service` (like in my examples above).

rhetoric101:
@djaffinito, I created PR #2462 and added your first suggestion but wasn't able to understand if the second suggestion applied to both snippets in the separate collapsers. Could you edit PR #2462 with that second set of changes?

rhetoric101:
This issue was resolved by PR #2591 (https://github.com/newrelic/docs-website/pull/2591)

PR #2462, opened 2021-05-26
Need to correct run command for ECS

rhetoric101:
Here is the request from GitHub issue 2461:

The suggested userdata snippets on Step 1, subsection 6 (both snippets) don't install the agent.

Specifically this section:

runcmd:
  - [ systemctl, daemon-reload ]
  - [ systemctl, enable, newrelic-infra ]
  - [ systemctl, start, --no-block, newrelic-infra ]
Should be:

runcmd:
  - [ yum, install, newrelic-infra, -y ]
  - [ systemctl, daemon-reload ]
  - [ systemctl, enable, newrelic-infra.service ]
  - [ systemctl, start, --no-block, newrelic-infra.service ]
For both "CentOS 7, RHEL 7, Amazon Linux 2" and "For CentOS 6, RHEL 6, Amazon Linux 1" snippets

Then in Step 2, we never mention that you need to add a runcmd to install the desired integration. Using the nginx example that the doc is currently using, it should say to add the following to the runcmd section:

  - [ yum, install, nri-nginx, -y ]
So the full runcmd section would be:

runcmd:
  - [ yum, install, newrelic-infra, -y ]
  - [ yum, install, nri-nginx, -y ]
  - [ systemctl, daemon-reload ]
  - [ systemctl, enable, newrelic-infra.service ]
  - [ systemctl, start, --no-block, newrelic-infra.service ]

### DIFF
--- a/src/content/docs/integrations/host-integrations/host-integrations-list/monitor-services-running-amazon-ecs.mdx
+++ b/src/content/docs/integrations/host-integrations/host-integrations-list/monitor-services-running-amazon-ecs.mdx
@@ -90,9 +90,10 @@ Here are instructions for changing EC2 launch configuration (taken from [Amazon 
          - newrelic-infra
          - nri-*
        runcmd:
+         - [ yum, install, newrelic-infra, -y ]
          - [ systemctl, daemon-reload ]
-         - [ systemctl, enable, newrelic-infra ]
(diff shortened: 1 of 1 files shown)


rhetoric101:
Closes issue #2461

rhetoric101:
This PR is replaced by #2591.

PR #2468, opened 2021-05-26
Chart refresh rates 

zuluecho9:
Moving content, related to changes made on recent jira DOC-6819 that I needed to follow up on. Moved 'Chart refresh rates' out of query limits doc and into its own doc.


rhetoric101 (approved):
@zuluecho9, these changes look good! Nice job removing that old redirect, too.

PR #2481, opened 2021-05-26
Edits to NRQL reference page for DOC-6863

dbarnesbrownNR:
Just publishing these edits that have been in my queue ready to go for a bit.

### DIFF
--- a/src/content/docs/telemetry-data-platform/understand-data/metric-data/query-metric-data-type.mdx
+++ b/src/content/docs/telemetry-data-platform/understand-data/metric-data/query-metric-data-type.mdx
@@ -256,9 +256,10 @@ For example, if you want to use a single value within a metric to do a compariso
     id="example-list-names-host"
     title="List all metric names for a particular host"
   >
-  This example query result will be the `average()` value function computed as total over `count`, for all values over 2. 
+
+  The average value of a metric is computed as `total` over `count`, so the following query would return metric data where the result of the `average()` value function is greater than 2.
     ```
-    FROM Metric SELECT uniques(metricName) WHERE %[total] / %[count] > 2
+    FROM Metric SELECT average(apm.service.transaction.duration) WHERE appName = 'MyApp' AND apm.service.transaction.duration[total] / apm.service.transaction.duration[count] > 2
     ```
   </Collapser>
 </CollapserGroup>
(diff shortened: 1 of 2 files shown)


src/content/docs/telemetry-data-platform/understand-data/metric-data/query-metric-data-type.mdx
```diff
@@ -235,6 +235,35 @@ If used in a query, `myNeatProcess.%.duration` will return results for all three
   </Collapser>
 </CollapserGroup>
 
+## Return results for invidual fields using `getField()`
```
urbiz-nr, line 238: Add anchor id

src/content/docs/telemetry-data-platform/understand-data/metric-data/query-metric-data-type.mdx
```diff
@@ -235,6 +235,35 @@ If used in a query, `myNeatProcess.%.duration` will return results for all three
...
+
+Metrics have multiple fields associated with them, and each field contains its own unique values. If you would like to extract one of those values using a query, you can do that with `getField()`. 
+
+For example, if you want to use a single value within a metric to do a comparison in a `WHERE` clause, you can do that with `getField(metricName, field)` or `metricName[field]`.
```
urbiz-nr, line 242: You've used the expression "you can do that with" in the previous sentence, I'd rewrite this.

src/content/docs/telemetry-data-platform/understand-data/metric-data/query-metric-data-type.mdx
```diff
@@ -235,6 +235,35 @@ If used in a query, `myNeatProcess.%.duration` will return results for all three
...
+    FROM Metric SELECT uniques(metricName) WHERE %[type] = 'gauge'
+    ```
+  </Collapser>
+
+  <Collapser
+    id="example-list-names-host"
+    title="List all metric names for a particular host"
+  >
+
+  The average value of a metric is computed as `total` over `count`, so the following query would return metric data where the result of the `average()` value function is greater than 2.
```
urbiz-nr, line 260: I'd use "returns" instead of "would return", it's a fact, not a possibility

urbiz-nr (approved):
See my comments, looks good otherwise!

rhetoric101 (approved):
@dbarnesbrownNR, your changes look good!

PR #2486, opened 2021-05-26
Errors inbox title fix

bradleycamacho:
Errors inbox doc title was incorrect, small fix.


rhetoric101 (approved):
@bradleycamacho, thanks for fixing this! You can merge it when you're ready.

PR #2487, opened 2021-05-26
attempting an image fix

mmfred:
What's new image for Pixie.


rhetoric101 (approved):
@mmfred, this filename correction fixed it in the preview!

PR #2491, opened 2021-05-26
Updated title from Error inbox to Errors inbox

nedl86:
### Give us some context

* Fixed title from Error inbox to Errors inbox (to be consistent with NR1 platform and rest of doc)


x8a (approved):

x8a:
Thanks for spotting this!

PR #2492, opened 2021-05-27
Update large-number-false-positive-security-vulnerabilities.mdx

tbradellis:
### Give us some context

This doc previously provided a very specific solution for a single vendor security scan.  The doc demonstrated how to suppress false positives for the single vendor.

However, this information about false positives applies across many security scan vendors.  I updated this to be a more generic reference point for when customers see alarming security scan violations coming from the java agent. This is a common question, so the answer should be more generic.  I left the previous instruction for suppression as an example of how one might need to go about doing such a thing.


x8a (approved):

x8a:
Thank you so much for these changes, @tbradellis! I did some rewording according to the documentation guidelines and we'll be publishing this later today.

PR #2500, opened 2021-05-27
Make SCIM and SAML set up more prominent

zuluecho9:
small prelim changes for DOC 6834.


rhetoric101 (approved):
@zuluecho9, I looked at these changes locally, and they look good!

PR #2503, opened 2021-05-27
chore: improve functionality of release notes

rudouglas:
I've used a Python script to pull more detailed data from GitHub, like usernames, and to allow us to customize more easily going forward

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,34 +18,122 @@ jobs:
       tag: ${{ steps.get-tag.outputs.tag }}
       last: ${{ steps.get-last-tag.outputs.last }}
       notes: ${{ steps.get-release-notes.outputs.notes }}
+      result: ${{ steps.py-get-updates.outputs.result }}
+      mails: ${{ steps.get-user-emails.outputs.mails }}
+      authors: ${{ steps.get-authors.outputs.authors }}
     steps:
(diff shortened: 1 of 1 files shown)


nr-kkenney (commented):
id be in favor of having this live in a python file rather than in the body of the workflow. feel free to grab another opinion on this, since this is just my personal opinion 🙈

rudouglas:
i'll take all opinions @nr-kkenney personal, professional or other wise 😂 This would probably be better ye, where would be the best place for this to live, just in the `github` folder?

.github/workflows/release.yml
```diff
@@ -18,34 +18,122 @@ jobs:
...
           
-      
+      - name: Install Python dependencies
+        id: install-python
+        run: |
+          python -m pip install --upgrade pip
+          pip install PyGithub
+          pip install invoke
+          
+      - name: Create Updates
```
nr-kkenney, line 75: haha. i should have commented on this so we could get a thread. i think moving this under .github is 👍  ive done .github/scripts for the demotron repo. im pretty open as to the exact path, but as long as its under .github, sounds good to me.

rudouglas: ha i don't know why you can't just make a thread without commenting 🙈  excellent ye that sounds much better, keep it all neat and tidy, thanks

PR #2510, opened 2021-05-27
Tweak readme to reduce word count and make it less "new"

austin-schaefer:
The site's not as new, so removing some of the language about how new it is.


mmfred (approved):

mmfred:
@austin-schaefer - this is ready to merge

PR #2513, opened 2021-05-27
feat: updates to release notes workflow

rudouglas:
And construct release notes
If the users email address is not public it parses their email and checks if that is a valid user, if not it just prints the Author name, if the user exists it prefaces it with `@` to tag them
Should look like this
Screenshot 2021-05-27 at 23 17 48

### DIFF
--- a/scripts/actions/release-notes.py
+++ b/scripts/actions/release-notes.py
@@ -1,64 +1,32 @@
 from github import Github
+from invoke import run
 import os
 import re
 
-from invoke import run
-
+# Get token from Workflow environment variable
 token = os.getenv('GITHUB_TOKEN', '...')
+
+# Initialize GitHub with token
 g = Github(token)
 
-uniques = {}
-releaseArray = ['merge to main', 'evening', 'morning', 'noon', 'release']
-
-emails = os.getenv('MAILS', '...')
-authors = os.getenv('AUTHORS', '...')
-emails = emails.split()
-
-authors = authors.split("%0A")
+# Create heading for Release Notes
 result = "## :rocket: What's new?\n\n\n"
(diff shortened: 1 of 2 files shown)


rudouglas:
The newest commit checks the diff between both `develop` and `main` and grabs info for those commits. It does this directly from the GitHub API now so all `usernames` are retrieved correctly, so they can be `@` mentioned in the notes

scripts/actions/release-notes.py
```diff
@@ -0,0 +1,32 @@
...
+# Compare diff between main and develop
+diff = repo.compare("main", "develop")
+
+# Loop through commits and add details to result
+for com in diff.commits:
+  line = "@" + com.author.login
+  line += " - " + com.commit.message.splitlines()[0]
+  line += " [->](" + com.html_url + ")"
+  result += line
+  result += "\n"
```
nr-kkenney, line 27: i would be in favor of using f-strings here. i think that would make it easier to understand some of these lines. can even reduce the total number of lines.
example:
```py
line += f"[->]({com.html_url})"
```

rudouglas: Excellent i\ll change that now

nr-kkenney: actually, sorry to make another request for change, haha. with all the lines, and variables, i think i would prefer something like:
```py
  result += "@{author} - {message} [->]({url})\n".format(
    author=com.author.login,
    message=com.commit.message.splitlines()[0],
    url=com.html._url
  )
```
i think that is probably an improvement in readability over just using f-strings, since the expected output is clearer.

nr-kkenney (approved):

Issue #2454, opened 2021-05-25
[Localization] Workflow submits already translated files for translation


roadlittledawn:
## Description

@jmiraNR reported we may be erroneously submitting files in `src/i18n/content/jp` for translation, when we should only be sending English files in `src/content` that have a `translate` field in their frontmatter.

Julian cited [this recent job](https://dashboard.smartling.com/app/projects/bf1be2489/account-jobs/bf1be2489:kqjsfpadl3km/files), and I can see we definitely have files with the `i18n` path. Need to investigate further to determine why / if we should be excluding those and if we write them back correctly when we create a PR.

## Expected behavior
Only submit English files in `src/content` marked for translation via their frontmatter to translation vendor. When complete and we create PR, ensure we write back to relevant path in `src/i18n/content`

### Acceptance Criteria

- [ ] Get Team access to Smartling (kris, liz, Alec, Ruairi) - @jpvajda 
- [ ] Filter out any markdown files that are a non-english language type.
- [ ] Don't add any translated file to the the translation queue

LizBaker:
Our workflow doesn't strip out the `translate: - jp` frontmatter in translated docs when it PRs new/changed files. This is why we suddenly had some japanese files get grabbed for translation, since we check for that frontmatter. I'll be looking into how to strip that out when deserializing, which should solve the issue (along with manually editing the 15 files which gained that frontmatter)

PR #2525, opened 2021-05-28
Fix translation workflow

LizBaker:
### Give us some context

Closes https://github.com/newrelic/docs-website/issues/2454
This PR adds a plugin to the unified processor which deserializes html sent back from our translation vendor. The plugin strips out both the `translate` and `redirects` fields from the frontmatter. If we leave the translate field then it's possible for files that have already been translated to get grabbed and sent to the vendor again.

It also manually removes the translate and redirects frontmatter from 15 translated files that came in with the last translation PR

## before plugin
Screen Shot 2021-05-28 at 2 49 13 PM

## after 🎉 
Screen Shot 2021-05-28 at 2 49 35 PM


scripts/actions/deserialize-html.js
```diff
@@ -67,6 +68,21 @@ const headingWithCustomId = (h, node) => {
   return result;
 };
 
+const stripTranslateFrontmatter = () => {
+  const transformer = (tree) => {
+    if (tree?.children[0]?.type === 'yaml') {
+      const frontmatterObj = yaml.load(tree.children[0].value);
+      delete frontmatterObj.translate;
+      delete frontmatterObj.redirects;
```
zstix, line 76: Do we want to remove redirects for localized pages? I would think that, if we have set up a redirect for a page in english, that we would also want that to be set up in other languges.

LizBaker: this was at the recommendation of @roadlittledawn, maybe he can expand on it?

roadlittledawn: yeah we do. we add redirects from English `.mdx` files to their localized equivalents rather than reading them directly from `src/i18n` files since they're always the same / we match them 1:1

zstix (approved):

jmiraNR:
Hello @zstix @LizBaker, What happens now with the job that had both English to translate files and Japanese files? Should I ignore it? proceed with it? Can I just "cancel" all the text/strings in Japanese and focus only on English? thank you.

PR #2541, opened 2021-06-01
nri-oracle: document usage of `tablespaces` argument

roobre:
### Give us some context

The OracleDB integration will fail to collect more than 200 tablespaces. When the user's environment has more than this amount, it is required to restrict collection using the (previously undocumented) `tablespaces` argument.

This PR documents this parameter and its usage.

No


bradleycamacho:
Thanks for the edit @roobre! I've made a small copy edit and will merge this as soon as the checks pass.

PR #2549, opened 2021-06-01
Delete "archive Locate exact metric names"

austin-schaefer:
### Give us some context

* This ol' stray doc exists 
* It gets literally 0 traffic
* It's about legacy/EoLed features


bradleycamacho:
Got weird build error when I first built, but I can't replicate it and the checks all passed. Approved!

bradleycamacho (approved):

PR #2564, opened 2021-06-02
Unnecessary arch params on Debian & Ubuntu

josemore:
### Give us some context

* We recently added support for ARM architecture. 
* We validated that it's actually not needed to determine the amd or arm64 architecture for Ubuntu and Debian since this will be automatically detected. 
* Removed unnecessary params to determine the right architecture in docs.


mmfred (approved):

mmfred:
Thank you for making the changes @josemore!

PR #2572, opened 2021-06-02
Fix sort order for Browser AMP doc

austin-schaefer:
It's weird having an install doc at the bottom of the category, after delete/disable docs


austin-schaefer:
Just looked at Gatsby preview and the sort order looks good there.

paperclypse (approved):
Looks good!

PR #2580, opened 2021-06-03
mysql and postgres doc update

lchapman4:
Added an entry for mysql installation on Windows. NR hosts the windows .msi file on our download site but we don't have directions for it in our docs. The docs explicity required a linux host which isn't true so I cleaned up that verbiage as well. For both files i just updated our links to the download site to default to https since we don't redirect there from http links.

### Give us some context


bradleycamacho:
Thanks for this edit @lchapman4! This should go live in the next hour or so.

PR #2582, opened 2021-06-03
Update dotnet agent installation instructions for Linux based Azure Web Apps

JcolemanNR:
I'm not sure how the formatting will all look on the main docs site... I tried to add less than/greater than symbol to an xml blob but it's not easy to see if it's working properly with the github preview tool.

### Give us some context

* I ran through the Dotnet agent Linux install instructions for Azure Web Apps and I'm not sure if Azure changed, or Visual Studio changed... but the steps required to get this going weren't clear from the existing documentation.
* I tried to add some missing less than/greater than symbols in sample XML fragments, but it's not clear if this will render properly on the main site.
* I added a new env var that I had to set to get things going
* I added a note about needing to manually copy the newrelic.config file out of the nuget package contents in to my application as visual studio was preventing me from editing the nuget checkout.


bradleycamacho:
@JcolemanNR Looks good to me, thanks for this addition! Merging in now.

PR #2584, opened 2021-06-03
Remove Decision UI's flapping (image and text content)

sjimenez-nr:
Addition to AI docs, and update of image. @zuluecho9 is approving it. 

Per https://newrelic.atlassian.net/browse/CSERV-1643

we are removing flapping statistic from Decision UI, development work is completed. we just need to update documentation to reflect it.


zuluecho9:
The CircleCI thing that didn't sign the agreement is a build tool of some sort on NR team.

PR #2590, opened 2021-06-03
Dm UI fix retention info

mmfred:
This is a small edit to clarify that data doesn't disappear when you change the number of days that info is retained.


barbnewrelic (approved):
Michelle F. and I chatted about this PR. Fine as is, go ahead and merge! May want to consider doing a more in-depth peer edit of the entire doc at some point.

PR #2605, opened 2021-06-06
Update use-tcp-endpoint-forward-logs-new-relic.mdx

dgcom:
There were extra lines before template definition which were corrupted copy of input definition.
Cleaned up to make configuration valid.


urbiz-nr:
Thanks for the update @dgcom!

urbiz-nr (approved):

PR #2617, opened 2021-06-08
Changed image title from filename to meaningful title.

rhetoric101:
As part of testing the migration of changes from the private docs site, I am fixing a poor image title.

Reviewer: please hover over the image to confirm that the title is no longer the filename.


barbnewrelic (approved):
I checked the image in localhost. The description not the png filename does indeed appear when I moused over it. Thanks for making a helpful fix!

PR #2618, opened 2021-06-08
fix: catch errors in release notes failure

rudouglas:
If there is an error getting any details from the commit object, the code will now catch these and continue with the loop instead of failing

I also refactored some of the `release.yml` workflow.
It will now grab the actual latest tag from a release rather than calculating it from the number of merges to `main`, so if a release fails, the next release won't look for a tag that failed to create,and will include all changes since the last release tag.


nr-kkenney (approved):
lgtm

PR #2619, opened 2021-06-08
clarify that a link requires NR employee access

austin-schaefer:
tiny clarification on link destination


barbnewrelic (approved):
Thanks for making that important clarification!

PR #2625, opened 2021-06-08
Fix image name to resolve broken link

alanwest:


urbiz-nr:
Thanks for fixing the image, @alanwest. Looks great!

urbiz-nr (approved):

PR #2627, opened 2021-06-09
change the link to point to a different section

lxierita:
The original links points to a section that doesn't really say much about what **data types** you can query


lxierita:
Close it for now. The section itself is fine, it turns out that the browser took me to the page rather than that particular section.

PR #2642, opened 2021-06-10
Corrected image title as a test of the private repo.

rhetoric101:
This is another test of transferring changes from the private docs site. This test is still a valid change we need to publish to the public site. I just added an image title.


urbiz-nr (approved):
Looks good @rhetoric101!

PR #2643, opened 2021-06-10
Inserted more meaningful OpenTelemetry image title as a test of the private repo.

rhetoric101:
This is another test of transferring branches from the private repo to the public repo. This image title fix is minor, but still a valid change that we should review and publish.


urbiz-nr (approved):
Nice, @rhetoric101!

PR #2651, opened 2021-06-10
Fixes some broken links

paperclypse:
They were getting lost in a muddle of deleted docs and misguided redirects.

This came in as a Docs hero request.


barbnewrelic (approved):
Thanks for fixing this!

PR #2660, opened 2021-06-11
fix(mssql): outdated metrics description

paologallinaharbur:
* What problems does this PR solve
mssql integration docs were not correct after the change https://github.com/newrelic/nri-mssql/commit/743e8ff32b3202a2e25b0bf09c221403ad03b265


barbnewrelic (approved):
Thanks for updating these docs!

PR #2670, opened 2021-06-11
Add broken links and certificate monitors to table

bradleycamacho:
This was requested by the Synthetics PMM. It adds the two new monitors to the list of monitors in our docs.


barbnewrelic (approved):
Looks fine, although I wonder if we really need to specify how many monitor types we offer. Seems like that would be an easy thing to miss updating if we added more.

PR #2673, opened 2021-06-11
Fixed link to fedramp settings for Infrastructure (under Log API)

Rilsta:
Adjusted link so the link for Infra will take you to the fedramp setting in Infra's config page.


barbnewrelic (approved):
Thanks for fixing that wonky link!

PR #2675, opened 2021-06-12
Workflow Tests: Add Slugs to Translation Queue

zstix:
## Description
In an effort to make our translation workflows feel more stable, I wanted to take some time to write proper tests for each workflow. This is the first in a series of PRs (one workflow at a time). My goal is to establish some helpful code references for things like mocking calls to our DB, faking a file system for use in tests, etc.

## Reviewer Notes
* I'm still adding some test cases (for edge-case scenarios)
* I think I am starting to uncover bugs with the code, but I wanted to review it with the team first

### DIFF
--- a/scripts/actions/__tests__/add-files-to-translation-queue.test.js
+++ b/scripts/actions/__tests__/add-files-to-translation-queue.test.js
@@ -1,0 +1,54 @@
+import AWS from 'aws-sdk';
+import addFilesToTranslationQueue from '../add-files-to-translation-queue';
+
+// Mock the aws-sdk so we can avoid calling our actual DB
+jest.mock('aws-sdk', () => {
+  const mockConfig = { update: jest.fn() };
+  const mockDocumentClient = { get: jest.fn(), update: jest.fn() };
+  const mockDynamoDB = { DocumentClient: jest.fn(() => mockDocumentClient) };
+
+  return { config: mockConfig, DynamoDB: mockDynamoDB };
+});
+
+// Singleton instance of our mock DB (used to generate responses)
+const mockDynamoDb = new AWS.DynamoDB.DocumentClient();
+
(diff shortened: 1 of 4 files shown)


scripts/actions/__tests__/add-files-to-translation-queue.test.js
```diff
@@ -0,0 +1,256 @@
+import fs from 'fs';
+import AWS from 'aws-sdk';
+import fetch from 'node-fetch';
```
aswanson-nr, line 3: I think we'll need to change these imports over to the old `require` syntax, since this will be running on node 12 in the Github runners.

scripts/actions/add-files-to-translation-queue.js
```diff
@@ -84,7 +85,20 @@ const main = async () => {
 
   await saveToTranslationQueue(key, 'set locales = :slugs', { ':slugs': data });
 
-  process.exit(0);
+  // If we are in a CI environment, return a success code
+  if (process.env.CI) {
```
aswanson-nr, line 89: Github Actions automatically set this to `true` and it doesn't look like you can override it. 
You can use `require.main === module` instead, it will evaluate to `true` if the file is executed by Node and `false` if the file is imported by another module. https://nodejs.org/docs/latest/api/modules.html#modules_accessing_the_main_module

jpvajda:
Moving back to in-progress as there are some suggestions, probably best to wait for @zstix and @aswanson-nr to return

jpvajda:
Just to confirm: is this the remainder of the work we need to complete before we can turn back on the [translation workflow job?](https://github.com/newrelic/docs-website/actions/workflows/send-content-to-translate.yml)

zstix:
I'm closing this because it's a little too stale. Going to leave the branch so that we can use it for reference.

PR #2679, opened 2021-06-14
chore: add infinite lineWidth to yaml deserializer

rudouglas:
This is to stop it converting single line strings to use `>-` when over 80 characters long

### DIFF
--- a/scripts/actions/deserialize-html.js
+++ b/scripts/actions/deserialize-html.js
@@ -74,7 +74,9 @@ const stripTranslateFrontmatter = () => {
       const frontmatterObj = yaml.load(tree.children[0].value);
       delete frontmatterObj.translate;
       delete frontmatterObj.redirects;
-      const frontmatterStr = yaml.dump(frontmatterObj).trim();
+      const frontmatterStr = yaml
+        .dump(frontmatterObj, { lineWidth: -1 })
+        .trim();
       tree.children[0].value = frontmatterStr;
       return tree;
     }


urbiz-nr (approved):

urbiz-nr:
Thanks for the update, @rudouglas!

jpvajda:
We probably want a engineer to approve this on the DevEn Team as well before merging.  I had a comment on the change as well.

scripts/actions/deserialize-html.js
```diff
@@ -74,7 +74,9 @@ const stripTranslateFrontmatter = () => {
       const frontmatterObj = yaml.load(tree.children[0].value);
       delete frontmatterObj.translate;
       delete frontmatterObj.redirects;
-      const frontmatterStr = yaml.dump(frontmatterObj).trim();
+      const frontmatterStr = yaml
+        .dump(frontmatterObj, { lineWidth: -1 })
```
jpvajda, line 78: By changing this to -1 are we now not enforcing any line width at all? Before it was 80 characters, but now it's no limit? I'm just wondering if we should still enforce a line width here.

jpvajda (changes requested):

scripts/actions/deserialize-html.js
LizBaker: it had no limit before, this is caused by the yaml parsing library we're using here which was introduced recently. i don't think there's any technical reason to enforce line width

jpvajda: Ah ok, thanks for the explaination @LizBaker

jpvajda (approved):

LizBaker (approved):
just tried it out, works great!

PR #2688, opened 2021-06-14
DOC-6943 Update DT Diagrams

kaileyhaynes:
Replaced old diagram images from DT's landing and intro docs with colorized & updated images.


bradleycamacho (approved):
This looks great, good job @kaileyhaynes !

PR #2691, opened 2021-06-14
Some slight changes around REST API admin

zuluecho9:
Changes around recent change where v1 users who aren't Owners or Admins no longer have access to REST API key.


rhetoric101 (approved):
@zuluecho9, these changes look good!

PR #2694, opened 2021-06-14
Update ruby agent configuration

tannalynn:
### Give us some context
This PR updates our autogenerated config documentation. This updates the wording on `browser_monitoring.auto_instrument` config option to clarify the default value, also we updated links to use markdown instead of html.
 I also added html comments to mark what sections are auto generated from the agent so it's easier to tell. (There used to be comments that showed the section was generated, but it seems like it went missing since the move to git for the docs site)

no


dbarnesbrownNR (approved):
All good, merging!

PR #2695, opened 2021-06-14
Trying to fix 404 from link on integrations auto-generated page.

rhetoric101:
Steps to reproduce:

1. Go to `https://docs.newrelic.com/docs/integrations/`
2. Scroll down and click on **Kubernetes plugin for Logs** and note the 404

Discussion:

The fix I chose for this, based on feedback from a colleague, was to simply remove the "path" line from `integrations.yml` for `Kubernetes plugin for Logs`. Now, if you go to  `https://docs.newrelic.com/docs/integrations/`, you should see the following if you scroll down the page, but `Kubernetes plugin for Logs` is no longer a link (just the child `Get Kubernetes log data`.

  Kubernetes plugin for Logs
    Get Kubernetes log data

### DIFF
--- a/src/data/taxonomy-redirects.json
+++ b/src/data/taxonomy-redirects.json
@@ -2353,5 +2353,11 @@
     "paths": [
       "/docs/serverless-function-monitoring/azure-functions-monitoring/get-started"
     ]
+  },
+  {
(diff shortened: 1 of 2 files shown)


urbiz-nr:
Tried locally, it works fine!

urbiz-nr (approved):

Issue #2288, opened 2021-05-13
Translation workflow failure.


nr-kkenney:

jpvajda:
moving to sprint 6 to measure impact...

PR #2696, opened 2021-06-14
chore: copy over images for translated docs

nr-kkenney:
* added logic to copy over images for translated docs. also added in
paths for skipping over already copied images, & ignoring directories
with no images.

Resolves: #2288

### DIFF
--- a/scripts/actions/fetch-and-deserialize.js
+++ b/scripts/actions/fetch-and-deserialize.js
@@ -1,6 +1,8 @@
 const AdmZip = require('adm-zip');
 const vfile = require('vfile');
 const { writeSync } = require('to-vfile');
+const path = require('path');
+const fse = require('fs-extra');
 
 const fetch = require('node-fetch');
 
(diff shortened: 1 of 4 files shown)


scripts/actions/fetch-and-deserialize.js
```diff
@@ -69,10 +107,9 @@ const fetchAndDeserialize = (accessToken) => async ({ locale, fileUris }) => {
   );
 
   createDirectories(files);
-
-  files.forEach((file) => writeSync(file, 'utf-8'));
+  writeFilesSync(files);
 };
 
 fetchAndDeserialize();
```
nr-kkenney, line 113: now that we have this imported into a test, this should be:
```javascript
if (process.env.CI) {
    fetchAndDeserialize();
}
```

nr-kkenney (commented):
whoops 🙈

scripts/actions/fetch-and-deserialize.js
```diff
@@ -69,10 +107,11 @@ const fetchAndDeserialize = (accessToken) => async ({ locale, fileUris }) => {
...
+  writeFilesSync(files);
 };
 
-fetchAndDeserialize();
+if (process.env.CI) {
+  fetchAndDeserialize();
+}
 
-module.exports = fetchAndDeserialize;
+module.exports = { writeFilesSync, fetchAndDeserialize };
```
zstix, line 117: If we change the export signature, I think we'll also need to change how it's imported in other files (`scripts/actions/check-job-prgress.js` comes to mind).

zstix (commented):
Love the tests ✅ 
Love the comments ✅ 

Just had one minor thing about `module.exports`.

Hopefully we can avoid doing this work in the future (by referencing the english image rather than duplicating the files). That said, I think it's probably out of scope for this PR.

scripts/actions/fetch-and-deserialize.js
nr-kkenney: oof. good catch.

aswanson-nr (approved):

PR #2724, opened 2021-06-16
Updating postgresql-monitoring-integration

carlossscastro:
Adding `db.maxconnections` metric.
This metric has been added and will be released in the next version of the integration, as soon as the this doc update is published


rhetoric101 (approved):
@carlossscastro, this docs update looks good! I will roll this into the noon publishing slot. That means that it may not be visible on the site until after 1:00 PM pacific today.

Issue #2323, opened 2021-05-17
Remove the translated MDX file when a locale is removed or file is deleted


zstix:
## Description
When a contributor removes a locale from the `translate` frontmatter, we should remove the translated file from the repository. If we don't the translated file will get more and more out of date with it's English counterpart. This will probably require adjusting [`add-files-to-translation-queue.js`](https://github.com/newrelic/docs-website/blob/develop/scripts/actions/add-files-to-translation-queue.js).

### Acceptance Criteria 

- [ ] When a contributor removes a locale from the `translate` frontmatter, we should automaticalluy remove the translated file from the repository.
- [ ] when the english version of a file is deleted, we want to remove the translated version of a file.
- [ ] Ensure this change is documented in our localization GDrive Doc `DevEn: Localization: how it works`

jpvajda:
This issue relates to the corner case raised about broken links on localized due to pages being removed from the english site it's a future problem for us to consider

https://github.com/newrelic/docs-website/issues/2749

PR #2726, opened 2021-06-16
Create check for out of date translated files

LizBaker:
This PR includes a workflow job which will add a required check on PRs to develop and main.
The check will find translated files that don't have an english counterpart, for instance: if an english file is deleted, any translations for the file should also be deleted.
Instead of deleting the files via the workflow we decided to make a failable check which will remind contributors to clean up any translated files that should no longer exist.

TODO:
- [x] remove previous WIP code in add-to-translation-queue
- [x] write tests
- [x] add job to required checks
- [x] settle on check name with docs friends

Closes https://github.com/newrelic/docs-website/issues/2323


scripts/actions/check-for-outdated-translations.js
```diff
@@ -0,0 +1,114 @@
...
+    .map((name) => doI18nFilesExist(name, ADDITIONAL_LOCALES))
+    .flat();
+
+  const orphanedI18nFiles = [...modifiedFiles, ...removedFiles];
+
+  if (orphanedI18nFiles.length > 0) {
+    orphanedI18nFiles.forEach((f) =>
+      // TODO: improve output
+      console.log(
+        `ACTION NEEDED: Translation without english version found-- ${f.replace(
```
aswanson-nr, line 90: Small nit, can you add a space after `found`?

scripts/actions/check-for-outdated-translations.js
```diff
@@ -0,0 +1,114 @@
...
+
+const fetchFilesFromGH = async (url) => {
+  let files = [];
+  let nextPageLink = url;
+
+  while (nextPageLink) {
+    const resp = await fetch(nextPageLink, {
+      headers: { authorization: `token ${process.env.GITHUB_TOKEN}` },
+    });
+    const page = await resp.json();
```
aswanson-nr, line 32: We might want to add some error handling around this. If a non-200 status code is returned, the response object will not have the `.json()` method. Erroring out and letting the catch handle it works, but we lose some visibility into the actual error that the API is returning.

Adding something like this before line 32 should do the trick.
```javascript
if (!resp.ok) {
  throw new Error(`Github API returned status ${resp.code} - ${resp.message}`);
}
```

aswanson-nr (approved):
Everything looks great! I added 2 small comments, but neither of them are required.

scripts/actions/check-for-outdated-translations.js
LizBaker: aw heck, I guess we lost that change at some point, gotta bring back mini arrow

aswanson-nr (dismissed):
🚀

scripts/actions/__tests__/check-for-outdated-translations.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+// Mock node-fetch so we can avoid calling out to GitHub
+jest.mock('node-fetch');
+
+global.process.exit = jest.fn();
+
+// Helper function to mock a response from Github
+const mockGithubResponse = (result) => {
+  fetch.mockResolvedValueOnce({
+    json: jest.fn(() => Promise.resolve(result)),
```
aswanson-nr, line 14: adding an `ok` field here should fix the unit tests

LizBaker: whoops! thanks :D

scripts/actions/check-for-outdated-translations.js
```diff
@@ -0,0 +1,118 @@
...
+  const i18nPrefix = path.join(process.cwd(), 'src/i18n/content');
+  const baseFileName = fileName.replace('src/content/', '');
+
+  return locales
+    .map((locale) => {
+      const filePath = path.join(i18nPrefix, locale, baseFileName);
+      const fileExists = fs.existsSync(filePath);
+      return fileExists ? filePath : null;
+    })
+    .filter(Boolean);
```
zstix, line 21: This might be simplified by using [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some):

```suggestion
  return locales
    .some((locale) => {
      const filePath = path.join(i18nPrefix, locale, baseFileName);
      return fs.existsSync(filePath);
    })
```

scripts/actions/check-for-outdated-translations.js
```diff
@@ -0,0 +1,118 @@
...
+    }, []);
+
+  const removedMdxFileNames = mdxFiles
+    .filter((f) => f.status === 'removed')
+    .map(prop('filename'));
+
+  // if a locale was removed from the translate frontmatter, we want to remove the translated version of that file.
+
+  const modifiedFiles = mdxFilesContent
+    .map((file) => {
```
zstix, line 77: If you want to avoid doing a [`map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) and then a [`flat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat), you're in luck: [`flatMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)!

zstix: And if you want some supplemental reading, `flatMap` (and functional programming Monads): https://mostly-adequate.gitbook.io/mostly-adequate-guide/ch09

scripts/actions/check-for-outdated-translations.js
aswanson-nr: The file names are being used later on in the script to provide output for the user. It's probably worth changing the function name to `findI18nFiles` instead of `doI18nFilesExist`

scripts/actions/check-for-outdated-translations.js
LizBaker: of course!

zstix (dismissed):

aswanson-nr (approved):

zstix (approved):

PR #2731, opened 2021-06-16
These collapser IDs were all the same

paperclypse:
That means it wasn't possible to link to any one of them directly.

(courtesy of @HenryTech)


rhetoric101 (approved):
@paperclypse, thanks for correcting these collapser IDs!

urbiz-nr (approved):

PR #2734, opened 2021-06-17
Sending logs from OpenTelemetry

alanwest:
@rhetoric101 Here's the logs doc w/ the structure we discussed.


barbnewrelic (approved):
I made a few minor copyedits and pushed as a separate commit. This new info is going to be very helpful, congrats!

PR #2747, opened 2021-06-17
Additional changes following initial review.

rhetoric101:
* Fixed link from intro to log management to this new doc.
* Made the advance-configuration topic into just a topic about endpoints by changing the filename, doc name, and removing the first sentence.
* Changed the nav file so it points to new endpoints file.


dbarnesbrownNR (approved):
Looks good!

PR #2752, opened 2021-06-18
Update create-nrql-alert-conditions.mdx

nedl86:
Loss of signal is now a standard part of the alerting platform and doesn't need to be requested, so removing the call-out.


urbiz-nr (approved):

urbiz-nr:
Thanks for helping us keep docs up-to-date, @nedl86!

PR #2810, opened 2021-06-23
update install steps to include required pieces

keegoid-nr:
Go functions should import:

- github.com/newrelic/go-agent/v3/newrelic
- github.com/newrelic/go-agent/v3/integrations/nrlambda

### Give us some context

* What problems does this PR solve? Dependency issues can happen on function invocation if missing dependencies.
https://github.com/newrelic/newrelic-lambda-extension/blob/main/examples/sam/go/main.go


mmfred (approved):

mmfred:
Thanks for updating the doc @keegoid-nr!

PR #2812, opened 2021-06-23
Remove references to Insights in Mobile HTTP errors

rkbodenner:
The UI elements still say "View in Insights", but this changes the doc to not otherwise refer to that EOL'ed product, and instead point to dashboards and the query builder.


mmfred (approved):

mmfred:
Looks good - thanks @rkbodenner!

PR #2828, opened 2021-06-24
Added clarity about creating baseline conditions

paperclypse:
Not so much how, as where you go.

I pulled the UI path from out of the image caption. I think this redundancy is ok.


austin-schaefer:
This PR addresses customer feedback CF-7216.

urbiz-nr (approved):

PR #2839, opened 2021-06-25
Call main function in outdated-translations workflow

LizBaker:
Just realized we aren't actually calling the function when executed from the command line 😅


nr-kkenney (dismissed):

LizBaker:
looks like the call to the github api isn't working correctly

nr-kkenney (approved):

PR #2861, opened 2021-06-29
Update microsoft-sql-server-monitoring-integration.mdx

carlossscastro:
Small change as we now support domain users as well


x8a (approved):
Thanks for the addition, @carlossscastro!

PR #2885, opened 2021-06-30
Update compatibility-requirements-java-agent.mdx

XiXiaPdx:
change overloaded term "support" to compatible

### Give us some context

Overloaded term "support". Caused confusion internally and potentially with customer.

no


paperclypse (approved):
I think this is a great change. I agree that "support" is an overloaded term. Thanks!

PR #2891, opened 2021-06-30
Fixing url for parent/child account doc

zuluecho9:
Small fix


barbnewrelic (approved):
Tested this in localhost and appears to work fine. Thanks for fixing this!

PR #2909, opened 2021-07-01
Fix malformed links

austin-schaefer:
This link extended over three ¶s. Probably an issue in the migration script.


paperclypse (approved):
Looks good to me!

PR #2916, opened 2021-07-01
Update users-roles-original-user-model.mdx

karayusuf:
### Give us some context

Restricted users have access to proactive detection configurations.
Update the original user documentation to indicate that.


paperclypse (approved):
Looks good!

PR #2926, opened 2021-07-01
DOC-6131: Lower-case "browser" everywhere it makes sense

austin-schaefer:
### Give us some context

Since the new pricing model launched, we no longer capitalize New Relic Browser as a product. Instead, it's a capability ("New Relic's browser monitoring" or "browser monitoring" or "browser agent" or similar).

We have a related jira in DOC-6131

### To-dos

- [x] Sweep for New Relic Browser
- [x] Sweep for just Browser in uppercase and change where appropriate
   * Note to future archaeologists: Changed lots of instances, but many instances are just atribute/event names.
- [x] Look for references to old Pro+SPA SKU
    * Ended up spinning this into a standalone jira DOC-7085
- [x] Check for links to https://newrelic.com/products
    * Done in #2926 
- [x] Consider sweeping for references to "enabling" rather than "installing" browser agent
- [x] Look for stray `''` in metaDescriptions
    * Turned out not to matter! See #2936 if you want more info

### Notes

- Did not update "original pricing/billing" docs
- Did not update docs for deprecated features like https://docs.newrelic.com/docs/browser/new-relic-browser/installation/monitor-amp-pages-new-relic-browser/
- Did not update any doc in the legal/licenses category

### Examples of pricing stuff

```Your New Relic Browser [subscription level](http://newrelic.com/browser-monitoring/pricing) and your [time picker](/docs/data-analysis/user-interface-functions/time-picker-setting-time-periods-view-data) settings determine what data you can view:```

PR #2936, opened 2021-07-02
Start removing stray `''` from metaDescription

austin-schaefer:
While working on #2926, I ran across a number of docs with malformed metadescriptions. This PR sweeps the site for these using this VS code regex search:

image


austin-schaefer:
After looking at this, I learned that the `''` syntax is translated under the hood into an HTML escape character for an apostrophe. For instance:

On https://docs.newrelic.com/docs/agents/go-agent/installation/update-go-agent/

The metadescription when you view page source looks like this:

```How to update New Relic APM&#x27;s Go agent so you can take advantage of the latest features, enhancements, and important security patches.```

Which Google correctly interprets as:

image

PR #2938, opened 2021-07-02
Remove handful of old permissions callouts for DOC-6138

austin-schaefer:
Doc-6138 asks for an audit of remaining permissions callouts.


austin-schaefer:
This regex turned up a few more. Asking @zuluecho9 for his thoughts on them and then will proceed.

image

zuluecho9:
This looks good, I didn't even know we still had these in docs, so good catch. This made me realize we have a lot of 'Account settings UI' mentions still in docs, and that's a v1 model UI page, so I'm creating a jira to take a look at those.

zuluecho9 (approved):

PR #2946, opened 2021-07-02
Fix some busted links from DOC-6322

austin-schaefer:
Fix for some busted stuff in DOC-6322


barbnewrelic (approved):
Tested the links and they appear to be fine now. Thanks!

Issue #2872, opened 2021-06-29
Manually run translation workflow job on Jul 1st


jpvajda:
Given we have disabled the automated translation workflow job that sends [new content to be translated](https://github.com/newrelic/docs-website/actions/workflows/send-content-to-translate.yml) due to the following bugs 
 
https://github.com/newrelic/docs-website/issues/2655
https://github.com/newrelic/docs-website/issues/2603

We'll need to run this manually on July 1st.

### References

- See the doc in`Google Drive` > `DevEn: Localization: how it works`

### Acceptance Criteria

- [x] On Jul 1st, kick of the send new content to be translate job
- [x] Ensure job runs successfully
- [x] Notify @jmiraNR when completed.

nr-kkenney:
@jpvajda @jmiraNR 
the files were queued up for translation have been submitted to smartling as three jobs. along the way, i created a few more jobs 🙈  but should have cancelled all except the ones we care about.

sorry that it took so long. ideally, this would have been a 1 point issue -- just run the workflow. but i had to run it quite a few times, and address awkwardness & problems I found along the way.

in particular:
* some of the docs queued for translation had been moved / renamed (after they had been added to the queue) which caused the workflow to fail. i had to either rename or delete those files in order to run the workflow.
* we had a bug in our code for files that failed. had to fix that in order to run this multiple times.
* this is more of an annoyance than a real problem -- had to create multiple jobs because we could not send all the files to smartling at once. this is due to how we send the requests. i think we can improve this.

also (these arent problems):
* i noticed we can have a japanese document with no english equivalent (due to the problem mentioned in bullet 1 above)
* i saw the pagination error in action 🙈  we already have an issue for that: https://github.com/newrelic/docs-website/issues/2727

jpvajda:
Thanks @nr-kkenney we can discuss at standup today, this is great detail to so highly appreciate the recap.

jmiraNR:
noticed the failed jobs and saw some new name of pages but with the explanation (renaming, moving, etc.) I get it. Thank you @nr-kkenney for the detailed info.

PR #2949, opened 2021-07-02
fixes for translation workflow & added extra logging output for debugging

nr-kkenney:
* added status code & text to output we receive when we run into an error
* fixed a bug that prevented us from saving failed uploads

Relates to: #2872


LizBaker (approved):
🐑 it

PR #2950, opened 2021-07-02
Redoing borked PR :)

dbarnesbrownNR:
If this all goes according to plan I will only need a light peer edit sometime today.

### DIFF
--- a/src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
+++ b/src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
@@ -1,0 +1,98 @@
+---
+title: Drop data using Prometheus remote write 
+tags:
+  - Integrations
+  - Prometheus integrations
+  - Install and configure remote write
+translate:
+  - jp
+redirects:
+  - /docs/
+---
+
+You can drop data you aren't interested in keeping by changing the `remote_write` section of the YAML config file.
+
+<Callout variant="tip">
+You can also drop remote write data using NerdGraph. For more information, see [Drop data using NerdGraph](/docs/accounts/accounts/data-management/drop-data-using-nerdgraph/).
+</Callout>
+
+## Drop entire metric data points from remote write integration [#drop-entire]
+
+If a target is sending a noisy metric that you don't want sent to New Relic, you can specify that New Relic should drop that data. 
+
+### Example 
+
+Let's say you don't want to recieve data for the metric `node_memory_active_bytes` from an instance running at `localhost:9100`. Using the `write_relabel_config` entry shown below, you can target the metric name using the `__name__` label in combination with the instance name.
+
+```
+remote_write:
+- url: https://metric-api.newrelic.com/prometheus/v1/write?prometheus_server=macbook-server-cluster
+  bearer_token: <redacted>
+  write_relabel_configs:
+    - source_labels: ['__name__', 'instance']
+      regex: 'node_memory_active_bytes;localhost:9100'
+      action: 'drop'
+```
+
+This tells Prometheus that you want to do some action against metrics with these labels. To limit which metrics with these labels are affected you must include some value for regex. By default this value is set to `.*` so will include all metrics. In this case, it will drop all metric data points coming out of Prometheus via remote write. 
+
+## Drop specific labels or attributes from data points  [#drop-specific]
+
+If a target is sending specific labesl or attributes you're not interested in receiving, you can drop these from the metrics you receive. 
+
+### Example
+
+Let's say one of your targets is sending a bunch of extra attributes you're not interested in receiving. These might include things like high cardinality attributes such as unique machine identifiers, JVM IDs, or similar. In this case, you need to change both the `remote_write` and the `scrape_configs` section of the YAML file. The result will look something like this. 
+
+```
+remote_write:
+- url: https://metric-api.newrelic.com/prometheus/v1/write?prometheus_server=macbook-server-cluster
+  bearer_token: <redacted>
+  write_relabel_configs:
+    - regex: 'extraLabelToRemove.*'
+      action: 'labeldrop'
+...
+scrape_configs:
+  # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
+  - job_name: 'node'
+    # Override the global default and scrape targets from this job every 5 seconds.
+    scrape_interval: 5s
+    static_configs:
+      - targets: ['localhost:9100']
+        labels:
+          group: 'production'
+          keepLabelName1: 'please-keep-me'
+          extraLabelToRemove: 'please-remove-me'
+          extraLabelToRemove1: 'please-remove-me'
+          extraLabelToRemove2: 'please-remove-me'
+          extraLabelToRemove4: 'please-remove-me'
+          extraLabelToRemove3: 'please-remove-me'
+          extraLabelToRemove5: 'please-remove-me'
+```
+
+## Prometheus or NerdGraph? [#prometheus-vs-nerdgraph]
+
+There are advantages to both dropping data using the method described on this page and using using [NerdGraph](/docs/accounts/accounts/data-management/drop-data-using-nerdgraph/). This section is intended to help you figure out which method is better for your specific needs and preferences. 
+
+### Considerations for the Prometheus config file method
+
+With this method, your dropped data never leaves the associated Prometheus instance. This is a valuable feature if bytes transferred is a cost consideration on the app hosting side.
+
+However, this method may be less appealing than the [NerdGraph](/docs/accounts/accounts/data-management/drop-data-using-nerdgraph/)option due to the following considerations: 
+- Maintained via config yaml files that need to be loaded onto each Prometheus instance (or via a shared storage mechanism) 
+- Requires access to Prometheus server, meaning that either:
+- - The server needs to be restarted
+- - Served must be be accessed at port with path `/-/reload` (assuming the server has lifecycle management enabled as described here in the Prometheus [configuration](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#configuration) docs. 
+
+### Considerations for NerdGraph data droping
+
+NerdGraph is a great option if you want to manage all your data dropping in a single place. It can also be updated easily via the API, and requires no restart or interaction with Prometheus. However, this method applies rules to all incoming data points. This means that you should set up your rules with careful consideration using `WHERE` filtering. 
+
+For more information, see [Drop data using NerdGraph](/docs/accounts/accounts/data-management/drop-data-using-nerdgraph/).
+
+
+## Learn more [#learn-more]
+
+- [Drop data using NerdGraph](/docs/accounts/accounts/data-management/drop-data-using-nerdgraph/)
+- [Send Prometheus metric data to New Relic](/docs/integrations/prometheus-integrations/get-started/send-prometheus-metric-data-new-relic/)
+- [Prometheus High Availability (HA)](https://docs.newrelic.com/docs/integrations/prometheus-integrations/install-configure/prometheus-high-availability-ha/)
--- a/src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
+++ b/src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
@@ -8,7 +8,7 @@ translate:
   - jp
 ---
 
-You can drop data you aren't interested in keeping by changing the `remote_write` section of the YAML config file.
+You can drop data you don't want to keep by changing the `remote_write` section of the YAML config file.
 
 <Callout variant="tip">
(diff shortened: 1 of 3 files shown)


src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
```diff
@@ -0,0 +1,96 @@
...
+title: Drop data using Prometheus remote write 
+tags:
+  - Integrations
+  - Prometheus integrations
+  - Install and configure remote write
+translate:
+  - jp
+---
+
+You can drop data you aren't interested in keeping by changing the `remote_write` section of the YAML config file.
```
kaileyhaynes, line 11: Line 11: I had to re-read "You can't drop data you aren't interested in keeping by changing..." a couple of times in order to understand what it was trying to say. This could be a nitpick.

I would suggest rewriting the first part of this sentence so there isn't two `-ing` words so close together. EX) `You can drop data you don't want by changing` or similar.

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
```diff
@@ -0,0 +1,96 @@
...
+You can also drop remote write data using NerdGraph. For more information, see [Drop data using NerdGraph](/docs/accounts/accounts/data-management/drop-data-using-nerdgraph/).
+</Callout>
+
+## Drop entire metric data points from remote write integration [#drop-entire]
+
+If a target is sending a noisy metric that you don't want sent to New Relic, you can specify that New Relic should drop that data. 
+
+### Example 
+
+Let's say you don't want to recieve data for the metric `node_memory_active_bytes` from an instance running at `localhost:9100`. Using the `write_relabel_config` entry shown below, you can target the metric name using the `__name__` label in combination with the instance name.
```
kaileyhaynes, line 23: Line 23: receive

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
```diff
@@ -0,0 +1,96 @@
...
+remote_write:
+- url: https://metric-api.newrelic.com/prometheus/v1/write?prometheus_server=macbook-server-cluster
+  bearer_token: <redacted>
+  write_relabel_configs:
+    - source_labels: ['__name__', 'instance']
+      regex: 'node_memory_active_bytes;localhost:9100'
+      action: 'drop'
+```
+
+This tells Prometheus that you want to do some action against metrics with these labels. To limit which metrics with these labels are affected you must include some value for regex. By default this value is set to `.*` so will include all metrics. In this case, it will drop all metric data points coming out of Prometheus via remote write.
```
kaileyhaynes, line 35: Line 35: "To limit which metrics with these labels are affected`,` you must include some value for regex."

kaileyhaynes, line 35: Line 35: I'm trying to figure out what is missing in the third sentence starting with `By default...`. Are you trying to say that the value is set to '.*' and will include all metrics?

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
```diff
@@ -0,0 +1,96 @@
...
+    - source_labels: ['__name__', 'instance']
+      regex: 'node_memory_active_bytes;localhost:9100'
+      action: 'drop'
+```
+
+This tells Prometheus that you want to do some action against metrics with these labels. To limit which metrics with these labels are affected you must include some value for regex. By default this value is set to `.*` so will include all metrics. In this case, it will drop all metric data points coming out of Prometheus via remote write. 
+
+## Drop specific labels or attributes from data points  [#drop-specific]
+
+If a target is sending specific labesl or attributes you're not interested in receiving, you can drop these from the metrics you receive.
```
kaileyhaynes, line 39: Line 39: labels

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+This tells Prometheus that you want to do some action against metrics with these labels. To limit which metrics with these labels are affected you must include some value for regex. By default this value is set to `.*` so will include all metrics. In this case, it will drop all metric data points coming out of Prometheus via remote write. 
+
+## Drop specific labels or attributes from data points  [#drop-specific]
+
+If a target is sending specific labesl or attributes you're not interested in receiving, you can drop these from the metrics you receive. 
+
+### Example
+
+Let's say one of your targets is sending a bunch of extra attributes you're not interested in receiving. These might include things like high cardinality attributes such as unique machine identifiers, JVM IDs, or similar. In this case, you need to change both the `remote_write` and the `scrape_configs` section of the YAML file. The result will look something like this.
```
kaileyhaynes, line 43: Line 43: Personal choice here. Since you're giving an example, I would use a `:` instead of a period for "The result will look something like this:"

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
```diff
@@ -0,0 +1,96 @@
...
+          extraLabelToRemove1: 'please-remove-me'
+          extraLabelToRemove2: 'please-remove-me'
+          extraLabelToRemove4: 'please-remove-me'
+          extraLabelToRemove3: 'please-remove-me'
+          extraLabelToRemove5: 'please-remove-me'
+```
+
+## Prometheus or NerdGraph? [#prometheus-vs-nerdgraph]
+
+There are advantages to both dropping data using the method described on this page and using using [NerdGraph](/docs/accounts/accounts/data-management/drop-data-using-nerdgraph/). This section is intended to help you figure out which method is better for your specific needs and preferences.
```
kaileyhaynes, line 73: Line 73: You repeat `using` twice. Remove one of them

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+## Prometheus or NerdGraph? [#prometheus-vs-nerdgraph]
+
+There are advantages to both dropping data using the method described on this page and using using [NerdGraph](/docs/accounts/accounts/data-management/drop-data-using-nerdgraph/). This section is intended to help you figure out which method is better for your specific needs and preferences. 
+
+### Considerations for the Prometheus config file method
+
+With this method, your dropped data never leaves the associated Prometheus instance. This is a valuable feature if bytes transferred is a cost consideration on the app hosting side.
+
+However, this method may be less appealing than the [NerdGraph](/docs/accounts/accounts/data-management/drop-data-using-nerdgraph/)option due to the following considerations:
```
kaileyhaynes, line 79: Line 79: Insert a space between the linked NerdGraph and option.

You also have the same NerdGraph hyperlink and destination a few paragraphs before this one. I would recommend keeping the first hyperlinked NerdGraph instance and removing the hyperlink here.

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+With this method, your dropped data never leaves the associated Prometheus instance. This is a valuable feature if bytes transferred is a cost consideration on the app hosting side.
+
+However, this method may be less appealing than the [NerdGraph](/docs/accounts/accounts/data-management/drop-data-using-nerdgraph/)option due to the following considerations: 
+- Maintained via config yaml files that need to be loaded onto each Prometheus instance (or via a shared storage mechanism) 
+- Requires access to Prometheus server, meaning that either:
+- - The server needs to be restarted
+- - Served must be be accessed at port with path `/-/reload` (assuming the server has lifecycle management enabled as described here in the Prometheus [configuration](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#configuration) docs. 
+
+### Considerations for NerdGraph data droping
```
kaileyhaynes, line 85: Line 85: dropping

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+NerdGraph is a great option if you want to manage all your data dropping in a single place. It can also be updated easily via the API, and requires no restart or interaction with Prometheus. However, this method applies rules to all incoming data points. This means that you should set up your rules with careful consideration using `WHERE` filtering.
```
kaileyhaynes, line 87: Line 87: The comma in the second sentence started with "It can also..." is not needed.

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
```diff
@@ -0,0 +1,96 @@
...
+### Considerations for NerdGraph data droping
+
+NerdGraph is a great option if you want to manage all your data dropping in a single place. It can also be updated easily via the API, and requires no restart or interaction with Prometheus. However, this method applies rules to all incoming data points. This means that you should set up your rules with careful consideration using `WHERE` filtering. 
+
+For more information, see [Drop data using NerdGraph](/docs/accounts/accounts/data-management/drop-data-using-nerdgraph/).
+
+
+## Learn more [#learn-more]
+
+- [Drop data using NerdGraph](/docs/accounts/accounts/data-management/drop-data-using-nerdgraph/)
```
kaileyhaynes, line 94: Line 94: This could be a nitpick too. Since you have the same hyperlink for `Data drop using NerdGraph` right before this one, I would remove one altogether. Readers can either find it under the Learn More tab or when reading "For more information..."

kaileyhaynes (changes requested):
Don't be overwhelmed by the amount of comments. Most are typos that need to be addressed :)

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
dbarnesbrownNR: Yes, and it's a convoluted sentence but to people dropping data this will make sense so I think it's okay. :)

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
dbarnesbrownNR: Good catch, I also moved that sentence to a new paragraph so it stands out as a prompt for the code block.

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
dbarnesbrownNR: Yup, took the top one out so it's in the pertinent section

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
dbarnesbrownNR: Changed this to "Considerations the NerdGraph method" so the header structure is parallel

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
dbarnesbrownNR: good catch!

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
dbarnesbrownNR: Yeah I waffled on that but I think I agree with you. Thanks!

dbarnesbrownNR:
This one has already been peer reviewed and just needs to be approved and moved to `ready for TW to merge` so I can get it merged in. Thanks!

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
dbarnesbrownNR: Fixed, thanks!

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
dbarnesbrownNR: Fixed!

src/content/docs/integrations/prometheus-integrations/install-configure/remote-write-drop-data.mdx
dbarnesbrownNR: Fixed!

urbiz-nr (approved):

PR #2977, opened 2021-07-06
Fixing typo

djaffinito:
Fixing a typo

### Give us some context

Fixing a typo in this:

> Use the inclusion and exclusion filters in CloudWatch Metric Stream is order to select which services / namespaces are being collected.

"is order" should be "in order".

Also changed "in CloudWatch Metric Stream" to "in the CloudWatch Metric Stream"

No


dbarnesbrownNR (approved):
Good catch, thanks!

Issue #2931, opened 2021-07-01
workflow fails attempting to start a translation for document that doesnt exist


nr-kkenney:
### Background
The current send translation content workflow / translation process allows for a doc to be queued up for translation, but not exist.

The scenario is:
1. a doc is edited. lets call it 'random.txt'.
2.  `random.txt` gets added to the translation queue.
3. before the next translation event (1st, 15th of every month) 'random.txt' is renamed (or deleted) to 'random2.txt'.
4. `random2.txt` gets added to the translation queue.
5. we will send `random.txt` and `random2.txt` for translation.

problem: we have a record in the translation queue which indicates 'random.txt' needs to be translated, but the document no longer exists.

### Acceptance Criteria

- [ ]  Don't attempt to translate the old file that was changed and only translate the new file. 
- [ ]  When sending things for translation, skip over documents that no longer exist.
- [ ]  Remove the record (from the translation queue) when a document is renamed or deleted.
- [ ]  Add tests for this scenario.

jpvajda:
Curious wouldn't these files still need translation, regardless of the name change @nr-kkenney ? I also want to ask @jmiraNR about this case, perhaps there a way to ensure if a file name changes we updated the file in queue, so the translation workflow always has the most current file names regardless of how often they change.

jpvajda:
Follow up question: in order to get our workflow enabled again, do these both need resolution first?

jmiraNR:
A file name or location change should not impact the eligibility of a page.

jpvajda:
@jmiraNR can you explain what "eligibility" means in this context?

nr-kkenney:
@jpvajda renaming a file does not change the fact that it needs a translation, but we are running into an edge case in this process that causes the workflow to fail. 

to be clear, renaming a file "should" and most likely will result in the file being translated. to be more explicit about the problem, we are tracking known changed files so that we know what to translate. if a user edits a doc "doc1.txt" that will be saved, and on schedule, the workflow will look at that doc and send it for translation. 

the specific case we may hit is:
1. user edits "doc1.txt". now doc1.txt is queued up for translation.
2. user renames "doc1.txt" to "doc2.txt", or changes its path, or deletes it. in the case of renaming, doc2.txt is now queued up for translation.

but there is only one doc in the repository, doc2.txt -- doc1.txt no longer exists. so we think we need to translate two documents, but there is only one document that exists.

jmiraNR:
[...]

PR #2982, opened 2021-07-06
enable translation workflow to handle 'sending' documents that dont exist

nr-kkenney:
This PR is for allowing the send translation workflow to handle files that are queued for translation, but dont exist due to being renamed or deleted in a separate edit. 

The way this is done is to ignore getting the content for those documents. This prevents them from being sent for translation, which also removes them from the queue, since only failed uploads are left on the queue.

Resolves: #2931

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -4,4 +4,5 @@ module.exports = {
   tabWidth: 2,
   semi: true,
   singleQuote: true,
+  arrowParens: 'always',
 };
--- a/scripts/actions/send-and-update-translation-queue.js
+++ b/scripts/actions/send-and-update-translation-queue.js
@@ -41,6 +41,7 @@ const getContent = (locales) => {
              * this step, it won't become a failed upload, and will then be
              * cleaned up from the queue.
              */
+            console.log(`Skipping over -- ${slug} -- since it no longer exists.`);
             return fs.existsSync(path.join(process.cwd(), slug));
           })
           .map(async (slug) => {
--- a/scripts/actions/send-and-update-translation-queue.js
+++ b/scripts/actions/send-and-update-translation-queue.js
@@ -28,19 +28,30 @@ const DOCS_SITE_URL = 'https://docs.newrelic.com';
  * @param {Object<string, string[]>} locales The queue of slugs to be translated.
  * @returns {Object<string, Promise<Page[]>>}
  */
-const getContent = (locales) =>
+const getContent = (locales) => {
   Object.entries(locales).reduce((acc, [locale, slugs]) => {
     return {
       ...acc,
       [locale]: Promise.all(
-        slugs.map(async (slug) => {
-          const mdx = fs.readFileSync(path.join(process.cwd(), slug));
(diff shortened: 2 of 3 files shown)


aswanson-nr (approved):
LTGM!

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -28,19 +28,31 @@ const DOCS_SITE_URL = 'https://docs.newrelic.com';
...
-          const html = await serializeMDX(mdx);
-          return { file: slug, html };
-        })
+        slugs
+          .filter((slug) => {
+            /**
+             * if a doc doesn't exist, it must have been renamed or deleted. in
+             * that case, it is safe to ignore. if we skip including a doc in
+             * this step, it won't become a failed upload, and will then be
+             * cleaned up from the queue.
```
zstix, line 42: Kinda nitpicky grammar thing, but if we're going to be using periods (`.`), then we should capitalize the first letter of each sentence.

.prettierrc.js
```diff
@@ -4,4 +4,5 @@ module.exports = {
   tabWidth: 2,
   semi: true,
   singleQuote: true,
+  arrowParens: 'always',
```
zstix, line 7: Is there a reason why you added this?

I'm not against it, just wondering if this came out of a discussion. Might lead to some unrelated lines getting checked in in future PRs.

zstix (commented):
One small comment and a question, but this looks great. Dat test tho....so good.

scripts/actions/send-and-update-translation-queue.js
nr-kkenney: 🙈

.prettierrc.js
nr-kkenney: this was driving me a bit crazy in the developer site, where eslint was recommending this change, but my file wouldnt autosave with parens, so i added this there.

i dont know if im just not using the right tools / setup, or if we would care about the same concerns for this repo (since it doesnt have the same commit check). but i figured i would add it here also 🙈

nr-kkenney: im cool with removing this though.

scripts/actions/send-and-update-translation-queue.js
zstix: It's a matter of preference. I personally like lots of comments in the code, but I know that it's not everyones cup of tea 🍵

.prettierrc.js
zstix: I just wanted to make sure that it was an intentional adjustment rather than a fix for your local setup. As long as future PRs wont have a ton of noise from this stuff getting updated, then I'm good with it 👍

scripts/actions/send-and-update-translation-queue.js
aswanson-nr: Tossing in my vote to keep the comment!

aswanson-nr (dismissed):

aswanson-nr (approved):

PR #2985, opened 2021-07-07
Update release.yml

rudouglas:
Added a while loop to check if a tag already exists to mitigate edge case of 2 tags getting created at the same time


nr-kkenney (approved):

rudouglas:
closing this as I have updated to the timestamp format in another PR

PR #2988, opened 2021-07-07
Increase max layout column of a widget

hredondo:
### What

Small fix in the limits of a widget. In this case the maximum column a widget can have.


src/content/docs/query-your-data/explore-query-data/dashboards/dashboards-api.mdx
```diff
@@ -106,4 +106,4 @@ All dashboard mutations offer a way to ask for errors when being executed. This
   </Collapser>
 </CollapserGroup>
 
-Keep in mind that these are expected errors that we are aware of in advance. You should also check for unexpected errors that will be returned in the standard [GraphQL errors field](https://graphql.org/learn/serving-over-http/#response).
```
hredondo, line 109: I don't know why this is recognized as changed. I edited the file directly in github.

x8a: I'd say we can ignore it since the content doesn't seem to have changed.

x8a (approved):
Thanks for the update!

PR #3035, opened 2021-07-09
feat: adding retry logic to translations with tests

rudouglas:


rudouglas:
A couple of questions on this:

 - How many times do we want this to attempt a retry before failing outright
 - How long a timeout do we want between each retry

rudouglas:
Just realised I may have missed a couple of instances of requests that also need retry logic, namely [uploadFile](https://github.com/newrelic/docs-website/blob/8ee2ad63de4b7bcd5534939bea271e8040960230/scripts/actions/send-and-update-translation-queue.js#L63) and [sendPageContext](https://github.com/newrelic/docs-website/blob/8ee2ad63de4b7bcd5534939bea271e8040960230/scripts/actions/send-and-update-translation-queue.js#L109)

nr-kkenney:
@rudouglas to answer your first question
1. i think just pick a starting point and if it doesnt work, we can tweak it down the line. really though, this is slightly dependent on the second point. i would just start in the range 3-10.
2. the libraries ive used in the past use exponential backoff with jitter. see here for an example: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

scripts/actions/utils/vendor-request.js
```diff
@@ -1,19 +1,53 @@
...
+    console.warn(
+      `[!] Error making request on attempt ${nthTry}/${MAX_RETRY}. Retrying in ${
+        DELAY_TIME / 1000
+      } seconds`
     );
-    throw new Error(JSON.stringify(response, null, 2));
+    // wait for delayTime amount of time before calling this method again
+    await waitFor(DELAY_TIME);
+    return makeRequest(url, options, nthTry);
   }
```
nr-kkenney, line 50: i will write some thoughts since i wanted to review this, but since my javascript knowledge is not the strongest, i think it would be worth grabbing another set of eyes 🙈 

1. im not a fan of the use of recursion here. i think a loop would be easier to read & understand whats happening.
2. it looks like youre using a try/catch as an if/else based on the fetch output, and that seems a bit awkward to me.
3. for the last line `return makeRequest` does that not need an await since the method is async?

nr-kkenney: im also still in favor of using a library, since hopefully someone has already worked out some of this complexity & has some of the features i mentioned in an overall pr comment, if you happen to find a good candidate. feel free to veto this though.

aswanson-nr: +1 to Kris's comment. Using a library for this logic reduces the amount of code and tests we have to maintain. @nr-kkenney found a library, https://github.com/mauricedb/polly-js, that could be perfect for this.

zstix: I'd love to grab some time in a campfire to chat about it, cause I actually really like the use of recursion here. I'd probably restructure things a little to reduce the logic, but this makes sense to my brain:

```js
const makeRequest = async (url, options, retriesRemaining = MAX_RETRY) => {
  try {
    // make an async call
  } catch (e) {
    if (retriesRemaining === 0) {
      // bail out
    } else {
      await sleep(DELAY_TIME);
      return makeRequest(url, options, retriesRemaining - 1);
    }
  };
}
```

zstix (commented):
Good dialog on this ticket. Happy to keep chatting about it.

scripts/actions/utils/vendor-request.js
```diff
@@ -1,19 +1,53 @@
...
-      `[!] Unable to make a ${options.method} request to ${url.href}. (${code})`
+/**
+ * @param {Object} options
+ * @param {String} url
+ * @param {Number} nthTry
+ * @returns {Object} data The result after making the request.
+ * @throws {Error} Will throw an error if the response "code" is not 'SUCCESS' after retrying
+ */
+const makeRequest = async (url, options, nthTry) => {
+  nthTry = nthTry || 1;
```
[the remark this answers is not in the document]
rudouglas: oooh nice thanks

rudouglas:
Actually don't merge yet, need to change the tests

nr-kkenney:
@rudouglas `cockatiel` was the library i found, and this was the example i came up with for how we would use it (that has the same functionality as your pr): https://github.com/newrelic/docs-website/pull/3219/files

PR #3056, opened 2021-07-12
fix: infra agent config args

varas:
Newer config format does not support `arguments` for CLI args.


urbiz-nr (approved):

urbiz-nr:
Thanks for updating our docs, @varas!

PR #3057, opened 2021-07-12
Updates to Java agent API annotations

meiao:
### Give us some context
These annotations were added some time ago to the Java agent but were not documented previously.

No


bradleycamacho:
Thanks for these updates @meiao! Merging in now

Issue #626, opened 2021-01-22
on the Attribute Dictionary  right rail add a free form search field.


jpvajda:
A user should be able to quickly search for an attribute using a free form search field, and when they execute a search it would return any matching attributes. 

<img width="384" alt="Screen Shot 2021-01-21 at 6 39 25 PM" src="https://user-images.githubusercontent.com/1395158/105434283-7b262500-5c18-11eb-9f2b-ddfd834d1619.png">

jpvajda:
This would be useful but definitely a stretch item for our MMF13.

austin-schaefer:
Request from duplicate issue #1281:

> Please restore attribute search capability to the Data Dictionary. In the current version you must select the event type first then choose the attribute. When planning optimal NRQL, it is helpful to know which attributes are shared by which event types. As such, attribute search very helpful.

zuluecho9:
There is probably hidden complexity to implementing this issue. Couple reasons: 1) some attributes are redundantly documented and have multiple entries despite having the same name. 2) Ideally, searching for an entry would land you on an attribute entry page that doesn't display the definition of the data type it's attached to at top, but I think currently an attribute entry will display a data type entry above it.

austin-schaefer:
Still very much a valid request.

bradleycamacho:
This has been requested again in #2289.

paperclypse:
I think this would also address this external issue that came in today #2351

paperclypse:
Would it be possible to index the data dictionary terms in our site search? That might be an easier, interim solution.

roadlittledawn:
@austin-schaefer || @jpvajda . if / when we add a free form search for attribute names, would we want to remove the dropdown to select a specific attribute?

jpvajda:
@roadlittledawn I'd say we'd keep the drop down as it helps users see what is available to them, or they can enter a search query... also interested in what @austin-schaefer thinks..

roadlittledawn:
@jpvajda i'll put a PR preview up for this so you can poke at it, BUT that attribute dropdown is only enabled when a single event is filtered on. otherwise it could potentially be a long list. and the list could have options/attributes that have the same names (because an attribute can be in multiple events OR there could be distinct attributes with same names that are because they mean something else within their event).

roadlittledawn:
here's that preview link: https://docswebsitedevelop-clangalbaattributesearch.gtsb.io/

austin-schaefer:
[...]

PR #3073, opened 2021-07-12
Support searching by attribute name in data dictionary

roadlittledawn:
## Description

Replaces the filter for selecting a specific attribute with a text input to search for attributes by name. Just matches on any instance of the search string in the attribute name. It lower cases the search string and the attribute names when matching.

Adds ability to apply filters by typing `enter`.

## Related issues / PRs

Closes https://github.com/newrelic/docs-website/issues/626

## Screenshot(s)

https://user-images.githubusercontent.com/2952843/125660278-2a5aed92-3364-4001-802c-9507f4e3255a.mp4

### DIFF
--- a/src/pages/attribute-dictionary.js
+++ b/src/pages/attribute-dictionary.js
@@ -172,208 +172,220 @@ AttributeDictionary.propTypes = {
 
 const pluralize = (word, count) => (count === 1 ? word : `${word}s`);
 
-const EventDefinition = memo(({ location, event, searchedAttribute }) => {
-  const filteredAttributes = searchedAttribute
-    ? event.childrenDataDictionaryAttribute.filter(({ name }) =>
-        name.toLowerCase().includes(searchedAttribute.toLowerCase())
-      )
-    : event.childrenDataDictionaryAttribute;
+const EventDefinition = memo(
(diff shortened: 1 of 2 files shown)


roadlittledawn:
i added a spinner while it's performing a search, buuut not 100% sure on the implementation because i had to use a `setTimeout()` in `useEffect()`. not sure that's ideal because it may prolong some searches a bit. i do like some visual cue that something is happening though since it sometimes takes ~3s to search (esp when rendering most of the events)

roadlittledawn:
looks like there's some prettier reformatting in the `attribute-dictionary.js` file ¯\_(ツ)_/¯

src/pages/attribute-dictionary.js
```diff
@@ -180,207 +170,219 @@ AttributeDictionary.propTypes = {
...
-        <div
+        <h2
           css={css`
-            position: relative;
+            position: sticky;
+            top: var(--global-header-height);
+            background: var(--primary-background-color);
+            padding: 1rem 0;
+
+            // cover up the right table border
```
zstix, line 206: Oh neat, I didn't realize we could do comments like this!

zstix (approved):
Nice work!!!

paperclypse:
YES! So cool!

PR #3081, opened 2021-07-13
DOC-6314 - Provider account ID now visible in the UI

x8a:


urbiz-nr (approved):
Tried locally, looks good!

PR #3082, opened 2021-07-13
Remove export single-paged dashboards

hredondo:
### What

Removing the single-paged use case to export dashboards as PDF. This is not valid anymore, and users should follow the same process regardless the amount of pages.

### Why?

After the homogenization of dashboards, avoid user errors when exporting single-paged dashboards.


urbiz-nr (approved):
Tried locally, looks good! Thanks @hredondo!

PR #3083, opened 2021-07-13
Adding benefits to AUM docs

zuluecho9:
Liaison work, we noticed we didn't have benefits in docs for AUM. Also reworded some stuff in auth domain doc, adding links to new overview SAML/SCIM doc.


urbiz-nr (approved):
Nice updates!

PR #3087, opened 2021-07-13
Add search by attribute (auto search)

roadlittledawn:
Draft PR for preview. Performs auto search as user types in attribute name search input

Branched off of https://github.com/newrelic/docs-website/pull/3073


jpvajda:
@roadlittledawn like you reported, there is definitely some wonky behavior here. like entering in characters is delayed, and choppy.  so unless we can improve this, I'd say we just go back to forcing the user to submit the search query via the search button.

PR #3089, opened 2021-07-13
Removing a nested nav in NR in practice

aswanson-nr:
# Summary
A quick test of removing the nested "New Relic One" nav from the "New Relic in practice" nav.


zuluecho9:
I think this is related, but not entirely sure, but just wanted to draw attention to it https://newrelic.atlassian.net/browse/DOC-7142

x8a (approved):

PR #3096, opened 2021-07-14
Update nrql-syntax-clauses-functions.mdx

nedl86:
Added information about using `OR` operator in conjunction with `FACET CASES`.


urbiz-nr (approved):
Tried locally, looks great!

PR #3098, opened 2021-07-14
Pixie beta users doc

mmfred:
This adds a doc for Auto-telemetry with Pixie beta users to upgrade their accounts for use with the generally available version.


urbiz-nr (approved):

urbiz-nr:
Looks good @mmfred!

nserrino (commented):
I don't believe linking and unlinking are necessary for beta customers to upgrade to the GA experience.

PR #3112, opened 2021-07-14
chore: sort top level nav list before comparison

aswanson-nr:
# Summary
When deciding which nav to show for a given page, the path is compared against the top level navs of the site. This was previously done in whatever order the query returned. This PR explicitly sorts the list so that we can always be sure which nav is used.

### DIFF
--- a/plugins/gatsby-source-nav/gatsby-node.js
+++ b/plugins/gatsby-source-nav/gatsby-node.js
@@ -277,7 +277,7 @@ const createNav = async ({ args, createNodeId, nodeModel, locales }) => {
 
   const allNavYamlNodes = nodeModel
     .getAllNodes({ type: 'NavYaml' })
-    .sort((a, b) => a.path.localeCompare(b.path));
+    .sort((a, b) => a.title.localeCompare(b.title));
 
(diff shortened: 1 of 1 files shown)


plugins/gatsby-source-nav/gatsby-node.js
```diff
@@ -275,13 +275,15 @@ const createNav = async ({ args, createNodeId, nodeModel, locales }) => {
     },
   });
 
-  const allNavYamlNodes = nodeModel.getAllNodes({ type: 'NavYaml' });
+  const allNavYamlNodes = nodeModel
+    .getAllNodes({ type: 'NavYaml' })
+    .sort((a, b) => a.path.localeCompare(b.path));
```
zstix, line 280: We might want to compare against the title rather than the path, since _technically_ the path isn't required (nav sections where the section title doesn't have a link):
Screen Shot 2021-07-15 at 11 49 01

aswanson-nr: Good call! I updated it to sort on title instead.

zstix (approved):

PR #3113, opened 2021-07-15
Remove chunky table from Prometheus doc

kaileyhaynes:
I removed a chunky table and replaced it with two clamshells to better organize the information and condense the page's length. 

Jira here: https://newrelic.atlassian.net/browse/DOC-6562


dbarnesbrownNR (commented):
This all looks good except I am not 100% certain if we should pull the recommendations at this time. I'll follow up by PM so we can get this merged.

PR #3127, opened 2021-07-15
Updated version to 7.1.1 per Java team.

rhetoric101:
We need to update the "What's new?" post with the new supported version. It is now 7.1.1.


paperclypse (approved):
Looks good!

PR #3132, opened 2021-07-16
Per-account trace rate limits

rhetoric101:
I initially made these per-account rate limiting changes over a month ago on the private site. While I waited for signoff, the branch got stale. So, to speed up the migration, I just manually moved these over when I received SME confirmation that we can push these out.

Here's what to review:

DOC #1 (Trace API general requirements and limits)

* Migrated four of the limits entries from the table over to New Relic data usage limits and policies. As I understand it, the remaining rows were either already in the main limits doc or the rows were unrelated to limits.
* The table had two miscellaneous items unrelated to limits, so I moved the HTTPS requirement as just a a modifier in the first section about endpoints. I dropped the obfuscation comment because I couldn’t find a better spot, and this appears in the troubleshooting section. Let me know if you have other ideas.
* Created a new Data limits section by combining the old limits section (minus the table) with the old section about exceeding span limits. 

DOC #2 (New Relic data usage limits and policies)

* Migrated four of the Trace API limits table rows over to this doc.
* Prefixed each DT-related limit with “Distributed tracing:”

DOC #3 (How distributed tracing works)

I inserted a link in the collapser for Trace rate limiting that goes to the licenses page (above) that talks about limits.


zuluecho9 (approved):

zuluecho9:
@rhetoric101 One thing I wondered: maybe we should put a link to the data limits in the 'How DT works' doc? I'll leave that in your court, that was the main thing I wondered, if people would be going there to find that out.

PR #3152, opened 2021-07-16
Update and rename new-relic-ios-3760.mdx to new-relic-ios-3761.mdx

jvanmetre:
Apple review got stuck in 'in review' for greater then 2 days, so I needed to reject and submit a new binary... this should be released by monday so update release date.


urbiz-nr (approved):
Thanks for the update @jvanmetre!

PR #3159, opened 2021-07-19
Update references to the entity-definitions repository

flisflis:
Update references to the entity-definitions repository after its graduation out of the newrelic-experimental org

### Give us some context

This PR fixes outdated links to the newrelic-experimental organization for entity type definitions. You can see an example of outdated link on the [Find and explore entities section of the What is an entity in New Relic? page](https://docs.newrelic.com/docs/new-relic-one/use-new-relic-one/core-concepts/what-entity-new-relic/#find)


urbiz-nr (approved):
Thanks for the updates, @flisflis!

PR #3160, opened 2021-07-19
chore: using gatsby@next

rudouglas:


jpvajda:
@rudouglas Thanks for testing this out! My thought is we wait until this feature is in a production version of Gatsby and is officially released before upgrading. it looks like it's targeted for Gatsby 3.11

PR #3172, opened 2021-07-19
Updated title for distributed tracing quick start

rhetoric101:
Reviewer: I updated the title so it makes more sense when you search for this topic.

Also, please build this locally and see if you can get the images to load. They would not load for me locally! Initially, I didn't change anything related to the images, but then I saw they had poor titles, and indentation. So, I went ahead and fixed these issues, but the images still won't load locally for me.


barbnewrelic (approved):
I checked this out in my localhost and it looks great. Title change is there compared to live site, and the images include a description when I moused over them. Go ahead and merge your PR whenever you are ready. Congrats!

PR #3174, opened 2021-07-19
Updated Infra related time limits

nathanweddle:
### Give us some context

All conditions now have TTL's applied to them. It is no longer allowed to have a condition without a TTL. The default value for infrastructure conditions is 24 hours. 0 (disabled) is no longer allowed. If a user sends 0, we will replace it with 24.
The Infra UI is being updated this week. The enable/disable checkbox is being removed and the default value will be 24. The user must use the default or change it to a pre-defined value.


barbnewrelic (approved):
Thanks for updating this doc. Helpful explanations of what happens and when!

PR #3175, opened 2021-07-19
Update completed translation workflow to use new database calls

### LINKED ISSUE #3078: Update workflow that handles completed translations (& making a PR)
## Description
Once we have a new database and utility functions for reading / writing to it (#3074), we need to update [the workflow](https://github.com/newrelic/docs-website/blob/main/.github/workflows/check-translations-and-deserialize.yml) that checks for the job status, downloads the new files, and makes a PR.

Much of the existing workflow should stay the same (deserializing, making the PR, etc).

## Needs Further Clarification
At some point in the workflow (and perhaps it is at this point), we need to remove any localized files that correspond to english files that were deleted (in PRs spawning this process). How we handle this is still unclear to me (-Z).

## Acceptance Criteria
* [x] Check the job status given a job ID in the new database
* [x] If a job has been completed, download the files, deserialize them, and make a PR
* [x] Update the job status in our database to be "in-review" (or something similar)
* ~Come up with a solution for english files that were deleted in the PRs related to these files (see above)~

elaguerta:
- Closes #3078 
- Changes scripts called in .`check-translations-and-deserialize.yml` to replace calls to old database functions with new functions in `scripts/actions/translation_workflow/database.js`
- At the end of the workflow, updates job statuses to "IN_REVIEW"

### DIFF
--- a/scripts/actions/remove-completed-batch.js
+++ b/scripts/actions/remove-completed-batch.js
@@ -1,4 +1,5 @@
 const { saveToTranslationQueue } = require('./utils/save-to-db');
+const { getJobs, updateJob } = require('./translation_workflow/database.js');
 const checkArgs = require('./utils/check-args');
 const { getAccessToken, vendorRequest } = require('./utils/vendor-request');
 const fetch = require('node-fetch');
@@ -10,10 +11,10 @@ const DOCS_SITE_URL = 'https://docs.newrelic.com';
  * Updates the "being translated" queue with the batches that are not done.
  * @returns {Promise}
  */
-const saveRemainingBatches = async () => {
+const setInProgressToDone = async () => {
   checkArgs(4);
 
-  const batchUids = process.argv[2].split(',').filter(Boolean);
+  const completedBatches = process.argv[2].split(',').filter(Boolean);
   const deserializedFileUris = process.argv[3].split(',').filter(Boolean);
 
   const code = await removePageContext(deserializedFileUris);
@@ -22,11 +23,10 @@ const saveRemainingBatches = async () => {
     console.log(`[!] Unable to delete all contexts`);
   }
 
-  await saveToTranslationQueue(
-    { type: 'being_translated' },
-    'set batchUids = :batchUids',
-    { ':batchUids': batchUids }
-  );
+  for (const batch_uid of completedBatches) {
+    const job = await getJobs({ batch_uid });
+    updateJob(job.id, { status: 'IN_REVIEW' });
+  }
 };
 
 /**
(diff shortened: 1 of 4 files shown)


scripts/actions/remove-completed-batch.js
```diff
@@ -22,11 +23,10 @@ const saveRemainingBatches = async () => {
...
   }
 
-  await saveToTranslationQueue(
-    { type: 'being_translated' },
-    'set batchUids = :batchUids',
-    { ':batchUids': batchUids }
-  );
+  for (const batch_uid of completedBatches) {
+    const job = await getJobs({ batch_uid });
+    updateJob(job.id, { status: 'IN_REVIEW' });
```
nr-kkenney, line 28: i think this will need an await. also, i believe status needs to be an integer since its a numeric value in the table.

if we want, we can change that. but it might be just as easy to create or refer to an enum somewhere (probably database.js) so that you can do something like:
```js
await updateJob(job.id,  { status: statuses.IN_REVIEW }
```

elaguerta: Thanks for the review. I can add an `await`, but is it necessary if nothing else depends on it? Also with `await`, might a long update or one that errors out block the remaining updates in a for loop?

I couldn't find the enum mapping for statuses. I did find in `/scripts/models/status.js` that Status.status looks like is will be a `String` in this scope. Here is a snippet of the Status structure:
 
```javascript
module.exports = (sequelize, DataTypes) => {
  const status = sequelize.define(
    'status',
    {
      id: {
        type: DataTypes.INTEGER,
        allowNull: false,
        primaryKey: true,
        autoIncrement: true,
      },
      status: {
        type: DataTypes.TEXT,
        allowNull: false,
        unique: true,
      },
    },
    {
      tableName: 'statuses',
    }
  );

  return status;
};

Comment (nr-kkenney) [reply scripts/actions/remove-completed-batch.js:28]: 1. for async, good call. i'll defer to you since my knowledge on js is limited. mainly i just want to make sure we capture errors if any occur. if we need await, and want the concurrency, i think we could alter that to use map instead with promise.all, and await that instead. if not, forget i said anything.

2. this is kind of my bad 🙈  you can ignore what i said here (the documentation i have for the method says numeric, but thats incorrect).

**Review by nr-kkenney — APPROVED** [review]

File: scripts/actions/remove-completed-batch.js (hunk 1 above)
Comment (zstix) [reply scripts/actions/remove-completed-batch.js:28]: Oh good point. We're not really `catch`ing any errors here (or logging successes). Perhaps we should have wrapped this in a `try` / `catch` block and done something like this for the success case:

```js
for (const batch_uid of completedBatches) {
  const job = await getJobs({ batch_uid });
  await updateJob(job.id, { status: 'IN_REVIEW' });
  console.log(`Status of job ${job.id} updated to ${job.status}`);
}
```

PR #3181, opened 2021-07-20
chore: update send to be translated workflow

rudouglas:
Notes:
Do we need to use the `status` database table for anything?


nr-kkenney:
@rudouglas i think statuses & locales probably look out of place, and are tables we could remove because they are simple maps & value storage. so instead of a separate table, we could just apply a value directly to the other tables. we could also have this live directly in code rather than a db table.

i prefer this approach for a couple reasons, although maybe the columns could be changed up:
1. the value is used & defined via the database, which i have a preference for over having this live in the code. so as an example, if you have a translation record with a "status" column, you can look up status in the db rather than have to look at the code (even if this is trivial).
2. i like a separate table over just defining an enum within the database. a separate table is just easier to work with and extend -- we can just add a new row to represent a new state.

while i do prefer this, i dont think the lift is too large either way if we prefer to not have this.

scripts/actions/translation_workflow/database.js
```diff
@@ -154,6 +154,18 @@ const addTranslationsJobsRecord = async (translationId, jobId) => {
   return record;
 };
 
+/**
+ * Method to get records from 'translations_jobs' table.
+ * @param {number} jobId - identifier of job
+ */
+const getTranslationsJobsRecords = async (jobId) => {
```
nr-kkenney, line 161: this works, but id say copy the same setup as the other get methods 🙈  where an object is accepted, and jobid (or translationId) can be a field on the object used to get results.

rudouglas: @nr-kkenney you are always biased 😂 I can't remember why i deviated from the others but i'm sure it was based on something 😛 i'll update that now

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -172,6 +180,14 @@ const sendContentToVendor = async (content, accessToken) => {
   const jobUids = jobsResponses.map((resp) => resp.translationJobUid);
   console.log(`[*] Successfully created jobs: ${jobUids.join(', ')}`);
 
+  const createTranslationJobRecord = jobUids.map((jobUid) =>
+    addTranslationsJobsRecord(jobUid)
+  );
```
nr-kkenney, line 185: i think you will need to await Promise.all for this.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -190,8 +206,14 @@ const sendContentToVendor = async (content, accessToken) => {
   });
 
   const batchResponses = await Promise.all(batchRequests);
+  const initialStatus = 'PENDING';
+  for (const response of batchResponses) {
+    addJob(response.job_uid, response.batch_uid, initialStatus);
```
nr-kkenney, line 211: this will also need await. additionally, the status we pass in should be integer value, not text.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -201,79 +223,60 @@ const sendContentToVendor = async (content, accessToken) => {
   });
 
   const fileResponses = await Promise.all(fileRequests);
-  const numSuccess = fileResponses.filter(({ code }) => code === 'ACCEPTED');
-
+  const successfulUploads = fileResponses.filter(
+    ({ code }) => code === 'ACCEPTED'
+  );
+  await updateStatus(successfulUploads, 'IN_PROGRESS');
```
nr-kkenney, line 229: status here will also need to be integer, not text.

rudouglas:
@nr-kkenney ahhh I think I get you now, so the name of the status (the `status` attribute in the table) would be pulled using the `id` from within the code, rather than having the string in the code itself, is that right? I think i was just confused on how this would be implemented in this particular step but if that's what you're saying it makes sense.

rudouglas:
also where can we see the integer values of that table, can we see them in AWS?

nr-kkenney:
@rudouglas you can ignore what i said about the integer values. i have comments in the code mentioning numeric, but the tables with foreign key to status are using text values, so i confused the two in my mind.

scripts/actions/send-and-update-translation-queue.js
rudouglas: yessir

rudouglas:
I merged the feature branch into this but i'm going to close it to create a new PR to the Feature branch

PR #3185, opened 2021-07-20
Clarifying Metric API timestamp units accepted

zuluecho9:
Small change, removing that timestamp must be in miliseconds and linking to that section for more info.


bradleycamacho (approved):
Looks good, feel free to merge

PR #3193, opened 2021-07-20
added a callout regarding terms of service.

mmfred:
added callout to Pixie docs on request from @toobagrrl


dbarnesbrownNR (approved):
Looks good!

PR #3194, opened 2021-07-20
Update user roles section

mmfred:
Updated some wording around roles according to style guidelines.


dbarnesbrownNR (approved):
Looks good to me!

PR #3203, opened 2021-07-21
Added note about health status updated by NRQL queries

urbiz-nr:
As reported in DOC-7179, health status can be updated by NRQL queries that include the FACET clause applied to entity GUIDs. I'm adding a little, temporary note to mention this much. Nothing too big since this behavior is to be changed this quarter.


barbnewrelic (approved):
Really appreciated your explanation of timeline. That made my review ez-pz. Looks just fine!

urbiz-nr:
This shouldn't be merged. See DOC-7179 for details.

PR #3205, opened 2021-07-21
[DOC-7133] Duplicate dashboards

urbiz-nr:
Searched for clone/cloned/cloning in dashboards docs and replaced for duplicate, as per new UI copy. See DOC-7133.


x8a (approved):
Looking great 👌🏻

PR #3206, opened 2021-07-21
Ru/send pending job

rudouglas:


rudouglas:
Original [PR](https://github.com/newrelic/docs-website/pull/3181) 
Seems to be failing on some of the tests, but there seems to be differences when testing locally

In GH:
```
[PASS scripts/actions/__tests__/check-for-outdated-translations.test.js](https://github.com/newrelic/docs-website/runs/3122647614?check_suite_focus=true#step:6:14)
[FAIL scripts/actions/__tests__/send-and-update-translation-queue.test.js](https://github.com/newrelic/docs-website/runs/3122647614?check_suite_focus=true#step:6:32)
```

Locally:
```
FAIL  scripts/actions/__tests__/check-for-outdated-translations.test.js
PASS  scripts/actions/__tests__/send-and-update-translation-queue.test.js (7.086 s)
```

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -172,6 +180,15 @@ const sendContentToVendor = async (content, accessToken) => {
   const jobUids = jobsResponses.map((resp) => resp.translationJobUid);
   console.log(`[*] Successfully created jobs: ${jobUids.join(', ')}`);
 
+  const queueTranslationsJobsRecord = jobUids.map((jobUid) =>
+    addTranslationsJobsRecord(jobUid)
```
nr-kkenney, line 184: i think this needs translationId as well.

rudouglas: oh, which id does that correlate to then, is it the translation id from the Translations table? I don't see how to correlate to that with the data we have at this point, i may need to rewrite this a bit but we should campfire 😬

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -190,8 +207,14 @@ const sendContentToVendor = async (content, accessToken) => {
   });
 
   const batchResponses = await Promise.all(batchRequests);
+  const initialStatus = 'PENDING';
+  for (const response of batchResponses) {
+    await addJob(response.job_uid, response.batch_uid, initialStatus);
```
nr-kkenney, line 212: id say just pass 'PENDING' directly here.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -201,79 +224,60 @@ const sendContentToVendor = async (content, accessToken) => {
...
-      : [];
-
-  await saveToTranslationQueue(key, 'set batchUids = :batchUids', {
-    ':batchUids': [...queue, ...batchUids],
-  });
+const updateStatus = async (successfulUploads, status) => {
+  for (const upload of successfulUploads) {
+    const batch_uid = upload.batchUid;
+    const job = await getJobs({ batch_uid });
+    const translationId = await getTranslationsJobsRecords({ jobId: job.id });
```
nr-kkenney, line 246: this should be:
```js
const translationId = await getTranslationsJobsRecords({ job_id: job.id });
```

to match the db column name. if it doesn't, this will throw an error.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -201,79 +224,60 @@ const sendContentToVendor = async (content, accessToken) => {
...
-      ? data.Item.batchUids
-      : [];
-
-  await saveToTranslationQueue(key, 'set batchUids = :batchUids', {
-    ':batchUids': [...queue, ...batchUids],
-  });
+const updateStatus = async (successfulUploads, status) => {
+  for (const upload of successfulUploads) {
+    const batch_uid = upload.batchUid;
+    const job = await getJobs({ batch_uid });
```
nr-kkenney, line 245: getJobs will return array, so i think you want:
```js
const [job] = await getJobs({ batch_uid });
```

if you are expecting only one matching result.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -201,79 +224,60 @@ const sendContentToVendor = async (content, accessToken) => {
...
-
-  await saveToTranslationQueue(
-    { type: 'to_translate' },
-    'set locales = :locales',
-    { ':locales': updatedLocales }
-  );
+const outputFailedUploads = async (failedUploads) => {
+  for (const failedUpload of failedUploads) {
+    console.log(`    [!] ${JSON.stringify(failedUpload)}`);
+  }
```
nr-kkenney, line 258: i think we will also want to mark this as a failure in the workflow in addition to printing out the failed files.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -41,7 +62,9 @@ const getContent = (locales) => {
              * this step, it won't become a failed upload, and will then be
              * cleaned up from the queue.
              */
-            console.log(`Skipping over -- ${slug} -- since it no longer exists.`);
+            console.log(
+              `Skipping over -- ${slug} -- since it no longer exists.`
+            );
```
nr-kkenney, line 67: this will log every time. whoops 🙈

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -54,6 +77,116 @@ const getContent = (locales) => {
...
+ * @param {string[]} locales a list of locales
+ * @param {string[]} jobUids a list of created JobUids
+ * @returns {Object[]} translation jobs records
+ */
+const createTranslationJobRecords = async (locales, jobUids) => {
+  const queueTranslationsJobsRecords = jobUids.map((uid, index) => {
+    return addTranslationsJobsRecord({
+      locale: locales[index],
+      uid,
+    });
```
nr-kkenney, line 115: this function takes translationId, and jobId as arguments.

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -54,6 +77,116 @@ const getContent = (locales) => {
...
+    )}`
+  );
+};
+
+/**
+ * @param {string[]} locales a list of locales
+ * @param {string[]} batchUids a list of created Batch uids
+ * @param {string} accessToken
+ * @param {Array<string[]>} pages an array of page arrays to be translated}
+ * @returns {(page: Page) => Promise<{code: string, slug: string, locale: string>}
```
nr-kkenney, line 177: this is no longer correct. i think you can just put Promise<{code : . . .

scripts/actions/send-and-update-translation-queue.js
```diff
@@ -147,134 +279,68 @@ const sendPageContext = async (fileUri, accessToken) => {
 };
 
 /**
- * Sends HTML content to the vendor by creating jobs, batches, and uploading
- * files. On success, this will return the batchUid for each locale.
- * @param {Object<string, Page[]>} content
- * @param {string} accessToken
- * @returns {Promise<{batchUids: string[], fileResponses: Object[]>}
+ * @param {(page: Page) => Promise<{code: string, slug: string, locale: string>}} uploadFileResponses
```
nr-kkenney, line 282: i think this just can be uploadFileResponse[]

nr-kkenney:
no longer needed since #3222 merged. closing.

PR #3222, opened 2021-07-22
bulk changes to send translation script

nr-kkenney:
### Description

* pending translations are sent to vendor for translation, and database records are created & updated appropriately
* added unit tests
* updated code docs in database.js
* moved sendPageContext and uploadFile methods for easier testing


zstix (approved):
There was a lot to review here. It looks good to me, but we might want to review it with the team so that everyone has context.

scripts/actions/__tests__/send-and-update-translation-queue.test.js
```diff
@@ -1,23 +1,314 @@
...
+const { vendorRequest, uploadFile } = require('../utils/vendor-request');
+const {
+  getReadyToGoTranslationsForEachLocale,
+  createJobs,
+  createBatches,
+  uploadFiles,
+} = require('../send-and-update-translation-queue');
 
-jest.mock('../serialize-mdx');
+console.log = () => {};
```
aswanson-nr, line 25: Instead of reassigning the function, you can just have Jest mock it out for you.
```suggestion
jest.spyOn(console, "log")
```

nr-kkenney: @elaguerta mentioned using `--silent` as an option when running the tests. i like that, so i will remove this line entirely, and to suppress console output just run the tests as: `yarn test --silent`

PR #3225, opened 2021-07-22
Updating HNR to include the no_trigger_on toggle

ryanv94:
Howdy Docs team - updating the example here for the Host Not Reporting condition type to include the optional no_trigger_on toggle. This enables the setting 'Don't trigger alerts for hosts that perform a clean shutdown'

### Give us some context

* The current Rest API for Infra Conditions docs don't reflect an important setting that is available in the UI, and via API. The API part is (until now) undocumented.


x8a (approved):
Thanks for the update!

PR #3233, opened 2021-07-22
Small changes to doc

HenryTech:
Process Running (PR) alert conditions are being migrated to use NRQL (Willamette pipeline). This move will mean that they can no longer tell if a process failed to start when a host is spun up. I have tried to remove all references to that sort of behavior, since PR conditions will change to being more "Process NO LONGER running."

I feel like these changes will not need to be applied until we move the PR condition types over to Willamette (happening soon, but has not happened yet). On the other hand, these don't feel like breaking changes to the document, so could be published immediately without any impact (I think).

Please do send me a DM in Slack (@shenry) if you have any questions about these changes.

### Give us some context


mmfred (approved):

mmfred:
Thanks for the fixes @HenryTech!

PR #3247, opened 2021-07-23
Update android-610.mdx

Sim-newrelic:
Fixed the issue we are seeing with crashes using API levels below 24. So writing a disclaimer to let our clients know that this may cause issues. 
It seems as though the New Relic Android Agent is still using the putIfAbsent method on the Map class, which is only available on API 24 devices: https://developer.android.com/reference/java/util/Map#putIfAbsent(K,%20V)


rhetoric101:
@Sim-newrelic, I am working on this PR. By the way, have you signed our contributor license agreement yet?

rhetoric101 (approved):
@Sim-newrelic, I moved the disclaimer out of the front matter because the page wasn't building with that. I moved it into the body and put it in a callout.

PR #3250, opened 2021-07-23
added a note for existing Pixie users

mmfred:
This PR adds a note for existing Pixie users who might expect to use existing API key.


rhetoric101 (approved):
@mmfred, I built this locally and it looks good!

PR #3252, opened 2021-07-23
Added verbose logs instructions for helm installs

mmfred:
This PR addresses the request in Issue 3183

I added this section: http://localhost:8000/docs/integrations/kubernetes-integration/troubleshooting/get-logs-version/#verbose-helm

Question: Do we need steps like 5-9 from the previous section on verbose logs using a manifest file?

### DIFF
--- a/src/content/docs/integrations/kubernetes-integration/troubleshooting/get-logs-version.mdx
+++ b/src/content/docs/integrations/kubernetes-integration/troubleshooting/get-logs-version.mdx
@@ -12,15 +12,15 @@ To generate verbose logs and get version and configuration information, follow t
 <CollapserGroup>
   <Collapser
     id="verbose"
-    title="Get verbose logs"
+    title="Get verbose logs for installations using a manifest file"
   >
     For the Kubernetes integration, the infrastructure agent adds a log entry only in the event of an error. Most common errors are displayed in the standard (non-verbose) logs. If you are doing a more in-depth investigation on your own or with New Relic Support, you can enable verbose mode.
 
     <Callout variant="caution">
-      Verbose mode significantly increases the amount of infor sent to log files. Temporarily enable this mode only for troubleshooting purposes, and reset the log level when finished.
+      Verbose mode significantly increases the amount of info sent to log files. Temporarily enable this mode only for troubleshooting purposes, and reset the log level when finished.
     </Callout>
 
-    To get verbose logging details:
+    To get verbose logging details for an integration using a manifest file:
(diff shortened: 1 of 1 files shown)


src/content/docs/integrations/kubernetes-integration/troubleshooting/get-logs-version.mdx
```diff
@@ -48,7 +48,28 @@ To generate verbose logs and get version and configuration information, follow t
...
+    <Callout variant="caution">
+      Verbose mode significantly increases the amount of info sent to log files. Temporarily enable this mode only for troubleshooting purposes, and reset the log level when finished.
+    </Callout>
+  To get verbose logging details for an integration using Helm:
 
+    1. Enable verbose logging:
+    ```
+    helm upgrade -n <namespace> --reuse-values newrelic-bundle --set newrelic-infrastructure.verboseLog=true newrelic/nri-bundle
+    ```
+    2. Gather logs from the pod connecting to [Kubernetes state metrics](#logs-pod-kubestatemetrics) as described below.
```
roobre, line 67: Depending on the issue being debugged, logs for other pods might also be interesting.

greenszpila: I understand that we enable verbosity only for the infra pod and then we get logs from it. `kubectl logs newrelic-infra-xxxxx` Do we have to enable verbosity for the other pods too? The verbosity flag seems not available for the infra operator pod.

roobre: Enabling logs for the infra agent is usually enough. The key point is that the infra agent `DaemonSet` may comprise more than one `Pod`, especifically one per node that is in the cluster. Sometimes it is useful to have the logs for those other pods as well! :)

mmfred: I created this confusion by overly truncating the information I was attempting to follow, from the verbose logs using manifest file section. I've updated the steps. Please take a look and if it looks correct now, I'll merge.

mmfred: @roobre and @greenszpila - You can see the build here: https://pr-3252.d1vx4w5vgyla4s.amplifyapp.com/docs/integrations/kubernetes-integration/troubleshooting/get-logs-version/#verbose-helm

PR #3254, opened 2021-07-23
Removed 11 translated files

mmfred:
@jmiraNR - this is to remove the 11 translated files. I'll follow up in slack.


jmiraNR (dismissed):
Approved. All matches. Thank you.

Issue #2139, opened 2021-05-05
[Docs] Anchor parameter is stripped out of redirected links


aswanson-nr:
### Related Gatsby Issue

https://github.com/gatsbyjs/gatsby/discussions/27582

### Release Date

> We merged gatsbyjs/gatsby#32334 and released it in `gatsby@next` so you could try it out there. It'll then be released on July 20th in 3.10

### Description

The redirects generated by the `gatsby-plugin-meta-redirect` plugin do not remove the trailing slash. This causes any `#anchor` parameters to be stripped out of the link and most likely applies to query parameters as well.

### Steps to reproduce
1. Go to https://docs.newrelic.com/docs/insights/new-relic-insights/using-new-relic-query-language/nrql-reference#functions)[https://docs.newrelic.com/docs/insights/new-relic-insights/using-new-relic-query-language/nrql-reference#functions
2. Notice that the page redirects
3. Notice that the `#functions` was removed from the URL and the page did not scroll down to that section

### Expected 

When a user clicks on the link in the example above, they should land on the Functions section of the page.

<img width="956" alt="Screen Shot 2021-05-06 at 9 25 51 AM" src="https://user-images.githubusercontent.com/1395158/117324562-26b45780-ae4d-11eb-947c-1e8a8ee7364f.png">

### Other information
* @jerelmiller submitted an issue to the plugin we use to generate these redirects to add an option to remove trailing slashes https://github.com/nsresulta/gatsby-plugin-meta-redirect/issues/11

### Acceptance Criteria

- [x] Test using Gastby next release 
- [x] If tests pass, upgrade to `Gatsby 3.10` when available.
- [x] Upgrade Developer SIte
- [ ] Merge Developer `develop` branch in to `feature/o11y-packs`
- [ ] Upgrade OSS site
- [ ] Upgrade private docs
- [ ] Upgrade Theme

austin-schaefer:
Thanks @aswanson-nr . There are a lot of redirects on the site and a lot of anchor links, so a fix here will be very helpful!

jpvajda:
from @zstix  I believe those redirects come from https://www.gatsbyjs.com/docs/reference/config-files/actions/#createRedirect in Gatsby. Not sure if we have the option to pass along the extra URL params

rudouglas:
Notes:
Not sure if this is an issue anymore, i've having trouble reproducing it. 

[...]

PR #3261, opened 2021-07-26
chore: upgrade gatsby to 3.10.1

rudouglas:


aswanson-nr (commented):
I'm still seeing the redirect behavior where it removes the anchor parameter. I don't think that needs to hold up this PR, since it's still worth upgrading. But I do not think it closes out https://github.com/newrelic/docs-website/issues/2139

aswanson-nr:
> I'm still seeing the redirect behavior where it removes the anchor parameter. I don't think that needs to hold up this PR, since it's still worth upgrading. But I do not think it closes out #2139

Now that I think more about it, this could be because the theme has not been updated to the new version of Gatsby as well.

rudouglas:
omg the first time i tried this it didn't work and then I went back and tried it multiple times and it was working, and now it no no work 🤦

rudouglas:
I tried several things to try to fix this:
 - Removing `forceTrailingSlashes` from the config
 - Remove `appendTrailingSlash` from the `createRedirect` function in gatsy-node
But this still seems to force a trailing slash and remove the parameter so there must be something i'm missing in the cde that's doing this

PR #3278, opened 2021-07-27
fix: some small typo

monitaure:
This PR simply fix what seems to be a very small typo from _Tt does not ..._ to _It does not ..._ on the _Security for synthetic monitoring_ page.


x8a (approved):
Thanks for spotting it and fixing it! 👏🏻👏🏻👏🏻

PR #3285, opened 2021-07-27
Remove facet Limit part of NRQL query rules and limits

adiosspandit:
Removing the facet limit ", and can only facet on one attribute. Queries that facet on more than one attribute won't work." Multiple facets are supported

### Give us some context
* Existing Doc is out of date
* Got a note from Product Manager that multiple facets are supported for outlier conditions

No.


rhetoric101 (approved):
@adiosspandit, thanks for making this change!

PR #3286, opened 2021-07-27
Some important edits to 'Access grant'-related docs

zuluecho9:
Some org/user mgmt liaison work, as I realized that our docs were pretty bad for a few reasons. Regarding review: recommend a light one considering I have a lot of things I want to change in these docs and have upcoming feature changes that will impact these docs pretty soon. 

Edits include: 

Made the tutorial procedure much more clear and reflective of how the UI is currently working. 
Cleared up some not obvious stuff with more explanation in that tutorial and in the tips section. 
Improved our definition of our roles, and how groups are defined.


rhetoric101 (approved):
@zuluecho9, these changes look good. I'll send you a few private comments, but I'm going to publish these.

PR #3291, opened 2021-07-27
Remove very old compatibility callout

austin-schaefer:
Remove a tip callout referencing a 2017+ version of Kafka


paperclypse (approved):
Looks good!

PR #3293, opened 2021-07-27
Sort order: Metric API first in list of MELT APIs

austin-schaefer:
Tiny pedantic fix


paperclypse (approved):
Looks good!

zuluecho9:
@austin-schaefer note that the probable reason that was in that order was because there were other Metric API docs towards the bottom of that category underneath.

austin-schaefer:
@zuluecho9 Oops! Should we un-do it? To me it still seems weird to have it in ELTM order. 

Perhaps we should make Metric into a category?

zuluecho9:
i think the deeper issue is how all those things are thrown in one category. Let's ignore it for now, prob fine.

PR #3294, opened 2021-07-27
There was a missing nav category path

paperclypse:
The New Relic Alerts nav category was missing a path in the nav yaml.

I'm hoping this fixes it.


barbnewrelic (changes requested):
I checked localhost. If I understand what you are trying to accomplish, I'm seeing the Alerts taxonomy below the Alerts and Applied Intelligence main taxonomy. I also see that the "Rules, limits, and glossary" level does not have a link, see screenshot. Intentional?
alerts-ai-taxonomy

paperclypse:
Not intentional! Just an oversight. This index page was completely missing before. I'll add the Rules, limits, and glossary link.

paperclypse:
Rules, limits, and glossary has a link now.

PR #3296, opened 2021-07-28
Explaining ARM64 is supported

josemore:
### Give us some context

* Adding a note to make it explicit that the infrastructure agent image on dockerhub is available on AMD and ARM architectures. Feedback from customers and partners.


x8a (approved):

x8a:
Thanks for adding this!

PR #3312, opened 2021-07-28
Adding dummy doc to fix sidebar issue

zuluecho9:
We were having a problem with the 'Manage data' doc ref addition in the accounts doc section; it was essentially stealing the location from the real location, so that when you clicked it, instead of going to the place it actually was located (in 'Manage your data' docs section) it was remaining in the accounts docs section, and that wasn't good at all. This is due to some known issues in how the sidebar/nav files work and we may fix that but in meantime I fixed this with a more old-fashioned dummy doc fix. 

Please merge this if it looks fine.

### DIFF
--- a/src/content/docs/accounts/accounts-billing/general-account-settings/manage-data.mdx
+++ b/src/content/docs/accounts/accounts-billing/general-account-settings/manage-data.mdx
@@ -1,0 +1,5 @@
+---
+title: Manage your data
+---
+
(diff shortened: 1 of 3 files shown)


paperclypse (approved):
It's not ideal, because you see the dummy doc briefly before the redirect works. But if this is the only way that we can solve this, I suppose it's all right.

If you haven't, please file a GitHub issue about this so that we don't have to do more dummy docs like this.

PR #3314, opened 2021-07-28
.NET Agent 8.x to 9.x migration guide

nr-ahemsath:
### Give us some context

This is the migration guide mentioned in https://newrelic.atlassian.net/browse/DOC-7023
The draft in GDocs is here: https://docs.google.com/document/d/1BCCgm_OAtt-1Eea-_o-Nb7mhJBuXYoOSsSim_HhMlog/edit#

I am not sure if I got the table formatting correct.  I copied the syntax from an existing .NET agent document, but the Markdown preview in VSCode doesn't look like what I'm going for.  If there's a better/more standard way to format this kind of information for our users, let me know.  (Also, if you know of a tool which can show me what these .mdx files look like when rendered in the docs site, that would also be great.)

### DIFF
--- a/src/content/docs/agents/net-agent/getting-started/8x-to-9x-agent-migration-guide.mdx
+++ b/src/content/docs/agents/net-agent/getting-started/8x-to-9x-agent-migration-guide.mdx
@@ -1,38 +1,40 @@
 ---
-title: .NET Agent 8.x to 9.x Migration Guide
+title: .NET agent 8.x to 9.x migration guide
 tags:
   - Agents
   - NET agent
   - Migration guides
-translate:
-  - jp
-metaDescription: 'This guide helps users of the .NET agent migrate from 8.x versions to 9.x'
-redirects:
-  - /docs/agents/net-agent/net-agent-api/net-agent-api-guide
-  - /docs/agents/net-agent/net-agent-api/guide-using-net-agent-api
+metaDescription: 'This guide helps New Relic .NET agent users migrate from 8.x versions to 9.x'
 ---
 
-## Summary
-
-This guide outlines the major changes between the 8.x and 9.x versions of the .NET Agent, issues that may be encountered while upgrading, and how to migrate successfully to version 9.x.
+This guide outlines the major changes between the 8.x and 9.x versions of the .NET agent, issues that you may encounter while upgrading, and how to migrate successfully to version 9.x.
 
 Main changes include:
 
-* Distributed Tracing is enabled by default
+* Distributed tracing is enabled by default
 * Removal of deprecated public agent APIs
 * Removal of deprecated configurable agent settings
 
 
-## Distributed Tracing enabled by default
+## Distributed tracing enabled by default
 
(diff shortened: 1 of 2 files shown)


src/content/docs/agents/net-agent/getting-started/8x-to-9x-agent-migration-guide.mdx
```diff
@@ -0,0 +1,418 @@
...
+title: .NET Agent 8.x to 9.x Migration Guide
+tags:
+  - Agents
+  - NET agent
+  - Migration guides
+translate:
+  - jp
+metaDescription: 'This guide helps users of the .NET agent migrate from 8.x versions to 9.x'
+redirects:
+  - /docs/agents/net-agent/net-agent-api/net-agent-api-guide
```
nr-ahemsath, line 11: I'm not really sure what the `redirects` for this doc should be so I just left the ones from the doc I copied to get this one started.

barbnewrelic: No redirects needed since this is a new doc. Thanks for checking!

src/content/docs/agents/net-agent/getting-started/8x-to-9x-agent-migration-guide.mdx
```diff
@@ -1,38 +1,40 @@
...
+All deprecated API methods have replacement API methods with equivalent functionality.  If you are using the .NET agent API in your application code, we recommend that you update the package reference in your project to the latest 9.x version of the agent API assembly before upgrading the .NET agent to 9.x. That way, you will get compile-time errors if your code is using any of the API methods removed in 9.x.
+
+* **8.x versions:** If you continue to use an 8.x or earlier version of the API assembly, and your code uses any of the deprecated APIs listed below, you will not get compile-time errors.
```
nr-ahemsath, line 36: I think the way this has been reworded is misleading, because of the distinction between the version of the agent API assembly referenced in a customer's code, versus the version of the agent being used to instrument the application created by compilation of that code.  These two versions are independent; it is possible for a customer to use an 8.x version of the API and use a 9.x version of the agent to instrument the app.

Here's the complete "two-by-two matrix" of possible combinations and what the user would experience when using any of the deprecated API methods being removed in 9.x in that case:

8.x API assembly, 8.x agent: no problem!  These methods aren't removed yet in the 8.x agent.
8.x API assembly, 9.x agent: the customer's app code will still compile, but the API methods will no longer have their intended effect, and runtime errors will be seen in the agent log.
9.x API assembly, 8.x agent: compile-time errors will prevent the customer's app from being built, so the agent behavior doesn't really matter.
9.x API assembly, 9.x agent: ditto.

I don't think we necessarily need to make this matrix explicit in the docs.  We're mostly concerned about the second scenario, where if the customer doesn't update their API assembly version to 9.x in their code before upgrading their agent to 9.x, they probably won't notice that their API methods are no longer having any effect.  We want them to end up in either scenario 3 or 4 so they notice the problem (because their app won't compile).  Maybe we should do a tip callout on the "we recommend..." part.

nr-ahemsath (commented):
I have feedback on just one part, but it's a bit of a wall of text...sorry.

src/content/docs/agents/net-agent/getting-started/8x-to-9x-agent-migration-guide.mdx
barbnewrelic: Per slack, we've agreed on a color-coded solution. I'll make the change now. Thanks for pointing this out!

barbnewrelic (approved):
Reviewed with contributor and approved via slack. Thanks so much for all your hard work on this important migration guide!

PR #3328, opened 2021-07-29
Sign up bash - Alba

x8a:


mmfred (approved):

src/content/docs/agents/nodejs-agent/index.mdx
```diff
@@ -19,13 +19,26 @@ redirects:
...
+  <ButtonGroup>
+    <ButtonLink
+    role="button"
+    to="https://newrelic.com/signup"
+    variant="primary"
+    >
+    Sign up for free
+    </ButtonLink>
+  </ButtonGroup>
+  No credit card required. Already have an account? [Login](http://one.newrelic.com/).
```
rhetoric101, line 32: @x8a, for some reason this page generated this: Gatsby.js development 404 page
There's not a page or function yet at /docs/agents/nodejs-agent/index

I tried various tricks but I wasn't able to fix this quickly. 

By the way, I don't think we need text if we have a button, so I think you can remove that text following the button.

rhetoric101 (commented):
@x8a, these all looked good, except for the nodejs-agent index page. For some reason, this is generating a 404. Could you take a look? I wasn't able to find a quick fix.

PR #3331, opened 2021-07-29
translation workflow fixes, touchups, testing

nr-kkenney:
* fixes for minor bugs
* updated code docs where I noticed they were incorrect
* added code for setting record state to completed
* added code for the deletion of records
* added testing materials and documentation


elaguerta (approved):
Looks great!

scripts/actions/translation_workflow/models/job.js
```diff
@@ -33,7 +34,7 @@ module.exports = (sequelize, DataTypes) => {
       },
       batch_uid: {
         type: DataTypes.TEXT,
-        allowNull: false,
+        allowNull: true,
```
LizBaker, line 37: why allow null here? shouldn't there always be a batch id?

nr-kkenney: yea, that was my original thinking as well. but i made this change it just made it harder to work with in the code, and slightly awkward to look at. if allow null is false, that means we can only create a job record if it has a value for batch_uid.

previously, this would mean that a job would have to be added to the table after we create a batch with the vendor, but we actually need to create a job with the vendor first, then a batch, so that was why i switched it.

so now:
1. create a job with vendor, create a job in the table with no batch yet.
2. create a batch with vendor, then add the batch_uid to the job.

LizBaker: ohhhh that makes total sense 👍

LizBaker (approved):

PR #3337, opened 2021-07-29
Change from test.com to example.com

vosechu:
Using a domain for examples is tremendously dangerous. `example.com` was created specifically for this purpose and is routable across the internet, so we should use it.


zuluecho9 (approved):

zuluecho9:
I'll be checking with security if this indicates that we should sweep entire docs site for this.

zuluecho9:
Will be creating jira to fix this in few other places on docs site.

PR #3343, opened 2021-07-30
Add buffer tune-up section to fluentd docs

danybmx:
### Give us some context

This PR adds more information about how to configure fluentd when forwarding logs to NR and also show the customer what is the default buffer configuration when using the newrelic fluentd output plugin.

No


x8a (approved):
Thanks for the contribution! 🙌🏻

PR #3346, opened 2021-07-30
fix: formatting of commands

keegoid-nr:
- line breaks for `docker run` commands needs `\`
- K8s commands don't typically use line breaks


bradleycamacho:
Thanks for the edit @keegoid-nr! Merging in now.

PR #3362, opened 2021-08-02
infra agent 1.19.0 release notes: recommends explanation

varas:
Infrastructure agent release notes update to add explanation on "recommends" dependencies.


urbiz-nr (approved):
Thanks for the updates @varas!

Issue #2655, opened 2021-06-10
☂️ Refactor translation workflows to handle status per-page, not per-job


roadlittledawn:
## Summary

We want to update our translation workflows so we only upload a file when it isn't already being actively translated. This ensures that we get a fully translated file back when the first job finishes.

In order to support this work, we need to update how we store the status of translations on a _per-file_ basis (rather than per-batch-job).

### Update the data in our queue
- For each active job, store what files are in it
- Do not submit same file in multiple jobs at once
- When adding file to queue, first check if it's in an active job
  - If yes, add to "Next Up" queue 

### Update how we submit files to be translated
- Only upload file for a job if it's not included in an ongoing job
- When a job finishes, check for any files in that job that might be queued up for another translation and upload those for new job at some point

## Acceptance criteria
- [x] #3074
- [x] #3075
- [x] #3076
- [x] #3077
- [x] #3078
- [x] #3080
- [x] Document the workflow system!
- [x] Close out all bugs this work resolves

### Resolves

https://github.com/newrelic/docs-website/issues/2603

## Helpful links
https://github.com/newrelic/docs-website/issues/2597

zstix:
I'm going to use this issue as an umbrella issue for this work. In order to do so, I'll need to rewrite the description a little (FYI)

PR #3369, opened 2021-08-02
minor fixes for translation

nr-kkenney:
These changes are some minor adjustments which we identified after we executed the send workflow. 

In particular:
* remove usage of `ERRORED` status. for this first run, we saw many files encounter this status due to failed uploads that were retryable. its likely we will encounter this scenario again, so rather than manually resetting state to PENDING, we would rather it just stay that way. in the future, we can figure out how to better incorporate the ERRORED state.
* upload files serially. we are hitting an api limit doing concurrent file uploads. the easiest path forward for now is to just upload them serially. this does decrease performance, but that is something we have plenty to give, especially if it reduces manual toil.
* provide the translation.slug if a file fails to upload just for context. this is because we encountered a serialization error w/o knowing which file it failed on.

Relates to: #2655

### DIFF
--- a/scripts/actions/send-and-update-translation-queue.js
+++ b/scripts/actions/send-and-update-translation-queue.js
@@ -158,23 +158,22 @@ const uploadFiles = async (batches, translationsPerLocale, accessToken) => {
   for (const batch of batches) {
     let successCount = 0;
 
-    await Promise.all(
-      translationsPerLocale[batch.locale].map(async (translation) => {
-        const fileUploadResponse = await uploadFile(
-          batch.locale,
-          batch.batchUid,
-          accessToken
-        )(translation);
+    const translations = translationsPerLocale[batch.locale];
+    for (const translation of translations) {
+      const fileUploadResponse = await uploadFile(
+        batch.locale,
+        batch.batchUid,
+        accessToken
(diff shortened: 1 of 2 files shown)


scripts/actions/send-and-update-translation-queue.js
```diff
@@ -158,27 +158,22 @@ const uploadFiles = async (batches, translationsPerLocale, accessToken) => {
...
+        accessToken
+      )(translation);
+
+      if (fileUploadResponse.code === 'ACCEPTED') {
+        await Database.updateTranslation(translation.id, {
+          status: 'IN_PROGRESS',
+        });
+        await Database.addTranslationsJobsRecord(translation.id, batch.jobId);
+        successCount += 1;
+      }
```
nr-kkenney, line 175: we probably want to wrap the body in a try catch so that each translation gets a chance to be uploaded, even if a previous translation encounters an error.

aswanson-nr: it looks like `uploadFile` will handle everything except for Fetch throwing on a network error. A Try/Catch here sounds perfect!

aswanson-nr (approved):

PR #3373, opened 2021-08-02
Add more API rate limiting details

dbetke:
Links to information on how to view your limits from the NRQL rate limits page.
Updates the callout in the Query system limits page to reflect that resource metrics now also include API query rate limits.


paperclypse (approved):
Thanks for helping out, @dbetke!

PR #3374, opened 2021-08-02
Add Infinite Tracing gem requirement for Sinatra

linbaker:
### Give us some context

In order to use Infinite Tracing with Sinatra apps (that are not also using Rails) the customer would need to manually call the New Relic gem AND the Infinite Tracing gem, or use `Bundler.require`.
The wording in my edit may be a little clunky now and may need rephrasing. The basics are, for customers with Sinatra apps:
* For Infinite Tracing, they  _must_ either call `Bundler.require` OR `require 'newrelic_rpm'` and `require 'newrelic/infinite_tracing'`
* If Infinite Tracing is not configured and the agent is greater than/equal to v6.11 (released June 2020), then including `require 'newrelic/infinite_tracing'` will do nothing so it won't hurt to include it even if the customer has no intention to use ∞T
* If the customer attempts to add `require 'newrelic/infinite_tracing'` with an agent version less than v6.11, it will cause an error when it looks for a gem that doesn't exist

Since this is an install doc, I assume we can exclude any mention of version and just assume the customer is installing a version of the agent less than a year old

Confirmed this edit with the Ruby engineering team in https://newrelic.atlassian.net/browse/GTSE-10264


paperclypse (approved):
I made some slight changes for clarity, @linbaker. If you feel like I haven't quite managed it, please do let me know. :) Thanks for helping out!

linbaker:
Much more clear, thank you @paperclypse!

PR #3375, opened 2021-08-02
Update PR and issue templates for week of Aug 9

austin-schaefer:
### Give us some context

* This PR updates issue and PR templates for the week of August 9
* Should not be merged until August 6
* @rhetoric101 is on point to review and merge


rhetoric101 (approved):
@austin-schaefer, we are not adding an announcement link for these template changes. Thanks for updating these templates!

PR #3381, opened 2021-08-03
Added note regarding headers for license keys

urbiz-nr:
As reported in #documentation.


zuluecho9:
@zuluecho9 is looking into this. This is a larger problem with the fact that apparently we didn't document that one could use license key for Metric API.

zuluecho9:
Gonna close this as I have a bunch of related work on this coming up shortly for release. Long story short: we had a problem in docs for a while apparently in that we didn't make it clear the license keys could be used for MELT API ingest and that doing that required use of the dedicated header. Upcoming docs work should clear all this up in another 1.5 weeks.

PR #3382, opened 2021-08-03
Increase max widgets and height limits

hredondo:
### What

We increased the max number of widgets per dashboard page and the max height of a widget.


urbiz-nr (approved):
Thanks for the update, @hredondo!

PR #3383, opened 2021-08-03
Integration disabled given Azure has retired the service

josemore:
### Give us some context

* Azure communicated the [service is no longer available](https://azure.microsoft.com/en-us/updates/azure-service-fabric-mesh-preview-retirement/).
* We had disabled the integration and we'll proceed to remove this documentation by end of the quarter.  


urbiz-nr (approved):

PR #3390, opened 2021-08-03
Added a reboot script

x8a:
Reboot script to easily clean out the cache when running into problems with local builds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,6 +133,7 @@
   "scripts": {
     "build": "gatsby build",
     "develop": "gatsby develop",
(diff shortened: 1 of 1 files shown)


paperclypse (changes requested):
I ran 'yarn reboot'. It successfully removed the node_modules, but then stalled out with an error at that point, because Gatsby is one of the node_modules that was removed. This means that it can't run `yarn clean`.

I think if you swap `yarn` and `yarn clean` it should work though:

`rm -rf node_modules && yarn && yarn clean && yarn start`

paperclypse (approved):
Yay! I love this.

PR #3406, opened 2021-08-04
Added yarn reboot to the style guide

x8a:
I added the `yarn reboot` script in our troubleshooting guide


paperclypse (approved):
I think it might still be useful to spell out the full set of commands related to kknuke/yarn reboot.

But I'm not super attached to that. :)

x8a:
> I think it might still be useful to spell out the full set of commands related to kknuke/yarn reboot.
> 
> But I'm not super attached to that. :)

They're still in the `package.json` file. I think I'll leave it as is.

PR #3424, opened 2021-08-05
adds GUID for JetBrains' dotCover profiler

bcanzanella:
Adds to documentation an issue I ran into trying to setup a New Relic agent on Windows. There was already an environment variable set from an old installation of JetBrains dotCover and it was preventing the New Relic agent from working -- this is its GUID.


mmfred:
Thanks for the added info @bcanzanella!

PR #3432, opened 2021-08-05
Clarify in the blog post that capture() is not GA

LukeCarrier:
### Give us some context

- The NRQL `capture()` function is not generally available despite the beta programme closing.
- The documentation is actively misleading, making no reference to the beta programme and failing to acknowledge that the beta programme has closed without the feature becoming generally available.
- This is frustrating for customers, so remove it.

Changes to docs only.


mmfred:
Thanks for your input @LukeCarrier - We have asked for a technical review from the team that handles NRQL. We'll get back to you with that input, most likely after a company holiday next week.

mmfred:
@LukeCarrier - I got confirmation that `capture()` did in fact GA. So I will close this because it removes documentation that we need. We'll follow up by removing any mentions of it as a beta product.

LukeCarrier:
@mmfred are you sure? This isn't what I was told [here](https://discuss.newrelic.com/t/capture-keyword-not-available/145701/3), and it doesn't seem to work as documented.

derekroller:
Howdy @LukeCarrier 

Regex capture has been fully GA'd and is available for all customers.  When the twitch stream you referenced was created, it was still beta only, but has seen been fully released.

PR #3450, opened 2021-08-06
Update secrets-management.mdx

ryanv94:
Attempting to make it clear that JSON is supported , and that the example provided is using obfuscated JSON. 

I struggled with this doc and eventually figured it out - I want to make it clear for other users.


urbiz-nr (dismissed):
Looks good, @ryanv94 - thanks for the update!

urbiz-nr (approved):

PR #3453, opened 2021-08-06
fix: release workflow breaking

rudouglas:
Escaped quotation characters
Run the github release command inside the script to eliminate issues with strings between Python and Bash


scripts/actions/release-notes.py
```diff
@@ -13,11 +13,14 @@
 result = "## :rocket: What's new?\n\n\n"
 
 # Get the Docs repo
-repo = github.get_repo("newrelic/docs-website")
+repo = github.get_repo("rudouglas/docs-website")
```
aswanson-nr, line 16: Reminder to update this before merging!

aswanson-nr (changes requested):
Looks good to me! Just saw one field that needs to be updated

scripts/actions/release-notes.py
```diff
@@ -32,7 +35,11 @@
   except AttributeError:
     pass
 
+# Limit the characters in the notes to 25000 as per api restrictions
+result = result[0:24999]
+
+# Encode the string to escape characters
+result = result.replace('"','\\"')
```
nr-kkenney, line 42: i think you may wanna swap these lines. if we replace one character with two, that could send us over the limit again.

scripts/actions/release-notes.py
rudouglas: 🤦

scripts/actions/release-notes.py
rudouglas: Ah yes i was trying to be clever and not have to replace characters if they were over a certain limit already 🥇

nr-kkenney (approved):

PR #3456, opened 2021-08-06
Adding note about optional dependency on Linux

josemore:
### Give us some context

* Starting with agent 1.19.0, the td-agent-bit dependency might not be installed by default on customer environments. Adding explanation and workaround until we determine if this will be changed. 


austin-schaefer:
Hi @josemore , going through our board looking for stale PRs and looked like this one is inactive. Closing this out but feel free to re-open if there's still active work to do!

PR #3482, opened 2021-08-12
Call out .NET Agent versions that support expected errors configuration settings.

vuqtran88:
Call out .NET Agent versions that support expected errors configuration settings.


urbiz-nr:
Thanks for adding this info, @vuqtran88!

urbiz-nr (approved):

PR #3494, opened 2021-08-16
Logging 5518 add api key query parameters log api

danybmx:
### Give us some context

This PR just update the log api documentation to indicate a new query parameter that the customer could use to authentication (this means, send their Api-Key in a query parameter)

No


paperclypse (approved):
Looks good!

PR #3497, opened 2021-08-16
update lookout example to use SQL style ‘yyyy-mm-dd HH:MM:ss’ timestamps

sdedovic:
### Give us some context

  * updates the docs to use SQL style timestamps. This is consistent with the current UI (when you enter an invalid timestamp format).
  * for example take a look at this: https://discuss.newrelic.com/t/absolute-date-range-selection/389/2


zuluecho9 (approved):

zuluecho9:
For a little more context on this, from PR submitter: 

The Lookout UI
a) only supports a subset of all NRQL, and
b) requires a specific fragment of a query
The editor, for which I am updating docs, is not meant for arbitrary queries a la the data explorer. This change to the docs is to reflect changes to the Lookout backend (only), where we removed support for a variety of timestamp formats, including this one we are discussing. We discussed this and are not worried about affecting existing users. I did some digging and saw many places where the timestamp is "yyyy-MM-dd hh:mm:ssZ" format so that is supported.

PR #3504, opened 2021-08-16
LOGGING-5959: Fix permissions for drop filter rules

bmcfeely:
https://newrelic.atlassian.net/browse/LOGGING-5959

### Give us some context

Some smart person discovered that we had a copy/paste error in our docs for drop filters, and had the incorrect permission listed.


paperclypse (approved):
Looks good!

PR #3517, opened 2021-08-17
Infrastructure agent - Update docker discovery snippet

brushknight:
Usually we will have ${discovery.private.ip} and only if public ip attached to container we could use ${discovery.ip}.

### Give us some context

* Update documentation


dbarnesbrownNR (approved):
Looks good, thank you!

PR #3519, opened 2021-08-17
Update python-graphQL.mdx

rudouglas:


roadlittledawn (approved):
built successfully on GC. approving and merging so tech docs can get the release PR moving

PR #3521, opened 2021-08-17
Made changes based on SOC2 audit results.

rhetoric101:
AWS Metric Streams and Errors Inbox are now SOC2 compliant.


rhetoric101:
The build test failed, so I just manually triggered a cache reset.

rhetoric101:
OK, I see that this check failure is due to a larger issue that came from Develop. DevEn is looking into a fix.

paperclypse (approved):
Looks good in my local build! I think you're good to merge

Issue #3458, opened 2021-08-06
Issue: Change your Apdex settings


adiosspandit:
* The Apdex Setting in UI is not always visibile
* Documenting under what conditions the setting is visible and editable will help
* If there is a particular role needed to be able to edit the setting that should also be added to this doc
*

### Anything else you'd like to share?
I am not sure if agent has to be actively reporting for that setting to be visible. 
I am an Admin Full User on an account but unable to see the setting.
Another user is sometimes able to see the setting but not able to edit it.
I am struggling to find the exact expected behavior for visibility and editability of the apdex setting

rhetoric101:
@adiosspandit, thanks for your questions about APDEX. I'm checking with subject matter experts to see if there is anything we can add to the doc to clarify when APDEX is visible.

rhetoric101:
@adiosspandit, I just confirmed that the Apdex settings are not visible for apps that are not reporting. I am going to add a note at the top of the Apdex settings document explaining that (https://docs.newrelic.com/docs/apm/new-relic-apm/apdex/change-your-apdex-settings/).

rhetoric101:
This is resolved in PR #3529.

PR #3529, opened 2021-08-17
Clarified when Apdex settings are visible

rhetoric101:
This is a PR to clarify that Apdex threshold settings are only visible when an app is actively reporting data.

Closes issue #3458.


paperclypse (approved):
Looks good!

PR #3536, opened 2021-08-18
Remove watermarks from release notes

austin-schaefer:
As the team has talked through watermarks, we're trying to move away from watermarks across the site. The release notes all include (pretty redundant) watermarks that indicate EoL status. Since this status is also indicated clearly in a callout, the watermark here is redundant.

Reviewer notes:

- I did this via find and replace
- I've tested this locally on a dozen or so pages and they all worked fine, but I'd like to poke at it more in the cloud preview where it's more responsive
- To test, just click around in some of the oldest agent release notes


paperclypse (approved):
I love how easy find/replace makes to take care of something like this. Looks good!

PR #3540, opened 2021-08-18
Remove banner from landing page

paperclypse:
Now that the event has started, we don't need the banner anymore.


urbiz-nr (approved):
Tried locally, works well.

PR #3556, opened 2021-08-19
Add in new version of the ios app

jvanmetre:
### Give us some context

* Adds in release note for the new version of the iOS Mobile app


urbiz-nr (approved):
Thanks for adding this release note, @jvanmetre!

PR #3557, opened 2021-08-19
Change values to match screenshots

HenryTech:
### Give us some context

### DIFF
--- a/src/content/docs/alerts-applied-intelligence/new-relic-alerts/get-started/your-first-nrql-condition.mdx
+++ b/src/content/docs/alerts-applied-intelligence/new-relic-alerts/get-started/your-first-nrql-condition.mdx
@@ -55,7 +55,7 @@ Give your NRQL condition a name that's meaningful to you. Your query's visualiza
 
 Set critical and warning thresholds to determine when you’re notified about your environment's performance. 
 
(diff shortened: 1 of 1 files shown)


dbarnesbrownNR:
Hi @HenryTech, it looks like the values are percent expressed as a decimal and not a percent, and 70% CPU usage makes more sense in the context of this example than 0.7% as a warning threshold. Thinking we should either skip this update or change to state explicitly that the example expresses percent as a decimal value in our UI. What are your thoughts?

HenryTech:
The example won't show the critical and warning threshold lines if the
values are not ridiculously low. Shawn K and I discussed this and figured
it's going to be easier to just stick with the example the way it is now,
and just use a very small threshold. Presumably, users will figure out that
they don't need to set their threshold this low.

To be clear, the percentage values are _not_ expressed as decimals in this
event type. CPU Percent of 70 looks like "70," not "0.7." However, if we
set the threshold to a reasonable level of 90%, the threshold markers do
not show up in the graph unless the metrics are up around that level
already.

dbarnesbrownNR (approved):
Approving changes as discussed in our sync, creating related issue to update screenshots and corresponding text values that correspond to more "real world" examples of CPU usage alerting.

PR #3568, opened 2021-08-19
DOC-7154 Revising the alerts taxonomy

paperclypse:
This PR is for generating the Gatsby preview for a proof of concept (which doesn't happen with a draft PR).

The taxonomy revision looks good, but I still need to add redirects and move all the files around.


paperclypse:
For some reason, these docs won't load in the Gatsby preview or the local build:
Create, edit, or find alert policies
Create conditions
Intro to NerdGraph API and Alerts
Use the API (section)
Baseline alert conditions

I imagine I made some pathing mistakes with these.

paperclypse:
Ready for a peer review! This is DOC-7154

mmfred:
@paperclypse This is working as expected. Just one thing: this link in the **Learn about Alerts** area has kind of weird behavior in the NAV (the page opens fine): Technical rules and limits 
It works, the Nav just collapses and the reader loses their place.

mmfred (approved):

paperclypse:
Thanks @mmfred!

PR #3572, opened 2021-08-20
Add instruction to restart APM service

linbaker:
Tested adding tags to APM agent through UI and NerdGraph, both did not take effect until I restarted my service and then only applied to new, incoming data.
I did check with Browser support team, this is not needed for Browser: https://newrelic.slack.com/archives/C7GUYJZ5E/p1629406376045100
Infra team did not get back to me so only including APM for now: https://newrelic.slack.com/archives/C7HGFBZMH/p1629406484141500


urbiz-nr (approved):

urbiz-nr:
Thanks for adding this clarification, @linbaker!

PR #3588, opened 2021-08-20
Updated NRQL condition topic

rhetoric101:
I made the following changes:

* In step 1, I rearranged the sentences, putting the overview sentence at the top and added sentence explaining that .7% is not a realistic value--just an example.
* Renamed step 2.
* Moved graph image out of step 2 into step 3 because it was out of sequence.
* Moved threshold definitions in step 3 to beginning and put them in a list and added image from step 2.

Please confirm that the percentages in the text now match the percentages in the images.


rhetoric101:
@paperclypse, would you have time to do a quick check of this update? I got SME feedback, so this should be pretty simple.

paperclypse:
Let's merge this one in after we merge in the related #3568

paperclypse (approved):
Looks good. I renamed this doc's filename and made a couple of other changes in #3568. It's possible you may get a merge conflict. If so, I'm happy to help out with that.

PR #3596, opened 2021-08-23
wording(agent): change brew uninstall package

bbbenja:
use `newrelic-infra-agent` like above


urbiz-nr (approved):
Thanks for the update, @bbbenja!

PR #3605, opened 2021-08-23
Fixed a broken URL

paperclypse:
It was using the wrong ID for the section it was linking to. I updated the URL to use the correct ID for the collapser.


paperclypse:
This was a Doc hero request.

zuluecho9 (approved):

PR #3621, opened 2021-08-24
Update spm-installation.mdx

smalsam-newr:
Updated post-build script to support Swift Package Manager release. When using SPM, frameworks are located in a different place than they were when using Cocoapods or manual installation. The script snippet now looks in that original place first. If it does not find the Python script it is looking for, it then checks the location where SPM would have the framework, before declaring an error.


rhetoric101 (approved):
@smalsam-newr, thanks for updating this document! I'll push it up to the site.

PR #3676, opened 2021-08-26
Fix possible bullet nesting issue in network docs

austin-schaefer:
Noticed what seems to be too-nested bullets in "Set up network flow data monitoring"

image


x8a:
hey @austin-schaefer! This is intended, as each device belongs to a different type of data.

zuluecho9:
@austin-schaefer @x8a I think we should discard/ignore this PR as I also noticed that this was building off old develop and not the new one (I made edit on Friday so this doc now has 'license key' not 'insert key', a recent change. @x8a, think Austin was drawing attention to also the second bulletpoint being indented, which I think I may have accidentally created. So maybe we just discard/ignore this PR and make a new one to fix that second bulletpoint indent?

PR #3696, opened 2021-08-27
Change to avoid conflation of span.kind with the New Relic concept of process boundaries.

rhetoric101:
This request can in from an SME.

### DIFF
--- a/src/data-dictionary/events/Span/span.kind.md
+++ b/src/data-dictionary/events/Span/span.kind.md
@@ -5,4 +5,4 @@ events:
   - Span
 ---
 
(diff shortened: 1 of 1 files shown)


barbnewrelic (approved):
I checked this in Gatsby build but not my localhost. Looks fine AFAIK. Sent you some screenshots 1:1 so you can see what I saw. Go ahead and merge if you're ready!

zuluecho9:
if it matters @rhetoric101: in future, I think a simple edit of a sentence like this, you can just push out yourself without a check, as long as it throws no required-check errors.

PR #3702, opened 2021-08-29
Lic dotnet

barbnewrelic:
This adds a single consolidated .NET logs in context doc to the taxonomy. (Its standalone taxonomy will be removed in a separate PR.) TW reviewer: Please check that this doc builds correctly. You do not need to focus on the text itself, as that will be handled in a separate Jira. Thanks!


barbnewrelic (commented):
Build works! I'm going to proceed with removing the Java and .NET leftover taxonomy so that doesn't confuse people.

PR #3711, opened 2021-08-30
Link fix

helenapm:
Fixing the link to the section that explains how to create a workload, which lives in a different page.


mmfred (approved):

mmfred:
@helenapm - thanks for the fix!

PR #3722, opened 2021-08-30
Lic java net cats

barbnewrelic:
Deleted and redirected the standalone logs in context categories for Java and .NET. Their docs have already been moved into the standalone docs https://docs.newrelic.com/docs/logs/enable-log-management-new-relic/configure-logs-context/java-configure-logs-context-all/ and https://docs.newrelic.com/docs/logs/enable-log-management-new-relic/configure-logs-context/net-configure-logs-context-all/


barbnewrelic (commented):
This PR is only to rebuild the taxonomy and it works fine, both on localhost and Gatsby, so I'm going to take the liberty of approving this so it's including in morning merge. More taxonomy work to come in a separate Jira.

PR #3726, opened 2021-08-31
Update auto-telemetry-pixie-data-model.mdx

rudouglas:
The redirects should have a leading slash, this was breaking one of our workflows but I also added a check to our workflow to catch this in future


urbiz-nr (approved):
Thanks for catching this @rudouglas!

PR #3727, opened 2021-08-31
chore: fix related resources leading slash

rudouglas:


scripts/actions/utils/related-resources-helpers.js
```diff
@@ -31,6 +35,7 @@ exports.getExcludedUrls = (frontmatter, siteUrl) => {
   return resources
     .map((resource) => resource.url)
     .concat(redirects)
+    .map((url) => appendLeadingSlash(url))
```
nr-kkenney, line 38: channeling my inner zack, i think you can do [point free](https://mostly-adequate.gitbook.io/mostly-adequate-guide/ch05#pointfree) syntax here.

this would be then:
```js
.map(appendLeadingSlash)
```

nr-kkenney (dismissed):

scripts/actions/utils/related-resources-helpers.js
rudouglas: oh hello

PR #3729, opened 2021-08-31
Fixing bad category URL

zuluecho9:
Someone changed something in folders so the category URL was no longer working. So this will remove path and make that category header not clickable and just an expander/collapser. Which is okay for now, better than the broken category link. We say now in style guide that it's acceptable to have a nonclickable category for categories that are just assemblies of docs that "live" in other locations. 

Also fixed a bad mobile-monitoring URL.


dbarnesbrownNR (approved):
Looks fine to me!

PR #3743, opened 2021-09-01
Fix the hostname attribute for win services

alvarocabanas:
- Fix the hostname attribute for win services.


bradleycamacho (approved):

bradleycamacho:
Thanks for the fix @alvarocabanas !

PR #3763, opened 2021-09-02
DOC-7263 - Implementing CrazyEgg

x8a:
Added crazyegg script to SEO component for the following pages:
- https://docs.newrelic.com
- https://docs.newrelic.com/docs/accounts/accounts-billing/account-setup/create-your-new-relic-account/

### DIFF
--- a/src/components/SEO.js
+++ b/src/components/SEO.js
@@ -11,11 +11,12 @@ const METADATA = [
 ];
 
 const crazyEgg = (location) => {
-  const homepage = 'https://docs.newrelic.com';
-  const signup =
-    'https://docs.newrelic.com/docs/accounts/accounts-billing/account-setup/create-your-new-relic-account';
+  const { pathname } = location;
+  
+  const homepage = '/';
+  const signup = '/docs/accounts/accounts-billing/account-setup/create-your-new-relic-account/';
 
-  if (location === homepage || signup) {
+  if (pathname === homepage || pathname === signup) {
     return (
       <script
         type="text/javascript"
--- a/src/components/SEO.js
+++ b/src/components/SEO.js
@@ -10,6 +10,22 @@ const METADATA = [
   },
 ];
 
+const crazyEgg = (location) => {
+  const homepage = 'https://docs.newrelic.com';
+  const signup =
+    'https://docs.newrelic.com/docs/accounts/accounts-billing/account-setup/create-your-new-relic-account';
+
+  if (location === homepage || signup) {
+    return (
+      <script
+        type="text/javascript"
+        src="//script.crazyegg.com/pages/scripts/0045/9836.js"
+        async="async"
+      />
+    );
+  }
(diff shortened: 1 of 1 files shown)


src/components/SEO.js
```diff
@@ -10,6 +10,22 @@ const METADATA = [
   },
 ];
 
+const crazyEgg = (location) => {
+  const homepage = 'https://docs.newrelic.com';
+  const signup =
+    'https://docs.newrelic.com/docs/accounts/accounts-billing/account-setup/create-your-new-relic-account';
+
+  if (location === homepage || signup) {
```
zstix, line 18: `location` is a little different from [`window.location`](https://developer.mozilla.org/en-US/docs/Web/API/Window/location) - it's not a plain string, but rather an object (you can see that in the prop-types below). Our [documentation](https://github.com/newrelic/gatsby-theme-newrelic/tree/develop/packages/gatsby-theme-newrelic#seo) on the `SEO` component doesn't specifiy the _exact_ shape of that object, but it does say that it should have a `pathname` field that we can use for this purpose.

Additionally, javascriprt will treat the `|| signup` as a separate condition on it's own and not attempt to verify if `signup` is equal to `location`. You need to supply `location ===` for both statements.

Here's the adjustments to the code that should handle both those things:

```suggestion
  const { pathname } = location;
  
  const homepage = '/';
  const signup = '/docs/accounts/accounts-billing/account-setup/create-your-new-relic-account/';

  if (pathname === homepage || pathname === signup) {
```

x8a: Thank you so much, @zstix! I wasn't sure about the location prop and whether it should be a relative path or an external link. I'll go ahead and merge your suggestion 👏🏻👏🏻👏🏻

jpvajda (approved):
I'm Ok that we merge this script  in, but I want to  run a lighthouse test after before / after deployment to understand impacts this script has on our site performance, for every script we add we should be doing that as a process @x8a  

I'll attach a lighthouse report to this PR and if you can plan on running this after deployment and reporting back the performance metrics that would be cool. 

Install the CLI here:
https://github.com/GoogleChrome/lighthouse

Run this command:
`lighthouse https://docs.newrelic.com --preset=desktop --view`

OR:
you can use this site https://web.dev/measure/

jpvajda:
Light house stats for the site without this change:

Screen Shot 2021-09-02 at 12 34 28 PM

Screen Shot 2021-09-02 at 12 33 47 PM

x8a:
@jpvajda, here's the report after deployment. Let me know if you want me to send you the full HTML report.
Screenshot 2021-09-06 at 10 32 40

jpvajda:
Interesting, it went up... I'm thinking the lighthouse testing is sometimes a bit subjective depending on a lot of factors. But Given this is in a range of `+ / - 10 pts`  I think it's good to know

Issue #3706, opened 2021-08-30
404 missing content / bad link for 'Mobile Apps' from the 'Instrument Everything' page


liamhurrell:
### Tell us what you need

* Bad link for 'Mobile Apps' from this page: https://docs.newrelic.com/docs/full-stack-observability/instrument-everything/
* This is the missing resource: https://docs.newrelic.com/docs/mobile/table-of-contents/

### Anything else you'd like to share?

Add other context like screenshots, links to other docs, and information about
your environment (operating system, application framework, etc.).

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/mobile/table-of-contents/
* **OS**: Mac OS 10.15.7
* **Browser**: Chrome 92.0.4515.159
* **Device**: Unknown

urbiz-nr:
Thanks for reporting this, @liamhurrell - we'll look into it.

urbiz-nr:
Closed #3706 via #3784.

PR #3797, opened 2021-09-06
Added mobile TOC link

urbiz-nr:
The mobile TOC link is broken (as reported in #3706), so I'm adding the redirect in the landing page.


urbiz-nr:
I think adding the redirect on a landing page is affecting another file, so I'm trying something else.

PR #3798, opened 2021-09-06
Fix wrong data in Varnish doc

alvarocabanas:
- Fix wrong data in Varnish doc


x8a (approved):
Thanks for adding this back, @alvarocabanas!

PR #3800, opened 2021-09-06
chore: improve mysql config and content

carlossscastro:
Redesigning the MySQL integration configuration section.
Add example configurations


x8a (dismissed):
Thanks for the changes!

PR #3802, opened 2021-09-06
Adding `yarn.lock` back to the repository

zstix:
It looks like the lockfile was removed in the following commit:

ae85abd5aca33c256062cd9048b002525ee7ab8b

We should _never_ be removing the lockfile. This file serves a very
crucial role in ensuring that the site functions in a consistent manner
on different machines. It's possible that this is why our recent pull
requests have stopped working.

For more information about lockfiles, see this article:

https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all/


x8a:
@thezackm, the Gatsby build has failed with the following error:
```
Could not parse import/exports with acorn: SyntaxError: Unexpected token/usr/src/app/www/src/content/docs/integrations/mlops-integrations/algorithmia-mlops-integration.mdx: Could not parse import/exports with acorn: SyntaxError: Unexpected token

For more details see https://gatsby.dev/issue-how-to
```

Not really sure what's wrong with the file.

nr-kkenney:
i saw this as well as was surprised it was both removed, and merged into develop.

now that im thinking about it, while this isn't necessarily something we should test for, our previous github builds may have caught such a change if they used yarn to install dependencies 🙈

LizBaker (approved):

PR #3805, opened 2021-09-07
fix: update language on step 4

thezackm:
### Give us some context

original language indicated we were polling for flow data; we do not poll, we receive flow data

no


x8a (approved):
Thanks for spotting this and making the edit! 💃🏻

PR #3818, opened 2021-09-08
Removing old beta watermark for Trusted Advisor integration

josemore:
### Give us some context

* We found that we never removed the beta watermark for this integration. 


x8a (approved):
Thanks!

PR #3846, opened 2021-09-09
feat: adding new dashboard to npm visualization instructions

thezackm:
### Give us some context

 - adding instructions on new `Data Ingest and Cardinality` dashboard
 - updating screenshots for dashboards so they all have the same look
 - re-ordered "start exploring..." section to put new dashboards ahead of lookout so the doc flows better for the reader
 - updated some verbiage to focus more on dashboards as the driving concept and not custom visualizations as a feature of dashboards

Note: The `Add the Network dashboards to your account` section needs some help; the `<figcaption>` tag in the ordered list seems to be breaking and I cannot see how to fix it. 

CC: @x8a 

no


x8a (dismissed):
Thanks for the edits and adding those screenshots, @thezackm!

x8a (approved):

PR #3855, opened 2021-09-10
Postgresql v4 config docs

alvarocabanas:
Redesigning the Postgresql integration configuration section.
Add example configurations


x8a (approved):
Thanks for the edits! 👏🏻

PR #3856, opened 2021-09-10
chore: improve couchbase config and content

carlossscastro:
Redesigning the Couchbase integration configuration section.
Add example configurations


x8a (approved):
Thanks!

PR #3860, opened 2021-09-10
Memcached config in V4 format

alvarocabanas:
Redesigning the Postgresql integration configuration section.
Add example configurations


paperclypse (approved):
Looks good! I've made some minor edits.

PR #3862, opened 2021-09-10
chore: improve HAProxy config and content

carlossscastro:
Redesigning the HAProxy integration configuration section.
Add example configurations


paperclypse (approved):
Looks good!

PR #3877, opened 2021-09-13
fix typo in oracle integration doc

carlossscastro:
Small fix in Oracle DB doc that still had a reference to Mysql by mistake


urbiz-nr (approved):
Thanks for the fix, @carlossscastro!

PR #3884, opened 2021-09-13
fix(User management): Invalid Markdown syntax was breaking this link

paperclypse:
The bold asterisks need to go outside of the URL markdown. This came in as a hero request.


zuluecho9:
just a heads up @urbiz-nr, in case it might relate to other work in that recent PR

zuluecho9:
I added some more stuff to this one, to fix a few 'account/user model' instances to just 'user model'.

PR #3888, opened 2021-09-13
chore: deprecate CAT in Node.js agent documentation.

michaelgoin:
Deprecate CAT in Node.js agent documentation.
  * Notes deprecation.
  * Recommends DT and shares transition doc.
  * Updates default config values for CAT and DT.


zuluecho9:
Made a few more general edits in there related to DT wording/formatting.

PR #3891, opened 2021-09-14
note about tags limitation for metric streams

josemore:
### Give us some context

* Added an explicit note explaining not all tags will be available for all metrics received in metric streams. We might improve this soon but worth mentioning in docs since it's a FAQ. 


urbiz-nr:

urbiz-nr (approved):

PR #3892, opened 2021-09-14
Removing beta and adding note about all AWS metrics supported

josemore:
### Give us some context

* This page is automatically generated based on conversion files we have for AWS API Polling integrations. We want to remove the beta callout and mention the full list of metrics should now be obtained from AWS (all metrics can be ingested with metric streams). 


urbiz-nr:

urbiz-nr (approved):

urbiz-nr (approved):

PR #3893, opened 2021-09-14
chore: Updating gRPC compatibility

meiao:
### Give us some context
Updating compatibility since the Java agent is no longer compatible with the latest gRPC release.

No


dbarnesbrownNR (approved):
All good, thank you!

PR #3918, opened 2021-09-15
chore: java configuration - removing conflicting duplicate entry

meiao:
### Give us some context
This configuration property had 2 entries and they were conflicting.
Removing the erroneous one.

No


bradleycamacho:
Thanks for the edit @meiao !

PR #3924, opened 2021-09-15
Remove extraneous word

TrainnosaurusRex:
Why only "digital"? I use New Relic custom events to monitor the speed of my carrier pigeons too!


austin-schaefer:
PR from my training account, closing

PR #3930, opened 2021-09-16
JMX : Add config V4 examples and reformat args

alvarocabanas:
Redesigning the JMX integration configuration section.
Add example configurations


urbiz-nr (approved):
Thanks for updating docs, @alvarocabanas!

PR #3940, opened 2021-09-16
Remove "need to change environment" bullet point

hunterbeezley:
While infrequent, when customers inevitably reach out to NR Support to move their II environment to another account, engineering has informed us (NR Support) that this is not actually possible and the customer needs to delete the environment via the incidentIntelligenceEnvironmentDeleteEnvironment NerdGraph API, then re-create it in the UI on the account that they want.

Since deleting the environment is not the same as moving the environment (there are configurations lost in deleting), I'm proposing that we simply take out mention of "moving" their environment completely.

### Give us some context


paperclypse (approved):
This change looks good to me @hunterbeezley. I didn't get a response when I pinged their hero, so I'm going to go ahead and approve this one.

If someone has a problem with it, we can always add it back in.

Thanks for helping out!

Issue #3724, opened 2021-08-31
Issue: Manage data retention


newtob-DB:
Hi,
The Metric Data Rollup language could be improved:
https://docs.newrelic.com/docs/telemetry-data-platform/manage-data/manage-data-retention/#metric-data-rollup

See this quote from the page:
"Then aggregated to 1-hour roll-ups for 90 days"

Oisin McColgan @newrelic suggests that it's not just 90 days their rolled up to 1-hour, he suggests this 1 hour roll up lasts for as long as the metrics are kept, most for 13 months, some indefinitely. That said, i'm not sure what these bullet points should say. Could you clarafy?

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/telemetry-data-platform/manage-data/manage-data-retention/
* **OS**: Unknown
* **Browser**: Unknown
* **Device**: Unknown

urbiz-nr:
Thanks for getting in touch, @newtob-DB! I'm sharing this with the team so we can work on improving and clarifying the docs.

bradleycamacho:
@urbiz-nr @mmfred Could one of you update the status of this issue? Thanks!

mmfred:
I'm working on this now. Kept getting pulled off to other things.

mmfred:
@newtob-DB - Thanks again for your request. I hope the updates pushed today are a bit clearer.

PR #3945, opened 2021-09-16
fix(data retention): added more detail to the metric rollup info.

mmfred:
This added detail fixes issue #3724


urbiz-nr (approved):
Looks good!

PR #3955, opened 2021-09-17
Update docs for infinite tracing FedRAMP support

toddwest:
### Give us some context

* New Relic has added the ability to select a FedRAMP compliant observer as of 9/17/2021 and these changes reflect that updated support, removing the old wording about it not being supported.


rhetoric101 (approved):
@toddwest, these changes look good!

PR #3978, opened 2021-09-20
chore: relaxing stale bot

jpvajda:
relaxes stalebot 🤖

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,7 +1,7 @@
 # Number of days of inactivity before an issue becomes stale
-daysUntilStale: 30
+daysUntilStale: 90
 # Number of days of inactivity before a stale issue is closed
-daysUntilClose: 10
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -1,7 +1,7 @@
 # Number of days of inactivity before an issue becomes stale
-daysUntilStale: 30
+daysUntilStale: 90
 # Number of days of inactivity before a stale issue is closed
-daysUntilClose: 10
+daysUntilClose: 105
```
LizBaker, line 4: I think this should probably be 15 instead of 105. it looks like the old logic was stale after 30 days and closed after 10 days on top of the original 30. so i'm thinking this set up would be close after 195 days which i don't think is the intention

aswanson-nr (approved):

bradleycamacho:
Nice 😎

PR #3986, opened 2021-09-21
Add Azure log forwarding docs

danybmx:
### Give us some context

We are adding documentation about how to configure log forwarding from Azure to NR for two use cases:
- Generic logs
- Azure Activity logs

The template and steps are more or less the same so we've added an optional step on the steps list to activate Activity Logs and all the rest are common steps.

No


src/nav/full-stack-observability.yml
```diff
@@ -2190,6 +2190,8 @@ pages:
                 path: /docs/logs/enable-log-management-new-relic/enable-log-monitoring-new-relic/vector-output-sink-log-forwarding
               - title: Stream logs from Heroku
                 path: /docs/logs/enable-log-management-new-relic/enable-log-monitoring-new-relic/heroku-log-forwarding
+              - title: Azure logs ARM template
+                path: /docs/logs/enable-log-management-new-relic/enable-log-monitoring-new-relic/azure-log-forwarding
```
danybmx, line 2194: Hi reviewer! This is the only real change on this file, the others are trimmed whitespaces from the end of the lines, I can put them back if you want.

x8a: No worries, I noticed that the diff are the spaces. It's not a problem 🙌🏻

x8a (approved):

PR #4000, opened 2021-09-21
chore: Java agent - adding missing JFR properties in config file

meiao:
### Give us some context
JFR feature was added to the agent some time ago, but the configuration template file was not updated with the respective params.
This adds those params.
Sorry for the removal of white space.

No


dbarnesbrownNR (approved):
Looks good to me. Thank you!

PR #4001, opened 2021-09-21
fix(user roles original user model): link broken by misplaced bolding

dbarnesbrownNR:
Couple of `**` in the wrong place, should be okay now.


paperclypse (approved):
Looks good!

PR #4041, opened 2021-09-23
fix(Networks): added OTLP per issue 4038

mmfred:
@GaPeppy - I added some info per your issue 4038. Can you take a look and see if this is doing what you requested. It's not an area I'm familiar with so if you have suggestions or edits, I'd love that.


GaPeppy:
There is a domain list at the top of article under the header "User-facing domains" for customers to possibly update URL filtering devices on their network.  This domain summary should include "*.nr-data.net".

I like the breakout of OTLP as its own area of support in the networks document.

mmfred:
Got review via slack.

PR #4069, opened 2021-09-27
update lambda cloud monitoring documentation

mkrichevsky:
Removed duplicated link (duplicate of lines 16-18)
Added a note regarding common issue for customers


x8a (approved):
Thanks for catching the duplication and adding the new callout, @mkrichevsky! 💃🏻

PR #4080, opened 2021-09-27
Fix awkward wording

keegoid-nr:
- fix wording


paperclypse (approved):
Looks good! Thanks for helping out.

PR #4089, opened 2021-09-28
Add intentions/recommendations for .NET app name API

JcolemanNR:
### Give us some context

This clarifies the intended and recommended use for the SetApplicationName API provided by the .NET Agent. The intention is to avoid some problematic usage seen by GTS.

NOPE! :)


dbarnesbrownNR (approved):
Looks good, thanks!

PR #4093, opened 2021-09-28
Add optional parameter and reformat

linbaker:
### Give us some context

Customer questioned weather the parameters for the `shutdown()` method were optional or not, our Github API documentation makes this more clear and includes a parameter not listed on the docs site:

http://newrelic.github.io/node-newrelic/docs/API.html#shutdown

Reformatted to match the Github format - although it has been awhile since I did any HTML so may need some fixes!

No


x8a (approved):
Thanks for the additions, @linbaker!

PR #4096, opened 2021-09-28
Update to recognize outside New Relic contributor

tbradellis:
This was significant work and goes a long way to improving the agent. We wanted to add special recognition for this update.


dbarnesbrownNR (approved):
Looks good to me!

PR #4116, opened 2021-09-29
Aporia Integration

iazam6:
I've added all the correct information from the Aporia technical docs except the images. I don't think i've formatted the code right though for each box. It's a bit confusing to me. The technical doc instructions are here (the first two paragraphs go in a different part of the website, which I'll be creating another pull request for) https://docs.google.com/document/d/1JBW39gJSMa36Tp9W1uqmsnZEbYcx-PDnWWudP-kjArM/edit#


barbnewrelic:
None of the other procedures in this doc have screenshots so I did not add them in from that google doc link. It took more than 2-1/2 hours of trying to fix the markdown and making the Gatsby build work, so I am not comfortable with making any other changes to this. But the good news is, the text is there in time for the release on Sept 30!

PR #4122, opened 2021-09-30
Morning merge to main

x8a:
A bit earlier to publish the Aporia docs.


src/nav/full-stack-observability.yml
```diff
@@ -2163,36 +2157,32 @@ pages:
...
+            path: /docs/logs/forward-logs/kubernetes-plugin-log-forwarding
+          - title: Logstash plugin
+            path: /docs/logs/forward-logs/logstash-plugin-log-forwarding
+          - title: Microsoft Azure logs ARM template
+            path: /docs/logs/forward-logs/azure-log-forwarding
+          - title: Vector plugin
+            path: /docs/logs/forward-logs/vector-output-sink-log-forwarding
       - title: Logs in context
         path: /docs/logs/enable-log-management-new-relic/configure-logs-context/configure-logs-context-apm-agents
         pages:
```
rudouglas, line 2188: I think maybe this needs to be altered to the new page, it is set up as a redirect currently so will result in page not being accessible

PR #4163, opened 2021-10-04
RHEL 6 end of life

josemore:
### Give us some context

* Given our general policy and based on Red Hat Enterprise 6 end of life, we no longer support it (officially). Agent and integrations would still work but tech support is not guaranteed.
* Oracle Linux is based on Red Hat so it's updated too.


x8a (approved):

Issue #4129, opened 2021-09-30
Issue: Users, roles, permissions (New Relic One user model)


roadlittledawn:
## Description

Short and simple one, BUT looks like this table header has inline style which makes it look weird in dark mode. Can we just remove OR use a more dark-mode-friendly color?
![2021-09-30_11-20-53](https://user-images.githubusercontent.com/2952843/135510003-825899f0-2335-4c2e-af74-ba3ee2182b9d.png)

mmfred:
@zuluecho9 - can you take a look at this feedback? I think it might be of interest to you.

zuluecho9:
Made it slightly lighter background color w/ https://github.com/newrelic/docs-website/pull/4167

PR #4167, opened 2021-10-04
fix(user type table): change bg color

zuluecho9:
Ignore, just testing something for now.


zuluecho9:
Did this work based on issue https://github.com/newrelic/docs-website/issues/4129

PR #4171, opened 2021-10-04
Update list of container orchestrators

keegoid-nr:
- Let customer know they will cause issues if used
- Link to Kubernetes CPM


paperclypse (approved):
Looks good!

PR #4178, opened 2021-10-05
Add benchmark example for ARM64

josemore:
### Give us some context

* Adding a new benchmark scenario running default infra-agent on default EC2 AL2 instance (AMD vs ARM) as a reference for customers adoption ARM instances.


urbiz-nr (approved):

PR #4184, opened 2021-10-05
fix(alerts): Replaced old image about signals

rhetoric101:
This was a documentation channel request about replacing a screen shot. 

I narrowed the screenshot and modified the caption.


barbnewrelic (approved):
Thanks for updating the screenshot. BTW that gif works great!

PR #4191, opened 2021-10-05
Updated links

halocameo:
Please check other links on this page as well.

### DIFF
--- a/src/content/docs/using-new-relic/welcome-new-relic/get-started/glossary.mdx
+++ b/src/content/docs/using-new-relic/welcome-new-relic/get-started/glossary.mdx
@@ -729,11 +729,11 @@ The length of time in seconds to wait for the aggregation window to fill with da
     id="metric"
     title="metric timeslice"
   >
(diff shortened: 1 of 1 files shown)


dbarnesbrownNR:
Hi @halocameo, thank you for these updates! However it appears all the redirects in place work for the content you edited, and we are going to be making some changes to parts of the site taxonomy this PR touches very soon. In light of that, I'm going to tentatively suggest we hold off on these changes pending confirmation from someone on the Docs team that we won't be overwriting the update with work in the pipeline for Oct release. Thanks!

dbarnesbrownNR:
Ah, gotcha @halocameo -- in that case I get the rationale so will try to balance with other considerations. Could you do me a favor and shoot me a message over Slack so we can discuss? I can't seem to figure out your New Relic identity from your GH handle and for privacy reasons we only use handles over GH.

dbarnesbrownNR (approved):
Approving so we have accurate h2 linking in advance of larger taxonomy updates. Thanks!

PR #4249, opened 2021-10-08
docs(Alerts): updates Alerts NerdGraph docs for notificationChannels

founddrama:
## Give us some context

Updates Alerts NerdGraph docs for notificationChannels:

  - adds notes about support for retrieving policies with channels
  - clarifies a confusing statement about what happens when dissociating channels from policies


rhetoric101 (approved):
@founddrama, thanks for adding the example query to list channels by policy, as well as for clarifying the removal of alert notification channels.

PR #4250, opened 2021-10-08
docs(Alerts): add NerdGraph example for searching Alert Policies by name

founddrama:
## Give us some context

The NerdGraph API for Alerts Policies recently added support for `name` and `nameLike` searches. This adds an example for using the `nameLike` search criterion.


rhetoric101 (approved):
@founddrama, thanks for adding that missing ID and adding this other content!

PR #4278, opened 2021-10-12
fix(Intro doc): Removed stray period

paperclypse:


rhetoric101 (approved):
@paperclypse, thanks for removing this bonus period!

PR #4313, opened 2021-10-14
feat: add customizable root nav bar

rudouglas:


plugins/gatsby-source-nav/gatsby-node.js
```diff
@@ -101,23 +101,22 @@ exports.onCreatePage = ({ page, actions }) => {
...
-        order: ['ASC'],
-      },
-    },
-  });
+const createRootNav = async ({ args, createNodeId, nodeModel }) => {
+  const { slug } = args;
+
+  const rootNavYamlNode = nodeModel
+    .getAllNodes({ type: 'NavYaml' })
+    .filter((node) => node.path === '/');
```
aswanson-nr, line 109: Since we have the `rootNav` boolean property on the nav yaml, can we use that instead of string based check?
```suggestion
    .filter((node) => node.rootNav);
```

plugins/gatsby-source-nav/gatsby-node.js
```diff
@@ -257,26 +256,9 @@ const createNav = async ({ args, createNodeId, nodeModel, locales }) => {
...
-          },
-        },
-      },
-      firstOnly: true,
-    },
-  });
-
   const allNavYamlNodes = nodeModel
     .getAllNodes({ type: 'NavYaml' })
+    .filter((node) => node.path !== '/')
```
aswanson-nr, line 261: Here as well
```suggestion
    .filter((node) => node.rootNav)
```

src/components/RootNavigation.js
```diff
@@ -3,35 +3,15 @@ import PropTypes from 'prop-types';
...
-  );
-
-  const pages = nav.pages.filter((page) => ![tdp, fso, ai].includes(page));
-
   return (
     <nav role="navigation" aria-label="Navigation">
-      <NavItem page={tdp} />
-      <NavItem page={fso} />
-      <NavItem page={ai} />
       <hr />
```
aswanson-nr, line 8: I think we want to get rid of this `<hr />`, it's adding an extra section break to the side bar.
```suggestion
```
Screen Shot 2021-10-14 at 1 37 22 PM

plugins/gatsby-source-nav/gatsby-node.js
rudouglas: I did think of that, but i was afraid that if someone added `rootNav: true` to a nav file it would muss things up, but i guess they could also change the `path` in it, i shall accept this change and remove any mention of `rootNav` from the other navs, we are setting it to false automatically if it's not found anyway

aswanson-nr (approved):

PR #4327, opened 2021-10-14
feat(whats-new): Add ability to mark What's New content as "featured"

TobiasDeardorff:
## Give us some context

This PR allows the specification of What's New content as "featured". This will enable us to handle these "featured" posts appropriately within the NR1 platform in the What's New nerdlet.

To test, add an `isFeatured: true` property/value to the frontmatter section of a post and verify that its value is returned by the api request that is updated in the `gatsby-config.js` in this PR.


TobiasDeardorff:
I can't see the build failure due to access, but it could be that I'm missing something on my end. We can sync up whenever you have the time. Thanks!

roadlittledawn:
@TobiasDeardorff  i think you'll need to move the README out of `src/content`. i think that may be causing the error in the build preview. i think we could probably add that info either to the root README or to the wiki.

roadlittledawn (approved):
LGTM from code perspective. preview looks good too. note that if not specified in frontmatter, it looks like this in the JSON:

```
isFeatured: null
```

PR #4338, opened 2021-10-15
fix(data dictionary): Add definitions for coreCount and processorCount

austin-schaefer:
Adds definitions for these two attributes, see DOC-6300 for more info.


austin-schaefer:
Pinged @josemore in Slack to get a review.

PR #4340, opened 2021-10-15
style(distributed tracing): Replace transparent icon images

austin-schaefer:
These transparent icons look great... But don't work in dark mode. We could change the colors, but then the icons would no longer correspond to the UI. As a workaround, I've just put a white background on each image. 

c.f. DOC-6924


austin-schaefer:
Rob pointed out this isn't the neatest approach, opening a fresh PR.

PR #4348, opened 2021-10-15
move create and delete tip

Charles-Mancuso:
Move delete tip in Monitor downtimes: Disable monitoring during scheduled maintenance times to make it similar to where the edit tip is located (below header, not above)

## Give us some context

Move create and delete tip to be consistent with where edit tip is located


bradleycamacho:
Ha, nice catch @Charles-Mancuso! I'll merge this in as soon as the checks complete.

PR #4371, opened 2021-10-18
Liz/move homepage nav items

LizBaker:
This looks a little crazy because my branch had more up to date commits so they're all listed here, but this adds the 3 featured nav items into the regular nav on the homepage and renames TDP and FSO in the homepage nav. Also updated the yml files of the 3 nav items to remove their icons


LizBaker:
just kidding, going back to the old PR

PR #4375, opened 2021-10-18
fix(nav files): change title of 'solutions & best practices'

zuluecho9:
DOC-7444. Changed title to 'Guides and best practices'. We think this is a better name for this section, especially if we are going to use it as a place to locate various general guides in future.


src/nav/new-relic-solutions.yml
```diff
@@ -1,4 +1,4 @@
-title: Solutions and best practices
+title: Guides and best practices
```
barbnewrelic, line 1: Or "Best practices guides"? Whatever you think is best!

PR #4384, opened 2021-10-19
fix(infra agent): add SLES 11.4 repository installation

rubenruizdegauna:


x8a (approved):
Thanks for the addition! 💃🏻

PR #4394, opened 2021-10-19
Add whats new id deletion logic

LizBaker:
We have a script to generate new whats new IDs, but it didn't include any cleanup for what's new posts that get deleted. 
the `nr1-announcements` end-point doesn't care about this because it only matches IDs to existing posts, but the `nr1-announcements/ids` endpoint grabs the entire ID list which includes the deleted posts. this is was causing issue in the whats new nerdlet since they rely on the `/ids` endpoint to know when to delete a post

### DIFF
--- a/scripts/actions/generate-whats-new-ids.js
+++ b/scripts/actions/generate-whats-new-ids.js
@@ -11,17 +11,23 @@ const generateWhatsNewIds = async () => {
   );
 
   return new Promise((resolve) => {
+    const currentWhatsNewPaths = [];
     vfileGlob('./src/content/whats-new/**/*.md').subscribe({
       next: (file) => {
         const slug = file.path
           .replace(/.*?src\/content/, '')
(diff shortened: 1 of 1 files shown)


roadlittledawn:
should the number of ID entries in [ids.json](https://docswebsitedevelop-lizwhatsnewcleanup.gtsb.io/api/nr1/content/nr1-announcements/ids.json) be equal to the number of posts in [nr1-announcements.json](https://docswebsitedevelop-lizwhatsnewcleanup.gtsb.io/api/nr1/content/nr1-announcements.json) ?

i see 104 IDs and 101 posts.

LizBaker:
@roadlittledawn yes! but i figured i'd let the action run the new script and commit the changes when it does rather than myself committing them. i guess there's no difference really, should i also commit the updated whats new file?

roadlittledawn (approved):

PR #4461, opened 2021-10-22
feat: add scripts for adding/removing redirects from files

rudouglas:
To add the slugs as redirects from all files in $DIRECTORY:

`yarn add-redirects $DIRECTORY`

To remove the slugs as redirects:

`yarn remove-redirects $DIRECTORY`

If $DIRECTORY is not specified it will default to all mdx files in `/docs/`


scripts/remove-redirects.js
```diff
@@ -0,0 +1,78 @@
...
+  const fileTree = await mdxToVfile(filepath);
+  const frontmatter = getFrontmatter(fileTree);
+  const yamlContent = getYamlFromFrontmatter(frontmatter);
+  const yamlJson = yamlToJson(yamlContent);
+  const frontmatterJson = yamlJson[0];
+  const { redirects } = frontmatterJson;
+
+  if (!redirects || !redirects.includes(slug)) {
+    return { skipped: slug };
+  }
```
zstix, line 26: A lot of the `processFile` function (and the `main` function) is pretty similar in both the `add-redirects.js` and `remove-redirects.js` files. We might be able to [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) up this file a little more by moving this logic to the `utils/edit-frontmatter.js` file.

Perhaps something like this?

```js
// function for processing one directory / filepath
const processDir = (processFn) => async (filepath) => {
  // do all the common stuff
  const editedFrontMatter = processFn(frontmatterJson, slug);
  // do the save stuff
  return { processed: slug };
}

// function for processing all the files (e.g. your "main" function)
const processFiles = async (processFn) => {
  // glob stuff
  const processPromises = filepaths.map(processDir(processFn)); // abbreviated, but you could write this out
  const processResults = await Promise.all(processPromises);
  // added / removed files
  const modifiedFiles = processResults.filter(result => result.processed); // cloud be simplified further with a "prop" helper function
  // print stuff
}
```

zstix (commented):
At first glance, this looks good!

I think there's an opportunity to consolidate some of this code so that we don't have to modify two different places in the future.

scripts/utils/docs-content-tools/handlers.js
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Adds slashes to beginning and end of directory if missing
+ * @param {String} directory - The directory to add slashes to
+ * @returns {String} The directory with slashes added if necessary
+ */
+const formatDirectory = (directory) => {
+  !directory.startsWith('/') && (directory = `/${directory}`);
+  !directory.endsWith('/') && (directory = directory.concat('/'));
+  return directory;
```
zstix, line 13: Kinda nitpicky: would you be willing to put these into `if` / `else` blocks? I think this might be a little easier to read:

```suggestion
  if (!directory.startsWith('/')) {
    return `/${directory}`;
  } else if (!directory.endsWith('/')) {
    return directory.concat('/');
  }

  return directory;
```

scripts/utils/docs-content-tools/handlers.js
```diff
@@ -0,0 +1,142 @@
...
+};
+
+/**
+ * Gets slug from path, relative to contents folder
+ * @param {String} filepath - The full filepath of the file
+ * @returns {String} The slug of the filepath
+ */
+const getSlugFromPath = (filepath) => {
+  return filepath.match(/content(.*?).mdx/)[1];
```
zstix, line 22: Is there a chance that this `match` wont find multiple results? If so then this might return an `undefined`.

scripts/utils/docs-content-tools/handlers.js
```diff
@@ -0,0 +1,142 @@
...
+};
+
+/**
+ * Reads a file and separates the content and frontmatter data
+ * @param {String} path - The full filepath of the file
+ * @returns {Object} Original path, slug from path, content, frontmatter data
+ */
+const separateData = async (path) => {
```
zstix, line 30: This doesn't actually need to be an `aync` function, since we're using `readFileSync` (notice the `Sync`) and `frontmatterGH` is synchronous:

```suggestion
const separateData = (path) => {
```

scripts/utils/docs-content-tools/handlers.js
```diff
@@ -0,0 +1,142 @@
...
+  const file = fs.readFileSync(path, 'utf8');
+  const slug = getSlugFromPath(path);
+  const separateData = await frontmatterGH(file);
```
zstix, line 33: `frontmatterGH` is synchronous and therefore does not need to be `await`ed.

```suggestion
  const separateData = frontmatterGH(file);
```

scripts/utils/docs-content-tools/handlers.js
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * Adds a redirect if the slug is not already in the redirects array
+ * @param {Object} data - slug, frontmatterData, rest
+ * @returns {Object} Edited frontmatterData, whether the redirect was added or skipped
+ */
+const addRedirect = ({ frontmatterData, slug, ...rest }) => {
+  if (frontmatterData.redirects.includes(slug)) {
+    return { frontmatterData, ...rest, skipped: slug };
+  }
+  frontmatterData.redirects.push(slug);
```
zstix, line 63: I'm pretty sure that [`Array.prototype.push`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push) modifies the original array, which might cause some weird side-effects. Here's what I mean by that:

```js
const me = { name: 'Zack', favNums: [1, 2, 3] };

const func = (obj) => {
  obj.favNums.push(4);
  return obj;
};

console.log(me); // { name: 'Zack', favNums: [1, 2, 3] };

// the result contains the `push`ed infromation
func(me); // returns { name: 'Zack', favNums: [1, 2, 3, 4] };

// the original array _also_ contains the `push`ed information
console.log(me); // { name: 'Zack', favNums: [1, 2, 3, 4] };
```

I'm not sure if that will cause a problem, but it has the potential to be a nasty bug that could be hard to track down. I'd suggest using a [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) (since it creates a new array) when you return:

```js
return {
  frontmatterData: {
    ...frontmatterData,
    redirects: [...frontmatterData.redirects, slug],
  },
  ...rest,
  added: slug,
};
```

scripts/utils/docs-content-tools/handlers.js
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * Removes a redirect if the slug is in the redirects array
+ * @param {Object} data - slug, frontmatterData, rest
+ * @returns {Object} Edited frontmatterData, whether the redirect was removed or skipped
+ */
+const removeRedirect = ({ frontmatterData, slug, ...rest }) => {
+  if (frontmatterData.redirects.includes(slug)) {
+    frontmatterData.redirects = frontmatterData.redirects.filter(
+      (path) => path !== slug
+    );
```
zstix, line 80: Since we're going to `filter` the redirects to remove the path anyway, I don't know if we need to bother checking if it exists. We can _probably_ remove the `if` statement.

scripts/utils/docs-content-tools/handlers.js
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Retrieves the user inputs from command line
+ * @returns {Object} The directory and action specified by the user
+ */
+const getUserInputs = () => {
+  const userInputs = process.argv.slice(2);
+  const actionIndex = userInputs.findIndex(
+    (arg) => arg === '--add' || arg === '--remove'
+  );
```
zstix, line 115: This might be a good opportunity to reach for [`Array.prototype.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) which returns true if at least one element passes the conditional statement:

```js
const hasAction = userInputs.some((input) => ['--add', '--remove'].includes(input));

if (!hasAction) {
  // complain to the user
}
```

zstix: Ah, but it looks like you probably want to find the first action...well then that's probably a case for [`Array.prototype.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead. Sorry I should have read down a few lines first.

```js
const actionInput = userInputs.find((input) => ['--add', '--remove'].includes(input));

if (!actionInput) {
  // complain to the user
}

const action = actionInput.replace('--', '');
```

Personally, I think that reads a little more clearly too "if there's no action input, complain". A little easier to follow than the `actionIndex === -1` stuff.

zstix (changes requested):
This looks really great. I'm a big big fan of how your broke down the logic into nice bite-sized bits of functionality.

I had a few style suggestions and at least one possible bug we'll want to avoid...but this is in pretty good shape otherwise!

scripts/utils/docs-content-tools/handlers.js
zstix: After chatting about this, it might make more sense to use `Array.prototype.map` instead of a `filter` here.

scripts/utils/docs-content-tools/handlers.js
```diff
@@ -0,0 +1,157 @@
...
+ * @param {String} directory - The parent directory to search
+ * @returns {String[]} All mdx files in the directory and sub-directories
+ */
+const getFilePaths = (directory) => {
+  if (directory.length === 0) {
+    // if user did not supply paths, default to all
+    return glob.sync(`${__dirname}/../../../src/content/docs/**/*.mdx`);
+  }
+
+  return glob.sync(`${__dirname}/../../../src/content${directory}**/*.mdx`);
```
zstix, line 117: Totally a style thing, not sure if this would work right.

```suggestion
  const subDir = directory.length ? directory : '/docs/';
  return glob.sync(`${__dirname}/../../../src/content${subDir}**/*.mdx`);
```

scripts/utils/docs-content-tools/handlers.js
rudouglas: yep

scripts/utils/docs-content-tools/handlers.js
rudouglas: Ok so the filter is actually to remove the slug from the redirects, then we also need to return an object with the data and a status result so i think it makes more sense to leave this

scripts/utils/docs-content-tools/handlers.js
```diff
@@ -0,0 +1,153 @@
...
+  if (!actionInput) {
+    console.warn(
+      '<!> No action specified, please append `--add` or `--remove` to the command'
+    );
+    process.exit(0);
+  }
+
+  const action = actionInput.replace('--', '');
+
+  const directoryInput = userInputs.find((input) => input !== actionInput);
```
aswanson-nr, line 135: Let's add a check here to ensure the directly actually exists.

scripts/utils/docs-content-tools/handlers.js
```diff
@@ -0,0 +1,153 @@
...
+ */
+const getUserInputs = () => {
+  const userInputs = process.argv.slice(2);
+  const actionInput = userInputs.find((input) =>
+    ['--add', '--remove'].includes(input)
+  );
+
+  if (!actionInput) {
+    console.warn(
+      '<!> No action specified, please append `--add` or `--remove` to the command'
```
aswanson-nr, line 128: We should add an example call to this script as part of the output. This will help demonstrate how to properly call the script.

scripts/utils/docs-content-tools/handlers.js
```diff
@@ -0,0 +1,153 @@
...
+};
+
+/**
+ * Globs an array of all mdx files in a directory + sub-directories
+ * @param {String} directory - The parent directory to search
+ * @returns {String[]} All mdx files in the directory and sub-directories
+ */
+const getFilePaths = (directory) => {
+  const subDir = directory.length ? directory : '/docs/';
+  return glob.sync(`${__dirname}/../../../src/content${subDir}**/*.mdx`);
```
aswanson-nr, line 113: Probably not worth changing at this point, but using path.join here would allow it to be file system agnostic

aswanson-nr (commented):
Just 2 small comments on input validation. I'll toss a commit in for those changes

zstix (approved):

PR #4493, opened 2021-10-23
chore(plugins): Remove all references to Plugins product

austin-schaefer:
This PR removes all references to New Relic plugins from across the docs site. See also DOC-7095


austin-schaefer:
Requesting @barbnewrelic as reviewer here since she's by far our top expert on plugins and EoLs

PR #4543, opened 2021-10-26
Nov 1 pricing changes

zuluecho9:
Removing AI/II as billing factor; AUM now available to Pro edition (not just Enterprise); AUM-provisioned user type now has 'basic user' as default (not full user). 

Changes detailed in DOC-7454.


paperclypse (dismissed):
I fixed a couple of technical issues. Otherwise, looks good!

PR #4607, opened 2021-10-28
feat(nerdgraph-tutorials): add anomaly detector configurations API tu…

ahorrocks2:
…torial

## Give us some context

I'm a NR engineer. We have a new API that will be customer facing (it is currently released in customer visibility behind a feature flag, available to a single customer). This PR adds a NerdGraph tutorial for it.


zuluecho9 (approved):

zuluecho9:
A note on this work: it's known we are lacking an overview of what 'anomaly detection' exactly means in New Relic and we'll try to get a doc for that later before this goes GA.

PR #4608, opened 2021-10-28
feat(drop-rules): Add beta feature documentation to drop attributes on aggregates

jonahaapala:
### What problems does this PR solve?

This is a beta feature we are planning to pilot that will give customers more control over their cardinality by allowing them to drop attributes off of rollup data but keep it on the raw data that can be queried in the near term.

### Add any context that will help us review your changes such as testing notes, links to related docs, screenshots, etc.

I tried to link to as many relevant docs as I could. We don't have much on the `RAW` keyword other than the [Tip on this doc](https://docs.newrelic.com/docs/data-apis/ingest-apis/metric-api/NRQL-high-cardinality-metrics/), which I do link to. I tried to have each section of the doc for the other drop rule types represented in this beta feature section so that once GA'd this should be pretty simply integrated into the format of the rest of the doc. I felt it was more clear to keep all the beta content together rather than sprinkle it throughout the document.


jonahaapala:
The link should already be updated. Any other changes? I think we are ready for final review!

dbarnesbrownNR (dismissed):
All good. Missing link was actually one I forgot to add on my first review. :)

dbarnesbrownNR (approved):

PR #4622, opened 2021-10-29
style(homepage): Put most root nav elements in alpha order

austin-schaefer:
Currently the root nav is in a bit of a mix of alphabetical and importance order. This PR puts it in alpha order except the Learn, New Relic One, and Guides categories. 

I've got two possible organizations, the other one is in #4624


austin-schaefer:
Going with alternate version #4624

PR #4624, opened 2021-10-29
style(homepage): Re-organize homepage left nav

austin-schaefer:
Alternate implementation of #4622

### DIFF
--- a/src/nav/root.yml
+++ b/src/nav/root.yml
@@ -42,7 +42,6 @@ pages:
     path: /docs/data-apis
   - title: Data dictionary
     path: /attribute-dictionary
-  - title: section-break
   - title: 'Release notes'
     path: '/docs/release-notes'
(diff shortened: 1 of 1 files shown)


barbnewrelic (approved):
It does seem odd to me that, if we are a data-driven/API company, the sections about data, api, release notes, are below the accounts section. Doesn't quite make sense to me. Also seems a bit odd that "what's new" is at the very bottom. Maybe in the next iteration we list the entire left nav in alpha order and insert section breaks after 5 or 7 items (just to give our eyes a chance to rest). Or think about which groupings really benefit from being at the top. Since this is an experiment with planned additional iterations, this is fine for now. Go for it!

austin-schaefer:
Thanks @barbnewrelic , and agree this would benefit from more iteration!

PR #4626, opened 2021-10-29
Moving 'query/charts' stuff into 'Data & APIs'

zuluecho9:
This was a very quick edit, just copying/pasting query/chart/dashboard-related docs into the 'Data and APIs' nav file. More work to come later for improving this section. 

DOC-7457


paperclypse (approved):
Looks good. All the redirects are good.

PR #4637, opened 2021-10-30
fix: broken link in CodeStream nav

bpmct:
This fixes a broken link in the side navigation of https://docs.newrelic.com/docs/codestream/codestream-ui-overview/


paperclypse (approved):
Good catch, @bpmct! Thanks for fixing that!

PR #4642, opened 2021-11-01
correct user authentication steps and examples

carlossscastro:
Better clarification on how to use SQL or Domain user to authenticate to the mssql instance.
Added missing example for domain user authentication

Signed-off-by: Carlos Castro <ccastro@newrelic.com>


paperclypse (approved):
Looks good!

PR #4661, opened 2021-11-02
fix(agent install pages): prepped for A/B test. Added buttons.

mmfred:
This PR prepares a handful of agent pages for A/B tests. I used buttons consistently in these pages for new account and guided install links.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2160,11 +2160,6 @@
   resolved "https://registry.yarnpkg.com/@sindresorhus/is/-/is-0.7.0.tgz#9a06f4f137ee84d7df0460c1fdb1135ffa6c50fd"
   integrity sha512-ONhaKPIufzzrlNbqtWFFd+jlnemX6lJAgq9ZeiZtS7I1PIf/la7CW4m83rTXRnVnsMbW2k56pGYu7AUFJD9Pow==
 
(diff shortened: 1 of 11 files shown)


yarn.lock
```diff
@@ -2160,11 +2160,6 @@
   resolved "https://registry.yarnpkg.com/@sindresorhus/is/-/is-0.7.0.tgz#9a06f4f137ee84d7df0460c1fdb1135ffa6c50fd"
   integrity sha512-ONhaKPIufzzrlNbqtWFFd+jlnemX6lJAgq9ZeiZtS7I1PIf/la7CW4m83rTXRnVnsMbW2k56pGYu7AUFJD9Pow==
 
-"@sindresorhus/is@^2.0.0":
```
x8a, line 2163: I think we shouldn't commit any changes to the `yarn.lock` file. Could you undo the changes?

x8a (dismissed):
It looks great! Thanks for the changes. Let's just check about the `yarn.lock` file.

yarn.lock
mmfred: I copied the yarn.lock file from develop, which is what the new changes (to 3.3.4) are on the right. This was the recommendation when the nvm upgrade started messing with that file.

x8a (approved):
Thank you so much!

PR #4687, opened 2021-11-03
Clarify callout variants

rhetoric101:
Our style guide had references to classes instead of variants and those classes had the wrong values and confused contributors. 

I removed the references to classes and just talked about variants.

I fixed a broken callout in alerts where a contributor had followed our style guide, which was incorrect.


urbiz-nr (approved):
Looks good, thanks for updating this!

PR #4713, opened 2021-11-04
Changing callout on infra-agent tarball install

josemore:
## Context

* This callout is outdated. The full standard agent is now supported on Graviton/ARM. Keeping the callout to highlight the custom installation via tarball is still not officially supported.


mmfred:
Thanks for the fix @josemore!

Issue #4703, opened 2021-11-04
Issue: Language agents and distributed tracing


tanaka-takayoshi:
## How can we make our docs better?

The example configuration for Java Agent has an invalid format. I confirmed the following format is working.

```
distributed_tracing:
  enabled: true
infinite_tracing:
  trace_observer:
    host: "YOUR_TRACE_OBSERVER_HOST"
```

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/distributed-tracing/enable-configure/language-agents-enable-distributed-tracing/
* **OS**: Mac OS 10.15.7
* **Browser**: Chrome 95.0.4638.54
* **Device**: Unknown

urbiz-nr:
Closed via #4727.

PR #4727, opened 2021-11-05
DT - Java agent config fix

urbiz-nr:
As reported in #4703.


urbiz-nr:
For some reason it's not rendering properly in my local build, so I'm reviewing this.

urbiz-nr:
Looks good now.

Issue #4726, opened 2021-11-05
Issue: Smart verbose mode description is misleading.


qryuu:
## How can we make our docs better?

>after an error is logged, only the 10 most recent debug messages are logged, and older logs are discarded.

With this description, it looks like the debug log is recorded after the Error is recorded in the generated Log File.
However, the actual behavior is as follows.
When the agent logs an error (calls the logger) it will flush on the disk the debug cache, then it will flush on disk the error message. 

https://newrelic.atlassian.net/browse/GTSE-12220
### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/infrastructure/infrastructure-troubleshooting/troubleshoot-logs/infrastructure-agent-logging-behavior/
* **OS**: Unknown
* **Browser**: Unknown
* **Device**: Unknown

urbiz-nr:
Closing this via #4728.

urbiz-nr:
See also #4729.

PR #4728, opened 2021-11-05
Rewriting smart verbose for Infra logs

urbiz-nr:
As reported in #4726.


urbiz-nr:
Would this description be more accurate, @qryuu?

urbiz-nr:
Closing this and carrying on in #4729.

PR #4740, opened 2021-11-05
Replace NRQL in GET example with working query

philweber:
The NRQL query in the existing example in invalid for most users (there is no SyntheticsEvent by default). If someone copies and pastes that code into their account, it will fail.


rhetoric101 (approved):
@philweber, thanks for catching this and making it better!

PR #4771, opened 2021-11-08
fix(user migration): add user assets

zuluecho9:
For user model migration docs, addition of user assets.


barbnewrelic (approved):
This looks fine, really like the bulleted list! Just one question: Right below that list it says, "It's possible that such a user will find themselves with both the original user record and new user records." Can we give any recommendations about what to do in this situation? For example, if you remove a contractor from your account, is there any way to know what other accounts they belong to, or does New Relic have a standard procedure to delete all contractor access to accounts?

barbnewrelic (approved):
Looks fine!

PR #4772, opened 2021-11-08
fix(aum): add info to okta about creating groups

zuluecho9:
Few things: 

For Okta doc, added info about creating groups in okta. 
More generally, added in AUM requirements a note about wanting to figure out groups on identity provider side before implementing AUM. 
Added info in Okta doc about recent change that they need to create a 'variable' field.


mmfred (approved):
Changes look great. No edits needed.

PR #4781, opened 2021-11-08
chore(codestream): Change line wraps to match rest of site, tweak formatting

austin-schaefer:
I made a LOT of formatting/style changes, but they basically all boil down to:

* Make all paragraphs into single lines, rather than multiple lines, to create a more consistent editor experience (@paperclypse guessing this was an artifact of the way CodeStream did it on their old docs site but LMK if I'm missing something)
* Use bold formatting everywhere for UI elements, where I noticed it
* Tweak list styles to more closely match other docs on the site
* Reformat a few lists and tables to use OL rather than UL

I've validated the formatting by checking each doc in Gatsby Cloud preview. I'd suggest reviewer do the same rather than trying to review line-by-line in the diff.

### DIFF
--- a/src/content/docs/codestream/codestream-integrations/msteams-integration.mdx
+++ b/src/content/docs/codestream/codestream-integrations/msteams-integration.mdx
@@ -5,97 +5,63 @@ description: "How to connect CodeStream to your Microsoft Teams channel."
 
 Connect New Relic CodeStream to your Microsoft Teams channel.
 
-When you post a codemark your teammates will get notified via the activity feed,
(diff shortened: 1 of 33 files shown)


paperclypse (dismissed):
Thanks for fixing those line breaks! I've clicked through every CodeStream doc. It's all still looking good! APPROVED

austin-schaefer:
Hey @paperclypse think your review went stale because I pushed up an additional commit for those code blocks we were talking about. Can you re-approve?

paperclypse (approved):
Looks good!

PR #4797, opened 2021-11-09
Improve documentation for .NET agent configuration

JcolemanNR:
## Give us some context

It was previously unclear as to whether or not the `sendDataOnExit` configuration option actually initiated a harvest cycle. This change clarifies the impact of this configuration option.


rhetoric101 (approved):
@JcolemanNR, thanks for making this update!

PR #4799, opened 2021-11-09
fix: update macos supported versions

josemore:
## Give us some context

* Aligning supported macOS version with our brew dependency, last 3 versions of macOS are supported.


rhetoric101 (approved):
@josemore, thanks for making these updates!

PR #4802, opened 2021-11-09
chore(using NR): Remove plugins redirects

austin-schaefer:
This doc has way, way too many redirects. See also #4491 for more context


rhetoric101 (approved):
@austin-schaefer, thanks for trimming these redirects!

PR #4812, opened 2021-11-10
Fix sample Dockerfile for .NET Agent Framework Apps

JcolemanNR:
## Give us some context

The existing docker example for windows .NET framework apps no longer worked and needed to be updated. Improvements include:
- Agent installer is downloaded directly from New Relic
- Agent installer is now deleted after installation
- The source docker image references the new microsoft mcr registry


rhetoric101 (approved):
@JcolemanNR, thanks for making these changes.

PR #4814, opened 2021-11-10
Improve windows .net core docker example

JcolemanNR:
## Give us some context

The Windows .NET Core docker example needed `powershell.exe` to be prepended to several commands.  The example now dynamically downloads the agent as well


x8a (approved):
Thanks for the updates!

PR #4815, opened 2021-11-10
feat(Split test): Added Split test for signups

x8a:


src/components/SignupsSplit.js
```diff
@@ -0,0 +1,109 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ButtonLink from './ButtonLink';
+import ButtonGroup from './ButtonGroup';
+import { Link } from 'react-router-dom';
+import { useTreatments, SplitContext } from '@splitsoftware/splitio-react';
+
+const SignupsSplit = (props) => {
+  const treatments = useTreatments(['signups']);
+  const treatment = treatments['singups'];
```
rudouglas, line 10: looks like this may be spelled wrong should it be `signups`? Also as per the linter we can change it to dot notation
```suggestion
  const treatment = treatments.signups;
```

src/components/SignupsSplit.js
```diff
@@ -0,0 +1,109 @@
...
+    <div {...props}>
+      <ButtonGroup>
+        <ButtonLink
+          data-tessen="stitchedPathLinkClick"
+          role="button"
+          to="https://newrelic.com/signup"
+          variant="primary"
+        >
+          Get an account
+        </ButtonLink>
```
rudouglas, line 23: Only a suggestion but seeing as we are duplicating a couple buttons here, we could break those out into their own variables and just include them in each group. So we'd have:

```js
const accountBtn = (
       <ButtonLink
          data-tessen="stitchedPathLinkClick"
          role="button"
          to="https://newrelic.com/signup"
          variant="primary"
        >
          Get an account
        </ButtonLink>
)
```

and then include it using `{accountBtn}` in the `btn` group

rudouglas (changes requested):

src/components/SignupsSplit.js
x8a: When I talked to @aswanson-nr he suggested to leave it as `'signups'`, since we're referencing the string. To avoid having to create a const for the string just to use it here.

src/components/SignupsSplit.js
x8a: Thanks! I'll create it 🙌🏻

src/components/SignupsSplit.js
rudouglas: hmm maybe, i'm not actually sure what format the `treatments` data is in so Alec might be right, though in that case we'll need to add an exclusion rule to ignore this linting error

src/components/SignupsSplit.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useTreatments, SplitContext } from '@splitsoftware/splitio-react';
+
+const SignupsSplit = ({
```
aswanson-nr, line 5: With the latest changes, this component becomes much more generally applicable. So let's rename it to something that doesn't relate to the test itself. Some ideas: `ABTester` or maybe `SplitTester`

src/components/SignupsSplit.js
```diff
@@ -0,0 +1,27 @@
...
+  testChildren,
+  defaultChildren,
+  treatmentWithConfig,
```
aswanson-nr, line 8: Along the lines of making this more general, let's have the split test name and event be props as well.
```suggestion
  testChildren,
  defaultChildren,
  testName,
  testEvent,
```

src/components/SignupsSplit.js
```diff
@@ -0,0 +1,27 @@
...
+}) => {
+  const treatments = useTreatments(['signups']);
+  const { treatment } = treatmentWithConfig;
```
aswanson-nr, line 11:
```suggestion
  const { treatment, config } = treatments[testName];
```

aswanson-nr: This is similar to what's going on here https://github.com/newrelic/gatsby-theme-newrelic/pull/503/files#diff-c3e9cf10d34504caf408d39e6042ae504aaa642bba82ab1778b62109e771ba8eL38-L44

src/components/SignupsSplit.js
```diff
@@ -0,0 +1,27 @@
...
+  const { isReady } = React.useContext(SplitContext);
+
+  return isReady && treatment !== 'off' && treatments === 'signups' ? (
```
aswanson-nr, line 14: `treatments` should be an object at this time, so I'm not sure this will ever be true. Adding a check for the 'control' treatment should be fine. That's when things haven't quite loaded yet.
```suggestion
  return isReady && treatment !== 'off' && treatment === 'control' ? (
```

aswanson-nr (commented):
This looks great! Just had a few comments around generalization and some props. One thing I noticed is that you'll need to add some handling around SplitIO's click events, so that we can track who uses what.

PR #4817, opened 2021-11-10
Add Newrelic CLI steps to convert integrations config files

carlossscastro:
This PR adds steps to use Newrelic CLI to convert config files from v3 to v4

Signed-off-by: Carlos Castro <ccastro@newrelic.com>


x8a (dismissed):
Thank you so much for the updates, @carlossscastro!

x8a (approved):

PR #4824, opened 2021-11-10
Remove spaces in whats new filename post

rhetoric101:
I renamed the TISAX "What's new" filename since I forgot to remove the spaces in the original. Although this file with spaces is still rendered correctly on the site, I have a diagnostic script that counts words in these directories and it fails on files with spaces.

I confirmed with DevEn that this shouldn't create a duplicate entry in the UI because they now have code to prevent that.


paperclypse (approved):
Seems to work!

rhetoric101:
@paperclypse, thanks for testing!

PR #4835, opened 2021-11-11
Remove Istio nav link since content directory is already gone

rhetoric101:
The left navigation link to Istio under "More integrations" wasn't removed when the Istio Adapter directory was removed. This is a clean-up task.


barbnewrelic (approved):
That was totally a "my bad" moment. Completely forgot about the yml file. Thx for fixing that!

PR #4838, opened 2021-11-11
fix: dedupe gatsby-plugin-sharp

aswanson-nr:
Gatsby-plugin-sharp version 3.14.1 has some issues when there were multiple version in the yarn.lock file. This upgrades the plugin to 3.14.3 which includes the fix for this issue. https://github.com/gatsbyjs/gatsby/issues/33557


nr-kkenney (approved):

aswanson-nr:
No longer relevant

PR #4839, opened 2021-11-11
add missing escape character

bkucey:

## Give us some context

The `add Browser monitor` won't work. The escape character is needed to continue the command to the next line.


x8a (approved):
Thanks!

PR #4844, opened 2021-11-11
infra-agent: hostname resolution internal in windows

rubenruizdegauna:
Added difference in hostname resolution between Linux,Macos and Windows


x8a (approved):
Thanks for the update, @rubenruizdegauna!

Issue #4431, opened 2021-10-20
[Docs] [Docs I/A] Apply new home page design


roadlittledawn:
## Summary

Apply design to `src/pages/index.js`.

I don't think any of this _needs_ to be in the theme.

For any design related questions, feel free to reach out to Àngel Ordax

### Questions

- [x]  icons: for time being, can we stick with feather instead of switching to eva based icons?

> `note` this ticket will not introduce a new icon library

### Notes on design
- `Popular content` section: Hard coded list of links. Tech docs team will update these periodically.
- `What's new` section: 3 most recent posts. Use `frontmatter.title` as secondary sorting criteria when multiple posts exist with same datestamp in `frontmatter.releaseDate`
- Tiles: Note the hover effect in the figma mock (on Data explorer)
- ~~Icons: Uses library called [Eva](https://akveo.github.io/eva-icons/#/). Uses SVG similar to feather.~~
- Links: Should use current teal on light/dark (not all design mocks may have been updated for that)

**Figma**
[Desktop (light mode)](https://www.figma.com/proto/0jsTPLtwz6Bljndkbr1COn/open-source-DS-%5BWIP%5D?page-id=56%3A39509&node-id=150%3A52221&viewport=501%2C48%2C1.05&scaling=min-zoom&starting-point-node-id=150%3A52221&show-proto-sidebar=1)

![2021-11-09_09-16-22](https://user-images.githubusercontent.com/2952843/140972707-9052cb61-4a8e-461d-b906-b84a652912c1.png)

[Desktop (light mode)](https://www.figma.com/proto/0jsTPLtwz6Bljndkbr1COn/open-source-DS-%5BWIP%5D?page-id=56%3A39509&node-id=171%3A40520&viewport=501%2C48%2C1.05&scaling=min-zoom&starting-point-node-id=171%3A40520&show-proto-sidebar=1)

[Mobile](https://www.figma.com/proto/0jsTPLtwz6Bljndkbr1COn/open-source-DS-%5BWIP%5D?page-id=141%3A43127&node-id=141%3A43617&viewport=501%2C48%2C1.01&scaling=scale-down&starting-point-node-id=141%3A43617)

## Acceptance criteria

- [ ] Home page updated per design mock spec
- [ ] Buttons in banner are instrumented via Tessen with an attribute/property that identifies it came from home page banner

## Helpful links

jpvajda:
@roadlittledawn thought by end of this week, this design may be finalized.

[...]

PR #4850, opened 2021-11-11
Homepage redesign

LizBaker:
Closes https://github.com/newrelic/docs-website/issues/4431

### Desktop
Screen Shot 2021-11-16 at 2 57 09 PM
Screen Shot 2021-11-16 at 2 57 48 PM

### Mobile
Screen Shot 2021-11-16 at 2 58 14 PM
Screen Shot 2021-11-16 at 2 58 24 PM


legnaleama:
It's looking good @LizBaker !

src/components/HomepageBanner.js
```diff
@@ -0,0 +1,224 @@
...
+
+const HomepageBanner = () => {
+  return (
+    <div
+      css={css`
+        max-height: 415px;
+        position: relative;
+        overflow: hidden;
+        border-radius: 0.5rem;
+        padding-top: calc(415 / 1252 * 100%);
```
clarkmcadoo, line 15: massive nit-pick, but could we set the calculation numbers as variables somewhere? maybe at the top of this file? That way it's easier to tell where the numbers came from. **Disclosure** I can def be talked out of this if this would be a huge pain and majorly divergent to the current pattern

```suggestion
        max-height: ${MAX_HEIGHT_DESKTOP}
        padding-top: calc(${MAX_HEIGHT_DESKTOP} / ${MAX_WIDTH_DESKTOP} * 100%);
```

src/components/HomepageBanner.js
```diff
@@ -0,0 +1,224 @@
...
+              @media screen and (max-width: 1500px) {
+                height: 30px;
+                font-size: 0.85rem;
+              }
+              @media screen and (max-width: 1100px) {
+                height: 25px;
+                font-size: 0.75rem;
+              }
+              @media screen and (max-width: 900px) {
+                height: 40px;
```
clarkmcadoo, line 175: We might look at making the buttons the same width < 900px width, but I'll def leave that up to @legnaleama

src/components/HomepageBanner.js
```diff
@@ -0,0 +1,224 @@
+import React from 'react';
+import { css } from '@emotion/react';
+import { Button, Link } from '@newrelic/gatsby-theme-newrelic';
+import bannerForeground from '../images/bannerForeground.svg';
+import relicsAtWork from '../images/photo-relics-at-work-01.png';
+
```
clarkmcadoo, line 6:
```suggestion
const MAX_HEIGHT_DESKTOP = 415
const MAX_WIDTH_DESKTOP = 1252
const TABLET_BREAKPOINT = 900
```

clarkmcadoo: just did those 3 as an example, but we'd set all of them here

src/components/HomepageBanner.js
clarkmcadoo: Just double-checked the designs, disregard this!

roadlittledawn:
i believe they want all sections from New Relic Integrations and down removed entirely. waiting for confirmation from Betsy

src/components/HomepageBanner.js
LizBaker: we definitely can, this is actually a weird hack to make the div the banner lives in respect a certain aspect ratio. so more like MAX_HEIGHT_BANNER/MAX_WIDTH_DESKTOP

src/pages/index.js
```diff
@@ -11,166 +11,125 @@ import {
...
 
 const HomePage = ({ data }) => {
   const {
     site: { layout },
+    allMarkdownRemark: { edges: whatsNewPosts },
   } = data;
 
   const { t } = useTranslation();
 
+  const latestWhatsNewPosts = whatsNewPosts.slice(0, 3).map((edge) => {
```
zstix, line 37: If we only need three, then we can add a `limit` to the graphql query, which should speed up the time it takes to build this page. I'll comment a suggestion down below too.

```suggestion
  const latestWhatsNewPosts = whatsNewPosts.map((edge) => {
```

src/pages/index.js
```diff
@@ -274,17 +246,43 @@ export const pageQuery = graphql`
         contentPadding
       }
     }
+    allMarkdownRemark(
+      sort: {
+        fields: [frontmatter___releaseDate, frontmatter___title]
+        order: [DESC, ASC]
+      }
+      filter: { fields: { slug: { regex: "/whats-new/" } } }
```
zstix, line 254: I mentioned it above in this file: if we only need three posts, let's limit the query. That should speed this up a little bit.

```suggestion
      sort: {
        fields: [frontmatter___releaseDate, frontmatter___title]
        order: [DESC, ASC]
      }
      filter: { fields: { slug: { regex: "/whats-new/" } } }
      limit: 3
```

src/pages/index.js
```diff
@@ -274,17 +246,43 @@ export const pageQuery = graphql`
...
+    ) {
+      edges {
+        node {
+          id
+          frontmatter {
+            title
+            releaseDate(formatString: "MMMM D, YYYY")
```
zstix, line 261: I don't actually think that we are using the release date on this page. We can probably remove it from the query (and the related propTypes).

```suggestion
```

zstix (commented):
This looks so much better!!! That said, I had a few comments that we might want to consider incorporating.

Also, one other small nitpick: I found that the integrations looked a little weird without a border. What do you think about us adding a 'lil something to help improve them feeling like "tiles":

Screen Shot 2021-11-16 at 17 05 53

legnaleama:
Hey @LizBaker, this is looking great, just a few amends: 

### Banner
The banner buttons aren't the same as in the Figma, both in style and hover effects. I'm aware I'm introducing some refactoring tasks here but I'd like to know what the long term implementation would be for those changes and/or if we could fit any at all in this PR. 

**Main differences**
- On light mode: both CTAs are using the dark mode version
- On both light and dark modes: Hover effect doesn't match the prototype in Figma

### CTAs
There seems to be a weird scaling going on, is this part of some Gatsby library? Ideally we should re-use the default button sizes (ie: SM, MD, LG) and tailor them to each breakpoint. In the past I've also used variants of those 3 scales for mobile and desktop independently but I think that is a bit overkill.
https://user-images.githubusercontent.com/3636549/142199005-54da63eb-86ff-4b3a-b54c-379c2e5f8de1.mov


### Mobile
**Layout issues on the banner**
- On mobile: The order of the buttons is reversed. Currently the secondary CTA ("Find your quickstart") is the first button. On the original design the free account CTA should be the main first action.

**Banner details**
- This is probably my fault but I noticed we use the same foreground image for desktop and mobile. I just went ahead an isolated the mobile version on Figma so it's easier to export. The previous one was also the wrong color on my side, sry.
- I also added a cropped version of the photo for mobile for better alignment with the decoration elements.
image

roadlittledawn (dismissed):
we can remove all sections from `New Relic integrations` on down

src/pages/index.js
LizBaker: oh perfect!

src/pages/index.js
LizBaker: left behind when i pasted the query 😅

src/pages/index.js
```diff
@@ -11,166 +11,125 @@ import {
...
-        <SectionDescription>{t('home.tdp.description')}</SectionDescription>
+      <Section layout={layout}>
+        <SectionTitle title={t('home.tdp.title')} />
         <DocTileGrid>
-          {tdp.tiles.map((link, idx) => (
+          {tdp.tiles.map(({ link, icon }, idx) => (
             <DocTile
               key={idx}
               title={t(`home.tdp.t${idx + 1}.title`)}
-              description={t(`home.tdp.t${idx + 1}.description`)}
```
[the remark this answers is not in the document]
LizBaker: ah nope, oversight

src/components/HomepageBanner.js
```diff
@@ -0,0 +1,224 @@
...
+                bottom: 100px;
+              }
+            `}
+          >
+            Create a free account
+          </Button>
+          <Button
+            as={Link}
+            to="https://developer.newrelic.com/instant-observability/"
+            css={css`
```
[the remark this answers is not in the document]
LizBaker: just gave it that variant (it's required anyway) but found the colors don't match the design so i'm overriding them for now

src/components/HomepageBanner.js
```diff
@@ -0,0 +1,224 @@
...
+            @media screen and (max-width: 900px) {
+              bottom: 3rem;
+              height: 15%;
+            }
+          `}
+        >
+          <Button
+            as={Link}
+            to="https://newrelic.com/signup"
+            css={css`
```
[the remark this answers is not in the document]
LizBaker: good idea! i was also considering using emotion's styled function but i'm not sure which is better

src/components/HomepageBanner.js
```diff
@@ -0,0 +1,224 @@
...
+        `}
+      >
+        <img
+          src={bannerForeground}
+          alt="banner foreground"
+          css={css`
+            position: relative;
+            height: 100%;
+            max-height: 415px;
+            z-index: 1;
```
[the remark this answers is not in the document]
LizBaker: gotcha, i think the sliver of pixels is now fixed with a new svg!

src/components/HomepageBanner.js
```diff
@@ -0,0 +1,224 @@
...
+              `}
+            >
+              introductory statement
+            </Link>{' '}
+            for a wider view on our platform's capabilities.
+          </p>
+        </div>
+        <div
+          css={css`
+            position: absolute;
```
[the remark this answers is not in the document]
LizBaker: very true, i started with separate elements and it became weirder from there, gonna see how a refactor might look

austin-schaefer:
Looking great @LizBaker ! One bit of feedback from me: Can we decrease the header size on `Welcome` and `Getting started`/`What's new`/`Popular content` sections? They look a bit too large to me relative to the rest of the copy and headers. 

image

austin-schaefer:
cc @legnaleama on my above comment re:header size

LizBaker:
@austin-schaefer definitely game if @legnaleama is!

src/components/HomepageBanner.js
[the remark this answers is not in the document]
zstix: FWIW, I'm still seeing the 'lil sliver with the new image:

Screen Shot 2021-11-18 at 14 55 03

src/components/HomepageBanner.js
```diff
@@ -0,0 +1,301 @@
...
+      css={css`
+        max-height: 423px;
+        position: relative;
+        overflow: hidden;
+        border-radius: 0.5rem;
+        padding-top: calc(423 / 1252 * 100%);
+        background: var(--color-brand-700);
+
+        @media screen and (max-width: ${BREAKPOINTS.SINGLE_COLUMN}) {
+          padding-top: calc(450 / 380 * 100%);
```
zstix, line 27: We might want to leave a comment on what's going on here (or abstract the aspect ratio into a constant at the top of this file).

src/components/HomepageBanner.js
```diff
@@ -0,0 +1,301 @@
...
+          <source
+            srcSet={bannerForegroundMobile}
+            media={`(max-width:${BREAKPOINTS.SINGLE_COLUMN})`}
+          />
+          <img
+            src={bannerForeground}
+            alt="banner foreground"
+            css={css`
+              position: relative;
+              height: 100%;
```
zstix, line 64: This _might_ solve the issue with the little sliver of background showing through.

```suggestion
              height: calc(100% + 1px);
```

src/components/HomepageBanner.js
```diff
@@ -0,0 +1,301 @@
...
+
+const HomepageBanner = () => {
+  return (
+    <div
+      css={css`
+        max-height: 423px;
+        position: relative;
+        overflow: hidden;
+        border-radius: 0.5rem;
+        padding-top: calc(423 / 1252 * 100%);
```
zstix, line 23: If we are positioning all the contents (just one div) with `position: absolute`, then we might not need to set the padding. If we're using the padding for the aspect ratio trick, then nevermind!

src/components/HomepageBanner.js
```diff
@@ -0,0 +1,301 @@
...
+              p {
+                padding-right: 1.5rem;
+                margin-bottom: 0;
+              }
+            }
+          `}
+        >
+          <h1
+            css={css`
+              color: white;
```
zstix, line 137: The theme has a color that's pretty close to the design for this header. We can probably remove this.

```suggestion
```

src/pages/index.js
```diff
@@ -406,103 +290,81 @@ const WelcomeTile = ({
...
-      ${welcomeTileStyles[variant]};
     `}
   >
-    <h3>{title}</h3>
-    <p>{description}</p>
-  </SurfaceLink>
+    <div
+      css={css`
+        display: flex;
+        padding-left: 2rem;
```
zstix, line 307: I think we want a tiny bit of padding on both the left _and_ the right.
```suggestion
        padding: 0 2rem;
```

zstix (approved):

zstix (approved):

legnaleama:
Let's go 🚀!! This looks great, let's take it live! Great work @LizBaker 💪 !

austin-schaefer:
@LizBaker : Pushed up three more commits, basically in order of importance in case any cause issues:

* fd0528a: Must-be-merged, fixes some broken links:
* a4f37f0: Non-fatal but important; tweaks naming and order for consistency with brand guidelines
* 1a9f3e2: Non-fatal, changes link destinations to point at docs rather than index pages

src/pages/index.js
```diff
@@ -264,6 +136,20 @@ HomePage.propTypes = {
...
+          whatsNewPosts: PropTypes.shape({
+            frontMatter: PropTypes.shape({
+              title: PropTypes.string,
+            }),
+            fields: PropTypes.shape({
+              slug: PropTypes.string,
+            }),
+          }),
+        })
+      ),
```
zstix, line 151: I ran the query in my local version of GraphQL and I got the following response:

```json
{
  "data": {
    "allMarkdownRemark": {
      "edges": [
        {
          "node": {
            "id": "80d638c5-d1e2-58c3-b4d7-98c3a8581cb2",
            "frontmatter": {
              "title": "Kubernetes Horizontal Pod Autoscaler (HPA) support"
            },
            "fields": {
              "slug": "/whats-new/2021/11/Whats-New-K8sHPA"
            }
          }
        },
```

Which makes me think that, if we're going to type this thing all the way down, we want the types to be something like this (`whatsNewPosts` > `node` and `frontMatter` to `frontmatter`):

```suggestion
      edges: PropTypes.arrayOf(
        PropTypes.shape({
          node: PropTypes.shape({
            frontmatter: PropTypes.shape({
              title: PropTypes.string,
            }),
            fields: PropTypes.shape({
              slug: PropTypes.string,
            }),
          }),
        })
      ),
```

LizBaker:
Waiting on Japanese translations for translations.json before merging

josephgregoryii (approved):

PR #4858, opened 2021-11-11
feat/npm new troubleshooting

thezackm:
adding new troubleshooting docs for network performance monitoring

CC @x8a


dbarnesbrownNR:
@x8a this ticket came in through an internal channel and @mmfred informed @thezackm you'd likely want to review. Thanks!

src/content/docs/network-performance-monitoring/troubleshooting/snmp-walk.mdx
```diff
@@ -11,63 +11,75 @@ redirects:
...
+### Capturing full SNMP walk [#capturing-snmpwalk]
 
-It is often desirable to capture the output of walking every OID available on your device(s). This output is especially used when creating new SNMP profiles for `ktranslate` and is a required artifact when [opening a profile request](https://github.com/kentik/snmp-profiles/issues/new/choose) on the GitHub repo.
+You may want to capture the output of walking every OID available on your devices. This output is used when creating new SNMP profiles for `ktranslate`, and it's a requirement to [open a profile request on GitHub](https://github.com/kentik/snmp-profiles/issues/new/choose).
 
-The primary differences in your command for this are changing the target to a root `.` and redirecting the output to a file that you can gather data from later. 
+The primary differences in your command for this are changing the target to a root `.`, and redirecting the output to a file that you can gather data from later. 
 
-**SNMP v2 Example:**
+**SNMP v2 and v3 example**
```
thezackm, line 75: @x8a - this is not a v3 example; if we need to include both I can provide a sample here but my assumption was since it was such a small change the reader could look at the previous v3 example to make their own edit there

src/content/docs/network-performance-monitoring/troubleshooting/snmp-walk.mdx
```diff
@@ -11,63 +11,75 @@ redirects:
...
 ```bash
 snmpwalk -v 2c -On -c $COMMUNITY $IP_ADDRESS . >> snmpwalk.out
 ```
 
-*SNMP v3 has the same required changes*
+The output of this command will be a file named `snmpwalk.out`, that lists every OID that the device responds to.
 
-The output of this command will be a file named `snmpwalk.out` that lists every OID that the device responds to. Note on devices with a large number of interfaces it is expected for this walk to take upwards of 10+ minutes to complete. 
+<Callout variant="tip">
+Devices with a large number of interfaces, this walk takes more than 10 minutes to complete.
```
thezackm, line 84: replace: 

`Devices with...` >> `On devices with...`

`this walk takes...` >> `SNMP walks can take...`

src/content/docs/network-performance-monitoring/troubleshooting/snmp-walk.mdx
x8a: I thought the same example works for both version but if it's not the case, let's add a specific one for v3 👌🏻.

thezackm:
```bash
snmpwalk -v 3 -On -l $LEVEL -u $USERNAME -a $AUTH_PROTOCOL -A $AUTH_PASSPHRASE -x $PRIV_PROTOCOL -X $PRIV_PASSPHRASE $IP_ADDRESS . >> snmpwalk.out
```

x8a (approved):

PR #4875, opened 2021-11-12
chore: Add new supported regions for Firelens log forwarding

jsubirat:
This PR adds some new regions to which we replicate our log forwarding images for Firelens.


x8a (approved):
Thanks for these additions!

PR #4887, opened 2021-11-13
feat: Add documentation for DelayedJob custom attributes

danielmklein:
Adding a section with code samples on how to add custom attributes to DelayedJob transactions by using a DJ plugin.

## Give us some context

With the New Relic Ruby agent, it is easy and well-documented to add custom attributes to transactions from pieces of code like Rails controllers, services, etc. However, it is not well-documented how to do the same from DelayedJob when one wants to access fields that are DelayedJob-specific, like `run_at`, etc.

This PR adds a section with code snippets demonstrating how to accomplish this.

The original discussion on this occurred in https://github.com/newrelic/newrelic-ruby-agent/issues/834, which led to this pull request.


x8a (approved):
Thanks for the addition, @danielmklein!

PR #4908, opened 2021-11-16
Unify wording around US and EU region

DrMurx:
The documentation uses "US region" and "EU region" most of the time. "US accounts" and "EU accounts" seems to have been used on this page only, and it might be misleading.


x8a:
Hi, @DrMurx! Thanks for your edits!

Let me check in with the writing team to hear what their thoughts are on this matter 👌🏻.

x8a:
Hey, @DrMurx! Thank you again for your contribution 👏🏻👏🏻👏🏻.

Unfortunately, this language is used across our documentation and we want to keep consistency with our terminology, so we won't be merging this change. However, feel free to open new PRs with suggestions you have 🙌🏻.

PR #4914, opened 2021-11-16
chore: updating v3 example syntax and small readability edit

thezackm:
updating the examples for v3 walks to support the proper output formatting.
also updated small sentence structure point on the last callout tip. 

cc: @x8a


x8a (approved):
Thanks again for your great contributions! 👏🏻

PR #4933, opened 2021-11-16
Adds cosmosdb support 

vuqtran88:
Update .NET agent compatibility matrix to include cosmosdb client library support.

### DIFF
--- a/src/content/docs/apm/agents/net-agent/getting-started/net-agent-compatibility-requirements-net-core.mdx
+++ b/src/content/docs/apm/agents/net-agent/getting-started/net-agent-compatibility-requirements-net-core.mdx
@@ -365,6 +365,54 @@ If your application is hosted in ASP.NET Core, the agent automatically creates a
       <tbody>
         <tr>
           <td>
+            Cosmos DB
+          </td>
+
+          <td className="fcenter">
+            <Icon
+              style={{color: '#328787'}}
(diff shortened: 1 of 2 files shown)


dbarnesbrownNR:
Hi @vuqtran88, I'm wondering if the info you are adding to the doc also exists in the Microsoft Azure docs themselves? If so, I'd suggest linking directly to those docs rather than listing out a long block of text within a table. Do you think this might work?

vuqtran88:
> Hi @vuqtran88, I'm wondering if the info you are adding to the doc also exists in the Microsoft Azure docs themselves? If so, I'd suggest linking directly to those docs rather than listing out a long block of text within a table. Do you think this might work?

Hi @dbarnesbrownNR , thanks for reviewing. Microsoft Azure does have the docs about their apis; but their docs is huge. I purposely listed out a set of api method calls, because they are methods called that we have tested. I thought this might give the customers better idea what they are expecting from this feature.

dbarnesbrownNR:
Taking my name off this one as I will not be available after Nov 19, but it will be picked up by another member of our team soon, @vuqtran88. Thanks for providing that clarity!

rhetoric101 (approved):
@vuqtran88, thanks for this Cosmos  DB update! We're having docs build issues, but when we get that fixed, we'll publish you changes.

PR #4960, opened 2021-11-17
Add upcoming PageViewTiming network and element attributes to data dictionary

aubreymasten:
## Give us some context

As part of our upcoming release, the Browser agent will now include:
* Information about the element that caused a largestContentfulPaint event
* Information about the current network connection when largestContentfulPaint and firstInteraction events occurred.

Both of these attributes give customers more context on the state of the application when these events occurred. Both new data sources also have equivalent MDN documentation, which is linked in each individual attribute description:
* Element information comes from existing LCP event we capture: https://developer.mozilla.org/en-US/docs/Web/API/LargestContentfulPaint
* Network information is captured from the NetworkInformation API and appended to the events when they occur: https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API

One last note is I removed what appeared to be duplicate text from the existing elementId and elementSize markdown files, if this was intentional let me know and I can replace it!

Related GitHub Issues:

https://github.com/newrelic/newrelic-browser-agent/issues/144
https://github.com/newrelic/newrelic-browser-agent/issues/145

### DIFF
--- a/src/data-dictionary/events/PageView/appName.md
+++ b/src/data-dictionary/events/PageView/appName.md
@@ -11,4 +11,4 @@ events:
   - PageViewTiming
 ---
 
-The name of the application that handled the request as shown in New Relic Browser.
+The name of the application that handled the request as shown in browser monitoring.
--- a/src/data-dictionary/events/PageView/pageRenderingDuration.md
+++ b/src/data-dictionary/events/PageView/pageRenderingDuration.md
@@ -6,4 +6,4 @@ events:
   - PageView
 ---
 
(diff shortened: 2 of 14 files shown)


zuluecho9:
@aubreymasten Hi Aubrey. When is this release meant to go out?

austin-schaefer:
Hi @aubreymasten , looks like this PR is marked as a draft. Is this ready for review? If not, do you know roughly when we're targeting for release?

aubreymasten:
@austin-schaefer, so sorry for the delay I'm not sure how this got lost in the mix! This is ready for review and release.

urbiz-nr (approved):
Thanks for the updates, @aubreymasten!

PR #4992, opened 2021-11-22
feat(Infra): Cleaning up Infra UI docs IA, adding new doc for Pegasus

urbiz-nr:
I've added the new doc for Pegasus and sorted old docs, they were all over the place.


urbiz-nr:
Carrying on in #4997.

PR #4995, opened 2021-11-22
Pegasus updates

urbiz-nr:
- Rearranged old docs
- New UI doc


urbiz-nr:
Need to merge the revert PR first - see #4994.

PR #5001, opened 2021-11-22
chore: replace remaining browser agent actions

rudouglas:


gatsby/on-route-update.js
```diff
@@ -1,10 +1,15 @@
 const onRouteUpdate = ({ location, prevLocation }) => {
   if (
-    window.newrelic &&
+    window.Tessen &&
     location.hash &&
     location.pathname !== prevLocation?.pathname
   ) {
-    window.newrelic.addPageAction('hash_request', { hash: location.hash });
+    window.Tessen.track({
```
aswanson-nr, line 7: Could the `useTessen` hook be used here instead?

aswanson-nr (approved):
Looks good to me! One question, but no need to change anything

gatsby/on-route-update.js
rudouglas: No it was giving out about that when I tried to use it, was saying it wasn't a react component or some nonsense

PR #5003, opened 2021-11-22
Replace old-fashioned, pre-v3 application creation method with current method

RichVanderwal:
This documentation never got updated years ago when v3 of the Go Agent was released.


paperclypse (approved):
Looks good to me!

PR #5008, opened 2021-11-23
Clark/2536 modify queue script

clarkmcadoo:
Updates the script adding files to the translation queue to include the project id corresponding to Machine Translation or Human Translation depending on the files' frontmatter.

This solution will loop over the set of supported languages to see if the frontmatter contains any of the supported locales, if so, the project id will be the Human Translated project id; otherwise, Machine Translated.


zstix (approved):
Really awesome work!!

rudouglas:
this is gorgeous great work @clarkmcadoo

PR #5021, opened 2021-11-23
Document percentile() and average() with FACET

rbayersdorfer-newrelic:
`percentile()` and `average()` have their own FACET functions and this needs to be specified in the docs.


rhetoric101 (dismissed):
@rbayersdorfer-newrelic, thanks for making these changes! I just added some formatting to a  reference to `t` and also replaced a single bullet with code block formatting for your example.

mmfred (approved):
Looks great @rhetoric101

Issue #3853, opened 2021-09-10
[Translation] Have translation workflow alert stakeholders when translation PR is created


nr-kkenney:
## Summary
Currently, when the translation workflow creates a PR as the last step of the translation process, the content team aren't alerted and aware that they have a PR they need to review. To make it clearer, the workflow should alert them when the PR is created so they can review it.

The action we currently use to generate PRs ([create-pull-request](https://github.com/peter-evans/create-pull-request)) has a `reviewers` option we can pass in to specify reviewers for the PR. It would look similar to [this line](https://github.com/newrelic/docs-website/blob/82c7feb816287c0788d0c9d567217b8bda6d25f3/.github/workflows/check-translations-and-deserialize.yml#L73), except `reviewers` rather than team.

On the feature branch, the deven team has already been added as a reviewer.

## Acceptance Criteria
- [x] ~Create a group in github and add relevant stakeholder users to it~
- [x] Translation workflow should be updated to add reviewer
  - add it to both jobs. one job is for machine translation, the other is human.

### Stakeholders
 - DevEn team
 - @jmiraNR 

<details>
  <summary>Previous Considerations</summary>

  Some quick ideas from the group:
* ~post to slack channel~
* add people as reviewers to the pr
* comment @ people in the pr
* other options not considered above

</details>

jpvajda:
solves this https://github.com/newrelic/docs-website/issues/4153

aswanson-nr:
If Github mentions are enough of a notification, then this https://github.com/marketplace/actions/issue-label-notifier action might be perfect!

jpvajda:
@aswanson-nr That approach will work as github sends email notifications to users which is a good way to get notified.

nr-kkenney:
@clarkmcadoo i think you mentioned groups in github. did you have an idea of what this would look like?

i don't see a grouping of people different to teams, and im not sure if we want another team or for julien to be added to our team.

nr-kkenney:
going to remove the team / group acceptance criteria since julien has been added as an individual reviewer, and we don't want to create a new team of just him, or us + him.

julien did mention another hire coming onboard and being able to review translation PRs, but its easy enough to just add them individually when the time comes.

PR #5027, opened 2021-11-23
chore: add stakeholders to translation prs as reviewers

nr-kkenney:
## Summary

Add Julien (stakeholder) as reviewer on translation PRs

Resolves: #3853

### DIFF
--- a/.github/workflows/check-translations-and-deserialize.yml
+++ b/.github/workflows/check-translations-and-deserialize.yml
@@ -70,6 +70,7 @@ jobs:
           branch-suffix: short-commit-hash
           base: develop
           delete-branch: true
+          reviewers: jmiraNR
           team-reviewers: developer-enablement
 
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


clarkmcadoo (approved):

clarkmcadoo:
@nr-kkenney, I think we talked about this for about 4 seconds, but adding a group for something like this in the future might be useful just to manage it via github instead of code changes.  What do you think?

nr-kkenney:
@clarkmcadoo i asked on the main ticket, but what is a group? i think I've only seen `teams` as a way to collect individual people, and I'm not 100% on making a new team for our org. But I'll double check.

clarkmcadoo:
Yeah, that's probably it.  Idk, we'd probably need another ticket for it cuz we'd have to figure out who to put on that team, then actually do it.

clarkmcadoo:
`https://github.com/orgs/newrelic/new-team` makes a new team, but lets talk about this a little bit more before we do anything, it was just an idea.

PR #5034, opened 2021-11-24
Add link from ticketing integrations to errors inbox

rhetoric101:
We have an old doc about ticketing integrations that shows up in a Google search for "New Relic Jira integration" but it has no information about Jira! 

This is a PR that adds a Jira section to this ticketing doc. It simply sends users over to errors inbox, which does talk about Jira. The goal is to improve SEO.


urbiz-nr (approved):
Looks good!

PR #5039, opened 2021-11-24
Explaining how we use each required infra-agent endpoint

josemore:
## Context
* Feedback from support, customers asking how the infra-agent uses the required backend endpoints. Added a simple explanation on each endpoint.


urbiz-nr (approved):

PR #5041, opened 2021-11-24
CloudWatch metric stream cost considerations

josemore:
## Context

* Multiple customers requested details about costs that should be considered when evaluating the AWS CloudWatch metric stream integration. Added a reference on all services required on AWS side and links to the pricing pages.


urbiz-nr (approved):
Thanks for updating the doc, @josemore!

PR #5044, opened 2021-11-25
Updated image for rate limits doc

urbiz-nr:
As requested in #documentation.


urbiz-nr:
Tried locally, it looks ok.

PR #5046, opened 2021-11-25
fix(AWS): note about global services

josemore:
## Context

* Certain AWS services are global (and not per-region), in those cases, their metrics will only be available if customers connect a stream in us-east-1 region (AWS design decision). Added a note explaining this.


urbiz-nr (approved):
Thanks for adding the callout, @josemore!

PR #5047, opened 2021-11-25
Document how to tail large amount of files using Infrastructure Agent Log Forwarder

jsubirat:
It adds documentation to explain how to forward a large amount of files using the Infrastructure Agent Log Forwarder when running on Linux machines.


urbiz-nr (approved):
Thanks for adding this section, @jsubirat - looks great!

urbiz-nr:
Tried locally, works well.

Issue #5042, opened 2021-11-24
Issue: .Net Performance Metrics


brunotag:
## How can we make our docs better?
What is the "name" of these perfmon metrics coming from APM? Which "entity" are they supposed to be from?

* Is something confusing?
How do we query for perfmon metrics coming from APM?

* Is something inaccurate or missing?
An example of query would be really helpful: that would immediately tell me whether or not this is working.

* Were you unable to complete a task? (What task?)
I can't tell whether or not the perfmon metrics are being collected by the New Relic Agent (probably not).

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/apm/agents/net-agent/other-features/net-performance-metrics/
* **OS**: Unknown
* **Browser**: Unknown
* **Device**: Unknown

brunotag:
Update: I figured those are all `FROM Metric SELECT apm.service....` metrics.
An example would help though.

urbiz-nr:
Hello @brunotag - glad you could figure it out! I'll pass this on to the team so they can add an example, as you suggest. Thanks for reaching out!

brunotag:
Actually, I did not.
This is what I was looking for, eventually found it by digging into APM stuff.
Maybe the docs could say "to see these metrics, go to More Views => Metric Explorer"
![image](https://user-images.githubusercontent.com/1215492/143510460-b513bdf3-2bc9-46b4-a618-4ec01f22a2cc.png)

I wasn't aware of a Metric Explorer inside APM

brunotag:
![image](https://user-images.githubusercontent.com/1215492/143510504-dec8f168-8e53-4427-9512-d2f669f3aed2.png)
Actually, this is what I was looking for.

No idea there was a "Metric Explorer" inside the APM UI, hidden in the "More Views" dropdown - and that page of the docs doesn't mention it.

urbiz-nr:
You're right - we'll add a note in the docs so users are more aware of this feature.

urbiz-nr:
@brunotag - #5049 fixes this. Thanks!

PR #5049, opened 2021-11-26
Added mention to metric explorer in Net agent doc

urbiz-nr:
As reported in #5042, added a quick note on the Metric explorer in APM.


x8a (approved):
👌🏻

PR #5052, opened 2021-11-26
Simplify documentation when tailing a large amount of files in the Infra Agent

jsubirat:
Small simplification when tailing a large amount of files using the Infra Agent, as these contents are now explained in more detail in the troubleshoot section.


urbiz-nr (dismissed):
Looks good, @jsubirat!

urbiz-nr (approved):

Issue #5057, opened 2021-11-29
Issue: Introduction to on-host integrations


shinji-newrelic:
## How can we make our docs better?
Link of the "list of available on-host integrations" is missing.
When I click this link on my browser, just blank page is shown. But I opened it as new tab it is redirected to "https://docs.newrelic.com/docs/infrastructure/host-integrations/".

I'm not sure why redirect doesn't work when just clicking case but the link should be updated to new url.

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/infrastructure/host-integrations/get-started/introduction-host-integrations/
* **OS**: Mac OS 10.15.7
* **Browser**: Chrome 96.0.4664.55
* **Device**: Unknown

urbiz-nr:
Thanks for letting us know, @shinji-newrelic - we'll fix this in #5059.

PR #5059, opened 2021-11-29
Fixing OHI list link

urbiz-nr:
As reported in #5057.


urbiz-nr:
Tried locally, link works.

PR #5060, opened 2021-11-29
feat: Exclusions file and tests

rudouglas:
This adds the ability to exclude files under specific paths, and of specific types which are configured using a YAML file.

This also adds tests to the `add-files-to-translation-queue` workflow as there weren't any before, and adds tests for excluding files/types


rudouglas:
oh darn looks like I also included a merge from `main` doh

PR #5061, opened 2021-11-29
feat: exclusion file and tests

rudouglas:
This adds the ability to exclude files under specific paths, and of specific types which are configured using a YAML file.

This also adds tests to the add-files-to-translation-queue workflow as there weren't any before, and adds tests for excluding files/types

### DIFF
--- a/scripts/actions/add-files-to-translation-queue.js
+++ b/scripts/actions/add-files-to-translation-queue.js
@@ -1,6 +1,5 @@
 const fs = require('fs');
 const path = require('path');
-const yaml = require('js-yaml');
 const frontmatter = require('@github-docs/frontmatter');
 const {
   getTranslations,
@@ -10,7 +9,8 @@ const {
 const { fetchPaginatedGHResults } = require('./utils/github-api-helpers');
 const checkArgs = require('./utils/check-args');
 const { prop } = require('../utils/functional');
-const { LOCALE_IDS, EXCLUSIONS_FILE } = require('./utils/constants');
+const { LOCALE_IDS } = require('./utils/constants');
+const { getExclusions } = require('./utils/helpers');
 
 const STATUS = {
   PENDING: 'PENDING',
@@ -33,20 +33,14 @@ const humanTranslatedProjectID = process.env.HUMAN_TRANSLATION_PROJECT_ID;
 const machineTranslatedProjectID = process.env.MACHINE_TRANSLATION_PROJECT_ID;
 
 /**
- * Loads the Exclusions yaml file and converts it to a JSON object.
- * @returns {Object} The Exclusions yaml file as a JSON object.
- */
-const getExclusions = () => {
-  return yaml.load(fs.readFileSync(path.join(process.cwd(), EXCLUSIONS_FILE)));
-};
-
-/**
  * Determines if a file should be included based on data from an exclusions file
  * @param {Object[]} fileData The files to check
  * @param {Object[]} exclusions The exclusions file
  * @returns {Object[]} The files that should be included
  */
-const excludeFiles = (fileData, exclusions) => {
+const excludeFiles = (fileData) => {
+  const exclusions = getExclusions();
+
   return fileData.filter(({ filename, locale, contentType }) => {
     const localeKey = Object.keys(LOCALE_IDS).find(
       (localeKey) => LOCALE_IDS[localeKey] === locale
@@ -54,7 +48,8 @@ const excludeFiles = (fileData, exclusions) => {
     return (
       !exclusions.excludePath[localeKey]?.some((path) =>
         filename.startsWith(path)
-      ) && !exclusions.excludeType[localeKey]?.some((type) => contentType === type)
+      ) &&
+      !exclusions.excludeType[localeKey]?.some((type) => contentType === type)
     );
   });
 };
@@ -105,9 +100,9 @@ const main = async () => {
   const changedMdxFileData = prFileData
     .filter((file) => path.extname(file.filename) === '.mdx')
     .filter((f) => f.status !== 'removed');
-  const exclusions = await getExclusions();
+
   const allLocalizedFileData = changedMdxFileData.flatMap(getLocalizedFileData);
-  const includedFiles = excludeFiles(allLocalizedFileData, exclusions);
+  const includedFiles = excludeFiles(allLocalizedFileData);
 
   const fileDataToAddToQueue = translationDifference(queue, includedFiles);
 
--- a/scripts/actions/add-files-to-translation-queue.js
+++ b/scripts/actions/add-files-to-translation-queue.js
@@ -47,12 +47,16 @@ const getExclusions = () => {
  * @returns {Object[]} The files that should be included
  */
 const excludeFiles = (fileData, exclusions) => {
-  return fileData.filter(
-    ({ filename, locale, contentType }) =>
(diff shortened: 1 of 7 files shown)


scripts/actions/add-files-to-translation-queue.js
```diff
@@ -35,6 +32,29 @@ const slugIntersection = (pendingFiles, filesToRemove) =>
 const humanTranslatedProjectID = process.env.HUMAN_TRANSLATION_PROJECT_ID;
 const machineTranslatedProjectID = process.env.MACHINE_TRANSLATION_PROJECT_ID;
 
+/**
+ * Loads the Exclusions yaml file and converts it to a JSON object.
+ * @returns {Object} The Exclusions yaml file as a JSON object.
+ */
+const getExclusions = () => {
+  return yaml.load(fs.readFileSync(path.join(process.cwd(), EXCLUSIONS_FILE)));
+};
```
nr-kkenney, line 41: i think this can just be moved into excludeFiles, or still be kept as a separate method but called directly from excludeFiles.

additionally, i think its cleaner & would allow for us to simplify the signature of excludeFiles, since we wouldn't need to pass `exclusions` parameter anymore.

scripts/actions/add-files-to-translation-queue.js
```diff
@@ -35,6 +32,29 @@ const slugIntersection = (pendingFiles, filesToRemove) =>
...
+/**
+ * Determines if a file should be included based on data from an exclusions file
+ * @param {Object[]} fileData The files to check
+ * @param {Object[]} exclusions The exclusions file
+ * @returns {Object[]} The files that should be included
+ */
+const excludeFiles = (fileData, exclusions) => {
+  return fileData.filter(
+    ({ filename, locale, fileType }) =>
+      !exclusions.excludePath[locale]?.some((exclus) =>
```
nr-kkenney, line 52: i would prefer a better name to exclus. usually the temp name of variables for anonymous functions is not something i nitpick, but this left me scratching my head a bit 🙈 

maybe something like `excludedPath`, `excludedDir` or something similar.

scripts/actions/add-files-to-translation-queue.js
```diff
@@ -35,6 +32,29 @@ const slugIntersection = (pendingFiles, filesToRemove) =>
...
+        filename.includes(exclus)
+      ) &&
+      !exclusions.excludeType[locale]?.some((exclus) => fileType === exclus)
```
nr-kkenney, line 55: same comment here as above

scripts/actions/add-files-to-translation-queue.js
```diff
@@ -52,10 +72,11 @@ const getLocalizedFileData = (prFile) => {
   const contents = fs.readFileSync(path.join(process.cwd(), prFile.filename));
   const { data } = frontmatter(contents);
   const checkLocale = getProjectId(data.translate);
+  const fileType = data.type;
```
nr-kkenney, line 75: this is just a musing that i have, not necessarily something to address since i think this would be a bit of effort. so you can just leave it 🙈 

we should rename this to something like `contentType` or `pageType` since file type has a meaning already and we are also using it to represent the team defined type in graphql.

scripts/actions/utils/docs-content-tools/i18n-exclusions.yml
```diff
@@ -0,0 +1,6 @@
+excludePath:
+  ja-JP:
```
aswanson-nr, line 2: I think we may want to stick with using `jp` to represent the `ja-JP` localization, since that's what writers will be using when they mark a page for human translation. 
```suggestion
  jp:
```

scripts/actions/utils/docs-content-tools/i18n-exclusions.yml
```diff
@@ -0,0 +1,6 @@
...
+    - /whats-new
+excludeType:
+  ja-JP:
```
aswanson-nr, line 5: Same comment as the `excludePath`

aswanson-nr (commented):
Looks great overall! I just think I found one small bug!

scripts/actions/add-files-to-translation-queue.js
rudouglas: The only trade-off with that is we would be running `yaml.load` and `fs.readFileSync`  for every file, would it not make more sense to only perform that action once?

scripts/actions/add-files-to-translation-queue.js
rudouglas: It's actually latin for exclude so 🤷 😂 excellent point

scripts/actions/add-files-to-translation-queue.js
rudouglas: EXCLUSIOR

scripts/actions/add-files-to-translation-queue.js
rudouglas: cool ye that makes sense

scripts/actions/add-files-to-translation-queue.js
```diff
@@ -35,6 +32,29 @@ const slugIntersection = (pendingFiles, filesToRemove) =>
...
+ * Determines if a file should be included based on data from an exclusions file
+ * @param {Object[]} fileData The files to check
+ * @param {Object[]} exclusions The exclusions file
+ * @returns {Object[]} The files that should be included
+ */
+const excludeFiles = (fileData, exclusions) => {
+  return fileData.filter(
+    ({ filename, locale, fileType }) =>
+      !exclusions.excludePath[locale]?.some((exclus) =>
+        filename.includes(exclus)
```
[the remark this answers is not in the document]
rudouglas: cool ye i think `startsWith` makes the most sense and just tell the content team to be explicit with their paths

scripts/actions/utils/docs-content-tools/i18n-exclusions.yml
rudouglas: I think i just changed that because that's how it was coming through in the object, but it makes more sense to keep it consistent with the way they do it now

scripts/actions/add-files-to-translation-queue.js
clarkmcadoo: if you run it like this, would it run more that once?:

`const excludeFiles = (fileData) => {
  const exclusions = getExclusions();

  return fileData.filter(({ filename, locale, contentType }) => {
    const localeKey = Object.keys(LOCALE_IDS).find(
      (localeKey) => LOCALE_IDS[localeKey] === locale
    );
    return (
      !exclusions.excludePath[localeKey]?.some((path) =>
        filename.startsWith(path)
      ) && !exclusions.excludeType[localeKey]?.some((type) => contentType === type)
    );
  });
};`

rudouglas: ahhh i misunderstood what you meant Kris, that does make complete and utter total sense

scripts/actions/utils/docs-content-tools/i18n-exclusions.yml
aswanson-nr: These are still in their `ja-JP` form. If I was overridden, feel free to ignore this comment.

aswanson-nr (approved):
Overall looks great! Just noticed one thing in the exclusions file

PR #5077, opened 2021-11-30
Remove disallowed components from sample alert queries

philweber:
NRQL alert conditions may not contain SINCE or UNTIL


urbiz-nr (approved):
Thanks for fixing the query, @philweber!

PR #5087, opened 2021-11-30
fix: add home page left nav info

bradleycamacho:
Add a section to our style guide on editing the home page left nav.


rhetoric101 (approved):
@bradleycamacho, thanks for adding this new section!

PR #5101, opened 2021-12-01
Update example (linux) Dockerfiles to use current .NET versions

nr-ahemsath:
The current Dockerfile examples for using the .NET agent in Linux reference `.NET Core 2.2`, which is pretty old and has been unsupported by Microsoft for a while now.  This PR updates the examples to use `.NET 6.0` which is the latest LTS version of the platform.  It also adds some comments to help our customers find the right container base images appropriate for their use case.

I tested these examples on my system and everything worked as expected.


urbiz-nr (approved):
Thanks for the updates, @nr-ahemsath!

PR #5102, opened 2021-12-01
Remove all references to OpenCensus per prod management

rhetoric101:
We no longer support OpenCensus, so I removed all references to it.


urbiz-nr (approved):
Looks good!

PR #5114, opened 2021-12-01
Small update with example for Varnish 6

carlossscastro:
Update for the Varnish integration with an example the new Varnish 6 version which requires some adjustments on the configuration


zuluecho9 (approved):

zuluecho9:
thanks @carlossscastro !

Issue #4204, opened 2021-10-06
[Machine Translation] Modify scripts to Download translations


roadlittledawn:
## Summary
><!> Please use the https://github.com/newrelic/docs-website/tree/feature/machine-translation feature branch

Similar to #2536 the interaction with the Smartling API is identical for both Machine/Human Translation. 
The deserializing method should also not change apart from the additional considerations.

In the current script called in the workflow [check-job-progress.js](https://github.com/newrelic/docs-website/blob/feature/docs-upgrades/scripts/actions/check-job-progress.js) we are picking up the `projectId` as an environment variable, so as we are creating a separate workflow for Machine Translation we can set the env variable there. This way we should just need to add a flag when running the script to tell it how to deserialize the data. Something like:

```bash
node ./scripts/actions/check-job-progress.js --machine
```

This will need to be passed through to...

### Fetching

We currently fetch translations using the $PROJECT_ID, so this shouldn't need to change in [fetch-and-deserialize.js](https://github.com/newrelic/docs-website/blob/feature/docs-upgrades/scripts/actions/fetch-and-deserialize.js). We will just need to add logic for when we pass through the `--machine` flag from `check-job-progress.js` when...

### Deserializing

When we fetch the completed translation doc, we will need to add `frontmatter` to each translated doc to signify that it has been Machine/Human Translated. Using the `--machine` flag we will be able to tell which frontmatter to add. (This will be needed to determine whether to show the Disclaimer as part of https://github.com/newrelic/docs-website/issues/2537)

E.G.
```yaml
translationType: machine/human
```

Currently, once we download a translated doc we then strip the `translate` key:value out of the frontmatter so the method for editing the frontmatter would be largely the same, ([see here](https://github.com/newrelic/docs-website/blob/develop/scripts/actions/deserialize-html.js))

### Accounting for `project_id`

The functionality for updating the tables with completed job status should not change, the `project_id` should already be in the tables and we are just updating the status column. But it would be good to double/triple check that.

For Downloading and Deserliazing, the relevant scripts are:
 - [check-job-progress.js](https://github.com/newrelic/docs-website/blob/feature/docs-upgrades/scripts/actions/check-job-progress.js)
 - [fetch-and-deserialize.js](https://github.com/newrelic/docs-website/blob/feature/docs-upgrades/scripts/actions/fetch-and-deserialize.js)
[...]

PR #5123, opened 2021-12-01
Download translation script updates

nr-kkenney:
## Summary

* For a run, only check one project at a time.  This means we run the script once for human translations, and once for machine translations
* add environment variable which indicates translation type when running script
* created new configuration module to be central place for configuration values for scripts. This represents a single place for setting & getting configuration values. We can also write JSDOCs for such pieces of code. I think this is a future facing improvement versus just using process.env.some_value.
* use new configuration module for type to add frontmatter when deserializing content.

## Links
Resolves: #4204


roadlittledawn (approved):
LGTM.

PR #5141, opened 2021-12-02
style(Serverless): Cleaned up some language, made it less mean

paperclypse:
This came in as a hero request


paperclypse:
@bradleycamacho The original requester wants to delete the last paragraph. What do you think?

bradleycamacho:
> @bradleycamacho The original requester wants to delete the last paragraph. What do you think?

I like the way you framed it. Letting the customer know that it can lead to higher ingest costs is valid as long as we don’t criticize cloudwatch itself. Depending on what @MattWhelan thinks, we can either use this new language or just remove the entire paragraph.

paperclypse:
I've deleted the last paragraph, @bradleycamacho. We can always add it back easily enough.

PR #5142, opened 2021-12-02
Update japanese homepage text

LizBaker:
This is an updated translation of the re-ordered/updated homepage text


src/i18n/translations/jp/translation.json
```diff
@@ -1,12 +1,18 @@
 {
   "404": {
     "statusCode": "404",
-    "errorMessage": "入力したURLは間違っているか、ページが削除された可能性があります。<2>ホームに戻る</2>"
+    "errorMessage": "入力したURLは間違っているか、ページが削除された可能性があります。&lt;2&gt;ホームに戻る&lt;/2&gt;"
```
roadlittledawn, line 4: we don't want the brackets escaped do we?

LizBaker: oh my goodness, nope, i'll fix those up

roadlittledawn: ah okay. looks like there's a couple instances of that happening

src/i18n/translations/jp/translation.json
```diff
@@ -16,29 +22,29 @@
     "intro": {
       "p1": "初めてご使用の場合、これら3つの手順に従い、アカウントを作成してください。（無料です）",
       "p2": "New Relic Oneの変更に対応する場合は、この<2>移行ガイド</2>を最初にご覧になるか、または<4>最新情報</4>をご確認ください。",
-      "p3": "テレメトリーデータプラットフォームやフルスタックオブザーバビリティ、アプライドインテリジェンスの詳細にスクロールします。または、プラットフォームの機能概要を把握したいなら、<2>New Relicの概要</2>を読み、<5>ソリューションとベストプラクティスガイド</5>を使用してください。"
+      "p3": "New Relicへのデータの取得、当社のプラットフォーム機能、オブザーバビリティソリューション、およびアラートツールの詳細については、スクロールしてご覧ください。あるいは、プラットフォームの機能の概要を把握するには、<2>New Relicの概要</2>を読み、<5>ガイドとベストプラクティス</5>を参照してください。"
```
roadlittledawn, line 25: technically i think we can remove the `p2` and `p3` fields from en and jp translation files since we don't use them anymore.

roadlittledawn (dismissed):
LGTM. though the preview for this seemingly vanished before i could check default related content ¯\_(ツ)_/¯

src/i18n/translations/jp/translation.json
LizBaker: good point, just removed those so we don't forget

roadlittledawn (approved):
:shipit:

PR #5143, opened 2021-12-02
fix(Docs homepage): Updated NRQL chart link

paperclypse:
The NRQL link in the popular content section was pointing to a less useful doc. It should be more useful now.

This came in as a hero request.


barbnewrelic (approved):
This looks fine in Gatsby build, thanks for fixing this!

PR #5146, opened 2021-12-02
call out for ingest increase

XiXiaPdx:
We had an internal customer issue. They were surprised at teh increase in ingest due to auto application naming. 

We discussed this as a team and decided it would be good to call this out to the customer. 

Can you help with creating the best call out possible? 

auto app naming will generate multiple apps for their one application. Each app generated would increase ingest.


bradleycamacho:
Hey there @XiXiaPdx, I made a suggested edit. Could you take a look and let me know what you think? Thanks!

XiXiaPdx:
> Hey there @XiXiaPdx, I made a suggested edit. Could you take a look and let me know what you think? Thanks!

:shipit:  🔥

Issue #5030, opened 2021-11-23
Improve local translation DB development experience


zstix:
## Description
At the moment, we have some [documentation and helper scripts](https://github.com/newrelic/docs-website/tree/develop/scripts/actions/translation_workflow/testing) to help engineers develop and test the translation workflow database. There is an opportunity to streamline this process a little and improve the development experience. Some potential improvements include:

* Configuring the `postgres` and `pgadmin` containers via [docker-compose](https://docs.docker.com/compose/)
* Adding a startup script (e.g. `yarn start:db`) that calls the docker-compose file
* Creating some [database migrations](https://sequelize.org/master/manual/migrations.html) rather than relying on copy / pasting SQL to bootstrap a database

zstix:
Just adding this here to get the wheels turning. We'll need to decide on the scope and write out some acceptance criteria.

nr-kkenney:
ive kinda been working on this in secret during #4730 🙈 

its pretty much done, at least a first attempt. will try

PR #5149, opened 2021-12-02
improvements for local development

nr-kkenney:
## Summary
* add docker-compose file for easier creation of local resources
* created database migration script & dockerfile which is used for running it with compose
* added yarn scripts for local setup & cleanup
* update readme
* removed no longer needed sql setup script -- migration script handles this for local development

## Examples / Usage
Now, from the root of the repo you can run `yarn db:start` and that will invoke docker-compose and create a postgres, pgadmin, and migration container. In about ~15 seconds, the database will be fully setup and allow you to begin testing with it.

You would see output similar to:
Screen Shot 2021-12-02 at 2 18 51 PM

If you want to cleanup an environment (and not reuse it later) you can run `yarn db:clean` and that will cleanup the docker resources.

## Links
Resolves: #5030


LizBaker (approved):
hooray for docker!

PR #5155, opened 2021-12-03
fix: update notes on querying Prometheus histogram data

jgals:
This PR updates our documentation on viewing and querying Prometheus histogram data to reflect the changes to the Prometheus OpenMetrics Integration (POMI) that landed in version 2.0.0 last July: https://github.com/newrelic/nri-prometheus/commit/66853c1551f97007b5a4e062fbaf2afe50d339ea.

Initially, I had attempted to update the docs to cover the behavior of POMI versions 1.2.0 through 1.5.0 versus 2.0.0 and later, but it became difficult to present all of the cases clearly. Given the time since this change in behavior, it seemed wiser to simplify the docs to reflect the behavior of only 2.0.0 and later, which matches how one would query Prometheus histogram data received via remote write.

cc @jsgithub1


urbiz-nr (approved):
Thanks for the updates, @jgals!

PR #5158, opened 2021-12-03
Clarify that logs in context isn't available in APM UI for PHP

rhetoric101:
We had a contributor mention that logs in context isn't available for PHP apps directly in the Logs tab of APM. It is only supported in the distributed tracing pages.


barbnewrelic (approved):
Thanks for allowing me to be so picky about the presentation of text. Looks great, feel free to merge anytime!

PR #5174, opened 2021-12-06
fix(infra): remove EAP callout

bradleycamacho:
Removes the EAP callout as request from PM. Pinging @urbiz-nr just so she's aware of this change.


barbnewrelic:
@bradleycamacho  I'll review after all the checks are done.

barbnewrelic (approved):
Your edit is fine. In future I recommend spelling out "early access program" in case readers don't immediately know what EAP means. :-)

PR #5186, opened 2021-12-07
Update actionText data dictionary entry

aubreymasten:
## Give us some context

This PR updates the data dictionary to better reflect the behavior of the actionText attribute. actionText is added to BrowserInteraction events and represents either the text of the element that was clicked to trigger that interaction, or the value set using the [actionText API](https://docs.newrelic.com/docs/browser/new-relic-browser/browser-agent-spa-api/actiontext-browser-spa-api/) call. I'm unclear on the exact historical context of when this behavior was introduced but for at least the last few years the actionText attribute has been present not just on the BrowserInteraction event, but all events that are considered to be a part of the interaction. Those events could be AjaxRequests, JavaScriptErrors or BrowserTiming events.

### DIFF
--- a/src/data-dictionary/events/BrowserInteraction/actionText.md
+++ b/src/data-dictionary/events/BrowserInteraction/actionText.md
@@ -3,6 +3,9 @@ name: actionText
 type: attribute
 events:
   - BrowserInteraction
+  - AjaxRequest
+  - JavaScriptError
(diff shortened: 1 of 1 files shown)


zuluecho9:
thanks @aubreymasten !

zuluecho9 (approved):

PR #5194, opened 2021-12-07
Changed wording

iazam6:
Legal team has decided every time we mention our product, we call it "ML Model Performance Monitoring". I have made this change here.

Note from @zuluecho9: we have convo in slack about this; may be a bigger question and may require a larger sweep/edit.


barbnewrelic (approved):
Thanks for making this edit! I also did a docs site search for other references to "ML" as zuluecho recommended but couldn't find any other that were specifically about the New Relic ML Model Performance Monitoring.

PR #5213, opened 2021-12-07
DoNotTranslate Component

### LINKED ISSUE #4203: [Component] Create DoNotTranslate component for content authors
Content authors want to be able to flag certain pieces / sections of content that should not be translated when we send it. 

You can add a class to an HTML element to tell Smartling to exclude it from translation:

```
<p class="notranslate">This paragraph won't be captured for translation</p>

<p>You have<span class="notranslate">5</span>items in cart.</p>
```

## Acceptance criteria
- [x] Verify with Julien this approach will work from his / Smartling's perspective
- [x] A new component can be used in .mdx files that outputs relevant element with the `notranslate` class
- [x] When we serialize MDX to HTML, children of the component are wrapped in relevant element with the `notranslate` class
- [x] When we deserialize HTML to MDX, it's converted to expected JSX syntax
- [x] After deserialization the convert MDX passes the `verify-mdx` script

## Helpful links
https://help.smartling.com/hc/en-us/articles/360000322914

josephgregoryii:
Component allows serialization for Smartling to ignore children inside component by adding prop `className="notranslate"`. 

closes #4203

### DIFF
--- a/scripts/actions/__tests__/deserialize-html.test.js
+++ b/scripts/actions/__tests__/deserialize-html.test.js
@@ -2,13 +2,11 @@ import deserializeHTML from '../deserialize-html';
 import serializeMDX from '../serialize-mdx';
 import fs from 'fs';
 
-test.only('deserializes DoNotTranslate', async () => {
+test('deserializes mdx with DoNotTranslate', async () => {
   const input = `
-  <DoNotTranslate>
-    <div>
-      Not all who wander are lost...
-    </div>
-  </DoNotTranslate>
+<DoNotTranslate>
+  Not all who wander are lost...
+</DoNotTranslate>
 `;
 
   const mdx = await deserializeHTML(await serializeMDX(input));
--- a/scripts/actions/__tests__/deserialize-html.test.js
+++ b/scripts/actions/__tests__/deserialize-html.test.js
@@ -2,9 +2,9 @@ import deserializeHTML from '../deserialize-html';
 import serializeMDX from '../serialize-mdx';
 import fs from 'fs';
 
-test('deserializes DoNotTranslate', async () => {
+test.only('deserializes DoNotTranslate', async () => {
   const input = `
-  <DoNotTranslate>
+  <DoNotTranslate className="notranslate">
     Not all who wander are lost...
   </DoNotTranslate>
 `;
@@ -43,7 +43,9 @@ This is an MDX doc
 test('deserializes simple components', async () => {
   const input = `
 <Button to="/docs/agents/ruby-sdk">
-  Go to Ruby docs
+  <Button to="/joe/clark">
+    Go to Ruby docs
+  </Button>
 </Button>
   `;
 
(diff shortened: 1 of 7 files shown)


scripts/actions/__tests__/deserialize-html.test.js
```diff
@@ -31,7 +43,9 @@ This is an MDX doc
 test('deserializes simple components', async () => {
   const input = `
 <Button to="/docs/agents/ruby-sdk">
-  Go to Ruby docs
+  <Button to="/joe/clark">
```
nr-kkenney, line 46: this particular change needed?

scripts/actions/__tests__/deserialize-html.test.js
```diff
@@ -31,7 +43,9 @@ This is an MDX doc
...
+    Go to Ruby docs
+  </Button>
```
nr-kkenney, line 48: this particular change needed?

roadlittledawn (commented):
Should we add two tests for `DoNotTranslate`, one for inline and block (that wraps other components and what not)?

scripts/actions/utils/serialization-helpers.js
```diff
@@ -42,13 +42,19 @@ const serializeTextProp = (h, node, propName) => {
...
+  );
+
+  return result;
 };
 
-const serializeJSValue = (value) =>
-  Buffer.from(JSON.stringify(value)).toString('base64');
+const serializeJSValue = (value) => {
+  return Buffer.from(JSON.stringify(value)).toString('base64');
+};
```
aswanson-nr, line 57: let's move this back to an implicit return. Just to keep things snazzy

scripts/actions/utils/serialization-helpers.js
```diff
@@ -98,10 +106,13 @@ const serializeComponent = (
...
   );
+
+  return result;
 };
 
-const getComponentName = (node) =>
-  node.name === null ? 'React.Fragment' : node.name;
+const getComponentName = (node) => {
+  return node.name === null ? 'React.Fragment' : node.name;
+};
```
aswanson-nr, line 115: Let's move this back to an implicit return as well. For the same reasons

aswanson-nr (commented):
Looks great! Would just love to see some more tests around serializing/deserializing:
* Around a code block
* Around a collapser
* Around 2 or more top level elements, such as the example below
```jsx
<DoNotTranslate>
  <h1>Blah</h1>
  <span>hello</span>
</DoNotTranslate>
```
* A deeply nested tree (3 or more level)

scripts/actions/__tests__/deserialize-html.test.js
josephgregoryii: nope 🙃 i can revert that

clarkmcadoo:
@aswanson-nr Not super worried about the codeblocks, they get `base64` encoded so they'll have no translations anyway.  But, I def got you on the Collapser and multiple html items in the same block 👍

rudouglas (approved):
👁️ :godmode: 🥈 👨‍🎨

PR #5216, opened 2021-12-08
Update configure-kubernetes-proxy.mdx

maxlemieux:
Add description of proxy settings for Pixie integration component of the nri-bundle chart.

Fix spelling of "Kubernetes".

Improve section headers, to clarify that each section of the docs relates specifically to the use of proxies with these integrations.


barbnewrelic (approved):
Thanks for adding this useful new content!

PR #5247, opened 2021-12-10
Adding one missing explicit delay for AWS ELB

josemore:
## Give us some context
* We detected the AWS ELB service also has an explicit delay of 5' which was not documented. Added.


x8a (approved):
Thanks for the addition!

PR #5254, opened 2021-12-10
chore: Java agent release notes 6.5.1 and 7.4.1

meiao:
## Give us some context

Documentation on the Java agent releases that fix the security vulnerability:
https://www.cve.org/CVERecord?id=CVE-2021-44228


paperclypse (approved):
Looks good!

PR #5268, opened 2021-12-13
chore: Removing time from release notes as it breaks date parsing for…

meiao:
… an internal system

## Give us some context
Times were added to the releaseDate field to have the releases show in a particular order.

Though this broke an internal system that expected to have only the date there.

So the time is being removed.


x8a (approved):
Thanks!

PR #5269, opened 2021-12-13
fix(Accounts): Fixed broken nav link

paperclypse:
The data center link in the left nav was broken. A bad path in the YAML file.


barbnewrelic (approved):
As long as you added the correct redirect to the original filename and checked your build that the change redirects to the new category, I'm assuming this is correct.

PR #5278, opened 2021-12-13
chore: Updates supported versions of .NET Agent NServiceBus

jaffinito:
## Give us some context

Note: We will not be releasing the agent till after the holiday break (like most everyone) so this is not in anyway urgent!

Updates our listed support for NServiceBus from 5.0 only to 5.0 or higher on both Framework and Core.

Resolves newrelic/newrelic-dotnet-agent#824


rhetoric101 (approved):
@jaffinito, thanks for making these changes.

PR #5280, opened 2021-12-13
chore: java-agent adding more info on recent release notes

meiao:
## Give us some context
Adding more info on the release notes related to CVE-2021-44228 about supported Java versions and workarounds.


rhetoric101 (approved):
@meiao, thanks for making these updates!

PR #5283, opened 2021-12-14
chore: java-agent - rectifying mitigation procedures

meiao:
## Give us some context
One of the proposed mitigation steps provided by log4j does not work for the Java agent.

Updating the release notes to reflect that.


rhetoric101 (approved):
@meiao, thanks for making these updates!

PR #5292, opened 2021-12-14
Add recommended SLIs for OTel services

helenapm:
We're adding the SLI queries for services that are instrumented with OpenTelemetry spans.


x8a (approved):
Looks great, thanks!

PR #5295, opened 2021-12-14
Update article to include release tag info

planteater:
The warning messages have updated copy that is more generic, to encapsulate the user of either build SHAs or release tags, so I updated the screenshots. Also updated the copy to not be build-SHA specific.

@paperclypse FYI


planteater:
@bradleycamacho - Think we can get this out this week? Thanks!

bradleycamacho:
Sorry for the delay @planteater, merging in now!

PR #5303, opened 2021-12-15
fix: use img html tag with import inside collapser

roadlittledawn:


barbnewrelic:
@roadlittledawn  Thank you for providing this solution! I looked at the Gatsby build, and all but the last 2 images are building successfully. Was this simply a test, or did you want us to finish fixing this doc. Thanks for confirming! cc @paperclypse

paperclypse:
I think I can follow the pattern @roadlittledawn laid out here to fix those last two images, @barbnewrelic

paperclypse:
@roadlittledawn I tried following your lead on this one to fix those remaining two images. I broke something. I'd love your help figuring out what happened here. Thanks!

roadlittledawn:
@paperclypse i'm not seeing a file for this `./images/NRQL-example-3.png`

src/content/docs/infrastructure/prometheus-integrations/view-query-data/translate-promql-queries-nrql.mdx
```diff
@@ -12,10 +12,21 @@ redirects:
   - /docs/integrations/prometheus-integrations/view-query-data/simulate-promql-nrql
 ---
 
+import promqlImgSrc from './images/PROMQL-query-2.png';
+
+import nrqlExampleSrc from './images/NRQL-example-3.png';
```
roadlittledawn, line 17: i think these should be named:
- NRQL-query-example-4.png
- NRQL-query-example-5.png

paperclypse: There are four images in this doc. We need to import all of them. 

The bigger issue seems to be that I can't seem to add this HTML+React syntax without getting a build error, either locally or on Gatsby cloud. I don't know how to move forward with this one.

Personally, I would prefer to be able to use our standard Markdown image syntax, instead of trying to introduce a new react component-based image syntax for images inside collapsers. From my perspective, this is a bug.

roadlittledawn: i believe the build error was due to wrong image file name is what i was saying

roadlittledawn: this can be used to work for the time being. long term we can try and get this into a sprint, but it's not clear to me why when MDX is parsed into HTML it doesn't use the right src path. i'm actually curious if using the webpack aliased path fixes this (what i brought up a couple days ago to reduce duplicating image files for same image)

paperclypse: Of course... Apparently I've lost my ability to correctly read filenames :D

paperclypse:
Let's see if my last commit did it. I still can't get this working in my local build, but I can't see anything wrong.

roadlittledawn:
my guess is this import needs to match the file name exactly:
```./images/promql-query-2.png```

file name:
```PROMQL-query-2.png```

roadlittledawn:
just pushed that up and will see what happens

roadlittledawn:
ironically, we must insert the right gatsby src path when serving / rendering in NR1 Help XP nerdlet

roadlittledawn:
@paperclypse && @barbnewrelic build succeeded. i'll let y'all review / approve / merge

paperclypse (approved):
Looks good

PR #5320, opened 2021-12-15
chore:create doc that coveres history and deleted docs

bradleycamacho:
This creates a style guide doc that covers accessing doc history and finding deleting files.


src/content/docs/style-guide/writing-docs/writer-workflow/github-history.mdx
```diff
@@ -0,0 +1,92 @@
...
+<CollapserGroup>
+  <Collapser
+    className="freq-link"
+    id="pr"
+    title="Finding the deletion PR"
+  >
+
+The easiest way to find a deleted file is to find the pull request that deleted it. There are multiple ways to accomplish this; here are some ideas:
+
+1. Use GitHub's search feature to search for relevany keywords. You can filter search results to only include PRs.
```
barbnewrelic, line 62: relevany :-)

barbnewrelic (commented):
This is fine. What I would like to know is, if I don't know what the filename is or even when it was deleted, how do I find it? In this case is it best to search Jira and hope that someone listed the PR that deleted it? Or copied the markdown and attached it to the Jira? That's what we used to do in Drupal.

My questions may be more useful for me than others because, for example, with EOL projects, I need to backtrack all over the place for what was documented when, and when was it deleted or replaced by something else. In these situations I often have no idea what was the original file or where it existed before our kazillion redirects. :-)

bradleycamacho:
@barbnewrelic I would recommend using the second approach in that case. As long as you know what the doc was or where it lived in the taxonomy, you can just keep going through release PRs until you find it.

PR #5325, opened 2021-12-15
Feat/machine translation script

clarkmcadoo:
Updated `add-files-to-translate` script to use flags to denote whether a `<url>` or `<directory>`.  Can also add a boolean flag `<machine-translate>` to denote using the script to add only files to be machine translated to the translations db.

*Example*

`yarn add-files-to-translate -d src/content/docs -mt`

### DIFF
--- a/scripts/actions/add-files-to-translation-queue.js
+++ b/scripts/actions/add-files-to-translation-queue.js
@@ -41,8 +41,8 @@ const translationDifference = (pendingFiles, prChanges) =>
       )
   );
 
-const humanTranslatedProjectID = 'HT_ID';
-const machineTranslatedProjectID = 'MT_ID';
+const humanTranslatedProjectID = process.env.HUMAN_TRANSLATION_PROJECT_ID;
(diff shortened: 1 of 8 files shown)


scripts/actions/add-files-to-translation-queue.js
```diff
@@ -26,11 +41,8 @@ const translationDifference = (pendingFiles, prChanges) =>
       )
   );
 
-const slugIntersection = (pendingFiles, filesToRemove) =>
-  pendingFiles.filter((file) => filesToRemove.includes(file.slug));
-
-const humanTranslatedProjectID = process.env.HUMAN_TRANSLATION_PROJECT_ID;
-const machineTranslatedProjectID = process.env.MACHINE_TRANSLATION_PROJECT_ID;
+const humanTranslatedProjectID = 'HT_ID';
+const machineTranslatedProjectID = 'MT_ID';
```
nr-kkenney, line 45: looks like we accidentally committed this. we will want to revert back to what we had before.

clarkmcadoo: oops

nr-kkenney (approved):

PR #5327, opened 2021-12-15
machine translation feature

nr-kkenney:
feature into develop. same as before, except with additional changes to the addition script, as well as a fix for exclusion.


nr-kkenney:
wrong branch 🙈

PR #5329, opened 2021-12-15
add(nerdgraph): new apm server side config docs

zuluecho9:
Per convo with unified API team and support, want to add detail about how to set server-side config for APM agents. Work done: 

Added a new 'Nerdgraph tutorial: APM config' doc in nerdgraph category, meant to be a place to add more APM examples in future.  
Added to the nerdgraph intro doc's list of tutorials and to nav file 
Added mention of this in server side config doc.


src/content/docs/apis/nerdgraph/examples/apm-config-nerdgraph.mdx
```diff
@@ -0,0 +1,93 @@
...
+      }
+    }
+    errors {
+      description
+    }
+  }
+}
+```
+
+And here's an example of what might be used for the `settings` query variables. Note that this uses an array, which may be helpful if you have a lot of entities you want to update.
```
baurfamily, line 55: These variables are fundamentally linked to the above code. The above code doesn't turn off the setting per se, the values in the settings JSON below is what determines that.

So, maybe (above): "Here is an example to update the value of useServerSideConfig" and (below) "Here is the JSON for the `$settings` variable to turn server side config off using the above query"

zuluecho9: i attempted to combine these.

PR #5331, opened 2021-12-16
docs: clarify the array of status codes are ints

brnhensley:
I had a customer ticket confused weather the codes should be an array of strings or nums, `["404","401"]` or `[404,401]` so I added that info. I also checked [the default value from the `lib/config/default.js`](https://github.com/newrelic/node-newrelic/blob/44b73f2b1fd549894474d1d37ab3b2f2d6e982ad/lib/config/default.js#L289) and it is an array with 404 in it, `[404]`, not just the int `404`. It's minor but I think that being explicit will help some people.

## Give us some context


x8a (approved):
Thanks a lot for specifying this!

PR #5333, opened 2021-12-16
chore: adding point release to the Java compatibility table

meiao:
## Give us some context
Adding the latest point release to the table.
6.5.1 is not compatible with Java 7.


x8a (approved):
Thanks for the addition!

PR #5345, opened 2021-12-16
fix(TDP nav): fixing bad order of intro docs

zuluecho9:
Attempting to fix order of API section of nav file


zuluecho9:
@Broseberg if you can check out the nav file for APIs section I've worked on , thought you'd be a good reviewer. I basically just added a general 'APIs' section and moved up the general API docs right at top of that section.

Broseberg (approved):
Good call, I think this structure make. much more sense.

Issue #5348, opened 2021-12-16
[Machine Translation] MT bug fixes, improvements


nr-kkenney:
## Summary

Our MT workflow ran on schedule and was [partially successful](https://github.com/newrelic/docs-website/runs/4542036697?check_suite_focus=true). At quick glance we encountered two errors:
* no serializer for `<strong>` elements.
* 401 unauthorized error. presumably because the token we used expired (which smartling says expires within 5-10 minutes, and Zack said he noticed it was around 5 minutes)

## Acceptance criteria
- [X] Cancel remaining MT jobs in Smartling
- [x] Clean up translations DB
- [x] Determine if we update auth to account for expiration or implement workaround (like batching via exclusions, etc)
- [x] Implement auth fix
- [x] Add serializer for `<strong>`, `<b>` elements
- [x] Run script to queue up MT files again
- [x] Trigger workflow to send content for translation again
- [x] Pause the workflow that checks smartling jobs and downloads/deserializes files through the holidays

## Helpful links
https://github.com/newrelic/docs-website/runs/4542036697?check_suite_focus=true

nr-kkenney:
we created a single job for ~1400 files awaiting authorization for machine translation. going to call this done.

PR #5353, opened 2021-12-16
5348 cache access token

nr-kkenney:
## Summary
This PR introduces a change to start caching the access token in getAccessToken. Additionally, getting the access token is now almost entirely encapsulated in vendor-request.js.

Now, most of the scripts don't have to even worry about access tokens -- getting and caching the token is handled in one spot. This is reflected with the changes to vendorRequest method signature, and removal of access token code usage.

We push the generation of access tokens deeper into the code, closer to where we actually need to use it. Rather than our high level scripts caring about getting & managing an access token, we leave that concern in our lower level vendor api code.

Now, every time vendorRequest is called, we call getAccessToken internally. getAccessToken will automatically generate & store valid access tokens with the caching change -- an initial request will generate a valid token, then subsequent requests will use the cached token. once the token expires, we will go out and get another one and repeat the cycle.

important to note -- the expiration for the token is set to 4 minutes in an attempt to give us a bit of leeway for using a token and not having it expire while we use it or attempt to use it.

## The Fix
How does this address the auth issues we ran into? 

Code that calls getAccessToken no longer needs to worry about the 5 minute expiration, since the internal caching is 4 minutes & retrieves a new token automatically.

## Links
Relates to: #5348

### DIFF
--- a/scripts/actions/check-job-progress.js
+++ b/scripts/actions/check-job-progress.js
@@ -6,7 +6,7 @@ const {
   getTranslationsJobsRecords,
 } = require('./translation_workflow/database.js');
 
-const { getAccessToken, vendorRequest } = require('./utils/vendor-request');
+const { vendorRequest } = require('./utils/vendor-request');
 const { fetchAndDeserialize } = require('./fetch-and-deserialize');
 const { configuration } = require('./configuration');
 
@@ -44,10 +44,9 @@ const log = (message, level = 'log', indent = 0) => {
 };
 
 /**
- * @param {string} accessToken
  * @returns {(batchUid: string) => Promise<Batch>}
  */
-const getBatchStatus = (accessToken) => async ({ batchUid, jobId }) => {
+const getBatchStatus = async ({ batchUid, jobId }) => {
   log(`Getting status for batch: ${batchUid}`);
 
   try {
@@ -55,7 +54,6 @@ const getBatchStatus = (accessToken) => async ({ batchUid, jobId }) => {
     const batchData = await vendorRequest({
       method: 'GET',
       endpoint: `/job-batches-api/v2/projects/${PROJECT_ID}/batches/${batchUid}`,
-      accessToken,
     });
 
     const { translationJobUid } = batchData;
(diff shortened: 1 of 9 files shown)


scripts/actions/utils/vendor-request.js
```diff
@@ -68,25 +69,35 @@ const makeRequest = async (url, options, nthTry = 1) => {
...
+    console.log('grabbing access token');
+    const { accessToken } = await makeRequest(url, options);
 
-  const { accessToken } = await makeRequest(url, options);
+    console.log('setting cached token');
+    cache.set('access_token', accessToken, 60 * 4);
 
-  return accessToken;
+    return accessToken;
+  }
```
josephgregoryii, line 100: Very clean solution

josephgregoryii (approved):

josephgregoryii:
Looks like there are some tests that are not passing. I think it is due to the length of arguments now that some of the function signatures have changed in some files.

josephgregoryii (dismissed):
Looks like there are some tests that are not passing. I think it is due to the length of arguments now that some of the function signatures have changed in some files.

scripts/actions/check-job-progress.js
```diff
@@ -74,7 +72,6 @@ const getBatchStatus = (accessToken) => async ({ batchUid, jobId }) => {
     const jobData = await vendorRequest({
       method: 'GET',
       endpoint: `/jobs-api/v3/projects/${PROJECT_ID}/jobs/${translationJobUid}`,
```
rudouglas, line 74: Do we not need any authorization for these api calls?

scripts/actions/check-job-progress.js
```diff
@@ -116,10 +113,7 @@ const main = async () => {
     });
 
     // get the status for all of the batch translation jobs
-    const accessToken = await getAccessToken();
-    const batchStatuses = await Promise.all(
-      batchUids.map(getBatchStatus(accessToken))
-    );
+    const batchStatuses = await Promise.all(batchUids.map(getBatchStatus));
```
rudouglas, line 116:
```
    const batchStatuses = test()
```

scripts/actions/check-job-progress.js
nr-kkenney: we do. but now that specific concern is handled entirely inside of vendorRequest.

rudouglas: ahhh right ye that makes sense to me noooow

rudouglas (approved):

PR #5380, opened 2021-12-18
Update configure-kubernetes-proxy.mdx

maxlemieux:
Removed configuration for newrelic-pixie chart proxy settings. Although this works for the newrelic-pixie chart, it turns out that this does not actually propagate to Pixie components, therefore it could give the wrong impression to readers.

True proxy support for Pixie is under consideration at this time, and we'll be able to document it when supported.


x8a (approved):
Thanks!

Issue #5187, opened 2021-12-07
Issue: Use tags to help organize and find your data


qryuu:
## How can we make our docs better?
The number of characters is listed in the tag limit, but is there a limit on the type of characters?

Is it only ASCII characters?
Can I use symbols and non-English characters?

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/new-relic-one/use-new-relic-one/core-concepts/use-tags-help-organize-find-your-data/
* **OS**: Unknown
* **Browser**: Unknown
* **Device**: Unknown

zuluecho9:
@qryuu thanks. I think we've addressed this with this PR. https://github.com/newrelic/docs-website/pull/5385 Much appreciated! Please let us know if it's still confusing.

PR #5385, opened 2021-12-20
fix(tags): adding allowed characters

zuluecho9:
Based on issue https://github.com/newrelic/docs-website/issues/5187, adding allowed characters in tag doc's requirements.


paperclypse (approved):
Works for me!

PR #5390, opened 2021-12-20
feat(alerts): New decision tree flowchart for aggregation methods

paperclypse:
I've created a flowchart image to provide a little guidance about what aggregation method to use.


bradleycamacho:
A few notes!

- The size of the diagram is a bit large. I would recommend looking into resizing it a bit if possible. If not, that's also OK since it's hidden in a clamshell!
- A big white background looks a bit ugly in dark mode, but that is probably a hard fix so I wouldn't worry about it unless you have a perfect and easy solution.
- `Does your data arrive in batches or intermittently` suggests that the options should be one or the other, not Yes or No. I'm honestly not sure which one refers to which.

Other than those two notes, I think it looks good!

bradleycamacho (approved):

Issue #4232, opened 2021-10-07
[Research] How to handle file path changes in `i18n` counterparts


roadlittledawn:
## Summary

Right now if a PR contains changes that include moving a file from directory to another (which is a status of `removed` from git's perspective), we check to see if the file has counterparts in the `src/i18n` language dirs and if it does, we throw a warning as part of the `validate-pr` workflow check.

We haven't really run into this case yet since that check was put in place. But we will. We should figure out how we want to handle these changes when the check fails.

## Outstanding questions
- Is this warning sufficient to kick off a human process for PR submitter / tech docs teammember to make equivalent change?
- Should a writer move/remove equivalent `i18n` file at all?
- Is it okay from a translation workflow perspective to manually change filepaths in `i18n`?

## Examples
This doc was moved and our translation workflow check probably wasn't in place, so when we it was moved and subsequent PRs made, we created a new file at the new path and never removed the old one from i18n.
* https://github.com/newrelic/docs-website/blob/develop/src/i18n/content/jp/docs/accounts/accounts-billing/account-setup/account-id.mdx
* https://github.com/newrelic/docs-website/blob/develop/src/i18n/content/jp/docs/accounts/accounts-billing/account-structure/account-id.mdx

## Acceptance criteria
- [x] Determine if warning is sufficient for tech docs / PR submitter to make equivalent change in `i18n`
- [X] If not, file ticket with some technical specs on how we can do this programmatically 

## Helpful links
https://github.com/newrelic/docs-website/issues/5127

roadlittledawn:
Long term, Julien said it may make sense to create a JSON file that records what files were moved in en / what should be removed in i18n, so he / his TBD teammate can go in and do that in subsequent PRs.

worth noting that since the human translations aren't added to site right away, we don't want to delete i18n files until the new i18n files come back from the file move in english.

roadlittledawn:
Chatted w/ @nr-kkenney about this a bit. Some notes:

Once MT is in place for Japanese, it's much more likely content folks will move files and the workflow check will warn them that equivalent `i18n` files need to be updated.

<details>
<summary>Short term solution</summary>
Content authors apply same file moves to corresponding `i18n` files.

This ensures that when PR is merged to `develop` and relevant files are added to translation queue, the English redirects and URL paths resolve to their Japanese translated equivalents.

[...]

PR #5398, opened 2021-12-21
Update workflow script to check for renamed files and invoke main

roadlittledawn:
## Description

Updates the script behind the workflow step that checks a PR for files that changed and may need to be applied to corresponding i18n files. Updates it to also check for status of `renamed` which indicates a file was moved.

Also updates the script to invoke `main()`, since I think this was never actually run 

## Related issues / PRs
Related to https://github.com/newrelic/docs-website/issues/4232

### DIFF
--- a/scripts/actions/check-for-outdated-translations.js
+++ b/scripts/actions/check-for-outdated-translations.js
@@ -103,6 +103,8 @@ const main = async () => {
   }
 };
 
-main();
+if (require.main === module) {
+  main();
+}
 
-module.exports = { checkOutdatedTranslations };
+module.exports = { main, checkOutdatedTranslations };
--- a/scripts/actions/check-for-outdated-translations.js
+++ b/scripts/actions/check-for-outdated-translations.js
@@ -34,12 +34,22 @@ const checkOutdatedTranslations = async (url) => {
     .reduce((files, file) => {
       const contents = fs.readFileSync(path.join(process.cwd(), file.filename));
       const { data } = frontmatter(contents);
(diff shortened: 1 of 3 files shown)


roadlittledawn:
The actual workflow failed. Not sure why it doesn't like the token / auth value, since this works when i use it directly and the `fetchPaginatedGHResults` works fine when used as part of other workflows

roadlittledawn:
Duh. Just needed to pull the default `GITHUB_TOKEN` from secrets like the other workflows 🤦

aswanson-nr:
I found some old issues around a [similar error](https://github.com/node-fetch/node-fetch/issues/401#issuecomment-565891390) and it appears to come from a newline being present at the end of the token. According to that comment, you can always tell because the error message is on the next line in the terminal output, which we are seeing as well.
However, I'm not quite sure how this could happen to this token value, since we use the same one in multiple places.

scripts/actions/check-for-outdated-translations.js
```diff
@@ -85,4 +103,6 @@ const main = async () => {
   }
 };
 
-module.exports = main;
+main();
```
nr-kkenney, line 106: i think we should wrap this in an if check similar to how we do in other files. otherwise, this executes automatically when you import which i don't think we want.

```js
if (require.main === module) {
  main();
}
```

roadlittledawn: done. and updated test to check for console.log being called 0 times

nr-kkenney (approved):

PR #5400, opened 2021-12-21
fix(APM): Fixes a broken image

paperclypse:
A side effect of moving some docs around.


bradleycamacho (approved):
Looks good in the preview. Approved!

PR #5409, opened 2021-12-22
chore(java-agent): bumping download link to latest Java 7 agent

meiao:
We released a new version of the agent and need to update a link to point to that release.


paperclypse (approved):
Looks good! Thanks!

PR #5419, opened 2021-12-23
chore: note about Azure cost

josemore:
## Give us some context

* GTS feedback: Added a note explaining Azure Monitor costs that need to be considered for Azure integrations.


urbiz-nr (approved):
Thanks for adding this info, @josemore!

PR #5424, opened 2021-12-23
ESA-625 Improved search examples

carl0FF:
In the spirit of better illustrating what can be done with entity search, the page has been reorganized a bit, and added examples that illustrate more capabilities of search


urbiz-nr (approved):
Thanks for the updates, @carl0FF!

urbiz-nr:
Tried locally.

PR #5429, opened 2021-12-24
chore: note about max memory defined by default

josemore:
## Give us some context

* When installing the infrastructure agent on Linux, a max memory limit of 1gb is enforced by service manager settings (systemd, upstart). GTS feedback received to document this default setting so that customers are aware.


urbiz-nr (approved):

PR #5432, opened 2021-12-27
fix(style guid nav): fixing typo

zuluecho9:
Fixing mistake in nav file


barbnewrelic (approved):
that's a great catch!

PR #5433, opened 2021-12-27
Update CPM Helm Repo Link

irishgordo:
* updating helm repo link to parity of instructions in NR1 UI

Resolves: docs/update-cpm-helm-repo-link

## Give us some context

* This just creates parity between the NR1 UI and the docs, as when "Create private location" call-to-action/button is clicked, a name + description have been filled out, and then the "Generate Key" button/call-to-action has been clicked - in the UI under the instructions of `helm repo add` the URL is different between the two, figured that the one in NR1 may be more accurate than the one given in the docs
**however**
I may be totally wrong with this edit 😅  - so feel free to close this if the NR1 UI actually needs it's URL to be updated instead of the docs 😅 
X-Reference Screenshot:
Screen Shot 2021-12-27 at 1 20 37 PM

Please also ignore this prior 'quick' page edit via:
https://github.com/newrelic/docs-website/commit/f842fb9661cfd3320ba51ffe49c9e69c66c57ac0
I had mistaken the commit message for a PR form 😅 😅 

These docs are awesome!  Props to everyone that's been able to make them what they are! 🎸 🚀 😄

### DIFF
--- a/src/content/docs/synthetics/synthetic-monitoring/private-locations/install-containerized-private-minions-cpms.mdx
+++ b/src/content/docs/synthetics/synthetic-monitoring/private-locations/install-containerized-private-minions-cpms.mdx
@@ -394,7 +394,7 @@ To start the CPM, follow the applicable Docker or Kubernetes instructions.
        * If you are copying the charts for the first time:
 
          ```
-         helm repo add <var>YOUR_REPO_NAME</var> https://helm-charts.newrelic.com
+         helm repo add <var>YOUR_REPO_NAME</var> https://helm-charts.newrelic.com/charts
          ```
        * If you previously copied the Helm charts from the New Relic Helm repo, then get the latest:
 


zuluecho9:
@irishgordo thank you! I'm going to assume you're correct on this, as I think we prob would have heard if UI instructions were wrong (more likely than hearing docs instructions were wrong). Thank you! And thanks for kind words.

zuluecho9 (approved):

keegoid-nr:
> @irishgordo thank you! I'm going to assume you're correct on this, as I think we prob would have heard if UI instructions were wrong (more likely than hearing docs instructions were wrong). Thank you! And thanks for kind words.

In this case, the UI instructions are wrong. Need to revert this PR and fix the UI instead.

@cleon-nr says:

> I’ve updated the CPM docs to point to the official NR Helm repo instead of our team’s own Helm repo. We created our own Helm repo as a stopgap to roll out Kubernetes CPM since the NR Helm repo wasn’t ready yet. However, now that we’ve been publishing to https://github.com/newrelic/helm-charts/tree/master/charts/synthetics-minion for at least 6 months, we’d like to phase out the old Helm repo.
> 
> The underlying change is that the redirect going from https://helm-charts.newrelic.com/charts to our team’s own S3 bucket will eventually be shutdown. Therefore, the charts will only be accessible by using https://helm-charts.newrelic.com (note the absence of /charts).

irishgordo:
@zuluecho9 my apologies!  Like I called out in the PR I wasn't entirely sure about this fix.  A revert sounds like a good plan as @keegoid-nr mentioned.

keegoid-nr:
@zuluecho9 This still needs to be reverted. Or should we submit a new PR to undo it?

PR #5434, opened 2021-12-27
docs: change getAgent to GetAgent

brnhensley:
`getAgent` is not the same as `GetAgent`, getAgent would throw an error. I also added c# language identifiers to the code blocks

## Give us some context

### DIFF
--- a/src/content/docs/apm/agents/net-agent/net-agent-api/iagent.mdx
+++ b/src/content/docs/apm/agents/net-agent/net-agent-api/iagent.mdx
@@ -14,7 +14,7 @@ redirects:
 
 ## Syntax
 
-```
+```cs
 public interface IAgent
 ```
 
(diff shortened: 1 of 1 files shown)


zuluecho9:
thanks @brnhensley. Can you explain the 'cs' addition for the syntax highlighting? i've heard that we're occasionally adding syntax highlighting so I think I get that but maybe you can explain a) what's connection between 'cs' and .NET? And do you think sweeping our other .NET agent API docs to add similar would be a good idea?

zuluecho9 (approved):

brnhensley:
Hi zuluecho9, thanks for approving this. I answered this question on slack, but for visibility I'll copy it here

> Can you explain the 'cs' addition for the syntax highlighting? i've heard that we're occasionally adding syntax highlighting so I think I get that but maybe you can explain a) what's connection between 'cs' and .NET? And do you think sweeping our other .NET agent API docs to add similar would be a good idea?

Syntax highlighting uses the file type extension or language name to identify the code language within a code block. .NET is written (primarily) in C#, and C# code is written in `.cs` files, so you can use markdown syntax highlighting for c# code by starting your codeblock with one of these two: 

`````
```cs
```csharp
`````

I had a conversation with Shawn in the #documentation slack channel asking if changes like this were welcome and I was told to go ahead. For now I’ve only been adding the syntax highlighting to docs I already want to make changes to but I would absolutely recommend using it broadly. You can use it for almost any type of code, you just need to learn what the type of file it goes into would be.

If you’d like to see what this does on the live site check out this page- https://docs.newrelic.com/docs/apm/agents/net-agent/net-agent-api/itransaction/#example-4

PR #5438, opened 2021-12-27
fix(Log management): LOGGING-6831: Log patterns is now "best effort"

bmcfeely:

## Give us some context

Currently an NR Integration Error is generated when Log Patterns is short circuited due to a lack of processing time available. This error isn't something customers can do anything about; they don't control the patterns logic. All this does is create confusion with customers, who in turn ask their sales team who then ask Logging about it.

We are going to internally keep track of patterns that are super slow and could be made better, and we should do more in terms of looking at accounts that have unacceptably high levels of Patterns failing. But customers don't need to see this -- pattern matching is a "best effort" process, it's OK if we can't match against the whole model.


zuluecho9:
this looks good to me @bmcfeely, thank you.

zuluecho9 (approved):

PR #5439, opened 2021-12-28
fix(Log management): LOGGING-6833: Grok types

bmcfeely:
We support typed attributes being extracted using Grok, but documentation does not make that clear

### DIFF
--- a/src/content/docs/logs/ui-data/parsing.mdx
+++ b/src/content/docs/logs/ui-data/parsing.mdx
@@ -103,6 +103,96 @@ Here's an overview of how New Relic implements parsing of logs:
   </tbody>
 </table>
 
+## Grok
+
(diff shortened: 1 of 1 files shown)


urbiz-nr (approved):
Thanks for adding this section, @bmcfeely!

urbiz-nr (approved):

aswanson-nr:
Hey @bmcfeely! I dug a bit deeper into why this is failing to build and it has something to do with the `<code>` elements within the table. If you swap those out for backticks, you'll get the same behavior as you should with a `<code>` element.

urbiz-nr:
Tried locally, works.

urbiz-nr (approved):

bmcfeely:
@urbiz-nr Thank you so much! Sorry for not replying earlier.

PR #5449, opened 2021-12-29
kubernetes: clarify <1.16 deprecation message

roobre:
## Give us some context

  * Previous notes stated that v1.16 was deprecated, which is not true.
  * Previous notes instructed to upgrade the integration rather than the cluster.


urbiz-nr (approved):
Thanks for the RN, @roobre!

PR #5483, opened 2022-01-04
fix: updating env var name for AWS SDK

thezackm:
updating the name of the environmental variable for AWS SDK within NPM container.


barbnewrelic (approved):
Thanks for making this update!

PR #5488, opened 2022-01-04
Added callout note for .net infinite tracing.

tehbio:
Added important callout to the .Net Agent Infinite Tracing configuration section about how spans might get limited.


barbnewrelic (approved):
Thanks for updating the config info!

Issue #5485, opened 2022-01-04
[Machine Translation] Most recent download / PR creation workflow failed for MT content


nr-kkenney:
## Summary
Most recent run of the job which PRs machine translated content failed, which can be seen [here](https://github.com/newrelic/docs-website/runs/4701678554?check_suite_focus=true). 

Error:
```none
Error: Invalid or unsupported zip format. No END header found
    at readMainHeader (/home/runner/work/docs-website/docs-website/node_modules/adm-zip/zipFile.js:107:10)
    at new module.exports (/home/runner/work/docs-website/docs-website/node_modules/adm-zip/zipFile.js:23:3)
    at new module.exports (/home/runner/work/docs-website/docs-website/node_modules/adm-zip/adm-zip.js:25:10)
    at fetchAndDeserialize (/home/runner/work/docs-website/docs-website/scripts/actions/fetch-and-deserialize.js:79:15)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Promise.all (index 0)
    at async main (/home/runner/work/docs-website/docs-website/scripts/actions/check-job-progress.js:139:5)
Error: Process completed with exit code 1.
```

## Acceptance Criteria
- [ ] Fix issue so job completes

nr-kkenney:
EOD progress update:

I merged a PR in to fix a couple other awkward things I noticed when attempting to fix the heart of this ticket. For this ticket specifically, Clinton and I now have a pretty good guess as to why this is failing and some next steps.

The request URI we are using to download our files is roughly ~177000 characters long, which is probably not cool for using with fetch -- @roadlittledawn did a quick search and saw a 80kb limit for URI length + request headers -- and why we aren't downloading anything successfully. Additionally, its likely that a number of the files even when downloaded are going to run into deserialization errors -- we saw this for at least 1 file, so that is the next thing to tackle.

But for tomorrow:
* split downloading files into multiple requests rather than all at once
* when deserializing files, don't fail one fail all. do as many as we can, and write out the errors.

nr-kkenney:
checklist of work for sanity:

- [ ] 0a. batch processing of file downloads.

- [ ] 0b. individual processing of file serialization (rather than as batches).

- [ ] 1a. For each file, download, deserialize, set DB record to complete if succesful. If error is encountered (only in deserialization?), set record to errored.

For each file, return its status -- errored or completed, etc. This allows us to aggregate the data.

[...]

PR #5497, opened 2022-01-04
fix: quick updates to fix bugs, suboptimal-ness of workflow

nr-kkenney:
## Summary

On route to tackling #5485 I noticed a couple other awkward things that I thought should be fixed immediately. 

### Problems
Problems I observed were:
1. jobs are being marked as completed too early.
2. the remove/cleanup script is cleaning up state for ALL completed operations, when it should be limited to its specific project (machine or human).

### Fixes
1. I've moved marking a job as completed to after the files have been deserialized, and the translations are marked as completed.
2. I've updated the cleanup script to only cleanup completed state for the specific project of the run.

### DIFF
--- a/scripts/actions/remove-completed-batch.js
+++ b/scripts/actions/remove-completed-batch.js
@@ -6,18 +6,20 @@ const {
 } = require('./translation_workflow/database.js');
 const { getAccessToken, vendorRequest } = require('./utils/vendor-request');
 const fetch = require('node-fetch');
+const { configuration } = require('./configuration');
 
-const PROJECT_ID = process.env.TRANSLATION_VENDOR_PROJECT;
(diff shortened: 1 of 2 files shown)


scripts/actions/remove-completed-batch.js
```diff
@@ -107,4 +109,4 @@ const removePageContext = async (fileUris) => {
   }, 'SUCCESS');
 };
 
-setInProgressToDone();
+setInProgressToDone(PROJECT_ID);
```
nr-kkenney, line 112: i made this an argument just for it to be clearer that this runs for a given project. 

Having it be part of the signature makes it explicit versus just having it be internal to the method.

clarkmcadoo (approved):

josephgregoryii (approved):

PR #5505, opened 2022-01-05
fix(jmx): clarification about jmxterm not bundled

josemore:
## Give us some context
* Updated JMXTerm troubleshooting section since it's no longer bundled with nrjmx.


rhetoric101 (approved):
@josemore, thanks for adding this clarification!

PR #5514, opened 2022-01-05
Update Possible Query Solution for checking otlp endpoint

wildcardlinux:
Adding another query option to possible Solutions troubleshooting:

FROM Log,Metric,Span SELECT count(*) where newrelic.source='%otlp%' SINCE 1 week ago TIMESERIES MAX LIMIT MAX

## Give us some context


rhetoric101:
@wildcardlinux, thanks for contributing this query! We try to put a label on each PR to keep track where the contribution came from. Are you with another company, or are you working at New Relic?

jgals:
Hey @rhetoric101, thank you for raising awareness on this PR internally. After chatting with our OpenTelemetry architect about this query a moment ago, I learned that all ingested OTLP data has `instrumentation.provider` set to `opentelemetry`, so the first query is sufficient, but the following query is also correct:
```
FROM Log, Metric, Span SELECT count(*) where newrelic.source LIKE 'api.%.otlp'
```
@wildcardlinux, I think we should prefer the above query rather than recommending two different queries to try. Would you be open to amending this PR to make that change?

rhetoric101 (approved):
@wildcardlinux, thanks for making these changes!

PR #5518, opened 2022-01-05
DSG for localize pages and release-notes

josephgregoryii:
## Related
https://github.com/newrelic/gatsby-theme-newrelic/pull/623


josephgregoryii:
This is still a WIP. PR created for preview

gatsby-node.js
```diff
@@ -265,11 +265,15 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
...
-        disableSwiftype: !i18nNode,
-      });
+      createPageFromNode(
+        i18nNode || node,
+        {
+          prefix: i18nNode ? '' : locale,
+          createPage,
+          disableSwiftype: !i18nNode,
+        },
+        !!i18nNode // defer if localized
```
roadlittledawn, line 275: i think we'd always want to defer these (localized pages w/english and localized pages with translated content). the check if it's an `i18nNode` in `disableSwiftype` is to ensure we're only indexing pages with translated content in the search engine.

josephgregoryii: This makes sense and clarifies the logic here then. I can fix that 👍🏻

gatsby-node.js
```diff
@@ -426,6 +431,7 @@ const createPageFromNode = (
         fileRelativePath,
         layout: 'basic',
       },
+      defer: defer,
```
roadlittledawn, line 434: i don't think we need to defer these. this is just a way to show a missing template page in local development in the case we can't find a template for a given MDX file type.

josephgregoryii:
Converted to draft to allow discussion around https://github.com/newrelic/gatsby-theme-newrelic/issues/620

gatsby-node.js
```diff
@@ -416,6 +421,7 @@ const createPageFromNode = (
         slugRegex: `${slug}/.+/`,
         disableSwiftype,
       },
+      defer: defer,
```
rudouglas, line 424: Could change to `defer,`

aswanson-nr:
Looks like most of the links on the homepage aren't working in the preview!

rudouglas:
This still needs the theme bump after merging https://github.com/newrelic/gatsby-theme-newrelic/issues/620

josephgregoryii:
Lets wait on merging for now to allow discussion around the fix

roadlittledawn (approved):
:shipit:

PR #5519, opened 2022-01-05
fix(errors-inbox): remove EU callout

bradleycamacho:
Remove EU callout now that the inbox is live in EU.


paperclypse (approved):
Looks good to me!

PR #5555, opened 2022-01-07
fix(user mgmt): updating for account settings UI

zuluecho9:
Small changes related to account settings UI changing: some things that were visible under account dropdown directly are now only found under Administration.


barbnewrelic (approved):
this looks fine. thanks for updating these docs!

barbnewrelic (approved):
This approach to pathing is so much easier to skim!

PR #5557, opened 2022-01-08
Add usage example to function_trace wildcarding docs.

umaannamalai:
## Give us some context

This PR adds a usage example for wildcarding to the function_trace config setting docs.


urbiz-nr (approved):
Thanks for this explanation, @umaannamalai!

urbiz-nr:
Works locally.

PR #5563, opened 2022-01-10
chore(java-agent): update jedis supported version

meiao:
## Give us some context
Jedis released version 4.0.0 a couple of weeks ago and our instrumentation does not apply to the new version.
So capping the supported version.


rhetoric101 (approved):
@meiao thanks for making this version update!

PR #5568, opened 2022-01-10
changes to machine translation for handling partial success, partial failure, batching

nr-kkenney:
## Summary

Key changes:
* files are downloaded in batches of 50. previously this was limitless, and we were running into errors because of this.
  * the batching change fixes the original bug, but I also added in a response check & console log for if we run into responses that aren't HTTP 200s.
* partial success & failure is now possible in the `fetch-content` jobs.
  * files are deserialized individually, and errors are caught and console logged out.
  * translations that download & deserialize correctly will be marked as `COMPLETED` in the DB. these will be subsequently cleaned up in the removal step / script.
  * translations that do not deserialize / throw during deserialization are marked as `ERRORED`. for the moment, these require manual updates to the file and we should treat these as NOT replayable / resubmittable files until fixed.
  * jobs that have all their translations complete are marked as `COMPLETED`. if at least one translation errors, the job is marked as `ERRORED` and we will treat that functionally as completed with errors -- we won't touch this job again, and fixes will be submitted in subsequent jobs.
  * the GitHub workflow will submit a PR if at least one translation is successfully completed, AND will fail if at least one translation errors.
* now, translations that are marked as `ERRORED` will prevent subsequent uploads to our translation vendor (of that particular file, locale) until remedied.
* added in another database helper method for updating translations using any field as input. the existing method uses translation id.

## Links
Resolves: #5485

### DIFF
--- a/.github/workflows/check-translations-and-deserialize.yml
+++ b/.github/workflows/check-translations-and-deserialize.yml
@@ -60,6 +60,7 @@ jobs:
           echo "::set-output name=commit::true"
 
       - name: Create Pull Request
+        id: create-pr
         if: steps.fetch-deserialize.outputs.successfulTranslations != 0
         uses: peter-evans/create-pull-request@v3
         with:
@@ -75,13 +76,13 @@ jobs:
           team-reviewers: developer-enablement
 
       - name: Remove Completed Batches From Queue
-        if: steps.fetch-deserialize.outputs.successfulTranslations != 0
+        if: steps.fetch-deserialize.outputs.successfulTranslations != 0 && steps.create-pr.outcome == 'success'
         id: remove-batches
         run: |
           node ./scripts/actions/remove-completed-batch.js
 
       - name: Fail workflow if at least one translation failed
-        if: steps.fetch-deserialization.outputs.failedTranslations
+        if: steps.fetch-deserialization.outputs.failedTranslations > 0
         run: exit 1
 
   fetch-machine-translated-content:
@@ -130,6 +131,7 @@ jobs:
           echo "::set-output name=commit::true"
 
       - name: Create Pull Request
+        id: create-pr
         if: steps.fetch-deserialize.outputs.successfulTranslations != 0
         uses: peter-evans/create-pull-request@v3
         with:
@@ -145,11 +147,11 @@ jobs:
           team-reviewers: developer-enablement
 
       - name: Remove Completed Batches From Queue
-        if: steps.fetch-deserialize.outputs.successfulTranslations != 0
+        if: steps.fetch-deserialize.outputs.successfulTranslations != 0 && steps.create-pr.outcome == 'success'
         id: remove-batches
         run: |
           node ./scripts/actions/remove-completed-batch.js
 
       - name: Fail workflow if at least one translation failed
-        if: steps.fetch-deserialization.outputs.failedTranslations
+        if: steps.fetch-deserialization.outputs.failedTranslations > 0
         run: exit 1
--- a/.github/workflows/check-translations-and-deserialize.yml
+++ b/.github/workflows/check-translations-and-deserialize.yml
@@ -51,7 +51,7 @@ jobs:
 
       - name: Commit changes
         id: commit-changes
-        if: steps.fetch-deserialize.outputs.batchesToDeserialize != 0
+        if: steps.fetch-deserialize.outputs.successfulTranslations != 0
         run: |
           git config --local user.email "${{ env.BOT_EMAIL }}"
           git config --local user.name "${{ env.BOT_NAME }}"
@@ -60,7 +60,7 @@ jobs:
           echo "::set-output name=commit::true"
 
       - name: Create Pull Request
-        if: steps.fetch-deserialize.outputs.batchesToDeserialize != 0
+        if: steps.fetch-deserialize.outputs.successfulTranslations != 0
         uses: peter-evans/create-pull-request@v3
         with:
           token: ${{ secrets.OPENSOURCE_BOT_TOKEN }}
@@ -75,10 +75,14 @@ jobs:
           team-reviewers: developer-enablement
 
       - name: Remove Completed Batches From Queue
-        if: steps.fetch-deserialize.outputs.batchesToDeserialize != 0
+        if: steps.fetch-deserialize.outputs.successfulTranslations != 0
         id: remove-batches
         run: |
           node ./scripts/actions/remove-completed-batch.js
+
+      - name: Fail workflow if at least one translation failed
+        if: steps.fetch-deserialization.outputs.failedTranslations
+        run: exit 1
 
   fetch-machine-translated-content:
     name: Fetch machine translated content
@@ -117,7 +121,7 @@ jobs:
 
       - name: Commit changes
         id: commit-changes
-        if: steps.fetch-deserialize.outputs.batchesToDeserialize != 0
+        if: steps.fetch-deserialize.outputs.successfulTranslations != 0
         run: |
           git config --local user.email "${{ env.BOT_EMAIL }}"
           git config --local user.name "${{ env.BOT_NAME }}"
@@ -126,7 +130,7 @@ jobs:
           echo "::set-output name=commit::true"
 
       - name: Create Pull Request
-        if: steps.fetch-deserialize.outputs.batchesToDeserialize != 0
+        if: steps.fetch-deserialize.outputs.successfulTranslations != 0
         uses: peter-evans/create-pull-request@v3
         with:
           token: ${{ secrets.OPENSOURCE_BOT_TOKEN }}
@@ -141,7 +145,11 @@ jobs:
           team-reviewers: developer-enablement
 
       - name: Remove Completed Batches From Queue
-        if: steps.fetch-deserialize.outputs.batchesToDeserialize != 0
+        if: steps.fetch-deserialize.outputs.successfulTranslations != 0
         id: remove-batches
         run: |
           node ./scripts/actions/remove-completed-batch.js
+
+      - name: Fail workflow if at least one translation failed
+        if: steps.fetch-deserialization.outputs.failedTranslations
+        run: exit 1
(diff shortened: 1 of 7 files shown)


scripts/actions/fetch-and-deserialize.js
```diff
@@ -53,63 +53,162 @@ const writeFilesSync = (vfiles) => {
...
+        path: newPath,
         extname: '.mdx',
-      }),
-    'utf-8'
+      });
+
+      createDirectories([temp]);
+      writeFilesSync([temp]);
+
+      return { ok: true, slug: `src/content/docs/${path}` };
```
nr-kkenney, line 174: will need to fix this path up. 
* missing `.mdx` extension
* will have double `/`

.github/workflows/check-translations-and-deserialize.yml
```diff
@@ -75,11 +75,15 @@ jobs:
           team-reviewers: developer-enablement
 
       - name: Remove Completed Batches From Queue
-        if: steps.fetch-deserialize.outputs.batchesToDeserialize != 0
+        if: steps.fetch-deserialize.outputs.successfulTranslations != 0
```
zstix, line 78: We probably also want to fail this step if we were unable to create a pull request (the previous step). Otherwise we might risk removing something from the DB and _not_ having the files committed to the repo.

.github/workflows/check-translations-and-deserialize.yml
```diff
@@ -141,7 +145,11 @@ jobs:
...
```
zstix, line 148: Same comment as the human-translated content: let's make sure we made a PR successfully before we clear the database.

scripts/actions/check-job-progress.js
```diff
@@ -100,6 +100,88 @@ const getBatchStatus = async ({ batchUid, jobId }) => {
...
+ * @param {Boolean} slugStatuses[].ok - Boolean flag signaling if translation deserialized. If it did, we can treat this translation as having completed.
+ * @param {String} slugStatuses[].slug - Slug representing file path translated.
+ * @param {String} slugStatuses[].jobId - Job id translation is associated with.
+ */
+const aggregateStatuses = (slugStatuses) => {
+  let totalFailures = 0;
+  let totalSuccesses = 0;
+  const jobStatuses = {};
+
+  slugStatuses.map(({ ok, jobId }) => {
```
zstix, line 114: Super nitpicky: If we're not returning anything, perhaps this could be a `forEach` or [`for..of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of).

zstix: If you really wanted to get fancy, this could all just be a reduce:


```js
const aggregateStatuses = (slugStatuses) =>
  slugStatuses.reduce((acc, status) => {
    const { ok, jobId } = status;
    const {  totalFailures, totalSuccess, jobStatuses } = acc;

    if (!(jobId in jobStatuses)) {
      jobStatuses[jobId] = { successes: 0, failures: 0 };
    }

    if (ok) {
      totalSuccesses += 1;
      jobStatuses[jobId].successes += 1;
    } else {
      totalFailures += 1;
      jobStatuses[jobId].failures += 1;
    }

    return {  totalFailures, totalSuccess, jobStatuses };
  }, { totalFailures: 0, totalSuccess: 0, jobStatuses: {});
```

Probably overkill tho.

scripts/actions/check-job-progress.js
```diff
@@ -100,6 +100,88 @@ const getBatchStatus = async ({ batchUid, jobId }) => {
...
+        { status: updateStatus }
+      );
+
+      console.log(`Translation ${record.id} marked as ${updateStatus}`);
+    })
+  );
+};
+
+/**
+ * @param {Object.<string, Object.<string, number>} jobStatuses - object whose top level keys are job ids, and whose values are { succeses: x, failures: y }, which describe failure and success count for that given job.
```
zstix, line 155: We could consider using a [JSDoc `@typedef`](https://jsdoc.app/tags-typedef.html) to define the shape of these status objects (success & failure counts).

scripts/actions/check-job-progress.js
```diff
@@ -100,6 +100,88 @@ const getBatchStatus = async ({ batchUid, jobId }) => {
...
+ */
+const updateJobRecords = async (jobStatuses) => {
+  await Promise.all(
+    Object.keys(jobStatuses).map(async (jobId) => {
+      const { successes, failures } = jobStatuses[jobId];
+
+      const records = await getTranslationsJobsRecords({
+        job_id: jobId,
```
zstix, line 163: If we call `jobId` `job_id` then this could just be `job_id` 😂

scripts/actions/check-job-progress.js
```diff
@@ -100,6 +100,88 @@ const getBatchStatus = async ({ batchUid, jobId }) => {
...
+      });
+
+      if (successes + failures === records.length) {
+        if (failures > 0) {
+          await updateJob(jobId, { status: 'ERRORED' });
```
zstix, line 168: We could use a constant like so:

```js
const STATUSES = {
  COMPLETED: 'COMPLETED',
  ERRORED: 'ERRORED',
  // ...
};
```

The advantage is that we can use our editors to complete STATUSES.ERRO_ and not rely on spelling stuff (which is always hard.

zstix (commented):
A few comments, I was only able to get through part of this review in the campfire.

.github/workflows/check-translations-and-deserialize.yml
```diff
@@ -75,11 +75,15 @@ jobs:
...
         id: remove-batches
         run: |
           node ./scripts/actions/remove-completed-batch.js
 
+      - name: Fail workflow if at least one translation failed
+        if: steps.fetch-deserialization.outputs.failedTranslations
```
roadlittledawn, line 84: Check for number here?

.github/workflows/check-translations-and-deserialize.yml
```diff
@@ -141,7 +145,11 @@ jobs:
...
 
       - name: Remove Completed Batches From Queue
-        if: steps.fetch-deserialize.outputs.batchesToDeserialize != 0
+        if: steps.fetch-deserialize.outputs.successfulTranslations != 0
         id: remove-batches
         run: |
           node ./scripts/actions/remove-completed-batch.js
+
+      - name: Fail workflow if at least one translation failed
+        if: steps.fetch-deserialization.outputs.failedTranslations
```
roadlittledawn, line 154: Also check for number here?

scripts/actions/check-job-progress.js
```diff
@@ -100,6 +100,88 @@ const getBatchStatus = async ({ batchUid, jobId }) => {
...
+          console.log(`Job ${jobId} marked as ERRORED`);
+        } else {
+          await updateJob(jobId, { status: 'COMPLETED' });
+          console.log(`Job ${jobId} marked as COMPLETED`);
+        }
+      } else {
```
roadlittledawn, line 174: if 100 files were uploaded as part of Job A. and 1 file was canceled, but the job was eventually completed. when we go to check the job status and download the files, we _may_ have a mismatch between the database and smartling data, because we filter out files that had a status of `CANCELED` a la:

https://github.com/newrelic/docs-website/blob/00d3ba65717b68384ffa03fda2045871e9021bb7/scripts/actions/check-job-progress.js#L63

scripts/actions/fetch-and-deserialize.js
```diff
@@ -53,63 +53,173 @@ const writeFilesSync = (vfiles) => {
...
 
-const fetchAndDeserialize = async ({ locale, fileUris }) => {
-  const response = await fetchTranslatedFilesZip(fileUris, locale);
+  // cleanup the last batch
+  if (currentBatch.length != 0) {
+    batches.push({ fileUris: currentBatch });
+  }
 
-  const buffer = await response.buffer();
+  return batches; // [ [{ locale: ja-JP, fileUris: ['hello_world.txt']}], [...], ... ]
```
roadlittledawn, line 80: Kris says this structure isn't accurate anymore. But I really appreciate seeing the shape of what we're working with, since it's easy to loose that context through out

.github/workflows/check-translations-and-deserialize.yml
nr-kkenney: I misspoke when I mentioned this to you yesterday 🙈 

I misunderstood how this would execute, but now I've got that sorted in my head. This step won't run if the previous step fails. We can actually see this in the logs of our currently failing action, so this should be good as is.

when the pr doesn't run, this won't run because they share a condition. if the commit step fails, or the pr step fails, the workflow doesn't reach this step and it won't run.

nr-kkenney: ^ that whole thing is correct, but im wrong again because I forgot what we talked about 🙈 

we wanted to make sure that if we accidentally change the workflow, that the cleanup step is always dependent on translations being completed, and the PR running successfully. we want to avoid the possibility of the PR not running, and cleaning up records which we might introduce with a typo in the condition, etc.

scripts/actions/check-job-progress.js
nr-kkenney: swapped to forEach. i did leave the implementation as is, but am now more convinced of using reduce 🙈

scripts/actions/fetch-and-deserialize.js
```diff
@@ -53,63 +70,175 @@ const writeFilesSync = (vfiles) => {
...
-      console.log(`[*] Deserializing ${path}`);
+/**
+ * @param {String} locale
+ */
+const extractFiles = (locale) => {
+  /**
+   * @param {AdmZip} zip - the downloaded zip containing batch of files.
+   * @returns {HtmlFile[]}
+   */
+  return async (zip) => {
```
aswanson-nr, line 156: Not sure why this is async, all of these zip methods appear to be synchronous

aswanson-nr (approved):

PR #5573, opened 2022-01-11
Add profile_trace API document

lrafeei:
This PR adds a profile_trace API document to the Python Agent docs


rhetoric101:
@lrafeei, thanks for making this contribution! It looks good! I just made a few small changes:

* I added a left navigation link. I noticed that there were some redirects at the top of the file, so was there a previous trace profiler that was deleted?
* Since we prefer not to have single-bullet lists, I pulled your wrapper discussion into the main sentence. I hope this is OK: `If you cannot use the decorator in your application, you can use the following call format: The wrapper form is `ProfileTraceWrapper`. It can be used to return a wrapped function without the use of a decorator.`

rhetoric101 (approved):
@lrafeei, thanks for adding this document! This looks good!

PR #5581, opened 2022-01-11
chore(java-agent): cap supported versions

meiao:
## Give us some context
Some updates on supported versions of libraries


urbiz-nr (approved):
Thanks for the update, @meiao!

PR #5619, opened 2022-01-12
Add a bug fix to .NET agent 9.3.0 release notes

nr-ahemsath:
Add a bug fix that we forgot to document in the .NET agent 9.3.0 release notes.


rhetoric101 (approved):
@nr-ahemsath, thanks for adding this bullet point!

PR #5634, opened 2022-01-13
chore: remove check for modified translations

rudouglas:
We can just remove the check for `modifiedFiles` here.

The script takes the english path of the file and checks if there is a corresponding locale file, in any of the specified locales

## removedFiles
If an english file is removed we also want to make sure any locale versions are removed, so we flag that

## renamedFiles
This is essentially deleting a file and recreating it in a different location, so we want to check any locale versions that match the original location are deleted/moved/renamed

## modifiedFiles

This is currently checking to see if the `translate` frontmatter has been removed. This is also essentially the same as deleting the file because we want to flag the locale file(s) to be removed too.

However, now we don't need to. If a file has `translate: jp` it will be `Human Translated` in japanese and `Machine Translated for everything else. If you remove `translate: jp` it just means you want this to be MT instead of HT, so you essentially don't need to touch any current HT locale files because they are going to be MT and replaced in the next run anyway.

*We also had 2 places were we were declaring locale id constants so I modified that slightly*

### DIFF
--- a/scripts/actions/check-for-outdated-translations.js
+++ b/scripts/actions/check-for-outdated-translations.js
@@ -1,6 +1,5 @@
 const fs = require('fs');
 const path = require('path');
-const frontmatter = require('@github-docs/frontmatter');
 
 const { fetchPaginatedGHResults } = require('./utils/github-api-helpers');
 const checkArgs = require('./utils/check-args');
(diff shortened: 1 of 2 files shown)


LizBaker (approved):
looks good ta me!

josephgregoryii (commented):
We love a good removal 👍🏻

josephgregoryii (approved):

Issue #5624, opened 2022-01-13
Issue: Compatibility and requirements for the Java agent


qryuu:
## How can we make our docs better?
https://docs.newrelic.com/docs/apm/agents/java-agent/getting-started/compatibility-requirements-java-agent/#jvm

There seems to be a conflict between Compatible agent versions and Not supported versions.

For example, in Java 5, all the versions listed in Compatible agent versions appear to be Not supported versions.

This means that ?
Potentially working versions = Compatible agent versions
Versions for which the technical support service has expired = Not supported

I think the wording is a little confusing.

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/apm/agents/java-agent/getting-started/compatibility-requirements-java-agent/
* **OS**: Unknown
* **Browser**: Unknown
* **Device**: Unknown

PR #5643, opened 2022-01-13
fix(java compat): make 'not supported' more clear

zuluecho9:
Work based on https://github.com/newrelic/docs-website/issues/5624. Making the 'Not supported' aspect of the compatibility table more clear.


paperclypse (approved):
I think this clarifies things

jasonjkeller (approved):

PR #5656, opened 2022-01-14
fix(NerdGraph): Use correct DashboardInput type

dvanoni:
Also update the `dashboard` query variable to match the mutation.

## Give us some context

While working through this example to import a dashboard, I noticed that the type of the `$dashboard` variable declared in the mutation did not match the expected type in NerdGraph. I also noticed that the `dashboard` property in the query variables JSON needed to be updated to match the variable in the mutation.

Ideally, it would probably be good to update the corresponding screenshot as well, but I wasn't sure if y'all wanted to handle that yourselves. 🙂


x8a (approved):
Thanks for the updates!

PR #5658, opened 2022-01-14
[DOC-7649] Redirects for auto-index pages

urbiz-nr:
Adding more redirects for removed auto-index pages. See DOC-7649.


paperclypse (approved):
I did a spot check on about 10 of these redirects. They all worked great!

PR #5667, opened 2022-01-14
chore: bugfixes, cleanup

nr-kkenney:
* remove async from extract method since its not using await keyword. simplify line which calls extract. 
* update console log to use record[0], since record is an array


roadlittledawn (approved):
LGTM

Issue #5404, opened 2021-12-21
Issue: Manage your dashboard


shinji-newrelic:
There is a chart type does not supported by exporter.
As instance, stacked bar chart is not supported and just printed as blank when exporting to PDF.

## How can we make our docs better?

Please add describe limitation or supported chart types.

https://docs.newrelic.com/docs/query-your-data/explore-query-data/dashboards/manage-your-dashboard/#dash-esport
https://docs.newrelic.com/docs/apis/nerdgraph/examples/export-dashboards-pdfpng-using-api/

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/query-your-data/explore-query-data/dashboards/manage-your-dashboard/
* **OS**: Unknown
* **Browser**: Unknown
* **Device**: Unknown

urbiz-nr:
Thanks for reaching out, @shinji-newrelic! I've spoken with the team and this is something they're working on improving. We'll assess how long before the changes take place, and update docs accordingly. Thanks!

urbiz-nr:
Jira'd as DOC-7632.

urbiz-nr:
Closed via #5683.

PR #5683, opened 2022-01-17
[DOC-7632] Detailing chart export limitations and issues

urbiz-nr:
As first reported in #5404, and further discussed with the team.


x8a (approved):
Great!

PR #5690, opened 2022-01-18
Prometheus fixes

urbiz-nr:
A few fixes and improvements:
* Adding auto-index redirect 
* Added lost troubleshooting doc in the nav file, and minor doc edits

Found these while starting work on DOC-7126.


x8a (approved):
Looking great!

PR #5692, opened 2022-01-18
[DOC-7126] Consolidating remote write docs

urbiz-nr:
As explained in DOC-7126, we had two intro docs for Prometheus remote write setup, so I've consolidated them.


x8a (approved):
Looking great!

PR #5695, opened 2022-01-18
Update Mona Doc

iazam6:
As requested by Mona, they have changed some instructions in their documents which I have changed on our end as well. This includes a step simplification and a tip addition.


x8a (approved):
Thanks!

PR #5712, opened 2022-01-19
Remove redundant redirect

LizBaker:
There was an issue with a left nav link not working because that mdx file had a redirect listed that matched the current working path, this removes it


rhetoric101 (approved):
@LizBaker, this worked like a champ locally for me! Thanks for fixing this!

PR #5720, opened 2022-01-19
MLops docs: remove 'insight key' refs and replace w/ license key

zuluecho9:
Most of work: editing MLOps docs to use 'license key' instead of 'insert key' or 'Insight key' or just 'API key'. 

(Also small fix of adjusting some synthetic monitor docs code formatting.)


zuluecho9:
@iazam6 One question I had was in this doc https://docs.newrelic.com/docs/mlops/integrations/datarobot-mlops-integration#connect, which references 'user key'. Should that be 'license key' also?

urbiz-nr (approved):
Thanks for fixing this, @zuluecho9!

PR #5725, opened 2022-01-19
Add more detail on otlp infinite tracing

jack-berg:
OTLP support was added to Infinite Tracing trace observers. This PR adds more detail on how to use it to the OpenTelemetry pages, and updates the status of the deprecated alternative exporters.


rhetoric101 (approved):
Jack, these changes look great! Thanks for making this process clearer!

PR #5747, opened 2022-01-20
Update aws-x-ray-monitoring-integration.mdx

maxlemieux:
Added example of checking for X-Ray Span data with NRQL, and clarified that Span is a reference to an event type.

## Give us some context

Clarifies where to find the X-Ray integration data and how to verify that the integration is working via NRQL.


zuluecho9:
@maxlemieux thank you. made a tiny edit (we just try to keep MELT data types separate conceptually and avoid referencing logs, spans, and metrics as 'events' when we can, even though we know that we also in some contexts refer to all NRDB objects as 'events').

zuluecho9 (approved):

PR #5753, opened 2022-01-21
Mention metric normalization rules for dropping data

wulczer:
The way to manage APM metric timeslice ingest is via metric normalization rules, not via NRQL drop rules.

Leave a pointer to the normalization rules docs in the overview page for dropping data using NerdGraph.


x8a (approved):
Thanks!

PR #5754, opened 2022-01-21
Fix typo in link to metric normalization rules page

wulczer:
A recent change added that link, but it has a typo in the document path.


x8a (approved):
Thanks!

PR #5774, opened 2022-01-24
add OR operator

mufaddal7:
It would make more sense to get the list of GUIDs for dashboard pages with `isDashboardPage:true`, without the proposed change it will only work for getting the GUID's of a dashboard that has multiple nested pages, it will not load GUID's of single dashboard pages.

### DIFF
--- a/src/content/docs/apis/nerdgraph/examples/export-dashboards-pdfpng-using-api.mdx
+++ b/src/content/docs/apis/nerdgraph/examples/export-dashboards-pdfpng-using-api.mdx
@@ -24,7 +24,7 @@ For example, You can generate static, snapshot versions of your New Relic dashbo
    ```
    {
     actor {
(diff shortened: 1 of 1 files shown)


paperclypse:
Hi @mufaddal7, thanks for your PR. If you would sign the contributor license agreement, that would be great. Thanks!

paperclypse:
I've asked our dashboards team to look this over. They think it looks good, but they wanted to double-check. Thanks for your patience!

paperclypse:
Hi @mufaddal7, thanks again for your PR. The dashboards team has looked over this doc and they realize that there are more issues that they'd like to tackle here. We've filed a ticket internally to improve this doc, including your suggestion, but in the meantime they've asked me to close this PR. 

Again, thanks so much for bringing this to our attention!

mufaddal7:
@paperclypse Does my suggestion make sense? and do you'll accept outside contributions?

PR #5800, opened 2022-01-25
Improvements on the setup flow

helenapm:
I'm updating the instructions to create SLIs because we're releasing significant improvements to the flow.


urbiz-nr (approved):
Thanks for the updates, @helenapm!

PR #5815, opened 2022-01-26
Removed the landing page type exclusion

jmiraNR:
## Give us some context

* Landing pages remain and need to be translated. Updating the exclusion file to allow for translation.

### DIFF
--- a/scripts/actions/utils/docs-content-tools/i18n-exclusions.yml
+++ b/scripts/actions/utils/docs-content-tools/i18n-exclusions.yml
@@ -11,5 +11,3 @@ excludePath:
     - src/content/docs/security/security-privacy/compliance
     - scripts/actions/__tests__/kitchen-sink.mdx
     - src/content/docs/infrastructure/manage-your-data/data-instrumentation/host-integrations-metrics.mdx
-excludeType:
-  ja-JP:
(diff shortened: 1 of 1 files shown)


scripts/actions/utils/docs-content-tools/i18n-exclusions.yml
```diff
@@ -13,4 +13,3 @@ excludePath:
     - src/content/docs/infrastructure/manage-your-data/data-instrumentation/host-integrations-metrics.mdx
 excludeType:
   ja-JP:
```
roadlittledawn, line 15: if empty i believe you can just remove the whole shebang

roadlittledawn (changes requested):

scripts/actions/utils/docs-content-tools/i18n-exclusions.yml
jmiraNR: I'd rather keep it so I have the template in case we need to exclude types.

roadlittledawn: since code runs on this, we should really remove this if not used.

roadlittledawn: we can add some doc on this so it's clear what options are available and how to use it.

roadlittledawn (approved):
TY! :shipit:

PR #5742, opened 2022-01-20
feat(K8): Moved content, created new sections and added doc placeholder

roobre:
Part of the work in the main branch https://github.com/newrelic/docs-website/pull/5715 for a kubernetes integration v3.

PR #5817, opened 2022-01-26
Add dockerfile for local development

roobre:
## Give us some context

Dependencies required to build docs locally are quite heavy and might clutter developer's machines. This Dockerfile yields a production-ready image based on `nginx` with the docs site built inside it.

Limitations:
- Docker image wil be big, around 3 GiB in size
- Dependencies will be cached, but intermediate builds unfortunately won't. This means that each build can take a long time, around `600s` in my local machine.

Developed while working on #5742

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.cache/
+node_modules/
+public/
+.dockerignore
+Dockerfile
+*.md
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+# This Dockerfile is a community contribution and is not maintained by the New Relic Docs team.
+# Use it at your own risk!
+#
+# docker build . -t docs-preview
+# docker run -ti --rm -p 8080:80 docs-preview
+# http://localhost:8080
+
+FROM node:16 as build
+
+WORKDIR /app
+
+# Pull dependencies
+ADD package.json yarn.lock .nvmrc /app/
+RUN yarn
+
+ADD . /app
+RUN yarn build
+
+FROM nginx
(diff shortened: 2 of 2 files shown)


bradleycamacho:
Wow, this is pretty sick @roobre! I'm going to pull DevEn into this so they can provide engineering thoughts.

aswanson-nr:
Hey @roobre! We do have tons of dependencies for this site, so I can see the benefits of a docker image. Although, all the dependencies are installed via `yarn`, so they are kept local to the repository itself. So I just wanted to get a better idea of the problem this solves. Is this intended for development workflows or deploying the site via Docker? I'm guessing production since the image is using `yarn build` and serving via Nginx.

roobre:
Hey @aswanson-nr, happy to elaborate. I made this to cover two use cases I personally had. The first one would be that I place the docker cache in a different filesystem, which is faster and with less redundancy than the one hosting the development files. The second one would be that this way is easier to deploy the preview to places, for example this image is hosted at nr-docs.roobre.es. This was useful to share the PR output when the gatsby cloud build fails.

As I said, I am aware this is fairly specific and it's possible that not many others have the same need :)

jpvajda:
@roadlittledawn can you confirm what we want to do with this PR?

aswanson-nr:
Hey @roobre! Since this is a large repository, we are hesitant merge in extra workflows that are not supported by the core developer enablement team. However, I totally understand your use case! Since this is not something we can merge into the docs site at this time, could I recommend that you fork the repository and check in the dockerfile there? That way it's easy to pull in new code from the site and keep the dockerfile up to date.

PR #5818, opened 2022-01-26
Remove unused related resources json files

LizBaker:
We removed the related resources script but still have the data files from it. This is some clean up 🧹


src/templates/docPage.js
```diff
@@ -109,12 +107,6 @@ const BasicDoc = ({ data, location, pageContext }) => {
             />
           )}
           <TableOfContents headings={headings} />
-          <RelatedResources
```
roadlittledawn, line 112: what happens if a content author adds related resources via frontmatter?

LizBaker: they become very saddened when it doesn't show up... 😂 just reverted that

roadlittledawn (approved):
thanks for the catch and clean up!

LizBaker:
close/open to retrigger build

roadlittledawn (dismissed):
:shipit:

roadlittledawn (approved):

PR #4949, opened 2021-11-17
Add initial docs for new externals feature

rhetoric101:
This is work for doc-7521 for documenting the new externals feature.

PR #5829, opened 2022-01-27
fix(externalservices):remove unpaired translation

bradleycamacho:
Removes unpaired translation not found in #4949


rhetoric101 (approved):
@bradleycamacho, that JP deletion looks correct!

PR #5837, opened 2022-01-27
[DOC-7683] JVMs: New UI updates

urbiz-nr:
See DOC-7683 for more details.


rhetoric101:
@urbiz-nr I am closing this because the work was actually for the OpenTelemetry UI that has a section about JVMs. We're covering this work in #6588.

PR #5841, opened 2022-01-27
fix(errors inbox): remove EU callout

bradleycamacho:
Errors inbox is now available in the EU. This removes a reference saying it is not.


paperclypse (approved):
Looks good

PR #5850, opened 2022-01-28
External services fast-follow changes

rhetoric101:
Reviewer, could you look at the following changes?

* In the introduction, I switched two sections around to highlight when you would use the feature.
* In the distributed tracing planning guide, I removed a collapser that had a warning that used to apply when you would enable DT. This is no longer applicable.
* In the limits topic, I removed an out of date comment about limit agreements for DT (per SME).


urbiz-nr (approved):
Looks good, tried locally. Great docs @rhetoric101!

rhetoric101:
@urbiz-nr, thanks for checking!

PR #5852, opened 2022-01-28
Clarified authentication options

nedl86:
Authentication has been a source of confusion for users trying to directly utilise the Log API. Clarified the two available options and restructured slightly to make it clearer, as well as tidying up the examples.


urbiz-nr (approved):
Thanks for the updates, @nedl86!

PR #5854, opened 2022-01-28
docs(ServiceLevel): SLO periods now include complete weeks

asanuy:
## Give us some context

Update the docs since now SLO periods in service levels only include complete weeks.

The changes include a new section in the release notes from the January 27 and a new question in the FAQs.


urbiz-nr (approved):
Thanks for the update, @asanuy!

PR #5904, opened 2022-02-01
Add important callout for endpoints not listed

keegoid-nr:
Confusion arises when customers see that:

- We have a list of non-compliant services where Synthetics is not listed.
- We have a list of compliant endpoints where Synthetics is also not listed.

This callout helps to clarify that this list of FedRAMP compliant endpoints is for services who's main endpoint is non-compliant. If a service's endpoint is not listed here and is not a non-compliant service, then it is compliant.

If this doc edit needs approval from New Relic Security, let's do that.


keegoid-nr:
Also updated the networking link to point to the new URL.
https://docs.newrelic.com/docs/using-new-relic/welcome-new-relic/get-started/networks/

rhetoric101 (approved):
@keegoid-nr, thanks for adding this callout!

Issue #5905, opened 2022-02-01
[Translation] Send content to be translated workflows failing


nr-kkenney:
## Summary
Both (human and machine) send content to translated workflows failed, with both errors appearing to the be the same. The machine translation workflow failure can be seen [here](https://github.com/newrelic/docs-website/runs/5015061811?check_suite_focus=true). 

The error in both cases is:
```
Run node ./scripts/actions/send-and-update-translation-queue.js
/home/runner/work/docs-website/docs-website/scripts/actions/serialize-mdx.js:10
const format = require('rehype-format');
               ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /home/runner/work/docs-website/docs-website/node_modules/rehype-format/index.js from /home/runner/work/docs-website/docs-website/scripts/actions/serialize-mdx.js not supported.
Instead change the require of index.js in /home/runner/work/docs-website/docs-website/scripts/actions/serialize-mdx.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/home/runner/work/docs-website/docs-website/scripts/actions/serialize-mdx.js:10:16)
    at Object.<anonymous> (/home/runner/work/docs-website/docs-website/scripts/actions/utils/vendor-request.js:6:22)
    at Object.<anonymous> (/home/runner/work/docs-website/docs-website/scripts/actions/send-and-update-translation-queue.js:7:39) ***
  code: 'ERR_REQUIRE_ESM'
***
Error: Process completed with exit code 1.
```

This is due to the rehype-format version upgrade i made going from 3 -> 4 as part of the changes for the [i18n utility script](https://github.com/newrelic/docs-website/pull/5851). The [release notes](https://github.com/rehypejs/rehype-format/releases) for rehype-format call this out 🙈 

Some things we could try (in no particular order):
* rollback that particular change (go back to version 3). see if that works.
* downgrade major version, upgrade minor. see if that works.
* use ESM syntax?

## Acceptance Criteria
- [x] Error above is resolved & workflows are able to run again w/o failing.

PR #5909, opened 2022-02-01
chore: downgrade rehype-format to fix workflow send content error

roadlittledawn:
## Description

Downgrades `rehype-format` back to 3.1.0 to keep using CommonJS `require()`'s.

Tested uploading via `scripts/actions/translation_workflow/testing/script.sh` and it completed successfully.

Not sure we're quite in a place to switch over to ESM.

## Related issues / PRs
Closes https://github.com/newrelic/docs-website/issues/5905

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -18,5 +18,5 @@ module.exports = {
   setupFiles: ['<rootDir>/loadershim.js'],
   setupFilesAfterEnv: ['<rootDir>/setup-test-env.js'],
   testRegex: '(/__tests__/.*|\\.(test|spec))\\.(ts|tsx|js)$',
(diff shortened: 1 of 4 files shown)


josephgregoryii (dismissed):

nr-kkenney (changes requested):
looks like the test failed. we should make sure those pass, and make sure the i18n script can be run locally.

which i believe is why i had to update the package version in the first place.

roadlittledawn:
@nr-kkenney yeah... the translation workflow worked when i tested locally (serialized MDX, uploaded, downloaded, deserialized). still not 100% clear why it fails the test though

nr-kkenney (approved):

PR #5926, opened 2022-02-02
fix: broken link to category page

bradleycamacho:
Fixes a broken link due to turning off index pages.

Reviewer, please double check that the new link is the best place to send users to. There wasn't a perfect location, and this seemed the best one.


urbiz-nr (approved):
Yep, that looks like the best redirect IMO, @bradleycamacho!

PR #5929, opened 2022-02-02
fix(style guide page): copy edits.

mmfred:
updates to released page.

### DIFF
--- a/src/nav/style-guide.yml
+++ b/src/nav/style-guide.yml
@@ -15,7 +15,7 @@ pages:
     pages:
       - title: Capitalization
         path: /docs/style-guide/capitalization/capitalization
-      - title: Product capability and feature capitalization and usage
+      - title: Product, capability and feature capitalization and usage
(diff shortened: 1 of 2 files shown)


mmfred:
This is ready for review. Should not have trouble building.

src/nav/style-guide.yml
```diff
@@ -15,7 +15,7 @@ pages:
     pages:
       - title: Capitalization
         path: /docs/style-guide/capitalization/capitalization
-      - title: Product capability and feature capitalization and usage
+      - title: Product, capability and feature capitalization and usage
```
bradleycamacho, line 18:
```suggestion
      - title: Product, capability, and feature capitalization and usage
```

bradleycamacho (approved):
One small change :)

src/nav/style-guide.yml
mmfred: I caught that right after I was all ready to go, so now we'll see what happens. I'm going to merge. Thanks!

PR #5937, opened 2022-02-03
Update include-java-agent-jvm-argument.mdx

jasonjkeller:
Update Java agent + Resin install instruction


urbiz-nr (approved):
Thanks for the updates, @jasonjkeller!

Issue #5927, opened 2022-02-02
What's new IDs are getting committed to develop instead of release branch


LizBaker:
Our update-whats-new-ids action is hardcoded to commit to develop. Now that we use release branches, any new IDs aren't picked up by the release containing the new post, but by a subsequent release. We want to update that to keep new IDs and posts together

AC
- [x] update the action to checkout the branch being PRed to main (checking out develop could add a whats new id for a post that doesn't exist on the release branch)
- [x] commit new IDs to the branch being PRed to main

PR #5948, opened 2022-02-03
Use release branch for whats new ID commits

LizBaker:
This updates the whats new ID generating workflow to commit new IDs to the branch PRing main instead of develop

closes https://github.com/newrelic/docs-website/issues/5927

### DIFF
--- a/.github/workflows/update-whats-new-ids.yml
+++ b/.github/workflows/update-whats-new-ids.yml
@@ -19,8 +19,7 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v2
         with:
-          ref: develop
-          persist-credentials: false
+          fetch-depth: '0'
 
       - name: Setup node.js
         uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


aswanson-nr (dismissed):
Looks good to me!

nr-kkenney (approved):

roadlittledawn:
@LizBaker i think the release branches have branch protection on them don't they? so we would need that remove protection part still right?

.github/workflows/update-whats-new-ids.yml
```diff
@@ -41,64 +40,11 @@ jobs:
...
-              repo: context.repo.repo,
-              branch: 'develop',
-              required_status_checks: null,
-              restrictions: null,
-              enforce_admins: null,
-              required_pull_request_reviews: null
-            })
-            console.log("Result:", result)
-
-        # Push directly to the `develop` branch so we get the changes included in the release PR
```
roadlittledawn, line 63: i think the release branches have some branch protection enabled on them, so we may need this i think?

roadlittledawn (changes requested):

.github/workflows/update-whats-new-ids.yml
LizBaker: @roadlittledawn ahh didn't realize that, thought they were just branched off develop and named anything but i see the daily/release* in the protection rules! i'll add that in

roadlittledawn (approved):
🤘

PR #5974, opened 2022-02-04
docs(Alerts): removing NRQL outlier conditions (which are EOL)

founddrama:
## Give us some context

* Outlier NRQL conditions are being EOL'ed as of 3/31/2022
* We have already locked outlier NRQL conditions from being created
* This removes them from the documentation

**NOTED:** There a couple of semi-obvious places in the **jp** translations where these could also be removed, but a couple other places where it wasn't immediately obvious to me how to remove them because the mentions were mid-setnence.


founddrama:
> I do not understand the messages re branch has conflicts that must be resolved. Do you happen to know what is the problem? I am not adept at resolving via command line and don't want to mess anything up. I can ask our engineers to troubleshoot this PR if nothing obvious leaps out at you. Thanks for any suggestions!

My initial attempt to reconcile the differences between this branch and the mainline develop branch produced a lot of noisy conflicts. I'll retry later this week (possibly as early as today) but if push comes to shove it's a self-contained enough changeset that it's low-effort to re-create it.

PR #5977, opened 2022-02-04
Update jvms-page-java-view-app-server-metrics-jmx.mdx

jasonjkeller:
Update JBoss EAP JVM page support


barbnewrelic (approved):
Wow, this doc is old. I remember when it was first created once upon a time! Thanks for updating it!

PR #6007, opened 2022-02-05
Specify condition when using per-signal env var

reese-lee:
- Adds new section called "A note about endpoints" that per Otel spec, when using a per-signal env var, it is required to set the endpoint URL ending in either `v1/traces` or `v1/metrics` (not doing so will result in a 404)
- Also fixes the Infinite Tracing row in the table
- Adds link to the new section in the config section above the table


x8a (approved):
Thanks a lot!

PR #6034, opened 2022-02-08
Revert "Use release branch for whats new ID commits"

LizBaker:
Reverts newrelic/docs-website#5948

we have an odd merge commit getting added by this action. we're not sure if it's causing stalled builds yet but i'm going to revert this for now. whats new IDs will go back to being committed to develop


roadlittledawn (approved):
LGTM :shipit:

PR #6035, opened 2022-02-08
Image Alias

josephgregoryii:
### Summary (WIP)
Files to check:
- a few mdx files (changing image imports FROM: `./images/example.png` TO: `images/example.png` AND `./images/...` TO `src/images/...` where applicable
- `gatsby-config.js` webpack config


gatsby-config.js
```diff
@@ -473,5 +473,13 @@ module.exports = {
...
       },
     },
+    {
+      resolve: `gatsby-alias-imports`,
+      options: {
+        aliases: {
+          images: path.resolve(__dirname, 'src/images/'),
+        },
+      },
+    },
```
aswanson-nr, line 483: This is only marginally less code than doing it ourselves and not pulling in another dependency. If you look at the 
 [source](https://github.com/luanbitar/gatsby-alias-imports/blob/main/gatsby-node.js) of this plugin, it's really only parsing the plugin config and outputting one line to the webpack config. I would recommend just following the docs for [webpack alias](https://github.com/luanbitar/gatsby-alias-imports/blob/main/gatsby-node.js) and put it directly into the config.

josephgregoryii: @aswanson-nr Great callout - I'll make that change :)

josephgregoryii:
Closing this branch to open a new one that only contains the code changes

PR #6043, opened 2022-02-08
chore: deal with errored translations

rudouglas:
This update deals with translations that have an `ERRORED` status in the DB:

 - No longer updates the translations status to `COMPLETED`
 - Logs all `ERRORED` status translations to the console
 - Only updates statuses for `ok` translations
 - Includes tests

Also updated the `send-an-update-translation-queue` file to:
 - Delete translations that have an `ERRORED` status from the DB
 - Includes tests

### DIFF
--- a/scripts/actions/check-job-progress.js
+++ b/scripts/actions/check-job-progress.js
@@ -58,9 +58,7 @@ const log = (message, level = 'log', indent = 0) => {
   const logIndicators = { log: '[*]', warn: '[!]' };
 
   const str = [
-    Array(indent)
-      .fill(' ')
-      .join(''),
+    Array(indent).fill(' ').join(''),
     logIndicators[level],
     message,
   ].join('');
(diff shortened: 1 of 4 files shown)


scripts/actions/check-job-progress.js
```diff
@@ -160,7 +175,10 @@ const updateTranslationRecords = async (slugStatuses) => {
 
   await Promise.all(
     slugStatuses.map(async ({ ok, slug }) => {
-      const updateStatus = ok ? StatusEnum.COMPLETED : StatusEnum.ERRORED;
+      if (!ok) {
+        return;
+      }
+      const updateStatus = StatusEnum.COMPLETED;
```
nr-kkenney, line 181: this is probably the only change we need -- it handles removing the errored state from being set. additionally, if everything is set to completed, all records will be deleted at the end.

we can still log the failed slugs, that seems prudent.

scripts/actions/check-job-progress.js
```diff
@@ -241,7 +259,11 @@ const main = async () => {
       )
     ).flat();
 
-    await updateTranslationRecords(slugStatuses);
+    const erroredStatuses = slugStatuses.filter(({ ok }) => !ok);
+    const okStatuses = slugStatuses.filter(({ ok }) => ok);
+
+    logErroredStatuses(erroredStatuses);
+    await updateTranslationRecords(okStatuses);
```
nr-kkenney, line 266: should be fine to leave this as it was, with everything passed in.

nr-kkenney (changes requested):
left a few comments. but i think we only need the change for setting translation records to completed.

PR #6065, opened 2022-02-09
New liaison model

austin-schaefer:
Update agile handbook to reflect new liaison model.

Requesting @Broseberg as reviewer for her perspective as manager, and @akristen for her "beginner's mind" perspective as someone newer to the team.

### DIFF
--- a/src/content/docs/agile-handbook/sprint-mechanics/liaisonships.mdx
+++ b/src/content/docs/agile-handbook/sprint-mechanics/liaisonships.mdx
@@ -4,18 +4,19 @@ template: basicDoc
 topics:
   - Docs agile handbook
 ---
-For large projects, we'll typically assign a particular tech writer to that project as a "liaison." The liaison’s job is to ensure that we get complete, consistent, and timely docs. 
+We're all responsible for the docs site. Every writer (anyone in the world, actually!) can edit any doc. When you see something that could be better, you should file an issue or, better yet, edit the doc!
+
+But those casual edits by themselves aren't enough to create a world-class product. Great docs need strong product knowledge, born of strong relationships with subject matter experts. That helps us understand user journeys and user experience (in the docs as well as the product), and it's particularly helpful as we design information architecture.
+
+To support that deeper expertise, we break our platform into a number of domain areas and assign each area a level of support:
+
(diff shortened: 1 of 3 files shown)


src/content/docs/agile-handbook/sprint-mechanics/liaisonships.mdx
```diff
@@ -208,39 +152,32 @@ However, a liaison is not the only author on a project. Liaisons should structur
...
-                Coordinate publication
+                Edit content
             </td>
             <td>
-                Liaison
+                Liaison or point person
             </td>
             <td>
-                When the time comes to release (whether that's beta, GA, limited release, or EoL), it's the liaison's job to coordinate with PM, Eng, and Product Marketing to ensure docs go out on time with other deliverables.
+                While many people might edit docs in an area, you're accountable as liaison or point person for the overall quality and structure of the content. That might mean volunteering for peer edit assignments, or tag-teaming with the hero to review an incoming pull request.
```
akristen, line 180: It could be good to reference (somewhere) the importance of cultivating and maintaining a _voice_ as part of the cohesive experience. Yes, we have a style guide to, well, guide the writers, but drawing attention to what goes into "quality" can empower liaisons to own that area.  Our style guide is a great starting point, but multiple writers can be within the parameters of those guidelines and still have our own unique voices explaining a feature/capability. "Voice" is also a tangible example of what goes into "quality," which is a vague and subjective term. 

A little nitpicky but something that came up for me is how much I should change my own natural voice when writing a doc—it's unclear how much creative control tws have with writing style, etc.

akristen: I actually want to modify "a little nitpicky," because I do think cultivating a strong, consistent voice within a given area is important for creating trust between reader/user and the feature they're using. 

Based on the rest of the doc, it does seem like there's some flexibility between who is contributing to a given area (e.g., liaisons work with different teams, other writers can swarm and support), such that the liaison should yield when appropriate. But at the very, very least, what makes a quality docs experience can be quite specific and nameable (that idea of empowering your writers to use their voice within style guidelines) without being restrictive/proscriptive (which the word "quality," vague and subjective as it is, can imply).

PR #6074, opened 2022-02-09
k8s/changes: add an upgrade from v2 section

roobre:
### What problems does this PR solve?

The page we dedicated to explain the changes for the new Kubernetes integration version did not have a section explicitly saying how to upgrade from v2. This PR adds this.

While this is slightly duplicated, as it is in fact the same command that can be found in the "install integration using helm" section, we thought it was worth adding it to encourage users to jump to action right away, rather than making them search _how_ to upgrade elsewhere.


x8a (approved):
Thanks a lot for the addition!

PR #6084, opened 2022-02-09
Update branch committed to by whats new ID workflow

LizBaker:
This is a take 2 to solve our whats new ID commit issue. Currently the workflow commits to develop but we'd like the IDs committed to the release branch. The previous attempt was adding an odd merge commit but this iteration specifies the checkout branch as the head and appears to have solved the issue in tests

## Related PRs
- https://github.com/newrelic/docs-website/pull/5948
- https://github.com/newrelic/docs-website/pull/6034

### DIFF
--- a/.github/workflows/update-whats-new-ids.yml
+++ b/.github/workflows/update-whats-new-ids.yml
@@ -19,8 +19,7 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v2
         with:
-          ref: develop
-          persist-credentials: false
+          ref: ${{github.head_ref}}
 
       - name: Setup node.js
         uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/update-whats-new-ids.yml
```diff
@@ -19,8 +19,7 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v2
         with:
-          ref: develop
-          persist-credentials: false
+          ref: ${{github.head_ref}}
```
roadlittledawn, line 22: interesting. so no `fetch-depth: '0'` but we add that ref. 

i'm curious why we don't need `persist-credentials`

LizBaker: the previous workflow had a ref of develop, i thought i could remove the ref and let it choose whichever branch triggered the PR but it turned out to do odd things. but it seems adding a ref specifying the head has solved it!

re: persist credentials, this persists github credentials throughout the workflow i guess? and removes them once it's finished. you can set it to false as we did before to opt out of that. i havent seen this is other checkout actions we've used so i didn't bother with it, but i can add it back

roadlittledawn: ah okay. cool cool. i just wasn't sure if that would affect the whole removing branch protection steps on the `daily-release/*` branch

PR #6086, opened 2022-02-09
fix(auth domain): edit to user upgrade UI

zuluecho9:
Org/users liaison work: new user upgrade UI now controls all aspects of user type upgrade requests (previously only controlled upgrades to full platform user).


barbnewrelic (approved):
This looks fine, thanks for another important edit to user management docs!

PR #6090, opened 2022-02-09
Fix broken configuration link in trace observer doc

rhetoric101:
The optional configuration step had a dead link in the trace observer setup. This will all get replaced shortly by a larger set of doc changes, but this is a short-term fix.

I am out of the office on 2/10/2022, but if this looks good, you can publish it for me.


urbiz-nr (approved):
Looking good!

PR #6091, opened 2022-02-09
whats-new external services post dated 02-09-2022

stamerman:
Can go out immediately as the feature is already live.

Post was written and approved by Scott Amerman


barbnewrelic:
Here is the screenshot that needs to be added to this What's New post.
Screen Shot 2022-02-09 at 2 38 51 PM

daynaslord (approved):
Looks good to me - thanks

PR #6094, opened 2022-02-10
Make core user post relative link absolute

rhetoric101:
I noticed that the final link in this post was relative, which means it fails to resolve in the product UI.

I am out of the office on 2/10/2022, but you can publish this if it looks correct.


urbiz-nr (approved):
Good catch!

PR #6101, opened 2022-02-10
[DOC-7650] Metric Streams callout

urbiz-nr:
See DOC-7650 for more information. Callout text approved by PM.


urbiz-nr:
Peer reviewer - there's 5 integrations that don't need the callout, it's no mistake!

src/content/docs/infrastructure/amazon-integrations/aws-integrations-list/amazon-cognito-monitoring-integration.mdx
```diff
@@ -11,6 +11,10 @@ redirects:
   - /docs/aws-aws_cognito
 ---
 
+<Callout variant="important">
+  Enable the [AWS CloudWatch Metric Streams integration](/docs/infrastructure/amazon-integrations/aws-integrations-list/aws-metric-stream/) to monitor all CloudWatch metrics from your AWS services, including custom namespaces. Individual integrations are no longer our recommended option.
```
x8a, line 15: I'd rephrase the last sentence to avoid passive voice. Something like, "We don't recommend installing individual integrations."
Could we also add context about why we don't recommend it anymore?

urbiz-nr: Thanks for the feedback. However, the PM wants this particular wording ;)

x8a (approved):

PR #6108, opened 2022-02-10
fix inaccuracy

ryanv94:
In the Original User Model all users are either authenticated via Username/Password, or via SSO. There is no option for a portion of users to auth one way and another portion to auth the other. 

the doc contained reference to "Read about an option to have users bypass SAML SSO if they use domains you own."  there is no way to bypass SSO. What users are bypassing is the requirement to verify their email address. Hopefully this PR clarifies that. :)


zuluecho9:
thanks @ryanv94 that's awesome, appreciated.

zuluecho9 (approved):

PR #6126, opened 2022-02-11
Update apm-config-nerdgraph.mdx to hoist guid to root level on update mutation.

nivgithub:
This is apparently related to a change that's coming out now with NerdGraph and effects are restrained just to this doc. - @zuluecho9


zuluecho9:
@nivgithub Could you give me a little more context for this change? What's the main problem you're fixing and do you think it may be related to other problems in our NerdGraph docs (if you know)? Are you associated with New Relic in some way (might be helpful for follow-up questions). Thank you for any more context, and we definitely appreciate the help.

zuluecho9 (approved):

PR #6140, opened 2022-02-11
docs: Updated code block identifier

wyat2:
With assistance from Brian Hensley we have updated some code blocks for better display.
Previously looked like terminal commands.

### DIFF
--- a/src/content/docs/apm/agents/nodejs-agent/api-guides/nodejs-agent-api.mdx
+++ b/src/content/docs/apm/agents/nodejs-agent/api-guides/nodejs-agent-api.mdx
@@ -865,7 +865,7 @@ Here is the structure for rules in New Relic's Node.js agent.
 
     This can also be set with the environment variable `NEW_RELIC_NAMING_RULES`, with multiple rules passed in as a list of comma-delimited JSON object literals:
 
-    ```json
+    ```js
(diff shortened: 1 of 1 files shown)


brnhensley:
Hello, I asked Andrew to fix a mistake I made earlier. In standard .md (like how GH interprets code blocks) the ```bash identifier will nicely color code pieces of code text, like so:

```bash
pattern: '[0-9]+',
pattern: '[0-9]+',
replace_all: true
replace_all: true
```

vs how we had it:

```
pattern: '[0-9]+',
pattern: '[0-9]+',
replace_all: true
replace_all: true
```

This was purely a visual improvement I tried to make, I did not remember that our NR docs website interprets the bash identifier differently and sticks it into a little console window with `$` and directly implies those are supposed to be terminal commands. So I asked Andrew to correct this for me. This code could maybe be made nice with the ```js identifier (like below), but I thought maybe it's not a good idea to experiment with the live site. If there is any way to test how something like that would look on the live site I'd love to know how I could try that.

```js
pattern: '[0-9]+',
pattern: '[0-9]+',
replace_all: true
replace_all: true
```


The ```json similarly added unexpected `( )`around the code and I asked him to remove that as well.

Thanks!

zuluecho9 (approved):

PR #6176, opened 2022-02-15
Gsanchez/fix/oracle docs

gsanchezgavier:

## Give us some context
- Metrics that are extended type are not correctly marked. this PR solves that.

 - It also removes the reference to the k8s intallation since this integration is not part of the bundle so it cannot be run in k8s with this instructions.


rhetoric101 (approved):
@gsanchezgavier, thanks for making all of these updates!

PR #6204, opened 2022-02-16
fix(partnership api docs): adding req callouts

zuluecho9:
DOC-7753. I didn't end up condensing any docs but did add callouts about availability to API docs. This category is pretty messy and I think for future project we can condense a lot of these and remove some.

### DIFF
--- a/src/content/docs/apis/nerdgraph/tutorials/provisions-your-subscriptions-nerdgraph.mdx
+++ b/src/content/docs/apis/nerdgraph/tutorials/provisions-your-subscriptions-nerdgraph.mdx
@@ -3,39 +3,34 @@ title: "NerdGraph tutorial: Partner and reseller subscriptions"
 metaDescription: 'For New Relic partners: create subscriptions for your accounts.'
 ---
 
-<Callout variant="important">
-  The Provisioning API is only available for [Partner accounts](/docs/new-relic-partnerships/partner-integration-guide/getting-started/partner-integration-requirements) on our [original pricing model](/docs/accounts/original-accounts-billing/original-product-based-pricing/overview-changes-pricing-user-model/#pricing-plans). 
-</Callout>
 
-The Provisioning NerdGraph API allows you, as a [New Relic partner](/docs/new-relic-partnerships/partner-integration-guide/getting-started/partner-integration-requirements) to create subscriptions for your accounts containing a more extensive range of New Relic products than the [Partner API](/docs/new-relic-partnerships/partner-integration-guide/partner-account-maintenance/partner-api).
+The subscription provisioning functionality of our NerdGraph API allows you, as a [New Relic partner](/docs/new-relic-partnerships/partner-integration-guide/getting-started/partner-integration-requirements), to create subscriptions for your accounts. And it gives you access to a more extensive range of New Relic products than the [Partner API](/docs/new-relic-partnerships/partner-integration-guide/partner-account-maintenance/partner-api).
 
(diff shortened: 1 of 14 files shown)


src/content/docs/apis/nerdgraph/tutorials/provisions-your-subscriptions-nerdgraph.mdx
```diff
@@ -3,39 +3,34 @@ title: "NerdGraph tutorial: Partner and reseller subscriptions"
...
+* You should **not** use NerdGraph to adjust your subscription provisioning unless specifically instructed by your New Relic representative.
+* Only available for [Partner accounts](/docs/new-relic-partnerships/partner-integration-guide/getting-started/partner-integration-requirements) that are on our [original pricing model](/docs/accounts/original-accounts-billing/original-product-based-pricing/overview-changes-pricing-user-model/#pricing-plans). 
 
-## Get started: endpoint and API key [#endpoint]
+NerdGraph's provisioning functionality and the [Partnership API](/docs/new-relic-partnerships/partner-integration-guide/partner-account-maintenance/partner-api) are compatible when it comes to the products supported by both (APM, Mobile, Insights, Browser, Synthetics, and Infrastructure). If you use NerdGraph to create a subscription for any of the newer products, you'll no longer be able to update the account using the Partnership API.
```
barbnewrelic, line 14: we no longer refer to these as products or capitalize them. see style guide https://docs.newrelic.com/docs/style-guide/capitalization/product-capability-feature-usage/

src/content/docs/apis/nerdgraph/tutorials/provisions-your-subscriptions-nerdgraph.mdx
```diff
@@ -3,39 +3,34 @@ title: "NerdGraph tutorial: Partner and reseller subscriptions"
...
 ```
 
 ## Product subscriptions for an account [#subscriptions]
 
-The Provisioning API does not allow you to upgrade or downgrade individual product subscriptions for an account. Instead, the API requires you to replace ([add](/docs/new-relic-partnerships/partnerships/partner-api/partnership-api-subscription-object#example-create)) the configuration for all product subscriptions for the account.
+NerdGraph doesn't allow you to upgrade or downgrade individual product subscriptions for an account. Instead, the API requires you to replace ([add](/docs/new-relic-partnerships/partnerships/partner-api/partnership-api-subscription-object#example-create)) the configuration for all product subscriptions for the account.
 
 <Callout variant="tip">
-  If any product configurations are not included, the Provisioning API automatically provisions the account with a free product (when available).
+  If any product configurations are not included, NerdGraph automatically provisions the account with a free product (when available).
```
barbnewrelic, line 33: you may want to ask ui-write team how to handle wording on line 33. we don't refer to them as products anymore but i'm not sure that "configurations for New Relic capabilities" is the right way to say this

barbnewrelic (changes requested):
just a couple comments. Your new callout is great for the rest of the docs. Thanks for taking care of this so quickly!

src/content/docs/apis/nerdgraph/tutorials/provisions-your-subscriptions-nerdgraph.mdx
zuluecho9: right, although i wasn't reviewing for this stuff and there is a ton of that in these docs (and in some sense it's not bad as this is referring to original pricing model, which still has those product classifications)

src/content/docs/apis/nerdgraph/tutorials/provisions-your-subscriptions-nerdgraph.mdx
zuluecho9: in this case, because this only is relating to the original pricing model, that model still has those as products so think it's fine

PR #6216, opened 2022-02-16
fix(dark mode): update screenshot

bradleycamacho:
Accidentally pushed his commit to a test branch instead of #6189.


barbnewrelic (changes requested):
I am confused by this PR. The original (cropped) screenshot still exists in https://docswebsitedevelop-darkmode83790.gtsb.io/docs/accounts/accounts-billing/general-account-settings/view-our-ui-dark-mode/  and there's a broken image at the top of this preview. Ideally, the screenshot being replaced would be a new one that shows the account dropdown with the options for Theme. Go to newrelic.com and you'll see what I mean. Good luck!

PR #6236, opened 2022-02-18
feat: add Korean translation functionality

rudouglas:
- Combines all declarations of `locale ids` into a constant located in [src/scripts/actions/utils/constants.js](https://github.com/newrelic/docs-website/blob/develop/scripts/actions/utils/constants.js) in format `{localeKey: locale}` e.g. `{ko: 'ko-KR'}`
     - This is to keep the `localeKey` and `locale` variable usage consistent across all scripts, as we need each of them for different purposes
 - [scripts/actions/check-job-progress.js](https://github.com/newrelic/docs-website/compare/feature/korean-translations?expand=1#diff-c1c3066f6308eccb7a065b31959adc9c3b2937f72889b25bdb64d4519a7abd6c) now accounts for multiple languages when marking translations as __COMPLETED__ 
 - Updates tests to account for new Korean language
 - Update the migration script so that the Test DB `locales` table includes `ko-KR`

### DIFF
--- a/scripts/actions/check-job-progress.js
+++ b/scripts/actions/check-job-progress.js
@@ -10,6 +10,7 @@ const {
 const { vendorRequest } = require('./utils/vendor-request');
 const { fetchAndDeserializeFiles } = require('./fetch-and-deserialize');
 const { configuration } = require('./configuration');
+const { LOCALE_IDS } = require('./utils/constants.js');
 
 const PROJECT_ID = configuration.TRANSLATION.VENDOR_PROJECT;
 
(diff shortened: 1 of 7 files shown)


scripts/actions/check-job-progress.js
```diff
@@ -175,14 +176,15 @@ const updateTranslationRecords = async (slugStatuses) => {
...
     slugStatuses.map(async ({ slug }) => {
-      const records = await updateTranslations(
-        { slug, status: StatusEnum.IN_PROGRESS },
-        { status: StatusEnum.COMPLETED }
-      );
-
-      console.log(
-        `Translation ${records[0].id} marked as ${StatusEnum.COMPLETED}`
-      );
+      Object.values(LOCALE_IDS).forEach(async (localeId) => {
```
nr-kkenney, line 179: i think we want to have locale be a property on the incoming slugStatus rather than iterate over every locale for every slug.

```
slugStatuses.map(async ({ locale,  slug }) => { . . . }
```

nr-kkenney: i think adding locale [here](https://github.com/newrelic/docs-website/blob/1277a589f91cd4e109cbd0fa84110dc2f9820b21/scripts/actions/check-job-progress.js#L253) would allow us to deconstruct it in this method.

scripts/actions/utils/constants.js
```diff
@@ -1,5 +1,7 @@
+// This should be in the format {localeKey: locale} across scripts
 const LOCALE_IDS = {
   jp: 'ja-JP',
+  ko: 'ko-KR',
```
roadlittledawn, line 4: we're now going to go with `kr` as the key

roadlittledawn (changes requested):
had a chat with Julien just now and to stay consistent with newrelic.com we're going to go with `kr` as the code.

PR #6244, opened 2022-02-18
Add documentation for new trace observer region in the EU.

rhetoric101:
This is a PR to introduce the new EU region and to differentiate it from the previous EU region that reports to the US data centers.

As a bonus, I changed the link from the Lambda tile in distributed tracing so it links directly to the page that discusses tracing with Lambda.


rhetoric101:
Temporarily closing to see if reopening it will kick off fresh build.

urbiz-nr (approved):
Nice @rhetoric101!

PR #6256, opened 2022-02-21
Add documentation to upgrade/downgrade Fluent Bit and support disclaimer

jsubirat:
# Description

Many of the support requests that the Logging team receives are related to the Infrastructure Agent's log forwarder. The log forwarder, based on Fluent Bit, allows the customer to define a totally customizable external configuration file, which is out of the scope of the support provided by New Relic. In this PR we include a PR informing of this, as well as instructions of how to upgrade/downgrade Fluent Bit to a particular version.


urbiz-nr (approved):
Thanks for the changes, @jsubirat!

PR #6257, opened 2022-02-21
Update Firehose documentation to reflect 1 MiB payload size

jsubirat:
This PR adds a special note on the buffer size limitation that needs to be explicitly configured when shipping Logs to New Relic using the Firehose integration.


urbiz-nr (approved):
Thanks for the updates, @jsubirat!

PR #6259, opened 2022-02-21
docs: add language identifiers

brnhensley:
I also made the indentation 4 spaces everywhere instead of a mix of 2 and 4. I also removed the unhelpful `<var>` stuff from codeblocks, I did that because it confuses the language syntax coloring where it's used.

## Give us some context


brnhensley:
Having the `<var>`s in the codeblocks makes them not able to color, like this image:
image

I personally think it would be better to avoid these for this reason, I just don't want to be the one to make style choices for the docs. Something like this would be just as effective in my opinion:

```cs
[Transaction(Web = true)]
public void Run()
{
    var uri = new Uri("YOUR_URL_HERE");
    NewRelic.Api.Agent.NewRelic.SetTransactionUri(uri);

    // your web task
}
```

urbiz-nr (approved):
Thanks for these changes, @brnhensley!

urbiz-nr:
> Having the `<var>`s in the codeblocks makes them not able to color, like this image: image
> 
> I personally think it would be better to avoid these for this reason, I just don't want to be the one to make style choices for the docs. Something like this would be just as effective in my opinion:
> 
> ```cs
> [Transaction(Web = true)]
> public void Run()
> {
>     var uri = new Uri("YOUR_URL_HERE");
>     NewRelic.Api.Agent.NewRelic.SetTransactionUri(uri);
> 
>     // your web task
> }
> ```

We'll look into this, thanks!

PR #6267, opened 2022-02-22
Remove reference to lambda

planteater:
We don't actually support open-in-IDE for lambda errors, so removed that reference. Also removed extraneous spaces in git-issues article.


planteater:
Last commit was to remove the PR comments feature, which has been removed from the product.

paperclypse (approved):
Looks good. Thanks for the update @planteater!

PR #6291, opened 2022-02-23
Query would not have worked and is not faceted

HenryTech:
Bad query example, since there was no actual FACET clause in the inner query. I edited it and also added quotes around the string value that must have quotes. I also used a real event type (SystemSample) for the example instead of a pretend one, as well as changing the attribute to match one on that event type.

## Give us some context


bradleycamacho:
@paperclypse just pinging you so you're aware of these changes!

bradleycamacho (approved):

PR #6308, opened 2022-02-24
Add a note that the endpoints are region-specific

reese-lee:
* We had a custy use the wrong endpoint for where their account is based, as it doesn't specify in the doc to use the endpoint for where your account is based, and were unable to see data in NR1.


barbnewrelic (approved):
I made a minor style guide edit and will merge this PR now. Thanks for helping to keep our docs up to date!

PR #6309, opened 2022-02-24
Add a check for correct endpoint based on region

reese-lee:
* Adds a check here about using region-specific OTLP endpoint. Complements [this PR](https://github.com/newrelic/docs-website/pull/6308) about the endpoints being region-specific.


barbnewrelic (approved):
Thanks for making the US/EU endpoint requirements clear!

PR #6324, opened 2022-02-24
Remove new material from OpenTelemetry JVM section and  delete "What's new" post

rhetoric101:
This PR removes the new material about JVMs from the OpenTelemetry UI page, and it also removes the related "What's new" post.

REVIEWER:

* Confirm that the JVM section no longer has any collapsers or procedural steps for navigating the UI (that's how you know it's the old material). YES
* Confirm that the "What's new" JVM  post from 2/17/2022 no longer appears in the list. YES -- THE FILE ITSELF IS NAMED 2-15 SO I'M ASSUMING THAT'S THE ONE YOU MEAN


barbnewrelic (approved):
Thanks for handling this! I learned a lot about WN process!

PR #6328, opened 2022-02-25
Incident Intelligence: Removing destination examples doc

urbiz-nr:
As requested in #6304.


urbiz-nr:
New url - https://docs-preview.newrelic.com/docs/incident-intelligence-destination-examples.

PR #6338, opened 2022-02-25
Add in recommendation to only use one tooling tech

reese-lee:
* This doc is not clear in saying that only ONE tooling tech (either OTel or an APM agent) should be used, and does not say that using both in any one service can cause issues. 
* This is to help clear up custy confusion, we've had several asks about it now.


reese-lee:
Actually I'm going to close this to re-edit with better wording and submit a new PR.

PR #6363, opened 2022-02-28
Small update to fix the layout on the Java Agent config file

kford-newrelic:
Fixing one line in the docs to correct layout


rhetoric101 (approved):
@kford-newrelic, this was a good catch! Thanks for fixing this!

PR #6367, opened 2022-02-28
feature: add flag to queue script to allow for targeting specific locale

clarkmcadoo:


scripts/actions/add-files-to-translation-queue.js
```diff
@@ -128,7 +133,17 @@ const main = async () => {
   const queue = await getTranslations({
     status: STATUS.PENDING,
   });
-  const fileDataToAddToQueue = translationDifference(queue, includedFiles);
+
+  const filterByLocaleOption = (file) => {
+    if (localeOption) {
+      return file.locale === localeOption;
+    } else return;
```
nr-kkenney, line 140: in the case a user doesn't supply a locale, you want everything returned.

i think this need to be a truthy valued returned by default.

scripts/actions/add-files-to-translation-queue.js
```diff
@@ -98,6 +102,7 @@ const main = async () => {
   const url = options.url || null;
   const directory = options.directory || null;
   const machineTranslation = options.machineTranslation || false;
+  const localeOption = options.locale || null;
```
nr-kkenney, line 105: looking at the usage below, do we need this separate variable here?

can we have:
```
if (options.locale)
```
on line 138

aswanson-nr (approved):

PR #6374, opened 2022-03-01
docs: fix default value for infra agent Storage sample rate

rogercoll:
The default `metrics_storage_sample_rate` value is 20


urbiz-nr (approved):
Thanks for fixing this, @rogercoll!

PR #6377, opened 2022-03-01
Firehose logs: clarification on the buffer size limitation and recommendation to use gzip

jsubirat:
This PR adds a small explanation on the buffer size limitation and the need to use GZIP compression to avoid having log messages coming from Firehose rejected.


tejunior (approved):

urbiz-nr (approved):
Thanks for the update, @jsubirat!

PR #6395, opened 2022-03-02
fix(AWS) how-to disable metric streams

josemore:
## Give us some context

* Explanations to help customers disable AWS CloudWatch metric stream integration (requires changes on their AWS account). Feedback from support cases.


barbnewrelic (approved):
Thanks for adding the CloudWatch procedure. Very helpful!

PR #6403, opened 2022-03-02
fix - style-guide - inclusive language

mmfred:
Another batch of changes for inclusive language fixes.


barbnewrelic (approved):
Good catch, thanks!

PR #6405, opened 2022-03-02
Logs API: updates about attribute types

bmcfeely:
- We support any value type for most attributes, updating text to reflect that
- Listing out supported value types
- Removing a now-obsolete restriction that attribute names not have space characters

## Give us some context

* Just something we noticed while investigating the data types we support


barbnewrelic (approved):
Looks great! Thanks for updating this API info!

PR #6417, opened 2022-03-03
fix(azure): note about 2-years secret expiration

josemore:
## Give us some context

* Azure application client secrets can now only be defined with an expiration of 2-years (previously, the "Never" option was available and recommended). Updated with a note explaining how to rotate the secret in the UI or API.


x8a (approved):
Thanks for the edits!

Issue #5992, opened 2022-02-04
"Flatten" the New Relic One category


austin-schaefer:
This ticket is part of the XD bug bash, Athena initiative, and FutureStack quality work.

## How can we make our docs better?

The [New Relic One category](https://docs.newrelic.com/docs/new-relic-one/use-new-relic-one/get-started/introduction-new-relic-one) has a vestigial `Use New Relic One` category containing all the "real" categories. Bumping everything out one level would be a nicer experience. Suggested structure in image form:

![image](https://user-images.githubusercontent.com/55203603/152613319-37389980-3b4c-4808-b36f-1131acdd418a.png)

austin-schaefer:
The key "install & configure" category is also far too deeply nested considering its importance

![image](https://user-images.githubusercontent.com/55203603/154773711-13acd687-8a06-41e8-a166-6b21d7f003ec.png)

austin-schaefer:
We're closing out all issues from 2022 and earlier. Feel free to re-open, or you can leave feedback via the "Was this doc helpful?" form in the right nav of any doc.

PR #6423, opened 2022-03-03
[DOC-7766] Flattening NROne IA

urbiz-nr:
See DOC-7766 and #5992 .


x8a (approved):
Wow! It looks much better now!

PR #6459, opened 2022-03-04
Added redirect for broken 404

### LINKED ISSUE #6456: Issue: 404
Super skeptical about New Relic now, because Azure feels like a 3rd class citizen to the point that the docs don't exist.

I had to dig pretty deep to find it at all, and ultimately the installer took me there: https://docs.newrelic.com/docs/agents/net-agent/azure-installation

And that's a 404. 

* Is something confusing?
* Is something inaccurate or missing?
* Were you unable to complete a task? (What task?)

* https://docs.newrelic.com/404/
* **OS**: Unknown
* **Browser**: Unknown
* **Device**: Unknown

paperclypse:
Closes #6456


barbnewrelic (approved):
I tested the redirect and it works, so I assume this is good to go!

PR #6481, opened 2022-03-08
fix(infra-agent): Debian 8 EOL reached

josemore:
## Give us some context

* We cannot support OS beyond the manufacturer EOL. Debian 8 LTS ended Jun-30-2020.


x8a (approved):
Thanks!

PR #6487, opened 2022-03-08
Add attribute limits info

reese-lee:
We have had a number of OpenTelemetry tickets where they experienced data being dropped due to exceeding our limits on attributes, but the numbers are not documented in the OTel section. This will make it clear that NR limits also apply to OTLP-sourced data, and include the exact numbers and links to the main limit docs for further perusal.


rhetoric101 (approved):
@reese-lee, thanks for adding this clarification!

PR #6520, opened 2022-03-10
fix(statsd): recommend latest instead of 2.0.0

josemore:
## Give us some context

* Updated docker image examples with "latest" tag instead of 2.0.0 to avoid updating docs for every release (2.1.0 is actually the latest now).


x8a (approved):
Thank you!

PR #1114, opened 2021-03-03
Remove windows installation instructions

arvdias:
### Tell us why

Windows support is not fully ready yet

PR #6529, opened 2022-03-10
fix(node agent): mention allow_all_headers is now also an envvar

jordigh:
Using `allow_all_headers` as an envvar was introduced in [#1114 of node-newrelic](https://github.com/newrelic/node-newrelic/pull/1114).


barbnewrelic (approved):
thanks for updating this Node.js config info!

PR #6546, opened 2022-03-11
fix(aws): removed unnecessary permissions

josemore:
## Give us some context

* As part of support case we confirmed some documented permissions are actually not needed.
* Also, removed the metric streams banner from these integrations as these are not replaced by CloudWatch metric streams.


x8a (approved):
Thanks a lot!

x8a (approved):

PR #6572, opened 2022-03-11
bug fixes for dashboard example JSON

keegoid-nr:
- fix `1e3` to get seconds from milliseconds
- sum first then divide


zuluecho9 (approved):

zuluecho9:
thank you much @keegoid-nr !

PR #6585, opened 2022-03-14
fix(node agent): clarify that proxy settings do not apply to Infinite Tracing

jordigh:
A client wanted to use the same proxy settings, but there's already a clear workaround for gRPC. This closes [#428 of
node-newrelic](https://github.com/newrelic/node-newrelic/issues/428).


zuluecho9 (approved):

zuluecho9:
thanks @jordigh !

jordigh:
Thanks for the edits, it reads much better that way.

PR #6589, opened 2022-03-14
Clarify how to avoid rate limiting

reese-lee:
This PR makes it clearer the config settings we recommend to avoid getting rate limited. It also adds more information to the attributes section on how to set max attribute lengths using span limits env vars. Both have been common issues with our OpenTelemetry users.


zuluecho9 (dismissed):

rhetoric101 (approved):
@reese-lee, thanks for making these changes!! They look great!

Issue #6592, opened 2022-03-14
Issue: Organize data with partitions


disqodesiree:
## How can we make our docs better?

In the first steps of this doc, it tells the reader to ensure they have the correct permissions and a partition plan but doesn't reference what those are.
[https://docs.newrelic.com/docs/logs/ui-data/data-partitions/#plan](url)

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/logs/ui-data/data-partitions/
* **OS**: Mac OS 10.15.7
* **Browser**: Chrome 99.0.4844.51
* **Device**: Unknown

zuluecho9:
thanks @disqodesiree. I'll work on that in https://github.com/newrelic/docs-website/pull/6594.

PR #6594, opened 2022-03-14
fix(logs): clarify roles/permissions

zuluecho9:
Did this due to issue https://github.com/newrelic/docs-website/issues/6592. Clarifying how roles/permissions work for logs and pointing to that requirements section from log partitions doc.


barbnewrelic (approved):
These edits are very helpful. Thanks!

PR #6601, opened 2022-03-15
fix: remove `ul` elements in favor of markdown

roadlittledawn:
## Description

Removes `ul` and `li` html tags in favor of markdown to fix this serialization error when sending file for translation.

https://github.com/newrelic/docs-website/runs/5547110228?check_suite_focus=true


x8a (commented):
It looks like there's a problem with a `<Callout>` in line 108 but I don't seem to spot it.

roadlittledawn:
@x8a i think it was the children of the `Callout` not being indented. just pushed that up so we shall see

nr-kkenney (approved):

PR #6602, opened 2022-03-15
Clarify when PMM review is necessary and add PMM to CODEOWNERS

rhetoric101:
It turns out that PMMs who submit "What's new" posts don't need a review from a fellow PMM.

I also replaced on PMM with another in the CODEOWNERS list of reviewers.


urbiz-nr (approved):
It's much more clear now, thanks!

PR #6614, opened 2022-03-15
Add .NET agent 9.6.1 release notes

nr-ahemsath:
Add release notes for .NET agent 9.6.1.


rhetoric101 (approved):
@nr-ahemsath, thanks for adding these release notes.

PR #6615, opened 2022-03-15
Remove accidental kr locale addition

LizBaker:
We accidentally merged the korean UI code before it was ready, this will revert that


roadlittledawn (approved):
🙈

PR #6617, opened 2022-03-15
Added callout about attributes config param

nedl86:
Attributes config param doesn't take effect when an external Fluent Bit configuration file is specified. Added callout explaining this.


rhetoric101 (approved):
@nedl86, thanks for adding this callout!

PR #6619, opened 2022-03-16
Add trace observer region list

rhetoric101:
We've never had a list of trace observer region, so this PR adds the list in two places in the trace observer doc:

* In the APM/OTEL section
* In the browser, mobile, and lambda section

This will take some work to maintain, but I think it's easier for readers to find the information right when they need it.

REVIEWER: 

This just needs confirmation that it is readable. I don't need a review of the who document.


urbiz-nr (approved):
Nice update, @rhetoric101!

PR #6622, opened 2022-03-16
Add a missing 's'

jlegoff:
## Give us some context

There is a missing 's' in a sentence


urbiz-nr (approved):
Thanks so much, @jlegoff!

PR #6623, opened 2022-03-16
Docs metrics adapter correct AccountID value path and modify chart links

alvarocabanas:
- We have moved the chart for the metrics adapter to its own repository.
- The nri-bundle installation keeps working as it was, but the links to the chart needed to be modified.
- We spotted also a wrong path for the accountID path.


x8a (approved):
Thanks!

PR #6665, opened 2022-03-18
feat(cloud-int): enforcing system limits

josemore:
## Give us some context
* We updated the way we handle metrics from new resources when we detect a large amount of new entities to be registered. We are publicly documenting the limits and behavior.


paperclypse:
Thanks for the PR, @josemore. Please let us know when you'd like us to help you get this published.

urbiz-nr (approved):

PR #6666, opened 2022-03-18
Capping the Apache Httpclient version

kford-newrelic:
In response to confirmation that existing instrumentation does not work for the newer 5.x client version


paperclypse (approved):
Looks good to me!

PR #6678, opened 2022-03-18
Update browser agent creds 

LizBaker:
We left the browser agent entity for the docs site in the nr1 dev experience account, now we are moving it to our docs eng account, this updates the info


roadlittledawn (approved):
:shipit:

PR #6686, opened 2022-03-21
Add collaboration page for making outlines

rhetoric101:
This is a style guide addition to help the collaboration between contributors and writers for making starting outlines.


zuluecho9:
We decided to close this, as we think maybe we are replicating basic github instructions, and we'll revisit if we keep running into problems with contributors not knowing how to contribute. thank you for your service, @rhetoric101 .

PR #6687, opened 2022-03-21
Licenses: fix broken links

urbiz-nr:
The licenses doc had a few links to the old automatic index pages, I'm fixing them here.


urbiz-nr:
Works well in the deploy.

PR #6690, opened 2022-03-21
fix(metrics): broken link to metric types

josemore:
## Give us some context

* Broken link to metric data types


urbiz-nr (approved):

PR #6700, opened 2022-03-21
Update heroku-log-forwarding.mdx

maxlemieux:
Add missing space in command to print Heroku drains in JSON format.

The existing command in docs will not run as it interprets the last part `--json` as part of the Heroku application name. Adding a space fixes this.


paperclypse (approved):
Looks good! Thanks for helping out!

PR #6708, opened 2022-03-22
chore(oma-datagov): Added two missing drop rule examples

JimHagan:
## Give us some context

* Added two missing examples of drop rule types that are important for data optimization.


urbiz-nr (approved):
Thanks for the update, @JimHagan!

PR #6725, opened 2022-03-22
Improve .NET agent install instructions for Linux

nr-ahemsath:
Two main changes to the .NET agent install instructions for Linux:

1. Switch to specifying `https://` URLs for the APT and YUM package manager configuration options instead of unsecure `http://` URLS.
2. Clarify the yum/rpm install instructions for rpm-based Linux distros (e.g. RHEL, Centos, etc.).  You can install using the package manager (YUM, which requires configuring the YUM repository) or with RPM (which requires downloading the .rpm file directly from the downloads site).  This mirrors the APT/DPKG (.deb) distinction that already exists for Debian-base distros (e.g. Ubuntu).

I hope I got the formatting right, please fix it if I didn't.


rhetoric101 (approved):
@nr-ahemsath, thanks for updating these instructions!

PR #6737, opened 2022-03-23
fix: removing unnecessary backticks

rudouglas:
Some back ticks `` in the code block that shouldn't be there


urbiz-nr (approved):
Thanks, @rudouglas!

PR #6740, opened 2022-03-23
chore: Adding lookout standalone experience to user type documentation

aeliaNR:
## Give us some context

*Note - I am a New Relic employee. Please feel free to ping me on Slack if there are any ambiguities.*

  * Lookout's standalone experience is not currently included in the detailed capability comparison table. This PR aims to explicitly cite the user and account requirements for Lookout. Lookout standalone is only accessible to Full users belonging to a Pro or Enterprise account. It is otherwise accessible via Explorer as already called out in the document.


zuluecho9:
Just a note: I'm looking into this. Changes to that table requires getting some more eyes on it.

PR #6786, opened 2022-03-25
fix(release note): user upgrade additions

zuluecho9:
For v2 user mgmt: more controls over specific upgrades (basic to core, core to full, basic to full; all paths)


paperclypse (approved):
Looks good to me

PR #6826, opened 2022-03-29
fix(aws): note about missing metrics

josemore:
## Give us some context

* From support tickets: customers expect a note explaining certain metrics might not be available in metric streams. There's no official list of metrics and these might be updated / implemented by AWS at any point of time so just keeping a reference of the namespaces where we've seen missing metrics.


x8a (approved):
Thanks!

PR #6832, opened 2022-03-29
fix: remove autoLinkHeaders from md files

clarkmcadoo:
Removing the plugin feature that was adding links to headers via gatsby and creating icons as links that didn't go anywhere:

Screen Shot 2022-03-29 at 10 54 45 AM

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -209,10 +209,6 @@ module.exports = {
               maxWidth: 850,
             },
           },
-          autoLinkHeaders,
-          // This MUST come after `gatsby-remark-autolink-headers` to ensure the
-          // link created for the icon has the proper id
-          'gatsby-remark-custom-heading-ids',
         ],
       },
     },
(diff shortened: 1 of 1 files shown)


gatsby-config.js
```diff
@@ -209,10 +209,6 @@ module.exports = {
               maxWidth: 850,
             },
           },
-          autoLinkHeaders,
-          // This MUST come after `gatsby-remark-autolink-headers` to ensure the
-          // link created for the icon has the proper id
-          'gatsby-remark-custom-heading-ids',
```
roadlittledawn, line 215: i think we may still want the `gatsby-remark-custom-heading-ids` plugin

roadlittledawn (changes requested):

gatsby-config.js
clarkmcadoo: ok, i can put that back

roadlittledawn (approved):
i still see heading links on MDX pages on front end. and the heading links are no longer in the whats new JSON.

PR #6841, opened 2022-03-30
fix F5 integration CA_FILE_FILE setting

carlossscastro:
our `CA_BUNDLE_FILE` config setting mentions a `USE_SSL` setting that does not exist. This PR removes that.


x8a (approved):
Thanks!

PR #6852, opened 2022-03-30
removed combining sli and updating synthetics guided flow

a-james-faria:
* Removed instructions on combining APM service levels as this has proven to create unnecessary complexity with the inability to determine which SLI would cause a degraded service level. 
* Updated input performance (synthetics) instructions to include the new guided flow just released.


barbnewrelic (approved):
Thanks for keeping this doc up to date!

PR #6854, opened 2022-03-30
AccountId() NRQL function (not yet ready for publish)

mlobiondo-newrelic:
This is waiting to be published until some time in April. This is documenting a NRQL function accountId() that allows you to return the accounts associated with the data queried. This is currently only for nerdgraph; query builder ability will come later (currently you can only run queries in UI query builder that are constrained by the account you're in but with nerdgraph you can cross account boundaries). 

…-website/commit/567d3b1ea5975e0e12fcdfb7c002087166bfe1c7

## Give us some context

### DIFF
--- a/src/content/docs/query-your-data/nrql-new-relic-query-language/get-started/nrql-syntax-clauses-functions.mdx
+++ b/src/content/docs/query-your-data/nrql-new-relic-query-language/get-started/nrql-syntax-clauses-functions.mdx
@@ -1833,11 +1833,7 @@ Use non-aggregator functions for non-numerical data in NRQL queries.
     className="freq-link"
     title={<InlineCode>accountId()</InlineCode>}
   >
-Use the `accountId()` function to return the [account ID](/docs/accounts/accounts-billing/account-structure/account-id) for queried events. This is useful when querying data that crosses multiple New Relic accounts.
-
-This function takes no arguments.
-
-Here are some example queries: 
+Use the `accountId()` function to return the [account ID](/docs/accounts/accounts-billing/account-structure/account-id) associated with queried data. This function takes no arguments. Here are some example queries: 
 
     <CollapserGroup>
       <Collapser title="Get the account ID for each Transaction event">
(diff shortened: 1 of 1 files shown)


mlobiondo-newrelic:
Hi @bradleycamacho, this is the new version of the accountId() PR (the changes are the same).  This still should not be merged yet.

zuluecho9:
For other tech writers: I did a review of this, for what it's worth. When this goes live, if it's not yet available to do this in query builder UI, I think we should add a note that this is not available in the UI, but only available when using NerdGraph queries. (As Mlobiondo said "you can use cross-account queries in dashboards via NerdGraph, JSON or Terraform, though not the Query Builder", and we maybe need to dig into that a little more.

mlobiondo-newrelic:
> For other tech writers: I did a review of this, for what it's worth. When this goes live, if it's not yet available to do this in query builder UI, I think we should add a note that this is not available in the UI, but only available when using NerdGraph queries. (As Mlobiondo said "you can use cross-account queries in dashboards via NerdGraph, JSON or Terraform, though not the Query Builder", and we maybe need to dig into that a little more.

@zuluecho9, the accountId() NRQL function described in this docs update is available via the Query Builder like any other NRQL function.  What's not available via the Query Builder is Cross-Account Queries, a related feature that will have its own documentation when it's GAed.  We don't want to mention cross-account queries in the public docs for accountId() since cross-account queries are only in beta at this point.

zuluecho9:
@mlobiondo-newrelic Is there much practical use of the accountId() function if they don't have the cross-account functionality? I guess that's what I'm not getting. Is there still a practical use to using it? I would think it would always be returning the same account IDs if they don't have cross-account ability. if that makes sense.

mlobiondo-newrelic:
@zuluecho9 It's not particularly useful when not used in a cross-account query.  It will always return the same account ID on a query run against a single account.  It's generally available now, though, while cross-account queries are not quite ready.

zuluecho9 (approved):

PR #6858, opened 2022-03-30
chore: add workflow to add relevant prs to docs eng beta board

roadlittledawn:
## Description

Adds a workflow to add relevant PRs to Docs Eng board.

Had to use different API than other related workflows to triage issues / PRs to other boards because the Docs Eng Board is a beta project board.

## Links
The gihutb doc [Automating projects (beta)](https://docs.github.com/en/issues/trying-out-the-new-projects-experience/automating-projects) was helpful

### DIFF
--- a/.github/workflows/pr-docs-eng-board-beta.yml
+++ b/.github/workflows/pr-docs-eng-board-beta.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
       - name: Get project data
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


roadlittledawn:
Looks like I'd need to create a github app because the `GITHUB_TOKEN` cannot access projects beta. 

> Note: GITHUB_TOKEN is scoped to the repository level and cannot access projects (beta). To access projects (beta) you can either create a GitHub App (recommended for organization projects) or a personal access token (recommended for user projects). Workflow examples for both approaches are shown below.

x8a (approved):
It all looks good but we as members of the org don't have the permissions to test it locally.

x8a:
Merging to test it works in GitHub.

PR #6865, opened 2022-03-31
Add eu endpoint to heroku log forwarding docs

danybmx:
Hi!

Just adding the EU endpoint for heroku logs since it's different to the US one.
This will allow EU customers to configure heroku drains correctly

Regards!


x8a (approved):
Thanks for the addition!

PR #6899, opened 2022-04-02
Update "Additional collector set up examples" link

reese-lee:
Updates the link to the "Additional examples" hyperlink to point directly at our Collector examples directory instead of the main examples repo, which can be confusing to navigate. This link takes the user directly to the two examples we have (at the moment).


rhetoric101 (approved):
@reese-lee thanks for making this easier for users with this direct link!

PR #6904, opened 2022-04-04
fix: change class to className

clarkmcadoo:


roadlittledawn:
did we test by running serialization locally on a file with `inline code` in the body and see if we get:

```
<code class="notranslate">TEXT HERE</code>
```

last i looked into it, i tried adding a handler for inline code (different than `InlineCode`) and when i did a test to serialize a file, could not get it to output with the class.

clarkmcadoo:
output for the new `<code>` element serializer:

```
<div data-type="component" data-component="code" class="notranslate">
      <p>agent.report_custom_element</p>
</div>
```

scripts/actions/__tests__/serialize-mdx.test.js
```diff
@@ -312,3 +312,13 @@ test("test <InlineCode> element serializes and adds 'notranslate' class to eleme
 
   expect(html).toMatchSnapshot();
 });
+
+test("test <code> element serializes and adds 'notranslate' class to element", async () => {
+  const mdx = `
+<code>agent.report_custom_element</code>
```
roadlittledawn, line 318: doesn't the MDX string here need to look like:
```
`agent.report_custom_element`
```

clarkmcadoo: we have instances of `<code>thing</code>` in MDX files, so i think we need both, i'll add a test

roadlittledawn (changes requested):
i checked out the PR and ran a test serialization on `src/content/docs/distributed-tracing/trace-api/report-new-relic-format-traces-trace-api.mdx`

here's what the serialized HTML of the paragraphs that had inline code (in backticks in the MDX file). 

```html
<li>Sending trace data in the expected format (in this case, our <code>newrelic</code> format).</li>
```

scripts/actions/__tests__/serialize-mdx.test.js
roadlittledawn: i only see one MDX file in `src/content/docs` that uses `<code>`. Backticks are def the most common way to add inline code. 

And good call to add a test.

clarkmcadoo:
Did a lot of testing, here are the results!:

```
  console.log
    input -
    `test`

      at Object.<anonymous> (scripts/actions/__tests__/deserialize-html.test.js:203:11)

  console.log
    html -
    <p><code class="notranslate">test</code></p>

      at Object.<anonymous> (scripts/actions/__tests__/deserialize-html.test.js:205:11)

  console.log
    mdx - `test`

      at Object.<anonymous> (scripts/actions/__tests__/deserialize-html.test.js:207:11)
```

scripts/actions/__tests__/serialize-mdx.test.js
clarkmcadoo: right right right, i was looking at a snapshot turns out

roadlittledawn (approved):
nice find on that little package! ran a little serialization myself locally and looks good! good stuff!

PR #6917, opened 2022-04-04
fix(guides): removing gone docs, fixing formats

zuluecho9:
DOC-7862. First batch of work for improving best practices/guides section and OM docs. Work done: 
• Removing links to docs deleted months ago
• Removing dummy doc links in guides section 
• Making it more clear that the two cloud migration docs were actually about cloud migration
• Some assorted reformatting and error fixing, as noticed


src/content/docs/new-relic-solutions/observability-maturity/introduction.mdx
```diff
@@ -40,79 +40,54 @@ import omaVDrivers from 'images/oma-v-drivers.png'
     id="upr"
     title="Uptime, performance, and reliability"
   >
-    ### Implementation guides
+### Implementation guides
```
paperclypse, line 43: Even though these are H3s, we may want anchor tags

src/nav/new-relic-solutions.yml
```diff
@@ -50,42 +49,28 @@ pages:
             path: /docs/new-relic-solutions/observability-maturity/innovation-growth/development-quality-implementation-guide
           - title: Release quality
             path: /docs/new-relic-solutions/observability-maturity/innovation-growth/release-quality-implementation-guide
-  - title: New Relic solutions
+  - title: Best practice guides
```
paperclypse, line 52: Couldn't add a comment above for some reason. 

What if we changed the nav title "Observability maturity" to something more active like "Achieve observability maturity"?

"Discover our best practices"
"Use our cloud migration solutions"

Solve common issues is already there :)

paperclypse (approved):
Looks good! Left a couple comments for you to think about, but otherwise I approve.

src/nav/new-relic-solutions.yml
zuluecho9: i agree. i'll think more about this one later. I'll be reviewing the docs more shortly.

Issue #6923, opened 2022-04-05
Wrong header name to specify the API Key


asllop:
The cURL examples to request alerts don't work. In the doc, the header name used to specify the API Key is `Api-Key`, but using this causes an error:

```
$ curl -X GET 'https://api.newrelic.com/v2/alerts_incidents.json' -H 'Api-Key:XXXXXXXXXX' -i 
HTTP/1.1 401 Unauthorized
Proxied-By: Service Gateway
Content-Security-Policy: frame-ancestors *.newrelic.com
Cache-Control: no-cache
Content-Length: 42
Content-Type: application/json
Date: Tue, 05 Apr 2022 10:49:49 GMT
Server: nginx
Status: 401 Unauthorized
X-Rack-Cache: miss
X-Request-Id: 7c602ba6d0e70b90ec1726b82b14fb3b
X-Runtime: 0.010482
X-Ua-Compatible: IE=Edge,chrome=1

{"error":{"title":"No API key specified"}}
```

Changing it by `X-Api-Key` works.

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/alerts-applied-intelligence/new-relic-alerts/advanced-alerts/rest-api-alerts/rest-api-calls-alerts/
* **OS**: Unknown
* **Browser**: Unknown
* **Device**: Unknown

PR #6927, opened 2022-04-05
docs(Alerts): Amend 'Api-Key' header to 'X-Api-Key'

founddrama:
## Give us some context

- In response to #6923 
- Amends Alerts REST API docs to refer to `X-Api-Key` (instead of `Api-Key`)
- Cross-referenced against the [API explorer](https://rpm.newrelic.com/api/explore)


x8a (approved):
Thank you for the update!

PR #6937, opened 2022-04-05
style(data types): Simplify language, add MELT mentions

austin-schaefer:
Mostly minor language tweaks, plus inserting several references to MELT to make it clearer that this doc covers the four key observability data types.

### DIFF
--- a/src/content/docs/data-apis/understand-data/new-relic-data-types.mdx
+++ b/src/content/docs/data-apis/understand-data/new-relic-data-types.mdx
@@ -1,5 +1,5 @@
 ---
-title: New Relic data types: Metrics, Events, Logs, and Traces (MELT)
+title: "New Relic data types: Metrics, Events, Logs, and Traces (MELT)"
 tags:
   - Ingest and manage data
   - Understand data
--- a/src/content/docs/data-apis/understand-data/new-relic-data-types.mdx
+++ b/src/content/docs/data-apis/understand-data/new-relic-data-types.mdx
@@ -1,11 +1,11 @@
 ---
-title: New Relic data types
+title: New Relic MELT data types
 tags:
   - Ingest and manage data
   - Understand data
 translate:
   - jp
-metaDescription: 'An overview of the types of data New Relic reports: metrics, events, logs, and traces (spans).'
+metaDescription: 'An overview of the MELT data New Relic reports: metrics, events, logs, and traces (spans).'
 redirects:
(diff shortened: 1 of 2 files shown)


src/content/docs/data-apis/understand-data/new-relic-data-types.mdx
```diff
@@ -59,6 +53,8 @@ Metrics are a strong solution for storing data long-term, and understanding tren
 
 Conceptually, "metrics" is a broad, general category. There are various ways New Relic measures and reports metrics but, in practice, when using the New Relic UI, you usually won't have to understand how exactly this happens. In our documentation, we typically will just refer to "metrics," regardless of how that data is reported, unless there's a reason you need to know more (like understanding [how to query your data](/docs/using-new-relic/data/understand-data/query-new-relic-data)).
 
+For a list of the metrics we collect, see [our documentation on metrics](/docs/telemetry-data-platform/ingest-manage-data/understand-data/metric-data-type).
```
zuluecho9, line 56: I don't think this is technically correct. that doc contains the types of metrics within the `Metric` data type. It's not including the other types of metrics we collect (e.g., metric timeslice data, which is still our main metric data source for APM, Browser, and Mobile, and not the metrics reported via events, like for infrastructure; and I'm not sure the types of metric data in this doc (gauge, count, and such) count as a 'list of metrics', as I'd maybe expect a 'list of metrics' to be more about the specific types of data). If  we're going to point to that doc, we might want to point to it from the 'dimensional metrics' entry. I'm not sure of the goal we're trying to accomplish with this addition.

src/content/docs/data-apis/understand-data/new-relic-data-types.mdx
```diff
@@ -1,11 +1,11 @@
 ---
-title: New Relic data types
+title: New Relic MELT data types
```
zuluecho9, line 2: I was told previously that we shouldn't use MELT like this in public docs, that it was more of an occasional marketing thing, and some people seemed conflicted on whether it even made sense marketing-wise. did that change? making sure. if so, should we be promoting "MELT" more widely?

src/content/docs/data-apis/understand-data/new-relic-data-types.mdx
austin-schaefer: This was an existing sentence from the doc that I re-arranged; happy to just delete it.

PR #6943, opened 2022-04-06
feat(statsd): add unix socket support

josemore:
## Give us some context

* Adding a note explaining a new capability in StatsD integration.


x8a (approved):
Thanks!

PR #6964, opened 2022-04-06
Add datasource meta tag

LizBaker:
Closes DOCENG-31 (jira)

Adds a dataSource meta tag when available from frontmatter

Screen Shot 2022-04-07 at 1 20 57 PM


clarkmcadoo:
Are we concerned at all about the build failing here? or is this part of the nastiness that went down yesterday?

caylahamann (approved):
💯

PR #6977, opened 2022-04-07
fix(Alerts): Amends Alerts-related navigation click paths

founddrama:
## Give us some context

Observed that some of the Alerts-related documentation had clickpaths that referred to old navigation labels that have since been updated. **I have amended the labels in these clickpaths to reflect the current state of the NR1 navigation for "Alerts & AI".**


x8a (approved):
Thanks for the updates!

PR #6982, opened 2022-04-07
fix broken link

keegoid-nr:
- fix link to pricing model details


barbnewrelic (approved):
Thanks for catching that!

PR #6997, opened 2022-04-08
Replace incorrect message limit on NoticeError api method

jaffinito:
## Give us some context

The comments for the `string message` parameter in the .NET Agent's NoticeError API page refers to a 1000 character limit.   This is not correct (it might have been 8 or so years ago). 

This updates the comment for that parameter to match what my testing has shown to be correct.  We will also be updating the in-code comments to reflect this as well.


x8a (approved):
Thanks a lot for the update!

PR #7004, opened 2022-04-08
Remove image of signup page

planteater:
Was going to replace it with one that includes the new Region drodown, but instead we'll just remove the image.


zuluecho9:
thanks @planteater .

zuluecho9 (approved):

PR #7013, opened 2022-04-08
Update Synthetics EU IP range listing

sshah-nr:

## Give us some context

    - Updating IP range listing for EU accounts


zuluecho9:
thanks @sshah-nr: just curious; do you happen to think this might be an issue in other synthetics docs? Or is it constrained to this doc?

zuluecho9 (approved):

sshah-nr:
Hey @zuluecho9 this change lists an additional URL under the EU accounts section and then yes transposes the URL for the existing link under the US accounts section. These links don't exist anywhere else so I'm not too concerned about it breaking elsewhere.

Also both the old and new urls point to the same location in S3 but I transposed it to standardize the format.

PR #7030, opened 2022-04-11
Fix 'applied intelligence' caps

zuluecho9:
decapped: 
applied intelligence
proactive detection
incident intelligence


paperclypse (approved):
Looks good! I didn't check out the build preview, but I assume that you've done that.

Issue #7005, opened 2022-04-08
Transaction.httpResponseCode documentation is missing deprecation notices


spencer-carver:
## How can we make our docs better?

* something is inaccurate or missing

[Transaction.httpResponseCode](https://docs.newrelic.com/attribute-dictionary/?event=Transaction&attribute=httpResponseCode) should indicate the property is deprecated in certain agents, such as the [nodejs agent](https://docs.newrelic.com/docs/apm/agents/nodejs-agent/attributes/nodejs-agent-attributes/) > 7.0.0.

I also don't see a mention on the docs page of the `http` attribute, which replaces it.

### Doc information (don't delete this section)

* https://docs.newrelic.com/attribute-dictionary/
* **OS**: Mac OS 10.15.7
* **Browser**: Chrome 100.0.4896.75
* **Device**: Unknown

PR #7031, opened 2022-04-11
fix(data dict): add missing http entries

zuluecho9:
Based on issue https://github.com/newrelic/docs-website/issues/7005. And on convo with node team. Work done: 

added two new attributes to data dictionary. 
Put a reference to the other attribute in the equivalent attributes in data dictionary
changed some example NRQL to use the newer attribute


barbnewrelic (approved):
Since your SMEs have already reviewed your updates, I'm sure this is good to go. Thanks for updating this info!

PR #7032, opened 2022-04-11
Correct .NET Agent Azure Site Extension paths

JcolemanNR:
Minor change to paths


JcolemanNR:
@zuluecho9 Good thinking. Fortunately no, the paths will still be the same for everything else. This change is specific to how the Azure Site Extension deploys the agent.

zuluecho9 (approved):

PR #7035, opened 2022-04-11
Fix name of log fowarding config element

nr-ahemsath:
Fix incorrect name of config element (for log fowarding, the name of the config element is just `forwarding`, not `logForwarding`) in main .NET agent config doc.


zuluecho9:
thank you @nr-ahemsath

zuluecho9 (approved):

PR #7045, opened 2022-04-11
Document environment variable options for configuring .NET agent logging features

nr-ahemsath:
Update the new application logging feature config options for the .NET agent to include the environment variable config options.  Thanks to @Lonli-Lokli for pointing out this gap in our documentation.

### DIFF
--- a/src/content/docs/apm/agents/net-agent/configuration/net-agent-configuration.mdx
+++ b/src/content/docs/apm/agents/net-agent/configuration/net-agent-configuration.mdx
@@ -3086,7 +3086,12 @@ request.headers.user-agent
 </Callout>
 
 ### Application logging [#application_logging]
-
+    
+    
+<Callout variant="important">
+These configuration options are only available for .NET agent version 9.7.0 or higher.
+    </Callout>
+    
 The `applicationLogging` element is a child of the `configuration` element. Use `applicationLogging` to configure instrumentation of your application's logging activity.
 
 There are three main sub-features:
@@ -3114,10 +3119,6 @@ NEW_RELIC_APPLICATION_LOGGING_FORWARDING_ENABLED="false"
 NEW_RELIC_APPLICATION_LOGGING_FORWARDING_MAX_SAMPLES_STORED="10000"
 NEW_RELIC_APPLICATION_LOGGING_LOCAL_DECORATING_ENABLED="false"
 ```
-
-<Callout variant="important">
-  This configuration option is only available in the .NET agent v9.7.0 or higher.
-</Callout>
 
 The `applicationLogging` element supports the following attributes and sub-elements:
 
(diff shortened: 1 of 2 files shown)


src/content/docs/apm/agents/net-agent/configuration/net-agent-configuration.mdx
```diff
@@ -3105,6 +3105,16 @@ For more details, see our documentation about [APM logs in context](/docs/apm/ne
 </applicationLogging>
 ```
 
+These features can also be configured via environment variables:
+
+```
```
zuluecho9, line 3110: @nr-ahemsath curious if you think it makes sense to include the values for these (="true") or if it'd be better to leave the values off if just giving examples.

src/content/docs/apm/agents/net-agent/configuration/net-agent-configuration.mdx
```diff
@@ -3105,6 +3105,16 @@ For more details, see our documentation about [APM logs in context](/docs/apm/ne
...
+NEW_RELIC_APPLICATION_LOGGING_ENABLED="true"
+NEW_RELIC_APPLICATION_LOGGING_METRICS_ENABLED="true"
+NEW_RELIC_APPLICATION_LOGGING_FORWARDING_ENABLED="false"
+NEW_RELIC_APPLICATION_LOGGING_FORWARDING_MAX_SAMPLES_STORED="10000"
+NEW_RELIC_APPLICATION_LOGGING_LOCAL_DECORATING_ENABLED="false"
+```
+
 <Callout variant="important">
```
zuluecho9, line 3118: you placed your addition in between the thing above and this callout so I'm curious if you have an opinion about where that callout should live.

src/content/docs/apm/agents/net-agent/configuration/net-agent-configuration.mdx
nr-ahemsath: I was back and forth on it myself; in the end I decided to parallel the XML config file example shown immediately above the new section by having some concrete values.  I could go either way.

src/content/docs/apm/agents/net-agent/configuration/net-agent-configuration.mdx
nr-ahemsath: All of the config options under `<applicationLogging>` are only available in 9.7.0+, so at the very least we should change the wording to be "These configuration options are only available in the .NET agent v9.7.0 or higher."

I could see it going before the XML config examples.  If you agree I can make the changes myself.

zuluecho9: I'll get it. Thanks!

zuluecho9 (approved):

PR #7066, opened 2022-04-12
Remove Linux system memory bug callout

nr-ahemsath:
A community member pointed out that we fixed the bug where the .NET agent did not report total system memory usage on Linux, and that we should remove this bug from the list of known issues on the compatibility and requirements page for .NET core applications.


rhetoric101 (approved):
@nr-ahemsath, thanks for updating this list! Be sure to thank for contributor who suggested this!

PR #7077, opened 2022-04-13
.NET agent 9.7.1 release notes, etc.

nr-ahemsath:
This PR has:

1. Release notes for .NET agent 9.7.1
2. An edit to the .NET agent 9.7.0 release notes to point users to use 9.7.1 if they install using the MSI installer
3. A few more updates to other .NET agent docs regarding the new logging features

cc: @jaffinito


barbnewrelic (approved):
Thanks for making all these updates!

PR #7079, opened 2022-04-13
feat(security): Updated intro to reflect new processes, removed FEDRA…

akristen:
Some updates in New Relic encryption methods have rolled out since this doc was last edited. After talking a SME, it was decided to:

- Shorten the intro
- Remove mentions of enhanced security
- Remove FedRAMP section

The biggest concern with this doc is letting customers know that it's no longer necessary to do extra procedures for "enhanced security," as they already receive that as soon as they engage New Relic for data storage. We are keeping the doc live because we will still provide support for existing customers using the two-tiered system.


paperclypse:
Great PR description, @akristen! Super helpful

paperclypse (approved):
Looks good! Left some comments in CodeStream

paperclypse (approved):
I totally dig this new intro! Nice work!

PR #7106, opened 2022-04-14
fix(style guide): Add color around agent vs integration

austin-schaefer:
Just adding a bit more context on the distinctions and usage. And changing quotes into code.

### DIFF
--- a/src/content/docs/style-guide/word-choice/usage-dictionary.mdx
+++ b/src/content/docs/style-guide/word-choice/usage-dictionary.mdx
@@ -34,7 +34,7 @@ For consistency across New Relic content, here are our general word usage guidel
 
     * Don't capitalize: use `the Ruby agent`, not `the Ruby Agent`.
     * Don't refer to a synthetics [`private minion`](/docs/synthetics/new-relic-synthetics/private-locations/install-configure-private-minions) as an `agent`.
-    * With regards to how `agents` relate `integrations`: An integration is anything not built by New Relic that we can inter-operate with. That could be something we monitor, an alerting service like PagerDuty, a service that pulls data from our APIs, etc. Agents are software that enable one or more  integrations. For example, our APM agents have integrations with various app frameworks and web servers and types of databases. One way to frame this is: "The integrations included with our agents are just some of [our many integrations](https://newrelic.com/instant-observability)" or "Our agent-based integrations..."
+    * With regards to how `agents` relate to `integrations`: An integration is anything not built by New Relic that we can inter-operate with. That could be something we monitor, an alerting service like PagerDuty, a service that pulls data from our APIs, etc. Agents are software that enable one or more integrations. For example, our APM agents have integrations with various app frameworks and web servers and types of databases. One way to frame this is: "The integrations included with our agents are just some of [our many integrations](https://newrelic.com/instant-observability)" or "Our agent-based integrations..."
   </Collapser>
 
   <Collapser
--- a/src/content/docs/style-guide/word-choice/usage-dictionary.mdx
+++ b/src/content/docs/style-guide/word-choice/usage-dictionary.mdx
@@ -32,9 +32,9 @@ For consistency across New Relic content, here are our general word usage guidel
 
     Style guidelines:
 
-    * Don't capitalize: use "the Ruby agent," not "the Ruby Agent."
+    * Don't capitalize: use `the Ruby agent`, not `the Ruby Agent`.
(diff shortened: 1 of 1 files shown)


src/content/docs/style-guide/word-choice/usage-dictionary.mdx
```diff
@@ -32,9 +32,9 @@ For consistency across New Relic content, here are our general word usage guidel
...
-    * With regards to how the concept of "agents" relate to the concept of "integrations": agents can be understood as containing various integrations. For example, our APM agents have integrations with various app frameworks and web servers and types of databases. One way to frame this is: "The integrations included with our agents are just some of [our many integrations](https://newrelic.com/instant-observability)" or "Our agent-based integrations..."
+    * With regards to how `agents` relate `integrations`: An integration is anything not built by New Relic that we can inter-operate with. That could be something we monitor, an alerting service like PagerDuty, a service that pulls data from our APIs, etc. Agents are software that enable one or more  integrations. For example, our APM agents have integrations with various app frameworks and web servers and types of databases. One way to frame this is: "The integrations included with our agents are just some of [our many integrations](https://newrelic.com/instant-observability)" or "Our agent-based integrations..."
```
paperclypse, line 37:
```suggestion
    * With regards to how `agents` relate to `integrations`: An integration is anything not built by New Relic that we can inter-operate with. That could be something we monitor, an alerting service like PagerDuty, a service that pulls data from our APIs, etc. Agents are software that enable one or more integrations. For example, our APM agents have integrations with various app frameworks and web servers and types of databases. One way to frame this is: "The integrations included with our agents are just some of [our many integrations](https://newrelic.com/instant-observability)" or "Our agent-based integrations..."
```

paperclypse (approved):
Looks good!

src/content/docs/style-guide/word-choice/usage-dictionary.mdx
```diff
@@ -32,9 +32,9 @@ For consistency across New Relic content, here are our general word usage guidel
...
-    * With regards to how the concept of "agents" relate to the concept of "integrations": agents can be understood as containing various integrations. For example, our APM agents have integrations with various app frameworks and web servers and types of databases. One way to frame this is: "The integrations included with our agents are just some of [our many integrations](https://newrelic.com/instant-observability)" or "Our agent-based integrations..."
+    * With regards to how `agents` relate to `integrations`: An integration is anything not built by New Relic that we can inter-operate with. That could be something we monitor, an alerting service like PagerDuty, a service that pulls data from our APIs, etc. Agents are software that enable one or more integrations. For example, our APM agents have integrations with various app frameworks and web servers and types of databases. One way to frame this is: "The integrations included with our agents are just some of [our many integrations](https://newrelic.com/instant-observability)" or "Our agent-based integrations..."
```
zuluecho9, line 37: this makes it sounds like the 'integration' is the third-party service or platform, and not the actual connection between New Relic and the other thing. I think this is an important distinction. I would say the 'integration' is just the conceptual linkage, not an actual specific thing. This may seem ambiguous, but I think there's necessarily some ambiguity here (for example, we may say we have an 'AWS integration', but then within that, we have assorted other AWS service integrations, so there can be various overlaps, and I think that's fine).

zuluecho9: am I to understand that we should be using backticks for everything and really not using quote marks around words for any reason? That is a bit counterintuitive to me so I want to make sure I'm getting that.

austin-schaefer: Integration: Your point makes sense to me as long as we can avoid ending up with complex/passive prose to describe the connection points. But I think you're right that the integration is the nebulous thing in the middle between us and the other system.

austin-schaefer: Quote marks:

* In general, I'm not a huge fan of quote marks in docs and prefer bold or code, but that's not an ironclad rule.
* In the usage dictionary specifically, though, the other definitions pretty much exclusively use backtick code formatting rather than quote marks. The intent there was to reduce ambiguity since a snippet of sample text might itself use code.

zuluecho9: I think the practical wording is like: "we have integrations". 

The integrations are the conceptual linkage between technologies, and therefore they are also our offerings. We are offering the linkage, in whatever form. So they take a practical, more concrete form even while they are largely conceptual. 

If this seems helpful/accurate to you, I could try to make a pass adding this to usage dictionary.

austin-schaefer: Yep, that makes sense to me! Thanks.

PR #7124, opened 2022-04-15
fix(definitions): improving 'integration'

zuluecho9:
Some improvements in glossary and usage dictionary regarding defining 'integration'.


paperclypse (approved):
Feels like a definite improvement!

PR #7128, opened 2022-04-15
fix(best practices guide): removed a collapser that is no longer rele…

homelessbirds:
John Bond reached out with some requests to remove certain sections and change the terminology.


paperclypse (approved):
Looks good!

PR #7136, opened 2022-04-15
feat(guided install): Moved to on-host integrations section, added redirect

paperclypse:
People will most likely find this doc from the link in the UI. We don't need placed so prominently in the New Relic One category.


paperclypse:
@x8a Let's chat about this real quick next week! It might feel like an odd move, but I think I can provide some helpful context :)

src/nav/infrastructure.yml
```diff
@@ -537,6 +537,8 @@ pages:
         pages:
           - title: Install on-host integrations
             path: /docs/infrastructure/host-integrations/installation/install-infrastructure-host-integrations
+          - title: Guided install overview
+          path: /docs/infrastructure/host-integrations/installation/new-relic-guided-install-overview
```
paperclypse, line 541:
```suggestion
          - title: Guided install overview
             path: /docs/infrastructure/host-integrations/installation/new-relic-guided-install-overview
```

x8a (approved):
Looking great!

PR #7137, opened 2022-04-15
chore: Use current name for setting

kaylareopelle:
## Give us some context
I'm not sure when this changed, but the setting is now called `autostart.denylisted_rake_tasks`

## Relates to
https://github.com/newrelic/newrelic-ruby-agent/issues/1087


paperclypse (approved):
Looks good! Thanks for helping out @kaylareopelle!

PR #7139, opened 2022-04-16
fix: update commands to new names

keegoid-nr:
- replace legacy commands with appropriate commands for new runtime


paperclypse (approved):
I'll defer to your expertise, @keegoid-nr :) looks good!

PR #7142, opened 2022-04-18
Update for the Atlassian-Jira integration doc

FullGC:
Hi!
I removed a comment related to our Atlassian-Jira integration, as we have fixed an issue we were having and it is not relevant anymore!


paperclypse (approved):
Looks good!

PR #7148, opened 2022-04-18
fix(oma-data-gov): Added MobileRequest to NrConsumption table

JimHagan:
I found one missing event type in the NrConsumption to Event table.  This was an glaring omission.


paperclypse (approved):
Looks good!

PR #7170, opened 2022-04-19
Update Python Agent Language to remove white/black list and replace with allow/deny list

lrafeei:
This PR updates the Python Agent's docs to reflect the changes in the agent's language regarding white/black listing.  Instances of this language were replaced with allow/deny listing (which results in a change in the configuration file documentation).


rhetoric101 (approved):
@lrafeei, these are important changes. Thanks for making them!

PR #7193, opened 2022-04-20
Limit what gets indexed in the data dictionary

LizBaker:
Right now we're filtering out the data dictionary page in the search because it has a lot of text and surfaces as a top result often when it shouldn't. This PR adds some content exclusion so we can test if the results come back better

### DIFF
--- a/src/pages/attribute-dictionary.js
+++ b/src/pages/attribute-dictionary.js
@@ -237,6 +237,7 @@ const EventDefinition = memo(
           </div>
         </h2>
         <div
+          data-swiftype-index="false"
           css={css`
             margin-bottom: 1rem;
           `}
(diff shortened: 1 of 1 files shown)


clarkmcadoo:
I'm trying to put in a search to get [this page](https://build-85d9e58f-4764-4f4e-8285-f7d2c664886d.gtsb.io/attribute-dictionary) to come up but haven't been able to, should i be trying something different?

caylahamann:
how would we be able to test that it doesn't index the different query params possibilities?

LizBaker:
we'd have to merge it and have swiftype re-index it, i'm not aware of any other way. it shouldn't cause any issues to do so though because we aren't exposing this page to search results at all right now

clarkmcadoo (approved):

PR #7209, opened 2022-04-21
fix screenshot widths for synthetics best practices

keegoid-nr:
- set widths to 1200 pixels to prevent blur
- remove markup


keegoid-nr:
Also continuing to improve scripted browser example

x8a (approved):
Wow, thanks for this huge update! Much appreciated 💃🏻💃🏻💃🏻!

PR #7222, opened 2022-04-21
Data retention: mark 'synth' as uneditable, explain pricing model diffs better

umanhay:
Work done: 
@zuluecho9 added some notes clarifying that the data retention doc is for New Relic One pricing model. Added an 'uneditable' note to synthetics entry. Assorted other changes related to explaining there are two data retention docs better. 

Original PR request: Removing Synthetics from this table as it is not an editable namespace and shouldn't be included here.


zuluecho9:
@umanhay I'm checking on this. Do you happen to know the context is for the removal? (not sure if you're a New Relic employee or not, sorry)

umanhay:
> @umanhay I'm checking on this. Do you happen to know the context is for the removal or when they removed this? Update: I learned that this info is placed here because, even though it's not editable in the UI, it was thought this was most relevant place to put this info. I'll work on an edit clarifying this isn't editable.

Hi @zuluecho9 Yep I'm on the team that owns the Data Retention UI/API. It's causing confusion to have it in this list. Can I ask why Synthetics specifically needs to be placed here at all? There are many namespaces, most of which are not editable.

zuluecho9:
@umanhay can you check out this edit I made? I added 'not editable' to the things in the table that aren't editable, and prefaced it with more context. The goal of that page was to display all namespaces, whether editable or non-editable, because that makes the most sense to place it. Curious what you think of this change and if there are other major namespaces we should document that we're missing.

umanhay:
@zuluecho9 Synthetics is not editable at all by Insights or NR1 customers. Those other namespaces you updated are all editable by NR1 customers, just not Insights customers (so maybe call that out instead?). 
As far as namespaces like Synthetics, which are not editable at all, maybe I can bring this discussion to my team/PM about if and what we'd include on this page? Then we can update this page based on that feedback.

src/content/docs/data-apis/manage-data/manage-data-retention.mdx
```diff
@@ -200,7 +200,7 @@ This table shows the default retention settings for the [namespaces](/docs/gloss
       </td>
 
       <td>
-        Log
+        Log (not editable)
```
umanhay, line 203: Can we update this to 'Logging' to match the UI?

zuluecho9:
@umanhay can I ask your name? sorry I can't remember it. was going to ping you in slack about this.

zuluecho9:
@umanhay I'm going to close this one just because it seems we need to do some more work here. I wanted to make sure you knew there was an 'original pricing model data retention' doc: https://docs.newrelic.com/docs/accounts/original-accounts-billing/product-based-pricing/overview-data-retention-components/. So the fact that there are two docs is part of the confusion here. (I think we used to have a note in this doc that it was for New Relic One pricing model but I think maybe I accidentally removed that thinking it was more aligned.) 

And then we need to decide how to address the goal of: a) we want to explain what the data retention periods are for as many namespaces as we can, even if not in the UI, and b) we want to make it clear which namespaces aren't editable (whether we do that for both user models in the same doc, or branch out to each). 

Happy to be in a meeting as i'm docs liaison on this.

umanhay:
> @umanhay I'm going to close this one just because it seems we need to do some more work here. I wanted to make sure you knew there was an 'original pricing model data retention' doc: https://docs.newrelic.com/docs/accounts/original-accounts-billing/product-based-pricing/overview-data-retention-components/. So the fact that there are two docs is part of the confusion here. (I think we used to have a note in this doc that it was for New Relic One pricing model but I think maybe I accidentally removed that thinking it was more aligned.)
> 
> And then we need to decide how to address the goal of: a) we want to explain what the data retention periods are for as many namespaces as we can, even if not in the UI, and b) we want to make it clear which namespaces aren't editable (whether we do that for both user models in the same doc, or branch out to each).
> 
> Happy to be in a meeting as i'm docs liaison on this.

@zuluecho9 I've pinged for you in the help-docs channel so we can continue this conversation. For this PR, can we at least take Synthetics out of that table?

zuluecho9 (approved):

PR #7225, opened 2022-04-21
Update with note to sign up for CNCF Slack account

reese-lee:
This came about from a user pointing out [here](https://github.com/newrelic/newrelic-opentelemetry-examples/issues/111) that the existing link to our CNCF Slack channel did not work for them, and Tyler H suggested that we make a note that users may have to first sign up for CNCF Slack to access that link. This PR adds that note and the sign-up link.


barbnewrelic (commented):
I'm assuming this link does not introduce any security vulnerabilities.

barbnewrelic (approved):
The frontmatter check failed but it is not required, so I'm going to go ahead and merge this PR.

PR #7230, opened 2022-04-21
fix: add banned functions for NRQL alerting

juholgado:

## Give us some context

  * The functions added are currently not supported for NRQL alerting. Conditions registered using these functions will yield incorrect results.


barbnewrelic:
@paperclypse  when you have a chance, could you take a look at this PR? I'm not clear why we would document something we don't support. Thanks!

barbnewrelic:
@juholgado Thanks for creating this PR, and congrats on being a first time contributor to the docs site! Typically we do not publish what we don't support, we try to focus on what we do support. One of our alerts PMs mentioned that this is a very rare use case. Can you please provide a bit more background about why we might need this info in the docs? Thanks for clarifying!

juholgado:
>  Can you please provide a bit more background about why we might need this info in the docs? 

We'll soon release a change that will prevent the functions called out here from being registered in an alert. We had the thought that it might be a better experience if there was information about the banned functions as opposed to only learning about which functions are banned when an error is returned upon registration.

However, if it conflicts too much with the principles of our public docs and the described behavior(relying only on the error returned on registration to inform customers that the functions are banned)is acceptable, I can close this PR.

barbnewrelic:
@juholgado  This totally sounds reasonable. Thanks for explaining! I'll proceed with reviewing and merging your PR.

PR #7235, opened 2022-04-21
chore: Update local decorating content

kaylareopelle:
## Give us some context 

This PR makes the language in the local decorating steps more precise. 

@barbnewrelic


barbnewrelic (approved):
Thanks for making this important update!

PR #7244, opened 2022-04-22
feat: add information about infra agent ProcessSample command name

rogercoll:
Adds some context of the command name in Linux environments.


x8a (approved):
Thanks!

PR #7271, opened 2022-04-25
fix: full file path not shown

rogercoll:
The characters "<" and  ">" seems that was preventing the code section to be rendered correctly. In addition, the whole "tip" was not rendered, neither the link.

Do you think this would solve the issue?


x8a (approved):
Thanks!

PR #7275, opened 2022-04-25
Fix to allow inline image classname to be applied

caylahamann:


roadlittledawn (approved):
Good catch! Just checked out several pages with different kinds of images in the build and HTML in JSON looks good (inline images in bulleted list, block level images, inline images in collapser titles).

the only other type of image i can think of (not sure if it's an issue or not) is images inside of `TechTile` components. Example [here](https://build-7c906767-821a-436a-96ae-5d5c3802062b.gtsb.io/docs/distributed-tracing/concepts/quick-start.json). HTML looks good, but not sure how the NR1 Help XP nerdlet processes those

caylahamann:
@roadlittledawn yeah that is a good point, that seems to be an issue with the image handler since it assumes that an inline image occurs if the parent has multiple children nodes, where in tech tiles they are inline images although they don't have any siblings. I think that's something that can be fixed in the nr1 help xp, just by checking the parent class is a tech tile

caylahamann:
@roadlittledawn a quick check made by changing the styles applied in the nr1 help xp shows that we can just change the classname for a tech tile icon to be for block-image and then the styles will be fixed

PR #7279, opened 2022-04-25
Sticky not really cache fix for data dictionary

caylahamann:
this is kind of a weird one but turns out it was a redux issue not a caching issue???
In the data dictionary service, we create nodes for all the data dictionary events and attributes. When they are updated, we want to make sure the redux store reflects that change. We can ensure that by running touchNode on all data dictionary files.

this section of the gatsby docs on node creation explains it: https://www.gatsbyjs.com/docs/node-creation/#freshstale-nodes


roadlittledawn:
yeah, ditto. good find!

PR #7312, opened 2022-04-26
Change data dictionary page type

LizBaker:
The page type for the data dictionary is currently being filtered out of search results, this PR updates it to be a non filtered type


caylahamann (approved):
this all looks good to me! this shouldn't cause any problems with the `attribute=` since you added in that no index attribute right?

PR #7316, opened 2022-04-26
chore: Add log event forwarding to high security table

kaylareopelle:
## Give us some context

I noticed this table did not yet include a mention of log events. @barbnewrelic


barbnewrelic (approved):
Oh good lord, how could I have missed this? Thanks so much @kaylareopelle !

PR #7317, opened 2022-04-26
Add reference to newer OTEL Java agent that simplifies JVM metric collection

rhetoric101:
The OpenTelemetry Java agent version 1.13.0 and higher now allows users to avoid workarounds to view JVM metrics.

This PR mentions that we recommend the newer agent.


rhetoric101:
This PR is replaced by #7951 because this branch is out of date (not the content).

Issue #6054, opened 2022-02-08
Issue: Upload source maps via API


quirosfelipe:
## How can we make our docs better?

* Is something confusing? A recent bug has been fixed for V1 users publishing source maps via API.  There is a sentence in the docs that must be changed  because will no longer apply after this fix. This is the [GTSE](https://newrelic.atlassian.net/browse/APPEXP-3171) for more info.
* Is something inaccurate or missing? The line that reads 'A [user API key](https://docs.newrelic.com/docs/apis/get-started/intro-apis/types-new-relic-api-keys#user-api-key) (before November 20, 2020, the [Admin API key](https://docs.newrelic.com/docs/apis/get-started/intro-apis/types-new-relic-api-keys#admin) was required; that will still work if already in place)' , should be changed for 'A [user API key](https://docs.newrelic.com/docs/apis/get-started/intro-apis/types-new-relic-api-keys#user-api-key)(this key MUST be used regardless of the user type. The Admin API key will no longer work.'
* Were you unable to complete a task? (What task?) N/A

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/browser/new-relic-browser/browser-pro-features/upload-source-maps-api/
* **OS**: Unknown
* **Browser**: Unknown
* **Device**: Unknown

pizzaz93:
@quirosfelipe and @bradleycamacho would the new text be added to this bullet point 
<img width="431" alt="image" src="https://user-images.githubusercontent.com/10904763/165180880-94783218-21a4-476c-9920-5a45cd712846.png">
?

quirosfelipe:
@pizzaz93  that's correct!

zuluecho9:
This is fixed in https://github.com/newrelic/docs-website/pull/7332. I opted to just say the user key is required (as noting that our older keys will no longer work for various tools/integrations is something that we could theoretically add in many places in docs).

PR #7332, opened 2022-04-27
Add note regarding Admin API Key

pizzaz93:
#6054


zuluecho9:
@pizzaz93 Hi Peretz. Can I ask what led you to wanting to make that clarification? Did you use the wrong API key for this? Any more details you can give me would be much appreciated. (One reason is that we avoid trying to add call-outs like this, as there are numerous amounts of call-outs like this we could theoretically make across the entire docs site for things like this. So trying to understand the problem a bit more.)

zuluecho9 (approved):

zuluecho9:
Thanks a lot, much appreciated @pizzaz93 !

PR #7344, opened 2022-04-27
fix(synthetics): Fix setProxyAdvanced examples

bpeckNR:
The setProxyAdvanced example was incorrect and would result in an err_no_supported_proxies error. The host, port, and scheme must be defined if using a non default port (varies by scheme) or a non default scheme (http).


zuluecho9:
thank you much @bpeckNR.

zuluecho9 (approved):

Issue #5993, opened 2022-02-04
Glossary has no left nav


austin-schaefer:
The poor [glossary](https://docs.newrelic.com/docs/glossary/glossary) has no left nav:

![image](https://user-images.githubusercontent.com/55203603/152613959-61a54cb2-53f5-4b10-9a6b-676a438ea37a.png)

zuluecho9:
Going to close this one. Going to move glossary into new improve 'guides and solutions' section.

PR #7348, opened 2022-04-27
Create left nav for Glossary

pizzaz93:
#5993, these initial lines are to ensure the nav is formatted correctly and compiles.


zuluecho9:
@pizzaz93 thanks a lot for the help Peretz. Very helpful. I'm actually working on something in this area now and am going to move that doc into a different category. appreciate all the help, it is hugely appreciated.

PR #7352, opened 2022-04-27
feat(style guide): new 'guides' doc

zuluecho9:
DOC-7924. Creation of a style guide doc about creating guide-type content (including OM docs)


zuluecho9:
Will close this as I feel it's so overlapped with existing guidance and guides are so rarely created and we can just send when needed to people creating guides.

PR #7363, opened 2022-04-28
Update docs with new winston-enricher signature

garbados:
Since v3.0.0, the function signature of `@newrelic/winston-enricher` has changed. It now requires passing Winston to the formatter before the formatter can be used.

Context:
- https://github.com/newrelic/newrelic-node-log-extensions/issues/36
- https://github.com/newrelic/newrelic-node-log-extensions/pull/33


garbados:
Welp. Not sure how I did that.

paperclypse (approved):
Looks good! Thanks for helping out, @garbados!

PR #7370, opened 2022-04-28
Small update to correct case of appName example

nedl86:
NRQL is case-sensitive. Updating example query to correct case of `appName` attribute.


x8a (approved):
Thanks!

PR #7371, opened 2022-04-28
Remove image reference

LizBaker:
An image was deleted along with its import but the image was still being referenced in the file


paperclypse:
Thank you!

PR #7375, opened 2022-04-29
fix(alerts): update event flow callout

baNROne:
* The allowable gap between data points, when using event flow, has been updated to 65 minutes.
* This change brings this page in line with the same information found [here](https://github.com/newrelic/docs-website/blob/develop/src/content/docs/alerts-applied-intelligence/new-relic-alerts/advanced-alerts/understand-technical-concepts/streaming-alerts-key-terms-concepts.mdx#L218).


x8a (approved):
Thanks!

PR #7400, opened 2022-05-02
chore(components): adding a comment

x8a:
Testing whether this ticket gets moved to the eng GH board.


x8a:
The workflow is working, so I'll close the PR.

PR #7452, opened 2022-05-05
Only add translation PRs to localization board

roadlittledawn:
## Description

Change label this workflow looks for to `translations` so only PRs back from translation vendors are added to the [board](https://github.com/newrelic/docs-website/projects/4) so we can better keep track of them.

### DIFF
--- a/.github/workflows/pr-triage.yml
+++ b/.github/workflows/pr-triage.yml
@@ -16,13 +16,13 @@ jobs:
       PROJECT_ID: 11873277
       PROJECT_NAME: Docs PRs and Issues
     steps:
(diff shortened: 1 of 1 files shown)


caylahamann (approved):
Is the only thing that changed the label??? 
it appears that way!
LGTM

caylahamann:
the build failed but this makes 0 changes to the build, so I feel like you could either a. just ignore it or b. run it again with a clear cache

roadlittledawn:
i did run wtih a clean cache and same result. just merged it because it doesn't have any bearing on gatsby stuffs so should be fine

PR #7462, opened 2022-05-05
chore: specify the maximum value for max_samples_stored

kaylareopelle:
## Give us some context

An upper limit was discovered for `application_logging.forwarding.max_samples_stored` This limit is now included in the description.


barbnewrelic (approved):
Thanks for updating this info!

PR #7470, opened 2022-05-06
feat(style guide): Removed semicolon from import statement 

paperclypse:
because we don't need it.


barbnewrelic (approved):
Good catch! I don't think any of us were using that trailing ;

PR #7496, opened 2022-05-09
chore(logs in context) - update Java config snippets

mnevilleoneill:
Updated Java config snippets to be more complete/transparent.


bradleycamacho (approved):
Thanks for the edits @mnevilleoneill !

PR #7504, opened 2022-05-09
[DO NOT MERGE] Try to repro stale cache in GC

roadlittledawn:
Trying to repro stale cache in gatsby cloud by moving a file and adding a redirect for it

this page moved from here

/docs/synthetics/new-relic-synthetics/pages/synthetics-results-access-individual-monitor-runs

to here
/docs/accounts/accounts/account-maintenance/synthetics-results-access-individual-monitor-runs

i left old page url in left nav.

in theory, any page within `/docs/synthetics` should have broken JS


roadlittledawn:
looks like redirect works and no broken pages

PR #7525, opened 2022-05-10
fix: remove dana from codeowners

clarkmcadoo:


rhetoric101 (approved):
@clarkmcadoo, this looks good to me. I don't know if you also need approval from an engineer, but I've added and removed names in this file.

roadlittledawn (approved):

PR #7536, opened 2022-05-11
what's new: quickstarts for k8s and prometheus

josemore:
## Give us some context

* Adding a "What's new" doc showcasing 6 new quickstarts published in Instant Observability for Kubernetes and Prometheus customers.


paperclypse (approved):
Looks good to me.

PR #7577, opened 2022-05-12
docs(Alerts): Add 3 new term threshold operators for static NRQL conditions

founddrama:
## Give us some context

- We (Alerts!) just released some additional threshold operators for static NRQL conditions
- This adds those operators to the exiting documentation
- Bringing it more/less in line with what's in the Terraform provider docs: https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/nrql_alert_condition


paperclypse (approved):
Works for me. Thanks!

PR #7591, opened 2022-05-13
docs(node.js) Add instrumenting already loaded module

keithmorris:
Provide additional information for instrumenting supported Node.js modules that have been already loaded using `newrelic.instrumentLoadeModule` for cases where a user may have to delay instrumenting. An example of this might be asynchronously loading and decrypting api keys from a remote location.


bradleycamacho (approved):
Thanks for the edit!

bradleycamacho (approved):

PR #7620, opened 2022-05-16
chore(heroku_log_forwarding) - Add deletion info

mnevilleoneill:
Added section covering how to delete Heroku syslog log drain token mappings via UI and API. Made a few minor capitalization changes throughout.


paperclypse (approved):
I made some edits so the build check would pass and to bring a couple things more in line with our style guide. Thanks!

PR #7637, opened 2022-05-17
chore:Updated bottom of the funnel analysis to refer the reader to th…

khickey-newrelic:
…e quickstart for dthe dashboard instead of to github

This is just a one-line change.  Now that a quickstart is available, readers should go there to get their dashboard. 

I ran up gatsby and checked the link work prior to commit.


x8a (approved):
Thanks!

PR #7694, opened 2022-05-19
Updated requirements to match docs on Puppet Forge

nedl86:
See https://forge.puppet.com/modules/newrelic/newrelic_infra


x8a (approved):
Thanks!

PR #7732, opened 2022-05-23
Increase font from 16 to 18px

LizBaker:
also added the antialiasing style zack suggested which stops the font from looking bold at all weights 

Screen Shot 2022-05-23 at 2 12 20 PM
Screen Shot 2022-05-23 at 2 12 29 PM

### DIFF
--- a/src/layouts/MainLayout.js
+++ b/src/layouts/MainLayout.js
@@ -60,6 +60,8 @@ const MainLayout = ({ data = {}, children, pageContext }) => {
       <Layout
         css={css`
           margin-top: 1rem;
+          -webkit-font-smoothing: antialiased;
+          font-size: 18px;
         `}
       >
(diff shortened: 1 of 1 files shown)


src/layouts/MainLayout.js
```diff
@@ -60,6 +60,8 @@ const MainLayout = ({ data = {}, children, pageContext }) => {
       <Layout
         css={css`
           margin-top: 1rem;
+          -webkit-font-smoothing: antialiased;
+          font-size: 18px;
```
caylahamann, line 64: this is so minor
```suggestion
          font-size: 1.125rem
```

caylahamann (dismissed):
approving but got like the teeniest suggestion

caylahamann (dismissed):

caylahamann (approved):

PR #7753, opened 2022-05-25
feat: add infra-agent installation instructions for RHEL 9

rogercoll:
Do not merge: RHEL 9 building files ready and tests performed on the infra-agent, but artifacts have not been released yet.

PR state will be changed once artifacts are available.


x8a (approved):
Thanks!

PR #7755, opened 2022-05-25
"What's New" FS Recap Blog

aalfano7:
This post should be the featured post in What's New until the end of day 6/30/2022.

This post should be published as soon as possible - time of day is less important. 

This does not need review by a PMM.


barbnewrelic:
@aalfano7  Your build failed. I'm going to make a minor copyedit and see if my commit pushes your build through.
Also, I am not sure we can keep a What's New post at top of list for an extended period of time, because the order is automatically generated. I'll have to check with the engineers about that.

aalfano7:
Thank you @barbnewrelic! We will need to use the "isFeatured" flag for this - if that helps. This was previously used for "FutureStack is back in-person at the Cosmopolitan of Las Vegas" up until 5/17.

barbnewrelic:
@aalfano7  Fixed the date field and now it builds just fine. I'll merge this PR today and create a followup PR to remove the flag end of June.

PR #7766, opened 2022-05-25
Removing exclusions for Korean site.

jmiraNR:
<!-- For Japanese readers: 

## Give us some context

*starting to translate the whole docs site in KR.


barbnewrelic (approved):
OK AFAIK THX!

PR #7768, opened 2022-05-25
docs: Update Solution in Surround Logs feature

Jasonoliva:
docs: Update Solution in Surround Logs feature

Step 4 solution in https://docs.newrelic.com/docs/logs/troubleshooting/find-issues-cause-or-impact-surrounding-logs/ is no longer accurate.

I worked with NR support (https://support.newrelic.com/s/case/5008W00000oaArHQAU/trouble-with-the-surrounding-logs-feature) to find a clearer way to explain this feature.

Also added a Step 5 so that it is clear to users what the expected behavior will be after clicking "Show surrounding logs"


Jasonoliva:
Closing to commit this on a different account https://github.com/newrelic/docs-website/pull/7769

PR #7771, opened 2022-05-25
Replace incident intelligence image

rhetoric101:
Based on SME feedback, we need to replace one image and remove another.


paperclypse (dismissed):
I think this looks ok. It's tough for me to know if those service names are generic or specific to New Relic...

paperclypse (approved):
Looks good!

PR #7740, opened 2022-05-24
removed `--devel`  as it is not needed anymore

greenszpila:
Please follow [conventional commit standards](https://www.conventionalcommits.org/en/v1.0.0/)
in your commit messages and pull request title.

## Give us some context

* What problems does this PR solve?
* Add any context that will help us review your changes such as testing notes,
  links to related docs, screenshots, etc.
* If your issue relates to an existing GitHub issue, please link to it.

PR #7780, opened 2022-05-26
Remove callout for --devel flag

rhetoric101:
This is a follow-up to PR #7740. We need to also remove the `--devel` callout.


barbnewrelic (approved):
thanks for keeping this doc up to date!

PR #7782, opened 2022-05-26
Fix(APM): Broken link

paperclypse:
The path was missing an initital `/`. This came in as a hero request.


barbnewrelic (approved):
a tiny fix makes a big difference. thanks!

PR #7795, opened 2022-05-26
docs: expermiment with razor/aspx lang identifiers

brnhensley:
I'm not sure this will work, these are not used often and my code editor doesn't recognize them. I also made some minor edits to page like capitalization and backticks.

Please follow [conventional commit standards](https://www.conventionalcommits.org/en/v1.0.0/)
in your commit messages and pull request title.

## Give us some context

* What problems does this PR solve?
* Add any context that will help us review your changes such as testing notes,
  links to related docs, screenshots, etc.
* If your issue relates to an existing GitHub issue, please link to it.

PR #7797, opened 2022-05-26
feat(browser): Add razor language syntax

paperclypse:
A quick followup to #7795 @brnhensley


paperclypse:
Although this one says it needs a Docs Engineering review, it doesn't really.

brnhensley:
Thanks for adding that, looks nice!
image

PR #7803, opened 2022-05-27
Remove hostmetrics to avoid doubledipping metrics

lazyplatypus:
## Give us some context

Right now `hostmetrics` being present in the collector will lead to harvesting metrics already collected by the infra agent, so I propose we remove it from the docs.


rhetoric101 (commented):
@lazyplatypus, I'll defer to you about this change. I'll go ahead and approve it.

PR #7810, opened 2022-05-27
feat(OHI): reworked docs

x8a:


barbnewrelic (approved):
Everything looks good to me! And thanks for changing the order of config after install! :-)

PR #7835, opened 2022-05-31
chore(k8s): Remove self-referential collapser links

austin-schaefer:
Remove strange self-referential links


barbnewrelic (approved):
Thanks for removing that unnecessary code!

PR #7838, opened 2022-06-01
Update manage-pixie-memory.mdx

maxlemieux:
Added missing `-` on `--set`

Command will not work without two dashes.


barbnewrelic (approved):
Thanks for fixing this!

PR #7853, opened 2022-06-01
fix: query typos

keegoid-nr:
- add charts
- update queries


bradleycamacho (approved):

bradleycamacho:
Looks good to me, thanks for the edit @keegoid-nr !

PR #7890, opened 2022-06-03
fix(retention): clarifying DP retention

zuluecho9:
Work done: 

Adding Data Plus column so data retention is more comparable. 

Finessing description of how Data Plus adjusts retention.


barbnewrelic (approved):
This looks fine!

PR #7891, opened 2022-06-03
fix(AI): Removed doc about unused API

paperclypse:
The AI team says that very few customers (they've been notified) use this API and we should delete the doc. I've added a redirect for the active incident event API doc.


barbnewrelic (approved):
thanks for cleaning out old docs!

PR #7903, opened 2022-06-06
OMA / SLM Update

a-james-faria:
Update to OMA SLM document; however I do not know why src/content/docs/accounts/original-accounts-billing/original-users-roles/users-roles-original-user-model.mdx is in my commit. Please help me resolve that file. I did not modify it. Thanks!


a-james-faria:
I didn't make changes to either of these. src/content/docs/new-relic-solutions/observability-maturity/uptime-performance-reliability/optimize-slm-guide.mdx
src/content/docs/service-level-management/create-slm.mdx

PR #7915, opened 2022-06-07
Infra-agent default config dir for macOS

josemore:
## Give us some context

* The default config dir and config file for macOS was missing in the main infra-agent config document.


barbnewrelic (approved):
thanks for adding MacOS info to this doc!

PR #7918, opened 2022-06-07
chore: Add translation nav script

LizBaker:
This is a script for generating a list of all nav titles to be sent for translation

### DIFF
--- a/scripts/getNavForTranslation.js
+++ b/scripts/getNavForTranslation.js
@@ -58,7 +58,7 @@ const getFlatNav = () => {
 
 const filterUniqueTitles = () => {
   const uniqueTitles = [];
-  const flatNav = getFlatNav(locale);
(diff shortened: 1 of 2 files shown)


scripts/getNavForTranslation.js
```diff
@@ -0,0 +1,86 @@
...
+    ];
+    if (file.pages) {
+      return readSubNavs(file.pages, appendedNav);
+    } else {
+      return appendedNav;
+    }
+  });
+};
+
+const getFlatNav = () => {
```
clarkmcadoo, line 46: Should this take a param if we're passing it in on line 61?

LizBaker: whoops nope, that was from before i made locale a global var

clarkmcadoo (approved):

PR #7945, opened 2022-06-08
fix: include cgroupsns field in privileges so examples work properly on cgroup v2

sigilioso:
Includes a required field in docker examples so nri-docker integration works when running inside a container and docker is on cgroup v2. This is required because, when running in cgroup v2, [default cgroup namespace mode is private](https://docs.docker.com/config/containers/runmetrics/#running-docker-on-cgroup-v2) and the integration cannot gather other containers metrics.

[nri-docker#131](https://github.com/newrelic/nri-docker/issues/131) was opened because those fields are missing.


barbnewrelic (approved):
Thanks for adding this to the docs!

PR #7951, opened 2022-06-08
Add recommended OTel Java agent version

rhetoric101:
This is a replacement for the older PR (#7317)  on the same topic since the branch became out of date (not the content).

The OpenTelemetry Java agent version 1.13.0 and higher now allows users to avoid workarounds to view JVM metrics.

This PR states that we recommend the newer agent.

This PR also includes a small wording change for the metrics-spans toggle memory.


barbnewrelic (approved):
Looks good to me!

PR #7961, opened 2022-06-08
fix(synthetics): making limits more clear

zuluecho9:
Liaison work, updating pricing and synthetics docs to make the synthetic check limits more clear. 

Also adding in data plus price point.


barbnewrelic (dismissed):
This all looks fine to me. I did notice the frontmatter for the new doc has a line for redirects (currently n/a) but the preview works fine so I guess it's OK to keep it. Merge away whenever you want!

zuluecho9:
@barbnewrelic yeah the `redirects` field works fine even if empty. thank you! (just waiting on SME review now)

PR #7980, opened 2022-06-09
Python Release Notes Corrections

TimPansino:

## Give us some context

* Slight corrections to release notes. Blacklist settings have never appeared in the agent, avoiding confusion by removing the mention of them.


zuluecho9:
@TimPansino do we technically call it 'allow list' now? or is it still 'whitelist' in the codebase?

TimPansino:
It's called `allowlist` everywhere in the codebase. There's a deprecated setting mapping for the old name `whitelist` but that's the only mention of it.

zuluecho9:
ah so am i understanding that when this says 'removes whitelist settings' is that referring to removing the 'white list' language?

PR #7991, opened 2022-06-09
Remove translated section-break

LizBaker:
we accidentally translated the "title" of section-break so it fell through our check and no `<br/>`s were added to the nav in korean

Screen Shot 2022-06-09 at 3 29 24 PM


zuluecho9:
@LizBaker do you need any sort of review on this? or fine without?

zuluecho9 (approved):

PR #7995, opened 2022-06-10
feat(aws-health): new fields and remove EOL

josemore:
## Give us some context

* We added the following new attributes to the AwsHealthNotification event: startTimestamp, endTimestamp and lastUpdateTimestamp.
* We are also taking the opportunity to remove the inventory section that has been EOL (we'll do the same for all integrations  in the next few weeks).


x8a (approved):
Thanks!

PR #7996, opened 2022-06-10
Infra-agent: update supported platforms

josemore:
## Give us some context

* Infra-agent is now compatible with RHEL 9, Ubuntu 22.04 and Amazon Linux 2022 (Beta).


x8a (approved):
Thanks!

PR #8012, opened 2022-06-13
Streamline and clarify obfuscation documentation

rhetoric101:
This is a revision that consolidates some sections to reduce duplication, and it clarifies UI and NerdGraph examples.

Here are some items I changed that should be checked in the review:

* Reorganized the introduction and the first section, swapping, deleting, and adding various sentences.
* Chopped up the section "Example: Basic process" and moved that content into other sections.
* Renamed "Checklist: Steps to obfuscate logs" as "Steps to obfuscate logs" and created child steps for the entire process, using the previous material.
* Review new "Step 1: Study the patterns of your logs."
* Review new "Step 2: Work with obfuscation expressions." Note that I inserted the full JSON in "Create an obfuscation expression" and in "Update an obfuscation expression."
* Look closely at "Obfuscation expression examples." I inserted examples for both the UI and NerdGraph. Please confirm that the escaping is correct in each., especially for "born date," "birth date," and "email address."
* Review new section "Step 3: Work with obfuscation rules" and confirm that this bullet is correct: "Define how you will capture sensitive data using NRQL" since I don't see any other references to NRQL. 
* The collapser "Reusing expressions in other rules is now at the bottom of the rules section (brought down from the old example section.
* CPU limits is now at the bottom. Is this appropriate here?


barbnewrelic (approved):
@rhetoric101  Thanks for chatting with me!

PR #8019, opened 2022-06-13
chore(redirects): Update jira shortlink

austin-schaefer:
Updates docs.newrelic.com/jira jira template shortlink for new jira instance


akristen (approved):
Thanks for the link fix! :)

PR #8030, opened 2022-06-14
fix: Update introduction-log-api.mdx

emartin-nr:
This fixes the example showing how message is treated when it is JSON. The example seems to indicate that the JSON attributes are extracted from the message (which is correct) and the message attribute is retained as a JSON string (which is not correct). So I just removed the JSON string from the "This will be treated as" part of the example. Thanks!


rhetoric101 (approved):
@emartin-nr, thanks for making this update!

PR #8031, opened 2022-06-14
statsd: added arm64 support

josemore:
## Give us some context

* StatsD integration is available on arm64 archictures.


rhetoric101 (approved):
@josemore thanks for making this update!

PR #8055, opened 2022-06-15
Use `minionId` changes to count restarts

keegoid-nr:
- change from `minionHostname` to `minionId`, which should work better for Kubernetes containers where the container name doesn't change.


bradleycamacho (approved):
Thanks!

PR #8061, opened 2022-06-15
Add requestcountpertarget to aws docs

### LINKED ISSUE #8033: ALB metric descriptions are truncated and misleading
[ALB Docs page](https://docs.newrelic.com/docs/infrastructure/amazon-integrations/aws-integrations-list/aws-albnlb-monitoring-integration/)

There is `requestCount` for both ALB loadbalancer itself and for the TargetGroup.  The NR docs leave out important details causing customer confusion.

Per the AWS docs:

**Loadbalancer** AWS metric name `RequestCount`

> The number of requests processed over IPv4 and IPv6. This metric is only incremented for requests where the load balancer node was able to choose a target. Requests rejected before a target is chosen (for example, HTTP 460, HTTP 400, some kinds of HTTP 503 and 500) are not reflected in this metric.
> 
> Statistics: The most useful statistic is Sum.

**TargetGroup** AWS metric name `RequestCountPerTarget`

> The average number of requests received by each target in a target group.
> 
> Statistics: The only valid statistic is Sum. This represents the average not the sum.

bradleycamacho:
Closes #8033

### DIFF
--- a/src/content/docs/infrastructure/amazon-integrations/aws-integrations-list/aws-albnlb-monitoring-integration.mdx
+++ b/src/content/docs/infrastructure/amazon-integrations/aws-integrations-list/aws-albnlb-monitoring-integration.mdx
@@ -115,7 +115,9 @@ New Relic Infrastructure Amazon integrations collect the following Amazon ALB da
       </td>
 
       <td>
-        The number of requests received by the load balancer. This includes requests over IPv4 and IPv6.
+    The count of requests processed over IPv4 and IPv6. Requests only increment this metric when the load balancer node was able to select a target. If a request is rejected before a target is choses, it is not reflected.
+    
+    Stats: Sum provides the most value.
       </td>
     </tr>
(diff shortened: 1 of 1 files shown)


GaPeppy:
I wasn't thinking of changing the schema field names (those are set by product engineering).  In Cloud Integrations, these metric values are simply copied from AWS.  But the current description in the NR docs is weak/vague.  So I was suggesting basically copying the metric description from AWS to enhance the detail provided by NR docs.

bradleycamacho:
Hey @GaPeppy, thanks for clarifying! I just had a conversation with our director and we're unsure if we can just copy over the direct definition from the AWS docs. I have written stronger (but still not the greatest) definitions in my latest commit instead.

I will be chatting with Legal soon to figure out what our stance on these definitions is. If they give us the go ahead to continue copying descriptions over, I'll come back and update these. Thanks!

PR #8069, opened 2022-06-16
AWS metric streams: notes about filters and useful queries

josemore:
## Give us some context

* Made cost considerations more prominent and added links to ways to filter data as needed. 
* Included query examples to break down the ingest data and number of metrics based on namespaces.


barbnewrelic:
The license check stopped for no apparent reason. I'm closing and reopening this PR to see if that helps.

barbnewrelic (approved):
Thanks for adding this helpful new info!

PR #8070, opened 2022-06-16
fix(mysql): slaveRunning has been reintroduced

paologallinaharbur:
Fix https://github.com/newrelic/nri-mysql/issues/100

Context:
We should remove from the callout in our mysql docs the deprecation of cluster.slaveRunning since [now we are calculating the metric ]( https://github.com/newrelic/nri-mysql/pull/89) in a different way.


barbnewrelic (approved):
Thanks for making this update!

PR #8081, opened 2022-06-16
.NET Agent - NuGet agent install instructions - add callout regarding logs folder permissions

nr-ahemsath:
A recent .NET agent support ticket involved a situation where the customer did not have any agent log files to help troubleshoot, and the cause ended up being that the agent's `logs` folder, as created when the agent is installed with the NuGet package, is created with default permissions, which means that only the user who built/published the app can write to the logs folder.  If their app is executed by a different username (a common pattern), the agent will not be able to write logs.  This PR adds a callout to the NuGet package install instructions warning of this issue.

~Initially setting PR status to "Draft" to get feedback from .NET agent team members before it's ready for docs team review~.


jaffinito (commented):
Looks good to me!

PR #8083, opened 2022-06-16
Additional reqs for CLM

planteater:
1. Added some additional requirements for code-level metrics.
2. Removed an extraneous section of the pull requests article that I completely overlooked in my previous update.


barbnewrelic (approved):
Thanks for updating these docs!

PR #8115, opened 2022-06-20
Update and sort list of languages

brnhensley:
The prism languages did not match the list of languages listed in the docs here: https://docs.newrelic.com/docs/style-guide/formatting/code-examples/#add-syntax-highlighting

I added the missing languages and alphabetized the list so that it could be searched easier in the future

I also updated the doc detailing the languages. Both locations contained languages not reflected in the other, they are synced together now.

## Give us some context


x8a:
Hey! After talking to the devs we've realised that these are all implicitly available so we don’t need to explicitly add them to the config file. See [prism's GH repo](https://github.com/FormidableLabs/prism-react-renderer/blob/master/src/vendor/prism/includeLangs.js).

When a code block contains a `<var>` tag, or other tags, the language highlighting gets blocked. So what you did in other PRs, which was to remove those tags, should fix the issue.

Let us know if you find a code block where the code highlighting is not working and it should.

Thanks a lot and I hope to see more PRs coming from you! 🙌🏻

brnhensley:
@x8a I did this because `go` doesn't work. My only clue why was that it wasn't in the prism config. See this doc: https://docs.newrelic.com/docs/apm/agents/go-agent/instrumentation/instrument-go-transactions/

PR #8123, opened 2022-06-21
Update record-handled-exception.mdx

ahuller:
Removed return section and changed headers to reflect that the function is actually void.

## Give us some context

The iOS agent sdk recordHandledException function is void in the code, so the docs should reflect that.


rhetoric101 (approved):
@ahuller thanks for clarifying this function!

PR #8124, opened 2022-06-21
Update recorderror-ios-sdk-api.mdx

ahuller:
function is void, so return values section should be empty.

## Give us some context

This function is void in the agent code and should be reflected as such.


rhetoric101 (approved):
@ahuller, thanks for removing this section about return values.

PR #8149, opened 2022-06-22
fix(user type): add logs detail

zuluecho9:
Add more detail for logs capabilities in user type table


barbnewrelic (changes requested):
s/be "Log management" not "Logs" but otherwise looks good to me!

zuluecho9:
thank you @barbnewrelic !

PR #8154, opened 2022-06-22
chore(Welcome to NR): Remove over-detailed content

austin-schaefer:
Remove some text that adds way too much detail and link fatigue


paperclypse (approved):
Looks good!

PR #8163, opened 2022-06-23
fix(aws): use dataPointCound for raw metrics

josemore:
## Give us some context

*Using dataPointCount() instead of count(*) when counting metric updates to avoid misleading results when querying over a long period of time.


x8a (approved):
👏🏻👏🏻👏🏻

PR #8175, opened 2022-06-23
Add support for translating whats new posts

LizBaker:
This adds full support for translating MD and specifically content in the whats-new directory

Outstanding work for later:
* We'll need to add any what's new post titles we want to be translated in the nav to the nav json in each language
* We'll need to batch translate all of 2022
* We need additional logic to show translated titles in the timeline

Example with disclaimer:

Screen Shot 2022-07-18 at 3 29 00 PM

Example whats new post in each locale:

Screen Shot 2022-07-15 at 11 08 27 AM

Screen Shot 2022-07-15 at 11 08 36 AM

Screen Shot 2022-07-15 at 11 08 19 AM


.github/workflows/add-slugs-to-translate-queue.yml
```diff
@@ -3,11 +3,12 @@ name: Add Slugs to Translation Queue
 on:
   pull_request:
     branches:
-      - main
+      - liz/translate-md
     types:
       - closed
     paths:
       - '**.mdx'
+      - '**.md'
```
LizBaker, line 11: specify path for whats new

roadlittledawn: still need to change to whats new path?

plugins/gatsby-source-nav/gatsby-node.js
```diff
@@ -55,7 +55,7 @@ exports.createResolvers = ({ createResolvers, createNodeId }) => {
             case slug === '/':
               return createRootNav(utils);
 
-            case slug.startsWith('/whats-new'):
+            case slug.includes('/whats-new'):
```
roadlittledawn, line 58: should we be extra cautious and check for `/whats-new/`?

roadlittledawn (commented):
Looking good!

The only other wrinkle I can think of is there are `iframe` tags used in whats new posts for videos. I have a feeling those aren't handled by default so we may have to add a handler in [handlers.js|https://github.com/newrelic/docs-website/blob/develop/scripts/actions/utils/handlers.js]

PR #8180, opened 2022-06-23
New capabilities doc

zuluecho9:
New capabilities doc, and related edits. Goal of this was to create a more detailed section on capabilities (currently we had no details on specific capabilities) and this would allow us to point to this doc from other docs when we need to explain various feature-related capabilities, instead of having to (as we do now) explain that in assorted, scattered places in other docs.


src/content/docs/accounts/accounts-billing/new-relic-one-user-management/user-capabilities.mdx
```diff
@@ -0,0 +1,275 @@
...
+tags:
+  - Accounts
+  - New Relic One user management
+metaDescription: An explanation of New Relic user capabilities and what they govern. 
+redirects:
+---
+
+import newRelicUserCapabilitiesUi from 'images/user-management-capabilities.png'
+
+A role, whether one of our [pre-built roles](/docs/accounts/accounts-billing/new-relic-one-user-management/user-management-concepts#roles) or a custom role, is a set of capabilities. You can think of a "capability" as a granular thing a New Relic user can do. Examples of capabilities are: the capability to view APM app settings, or modify alert conditions, or manage data retention settings.
```
akristen, line 12: Is there a limit on how many capabilities one role has? Example: An org has 2 teams, Team A and B. Team A has one role with X capabilities, but team B has a separate role with Y capabilities?

Could be helpful to give an example situation in 1-2 sentences that makes concrete what this means in practice.

src/content/docs/accounts/accounts-billing/new-relic-one-user-management/user-capabilities.mdx
```diff
@@ -0,0 +1,275 @@
...
+* Our newer user model: [learn about roles](/docs/accounts/accounts-billing/new-relic-one-user-management/user-management-concepts). 
+* Our original user model: [learn about roles](/docs/accounts/original-accounts-billing/original-users-roles/users-roles-original-user-model). 
+
+## Some important tips for using capabilities [#important-points]
+
+Some important points for understanding and using capabilities: 
+
+* **Capabilities can change.** The capabilities we expose in the UI are subject to change and reorganization. The capabilities in this doc were last updated June, 2022.     
+* **Some capabilities have overlapping functionality.** That is why selecting some capability checkboxes will automatically check or uncheck other boxes.
+* **Not all capabilities are exposed.** Some roles have hidden capabilities that are not visible in the UI and not available for addition to a custom role. For example, the **Organization manager** role has capabilities related to managing and configuring organization-level settings, but those are not available for addition to a custom role.
```
akristen, line 34: Is there a list of roles that are not "exposed" in addition to organization manager? Is it in the UI that you can reference? (If you don't think it's helpful or you think it might be redundant, disregard.)

src/content/docs/accounts/accounts-billing/new-relic-one-user-management/user-capabilities.mdx
```diff
@@ -0,0 +1,275 @@
...
+* **Some capabilities have overlapping functionality.** That is why selecting some capability checkboxes will automatically check or uncheck other boxes.
+* **Not all capabilities are exposed.** Some roles have hidden capabilities that are not visible in the UI and not available for addition to a custom role. For example, the **Organization manager** role has capabilities related to managing and configuring organization-level settings, but those are not available for addition to a custom role. 
+* **Capabilities don't affect querying.** Most capabilities apply to New Relic UI experiences and not to querying data. For example, if your capabilities restrict you from accessing the APM UI, you can still query APM-related data using the query builder if you have access to that account. If you require more firm data boundaries for some projects or users, one way to do this is by [adding accounts](/docs/accounts/accounts-billing/account-structure/add-accounts).
+
+## Capability definitions [#capability-definitions]
+
+Here's a composite screenshot of the capabilities we expose in the capabilities UI for the purpose of adding to custom roles.
```
akristen, line 39: I like the screenshot, but this chunk of text could probably be more interactive and answer more questions. Maybe instead of "Here's a composite screenshot" something like:

"This screenshot shows all of the possible capabilities that are viewable by all users. These capabilities can be applied to custom roles in any combination your org needs."

This could also be a good spot to explain which capabilities are **not** exposed here.

akristen: Emphasis here on calling out "all possible capabilities" and "all users." Just so it's more obvious for the audience who and what this pertains to, even if a little repetitive.

src/content/docs/accounts/accounts-billing/new-relic-one-user-management/user-capabilities.mdx
```diff
@@ -0,0 +1,275 @@
...
+
+Here's a composite screenshot of the capabilities we expose in the capabilities UI for the purpose of adding to custom roles. 
+
+<img
+  title="New Relic user capabilities screenshot"
+  alt="New Relic user capabilities UI screenshot"
+  src={newRelicUserCapabilitiesUi}
+/>
+
+To learn more about these capabilities, select a category below.
```
akristen, line 47: Just to clarify: these collapsers are exhaustive for each feature/product? i.e. you're listing all the relevant capabilities as they pertain to all of NR's features? If yes, I'd err on the side of being more obvious than not here, so there's no doubt that all the relevant info is accounted for/represented. :)

akristen (dismissed):
Good job on the doc! Everything makes sense-- I went ahead and added notes about clarifying some vague language, but the doc is good to go otherwise. :)

src/content/docs/accounts/accounts-billing/new-relic-one-user-management/user-capabilities.mdx
zuluecho9: there's a relatively small set of capabilities we expose in the capabilities UI. it's far from all the stuff you can do in NR. it's just what they decide to give customers control over for the purposes of creating new custom roles. so it's far from all, in other words. for example, they recently added a capability for controlling who can view/manage a user key, so that hadn't been available for a long time and might have gone for a while without being available. If it's not exposed as a capability, it just means it's probably a fundamental capability tied to all users, or perhaps just to full platform users. It's complicated, basically. but I might explain these things a little better, thanks.

PR #8223, opened 2022-06-28
Ruby CodeStream integration: YAML syntax fix

fallwith:
for enabling CodeStream in the Ruby agent, a colon should be used to
separate the 'code_level_metrics.enabled' configuration key from the
'true' or 'false' configuration value.


rhetoric101 (approved):
@fallwith thanks for catching this error!

PR #8234, opened 2022-06-28
fix: add tip that capture() regex must match entire input

mlobiondo-newrelic:

## Give us some context

Customers (internal and external) are often confused that capture() doesn't work as expected, because their regular expressions are designed to only match part of the attribute value and they must match the whole thing.


rhetoric101 (approved):
@mlobiondo-newrelic thanks for adding this clarification!

PR #8235, opened 2022-06-28
Java agent compatibility page updates

kford-newrelic:

## Give us some context

* This PR fixes some small issues with version ranges previously mentioned, to make the documentation more accurate. Also added references to some instrumentation that was recently added but not previously updated.


rhetoric101 (approved):
@kford-newrelic, thanks for making these updates!

PR #8252, opened 2022-06-29
Clarify Wistia and YouTube options with Video components

rhetoric101:
I realized that I hadn't updated the instructions for embedding videos as I thought I had. So, here's an improvement:

* Added instructions for upload videos to Wistia and getting the ID.
* Divided the YouTube instructions into iframe and Video sections.
* Added some basic instructions about using the new `vertical` property for images.

### DIFF
--- a/src/content/docs/style-guide/images/embed-videos.mdx
+++ b/src/content/docs/style-guide/images/embed-videos.mdx
@@ -74,7 +74,7 @@ Here's how it is displayed:
 
 ### Use our Video component [#video-comp]
 
-You can use the `Video` component in the same way you use it for videos hosted on Wistia. Copy the video ID from YouTube and insert it into the `Video` component. Note that you can also add properties to control the video (see [Sizing videos](#sizing)).
+You can use the `Video` component in the same way you use it for videos hosted on [Wistia](#wistia-options). Copy the video ID from YouTube and insert it into the `Video` component. Note that you can also add properties to control the video (see [Sizing videos](#sizing)).
 
 Here's an example without any additional properties:
 ```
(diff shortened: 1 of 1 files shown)


src/content/docs/style-guide/images/embed-videos.mdx
```diff
@@ -13,37 +13,122 @@ redirects:
...
+
+### Use our Video component [#video-comp]
+
+You can use the `Video` component in the same way you use it for videos hosted on [Wistia](#wistia-options). Copy the video ID from YouTube and insert it into the `Video` component. Note that you can also add properties to control the video (see [Sizing videos](#sizing)).
+
+Here's an example without any additional properties:
+```
+<Video
+  type="youtube"
+  id="<var>04JP0ky_hjI/var>"
```
caylahamann, line 83:
```suggestion
  id="<var>04JP0ky_hjI</var>"
```

src/content/docs/style-guide/images/embed-videos.mdx
```diff
@@ -13,37 +13,122 @@ redirects:
...
+  id="04JP0ky_hjI"
+/>
+
+## Sizing videos [#sizing]
+
+When you use the `<Video>` component you can add properties to control how it is displayed.
+
+### Width [#width]
+
+We recommend that you only set width and not height. Here's an example:
```
caylahamann, line 99: so it is not possible to set a `height` on the video, so it is only possible to set a `width` so even if users tried to set a height it would do nothing

rhetoric101: @caylahamann thanks! I'll clarify this in the doc!

PR #8253, opened 2022-06-29
Update theme and add sign up config

caylahamann:

## Give us some context

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -161,6 +161,15 @@ module.exports = {
             }),
           },
         },
+        signup: {
+          environment: process.env.ENVIRONMENT || 'staging',
+          signupUrl:
+            process.env.SIGNUP_URL ||
+            'https://signup-receiver.staging-service.newrelic.com',
+          reCaptchaToken:
+            process.env.RECAPTCHA_TOKEN ||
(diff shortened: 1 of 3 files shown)


gatsby-config.js
```diff
@@ -161,6 +161,15 @@ module.exports = {
             }),
           },
         },
+        signup: {
+          environment: process.env.ENVIRONMENT || 'staging',
```
roadlittledawn, line 165: we need to add this env var to gatsby cloud. should it be `production`?

roadlittledawn: looks like it should be `production`. but please double check me

caylahamann: yes it should be production, but don't we already have this set in the site??

caylahamann: i just added another variable

roadlittledawn: ideally we'd operate from one env var to indicate waht env we're in.

not 100% what would happen if window._nr_signup.environment is set to `development` though (which is the only other value we use for `GATSBY_NEWRELIC_ENV`.

in any case, we can proceed and dig into that further as a potential refactor if needed

roadlittledawn (approved):

PR #4900, opened 2021-11-15
Added '-r newrelic' startup script to install docs

tytyhibye:
Issues around module load order are a high ticket driver and adding this require flag to their scripts can safeguard against this common issue caused by Babel. I threw it in a callout but you can use whatever layout you deem best.

Please follow [conventional commit standards](https://www.conventionalcommits.org/en/v1.0.0/)
in your commit messages and pull request title.

## Give us some context

* What problems does this PR solve?
* Add any context that will help us review your changes such as testing notes,
  links to related docs, screenshots, etc.
* If your issue relates to an existing GitHub issue, please link to it.

PR #8262, opened 2022-06-30
Update default Node.js agent install require instructions 

### LINKED ISSUE #3094: Issue: Install the Node.js agent
### Tell us what you need

Currently, step 5 of the installation says:

`Add require('newrelic'); as the first line of your app's main module.`

This does work, but many Node.js users compile/transpile their JavaScript and can't control the final require order or have confusion on where to place this require statement. In these cases, the work around is to load the `newrelic` module in the start script with the [Node command line option -r](https://nodejs.org/api/cli.html#cli_r_require_module). For example, the start up script would look like `node -r newrelic ./dist/server.js`. Then requiring or importing the 'newrelic' module in the application code is only necessary when using the API. 

While both methods work, the `-r` installation method is more reliable. Last month, over 30% of tickets had an instrumentation issue that was resolved by switching from using `require('newrelic');` to load the agent to `-r newrelic` in the start script.

This is a documented work around in the agent Github README:

* https://github.com/newrelic/node-newrelic#:~:text=if%20you're%20compiling%20your%20javascript%20and%20can't%20control%20the%20final%20require%20order%2C%20the%20node.js%20agent%20will%20work%20with%20node's%20-r%2F--require%20flag.

This is also the method guided install uses so it is well documented/tested. 

We discussed this in GTS/Engineering pairing, initially the thought was to include this as an alternate install method and use the same verbiage as the agent README that states what situations you should use `-r newrelic` and that every other situation you should use `require('newrelic');`. But since this is such a common issue and it is a well tested and documented 'workaround', the consensus was that we should offer the `-r newrelic` method as the primary install method with `Add require('newrelic'); as the first line of your app's main module.` as an alternative if customers are not able to use `-r newrelic` in their start script. 

Potential verbiage:

Add `-r newrelic` to your start script to load the agent with the [Node command line option -r](https://nodejs.org/api/cli.html#cli_r_require_module). For example, the start up script would look like `node -r newrelic your-program.js`.

Since this is a change to crucial Node.js agent docs, approval from Node.js PM may be needed. Node engineering team and support team are familiar with this request so reach out to #gts-nodejs or #node-agent with any questions! Thank you Docs Team!

* https://docs.newrelic.com/docs/agents/nodejs-agent/installation-configuration/install-nodejs-agent/
* **OS**: Mac OS 10.15.7
* **Browser**: Chrome 91.0.4472.114
* **Device**: Unknown

pizzaz93:
Fixes #3094

### DIFF
--- a/src/content/docs/apm/agents/nodejs-agent/installation-configuration/install-nodejs-agent.mdx
+++ b/src/content/docs/apm/agents/nodejs-agent/installation-configuration/install-nodejs-agent.mdx
@@ -62,15 +62,16 @@ To install the Node.js agent:
 
    * Customize the `license_key` setting with [your license key](/docs/apis/intro-apis/new-relic-api-keys/#ingest-license-key).
    * Customize the [`app_name`](/docs/agents/nodejs-agent/installation-configuration/nodejs-agent-configuration#app_name) setting with one or more [meaningful app names](/docs/apm/new-relic-apm/installation-and-configuration/naming-your-application).
-6. Add `require('newrelic');` as the first line of your app's main module.
-
-<Callout variant="important">
-  If you are using [Babel](https://babeljs.io/docs/en/index.html) or a similar transpiler you can safeguard against any issues related to module load order by utilizing the [Node command line option '-r'](https://nodejs.org/api/cli.html#-r---require-module) to preload the 'newrelic' module at application startup. For example, if your application's entry point is `./dist/server.js` then you would use the require flag like so:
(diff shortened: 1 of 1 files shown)


x8a:
Hey, @pizzaz93! Thanks for your PR!

I did a bit of digging since the document already had the information, and I found the [PR that fixed the issue](https://github.com/newrelic/docs-website/pull/4900) back in last November.

I'll close this PR and the original issue, since it's fixed in the public docs already.

Thanks a lot for your collaboration!

PR #8272, opened 2022-06-30
Fix deserialization formatting bug

roadlittledawn:
## Description

Without a line between the paragraph and the table, the serialized HTML nests the `table` inside a `p` like so:

```html
<p><strong>Arista eAPI</strong>
<table data-type="component">
...
</table>
</p>
```

Adding the line closes the `p` tag like so which puts `table` on its own when deserializing it from serialized html:

```html
<p><strong>Arista eAPI</strong></p>
<table data-type="component">
...
</table>
```

converts to

```
  <Collapser
    id="api-polling-config"
    title="オプションのAPIポーリング構成"
  >
    **Arista eAPI**

    <table>
      <thead>
```

### DIFF
--- a/src/content/docs/network-performance-monitoring/advanced/advanced-config.mdx
+++ b/src/content/docs/network-performance-monitoring/advanced/advanced-config.mdx
@@ -1109,6 +1109,7 @@ global:
     title="Optional API Polling configurations"
   >
     **Arista eAPI**
+    
     <table>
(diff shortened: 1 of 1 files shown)


caylahamann (dismissed):

roadlittledawn:
too many concurrent builds. triggered build via gatsby cloud UI

PR #8280, opened 2022-06-30
chore(java): Remove outdated update info

austin-schaefer:
Removes a whole slew of info about upgrading from old versions of the agents. All of these agent versions are at least 3.5+ years old, some are even 5+ years old.


paperclypse (approved):
Works for me! :D

PR #8281, opened 2022-06-30
improve readability of performance analysis charts

keegoid-nr:
- Change to billboard style, which is easier to read for input into the Google Sheet.
- Set a 2 day period on performance data to get more accurate results since some monitors may run infrequently.


x8a (approved):
Thanks!

PR #8295, opened 2022-07-01
Update aws s3 metric filter instructions 

### LINKED ISSUE #4465: Issue: Amazon S3 monitoring integration

* Is something confusing? Yes
* Is something inaccurate or missing? Yes

Under [Amazon S3 data for request metrics](https://docs.newrelic.com/docs/integrations/amazon-integrations/aws-integrations-list/aws-s3-monitoring-integration/#provider-S3BucketRequests) we have the following:

`AllRequests` | The total number of HTTP requests made to an Amazon S3 bucket, regardless of type. If you’re using a metrics configuration with a filter, then this metric only counts HTTP requests from objects which meet the filter's requirements.
-- | --

This perhaps needs a rewrite to clarify that at least one bucket metric filter is actually required for publishing metrics to Cloudwatch, which we are then able to poll. In the AWS S3 console, when configuring bucket request metrics, the following message will show up if no metric filters have been created:

>You don't have any filters
To view request metrics, create a filter.

So the verbiage in the docs about "If you're using a metrics configuration with a filter" is off, as it's required there isn't really an "if" about it unless talking about "if it has been configured properly".

 Also, there is the possibility that a customer creates multiple bucket metric filters for bucket requests. If they have, then they will need to have one of the filters set to the scope of `This filter applies to all objects in the bucket`, and named `EntireBucket`. Otherwise, no bucket request metrics are gathered.
 
 [AWS Docs that covers creating a s3 bucket metrics config for all objects](https://docs.aws.amazon.com/AmazonS3/latest/userguide/configure-request-metrics-bucket.html)
 
 Note that maybe we should reference step 7, when creating the bucket metrics filter, that it needs to be named `EntireBucket`. It is also important to underline that this should be the ONLY metrics filter for all objects in the bucket as any additional ones may cause the metrics to not report properly (based on my testing in my lab).

Please see internal GTSE-11937 for more info, or reach out to me in Slack if further clarification is needed.

Possibly helpful for illustrating the config in s3:
s3BucketMetricFilterEntireBucket

* https://docs.newrelic.com/docs/integrations/amazon-integrations/aws-integrations-list/aws-s3-monitoring-integration/
* **OS**: Mac OS 10.15.7
* **Browser**: Chrome 93.0.4577.82
* **Device**: Unknown

pizzaz93:
Fixes #4465

### DIFF
--- a/src/content/docs/infrastructure/amazon-integrations/aws-integrations-list/aws-s3-monitoring-integration.mdx
+++ b/src/content/docs/infrastructure/amazon-integrations/aws-integrations-list/aws-s3-monitoring-integration.mdx
@@ -121,7 +121,7 @@ To collect these metrics, you must enable request metrics for your S3 bucket:
       </td>
 
       <td>
-        The total number of HTTP requests made to an Amazon S3 bucket, regardless of type. When using a metrics configuration with a filter, then this metric only counts HTTP requests from objects which meet the filter's requirements. If creating multiple bucket metric filters then a filter needs to be created and scoped to all the objects in the bucket as detailed in these [AWS instructions](https://docs.aws.amazon.com/AmazonS3/latest/userguide/configure-request-metrics-bucket.html). Note that **only** that filter should be scoped to all objects in the bucket. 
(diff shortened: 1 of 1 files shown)


zuluecho9:
thank you @pizzaz93. you are the best, or at least one of the best.

zuluecho9 (approved):

pizzaz93:
> thank you @pizzaz93. you are the best, or at least one of the best.

Thanks 🙏 :)

PR #8299, opened 2022-07-01
Embeddable pages 

caylahamann:
- [x] File follow up to implement embed pages on all MDX file pages on docs / remove JSON files plugin `gatsby-plugin-generate-doc-json 

- adds `embedDarkMode` query param set to `enabled` or `disabled` in order to render embed in dark or light mode


gatsby-node.js
```diff
@@ -219,6 +219,50 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
...
   });
 
+  const createEmbed = (node) => {
+    const {
+      frontmatter: { redirects },
+      fields: { fileRelativePath, slug },
+    } = node;
+
+    if (
+      fileRelativePath.includes('src/content/docs/release-notes') ||
```
roadlittledawn, line 229: interesting. release note content does get surfaced in `nr1-help-xp`. buuuut let's be sure to exclude them from results via the docs library service before we implement the iframe in nr1-help-xp

gatsby-node.js
```diff
@@ -219,6 +219,50 @@ exports.createPages = async ({ actions, graphql, reporter }) => {
...
+            '/'
+          )}`
+        );
+        createRedirect({
+          fromPath: path.join(fromPath, 'embed', '/'),
+          toPath: pagePath,
+          isPermanent: true,
+          redirectInBrowser: true,
+        });
+      });
```
roadlittledawn, line 262: do we need to create redirects for embed pages also? 

if when DLS gets a result and checks for redirect on the standard page first, THEN gets the new URL and appends `/embed/` onto the end, i don't think we need redirects right?

caylahamann: if we see that the only use for the `/embed/` is to be used for the DLS and no other potential use cases in the future, then we can remove this

gatsby-node.js
caylahamann: well if they get surfaced then I can remove this

gatsby-node.js
roadlittledawn: for now, yes, we only need to support embed pages via DLS. and we can document that. best spot i think is likely our site architecture wiki page

clarkmcadoo (approved):

PR #8333, opened 2022-07-06
Update advice on cadence alerts for clock skew

danielgblanco:

## Give us some context

Cadence should still be the recommended aggregation method when we consider clock skewed events that the New Relic user has no control of, as a single event with a skewed timestamp can trigger aggregations on windows that do not have a representative dataset available. This happens with `PageAction` but in reality could happen with any event instrumented by untrusted devices.


danielgblanco:
Thanks for the quick review. That's it signed :)

PR #8348, opened 2022-07-06
Minor eol sweep

austin-schaefer:
Removes some very old EoL-related copy from APM agent docs


barbnewrelic (dismissed):
Just one suggestion, then feel free to merge whenever you want. No second review required. Thanks for taking care of this!

akristen (approved):
Good catch... 2015 was certainly a different time!

PR #8365, opened 2022-07-07
Add links to sample instrumentation in setup

rhetoric101:
Since our setup guide targets users who want to instrument their own services right away, I thought we should offer an alternative track for people who want to instrument a sample service before working on their own.


barbnewrelic (approved):
Really helpful tip!

PR #8373, opened 2022-07-07
Fix typo in memory_limiter configuration stanza

rkbodenner:
The correct config stanza name appears in the sample code block elsewhere in this doc.


akristen (approved):
Good catch! Thanks a bunch~

PR #8374, opened 2022-07-07
Shorten local dev time

LizBaker:
This adds a new env var, `BUILD_I18N`, which determines whether or not to source/create all the i18n pages. The default is _not_ to build them for the local dev server. The var has been added to the develop and main sites on gatsby cloud as true. This cuts down local server spin up time from 11 min to 6 min (on my machine). 

I also added logic and an env var called `BUILD_FOLDERS`. You can use this to specify which folders under the docs directory you would like to build when running locally. It can be any of those top level folders (docs/*) separated by commas. 
ex: `BUILD_FOLDERS=accounts,codestream yarn start` would build only the content inside the accounts and codestream folders. This results in the local server starting in as little as 2 minutes 🎉 

I also decided to ignore the whats-new folder since I figured if someone is specifying the folder they're working on they probably don't care about what's new-- but we can change that if that doesn't sound good


gatsby-config.js
```diff
@@ -207,13 +223,21 @@ module.exports = {
       options: {
         name: 'translated-content',
         path: `${__dirname}/src/i18n/content`,
+        ignore:
+          process.env.BUILD_I18N === 'false'
```
clarkmcadoo, line 227: nit-pick, but the not-logic always gets me, would the suggestion work?

```suggestion
          process.env.BUILD_I18N ? [] : [`${__dirname}/src/i18n/content/*`]
```

LizBaker: i think env vars are always strings unfortunately so we would have to change it to a boolean first to do this 🤷‍♀️

clarkmcadoo: omg yeah my b

clarkmcadoo (approved):

LizBaker:
build weirdness, going to try close/open

PR #8377, opened 2022-07-08
Improve SyntheticRequest page

keegoid-nr:
- add charts and organize page
- set default `accountId` to 0.
- replace `location` with `locationLabel` for readability


akristen (approved):
Great changes @keegoid-nr. Thanks for contributing!

PR #8380, opened 2022-07-08
Updated Impact of Inspected Count rate limit

ganeshmuralidhar:
We gathered information around the impact of Inspected Count hitting the rate limit within an account and how will different components of NewRelic be impacted from a customer experience point of view. Here are the details:
https://newrelic.slack.com/archives/C01508Q2879/p1656433937163829
https://newrelic.slack.com/archives/C01508Q2879/p1657187781242109

## Give us some context

* What problems does this PR solve? - Better clarity in documentation

### DIFF
--- a/src/content/docs/query-your-data/nrql-new-relic-query-language/get-started/rate-limits-nrql-queries.mdx
+++ b/src/content/docs/query-your-data/nrql-new-relic-query-language/get-started/rate-limits-nrql-queries.mdx
@@ -107,7 +107,12 @@ For our [original data option](/docs/accounts/accounts-billing/new-relic-one-pri
   </tbody>
 </table>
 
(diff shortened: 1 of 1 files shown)


x8a (approved):
Thanks!

zuluecho9:
@ganeshmuralidhar Wanted to ask a few more questions on this: 

What was the specific problem that led to this change? Could you give me a little more info on the problem that you were trying to solve with this? 

i noticed you only added this to the 'original data option'. I wanted to make sure: are you sure that this doesn't apply to the new Data Plus option? (this one? https://docs.newrelic.com/docs/query-your-data/nrql-new-relic-query-language/get-started/rate-limits-nrql-queries/#data-plus) 

Your addition sounds somewhat like you are trying to define exactly what the query limits apply to, which is something we try to do here: https://docs.newrelic.com/docs/query-your-data/nrql-new-relic-query-language/get-started/rate-limits-nrql-queries/#query-count-limits. So I'm curious if you see your addition overlapping the definition there. If it is, maybe we can just better define that and point to it. 

This sounds like mainly an alerting consideration: do you think this might be better suited to be an addition in the alerting area? (or perhaps in addition to?) If so, do you know of a place in alerting section where something like this might live? Maybe in here? https://docs.newrelic.com/docs/alerts-applied-intelligence/new-relic-alerts/learn-alerts/rules-limits-alerts/

PR #8411, opened 2022-07-12
Update to Korean human translation workflow.

jmiraNR:
Added Korean tag for human translation.


barbnewrelic (approved):
Our Korean readers thank you for these updates!

PR #8428, opened 2022-07-12
Update local setup doc

LizBaker:
This is a documentation update for building a smaller part of the docs site locally


barbnewrelic (approved):
Thanks so much for adding that helpful info about making local builds faster!
I made a few other cosmetic edits to existing text because I can't help myself. :-)

PR #8451, opened 2022-07-13
Remove un-used timingPrecision config attribute from .NET docs

JcolemanNR:
We received a customer question about the the `timingPrecision` configuration attribute and realized it doesn't actually do anything after reviewing code. To minimize customer confusion, we should remove this from our published docs.


rhetoric101 (approved):
@JcolemanNR thanks for removing this collapser!

PR #8460, opened 2022-07-13
fix(telemetry sdks): remove .NET telemetry sdk from list

jackdelamotte:
Per [this issue](https://github.com/newrelic/newrelic-dotnet-agent/issues/1047), the .NET Telemetry SDK has been deprecated and should no longer be used. This change removes it from the list of available libraries on [this](https://docs.newrelic.com/docs/data-apis/ingest-apis/telemetry-sdks-report-custom-telemetry-data/) docs page.


rhetoric101 (approved):
@jackdelamotte thanks for removing the .NET row of this table!

PR #8474, opened 2022-07-14
fix(Browser summary): Rewrite to reflect current UI

barbnewrelic:
For NR-18640


rhetoric101 (dismissed):
@barbnewrelic I think this is a big improvement! Thanks for tackling this page! I just inserted some minor suggestions.

PR #8484, opened 2022-07-15
Changes to .NET agent documentation for major version 10.0.0 release

nr-ahemsath:
This PR has changes to a bunch of the .NET agent documentation to support our upcoming major version release (10.0.0), mostly caused by the fact that the name (and install path) of our Linux agent installation packages is changing from `newrelic-netcore20-agent` to `newrelic-dotnet-agent`.

We have release agent version 10.0.0 as of 7/19/2022 so this should be merged as soon as practicable.


rhetoric101 (approved):
@nr-ahemsath thanks for making all of these updates!

PR #8505, opened 2022-07-18
Deleting anomaly detection

zuluecho9:
Deleted anomaly detection page, and added redirect to external-redirects file to point at new, upcoming doc on beta site. 

Also deleted references to this feature in a couple other docs.


homelessbirds (approved):
I agree that some of these changes are awkward but I think they're the best we can do for now. We're adding a new product so we're going to be adjusting these again and I think they're good enough for now

PR #8524, opened 2022-07-19
feat(style guide): Clarifying the difference between public I/O and add data in the UI (NR-23312)

paperclypse:
A style guide update that attempts to clarify (for our writers) the difference between the public I/O page and the Add data page in the UI. 

I can add more info here about why I made this change, for example.

Reviewer: This is a minor change. What do you think?

I want to know if I've succeeded in clarifying this difference.

### DIFF
--- a/src/content/docs/style-guide/writing-strategies/install-docs-guidance.mdx
+++ b/src/content/docs/style-guide/writing-strategies/install-docs-guidance.mdx
@@ -1,5 +1,5 @@
 ---
-title: How to document install methods
+title: How to document our install methods
 topics:
   - New Relic only
   - Basic style guide
@@ -74,7 +74,7 @@ CodeStream has its own installation method.
 
 [Guided install](/docs/new-relic-one/install-configure/new-relic-guided-install-overview/) provides a single command you can run on your system to install New Relic and start monitoring a variety of things.
 
-At present, guided install installs the infrastructure agent, configures a certain set of integrations, and enables sending logs. It also detects integrations on your system and prompts you to choose whether to monitor those integrations or not.
+At present, guided install installs the infrastructure agent and enables sending logs. It also detects integrations on a system and prompts you to choose whether to monitor those integrations or not. If guided install detects an application of a supported language, it can install the APM agent for that language.
 
 For the docs, we should send people to guided install whenever it's appropriate to do so. That is, when guided install is likely to lead to a successful outcome.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/style-guide/writing-strategies/install-docs-guidance.mdx
```diff
@@ -74,7 +74,7 @@ CodeStream has its own installation method.
 
 [Guided install](/docs/new-relic-one/install-configure/new-relic-guided-install-overview/) provides a single command you can run on your system to install New Relic and start monitoring a variety of things.
 
-At present, guided install installs the infrastructure agent, configures a certain set of integrations, and enables sending logs. It also detects integrations on your system and prompts you to choose whether to monitor those integrations or not.
+At present, guided install installs the infrastructure agent and enables sending logs. It also detects integrations on a system and prompts you to choose whether to monitor those integrations or not. If guided install detects an application of a supported language, it can install the APM agent for that language.
```
akristen, line 77: Might consider rewording to avoid 'install installs'. This isn't a great example but something small like:

At present, when someone opts for the guided install, they get the infrastructure agent with logs yadda yadda

src/content/docs/style-guide/writing-strategies/install-docs-guidance.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: How to document install methods
+title: How to document our install methods
```
akristen, line 2: Based on this title, I thought the doc would include something more action-oriented for tech writers rather than an information overview for install methods. 

I'm wondering if we can add another ticket (or continue to work on this doc) that clarifies some ambiguous action items here, e.g.:

- "For the docs, we should send people to guided install whenever it's appropriate to do so. That is, when guided install is likely to lead to a successful outcome."

How can a writer who has never documented install methods make this call? Maybe we can add something about always reaching out to engineers to clarify why the guided install is structured the way it is. Something else that COULD be helpful (not sure if we want it documented) is maybe some context on guided install being a work in progress (WIP) so writers know why exactly we should cross reference with SMEs for this especially.

- "They also provide the most straightforward installation path for standard use cases.

This method is less effective for customers who have more complicated or obscure things they're trying to monitor or non-standard systems."

Same comment as above. Might be worth getting to the bottom of the WHO is guided install best for -- I've heard mixed stories between different engineering teams + our tech writer team. Also, "straightforward install path" almost contradicts the "most successful" qualifier in the section above. If manual install is the most straight forward, then why bother with guided install (from the POV of a writer checking out this page). Could be worth teasing this apart, or if we don't have those answers right now, reiterating who our best resources are to get the information we need.

- "That being said, the current state of the manual installation docs is inconsistent and, in some cases, like the APM agent docs, bewilderingly complicated." 

Part of me wants to push against this but I'm not sure why/how LOL. Maybe we can come up with a more diplomatic way to say these are works in progress, or we're hoping to phase them out, or they're mostly there if the other two fail. ("Method of last resort" kind of drives at this, but could be helpful to say it more outright. AFAIK, the docs install is good if you're having troubleshooting errors and you want more hands-on to the individual files to figure out what's failing the install? Idk)

akristen: Didn't realize how long this comment is. Just some food for thought!

akristen (dismissed):
I'm approving the changes because your actual edits are good updates! Re: the longer comment, add sentences as you see fit but wanted to give you my general thoughts while you're working on this doc. :)

src/content/docs/style-guide/writing-strategies/install-docs-guidance.mdx
paperclypse: This is a great comment, @akristen ! A lot of good food for thought here. I have a ticket to try to uncover some of the answers to these questions. I'll drop a link to your comment in that ticket.

Unfortunately, I don't know some of the answers to your great questions!

paperclypse: I think you're right thought about being more diplomatic. :D I'll rewrite that sentence.

Issue #8542, opened 2022-07-21
Issue: Install .NET agent for Linux - Package Name Change


Nateb1121:
## How can we make our docs better?

* Is something inaccurate or missing?
As of July 19th the .NET agent's package name was changed which invalidates this documentation. 

For more information on the name change please see https://docs.newrelic.com/docs/apm/agents/net-agent/getting-started/9x-to-10x-agent-migration-guide/#upgrading-from-newrelic-netcore20-agent-to-newrelic-dotnet-agent 

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/apm/agents/net-agent/installation/install-net-agent-linux/
* **OS**: Unknown
* **Browser**: Unknown
* **Device**: Unknown

rhetoric101:
This is resolved by #8545

nr-ahemsath:
@Nateb1121 We have a docs change PR in progress that will address this: https://github.com/newrelic/docs-website/pull/8484

PR #8545, opened 2022-07-21
Update .NET Agent Package Name

Nateb1121:
* Updates package name references with the new package names for the v10.x agent via a `sed s/newrelic-netcore20-agent/newrelic-dotnet-agent/` 
* Adds a notation about the agent name change for customers who might not have noticed the package change

Closes https://github.com/newrelic/docs-website/issues/8542 but will need validation from the .NET agent support team before this is published for customers. I've only validated the tarball install.


nr-ahemsath:
@Nateb1121 We have an internal docs change PR in progress which will address this issue: https://github.com/newrelic/docs-website/pull/8484

I'm going to go ahead and close this PR.

PR #8547, opened 2022-07-21
Node.js Agent v8.16.0 Release Notes

jordigh:
Node.js Agent v8.16.0 Release Notes


rhetoric101 (approved):
@jordigh these release notes look good. Thanks for adding these!

PR #8548, opened 2022-07-21
chore: updates Node agent log forwarding configuration to be enabled by default

michaelgoin:
Updating the Node.js agent configuration docs to reflect that log forwarding is now enabled by default.


bizob2828 (approved):

rhetoric101 (approved):
@michaelgoin thanks for updating this configuration default.

PR #8553, opened 2022-07-21
New .NET agent release notes versioning

nr-ahemsath:
This PR makes two changes, both around the .NET agent's adoption of Semantic Versioning (SemVer), which requires that we use only three numbers (major version, minor version, and patch version - e.g. 10.0.0) instead of the four numbers we're currently using:

1. Change our description of the latest release from `10.0.0.0` to `10.0.0`
2. Rename the .mdx file to use a filename pattern that makes it easier to read what agent version these release notes are for (I cribbed this from the Node agent's release notes)

My only concern about this is that these changes don't break the way the release notes appear or are ordered when somebody browses to https://docs.newrelic.com/docs/release-notes/agent-release-notes/net-release-notes


rhetoric101 (approved):
@nr-ahemsath I added the redirect. This works for me locally!

PR #8570, opened 2022-07-22
fix(Go): update configuration instructions

nr-swilloughby:
## Fixed
 * Agent configuration instructions showed examples with hard-coded license key embedded in the project source code. This is not recommended for security and privacy reasons. This was updated to explain the risk and to show how to get that value from the runtime environment instead.
 * Added additional instructions to show how one may configure a wide variety of agent parameters from the environment.


zuluecho9:
thanks a lot @nr-swilloughby !

zuluecho9 (approved):

PR #8587, opened 2022-07-25
Add react logo

LizBaker:
Can be used in techtiles as 'logo-react'

I also deleted some image imports in the apm intro since they weren't being used


clarkmcadoo (approved):

roadlittledawn:
nice find on logos not being used in that MDX file 🧹

PR #8634, opened 2022-07-27
fix(node agent): document the new grpc.record_errors configuration option

jordigh:
This was introduced [in version 8.17.0 of the Node.js agent](https://github.com/newrelic/node-newrelic/pull/1291/files).


barbnewrelic (approved):
Thanks for adding this new info!

PR #8655, opened 2022-07-28
Update manage-monitor-runtimes.mdx

ryanv94:
0.6.x is not the latest.

## Give us some context

* Clarification of the manage runtime docs - the latest runtime is not 0.6.x (chrome 72). The latest is Chrome 100.


akristen (approved):
Good catch about our runtimes!

PR #8658, opened 2022-07-28
fix(SAP): add limits

zuluecho9:
We had a request to add limits to SAP doc. The limits are based on pricing edition.


barbnewrelic (commented):
this is fine. thanks for adding this info!

PR #8672, opened 2022-07-28
Chore: Update network port behaviors to match changes in guided install

Mesverrum:
## Give us some context

Product team is updating the guided install to make it so that some steps we previously called out as optional are now the default setup.


akristen (approved):
Thanks for getting this doc up to date!

PR #8690, opened 2022-07-29
Fix - Added quotations to password value

bertomoore:
Strings with special characters need to be surrounded by quotes in YAML:
https://yaml.org/spec/1.2.2/#53-indicator-characters

Improvement to documentation

Many passwords use special characters, and not including the quotations in the documentation may cause confusion


rhetoric101 (approved):
@bertomoore thanks for adding these quotation marks!

PR #8710, opened 2022-08-01
Add AppInfoConfigOption

LizBaker:
Closes 37033

### DIFF
--- a/src/components/AppInfoConfigOption.js
+++ b/src/components/AppInfoConfigOption.js
@@ -1,0 +1,60 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { SelectInLine } from '@newrelic/gatsby-theme-newrelic';
+import MDXContainer from './MDXContainer';
+
+const AppInfoConfig = ({
+  optionType,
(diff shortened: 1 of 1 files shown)


src/components/AppInfoConfigOption.js
```diff
@@ -0,0 +1,60 @@
...
+        }
+      >
+        {select.options.map((option) => (
+          <option key={option.label} value={option.value}>
+            {option.displayName}
+          </option>
+        ))}
+      </SelectInLine>
+
+      {mdx && <MDXContainer body={mdx} />}
```
caylahamann, line 38: would we be able to destructure the `body` property off of mdx, because likely we will pass in the entire object, so it's really `mdx.body` that would get passed down to the `MDXContainer`

caylahamann (commented):
just one suggestion!

src/components/AppInfoConfigOption.js
LizBaker: yep!

caylahamann (approved):
Thank you!

PR #8737, opened 2022-08-02
Clarify in style guide about deleting/renaming image files

zuluecho9:
Wanted to clarify in style guide about how deleting/renaming images works


src/content/docs/style-guide/images/screenshots-images.mdx
```diff
@@ -270,3 +271,11 @@ When using annotations:
     />
   </Collapser>
 </CollapserGroup>
+
+## Delete or rename an image file [#delete-rename-image]
+
+Because our translated i18n directory mdx files reference the same image files that our English-language mdx files do, when image files are deleted or renamed, this can cause a conflict error. Here are some tips for deleting or renaming an image file:
```
akristen, line 277:
```suggestion
Our translated i18n directory of mdx files and our English-language mdx files reference the same image folder. When you delete or rename an image file in the English docs, a conflict error is created in the translated files. Here are some tips for deleting or renaming an image file: 
```

akristen: idk if I worded this correctly re: error created in the translation files. maybe "error is created between the English and translated files"? Not sure if that's important but if it is then maybe teasing that verbiage out

PR #8807, opened 2022-08-05
Add link to logstash plugin in github

danybmx:
This PR just adds a link to the logstash plugin in the github repository to help the users to know which plugin are they going to use.


barbnewrelic (approved):
Thanks for adding that link. Very helpful!

PR #8818, opened 2022-08-06
fix(docs): update iOS requirements

nidelson:
Reading the New Relic React Native Agent installation documentation, I have verified that iOS 10 SDK or above is required.

But on the iOS Agent compatibility and requirements page it is still like the iOS 9 SDK being the bare minimum.

As one depends on the other, it became clear that iOS 9 is the bare minimum.

> iOS 10 -> iOS Agent -> React Native Agent

PS: Happy to be contributing :)


akristen (approved):
Thanks for updating these changes!

Also, when you update our English docs, you **won't** need to do the same changes in the translated files. You can read more about our translated docs here, if you're curious (it's pretty cool!): https://docs.newrelic.com/docs/style-guide/writing-docs/docs-translation/

nidelson:
> Thanks for updating these changes!
> 
> Also, when you update our English docs, you **won't** need to do the same changes in the translated files. You can read more about our translated docs here, if you're curious (it's pretty cool!): [docs.newrelic.com/docs/style-guide/writing-docs/docs-translation](https://docs.newrelic.com/docs/style-guide/writing-docs/docs-translation/)

Very coll!
I hadn't thought about the automated biweekly process for synchronizing translations. Top!

PR #8821, opened 2022-08-09
Update some PHP docs

brnhensley:
Cleaned up code that was running together on single lines, made a new codeblock out of text and added lang identifiers. This should all just improve readability.


akristen (approved):
Looks good!

PR #8852, opened 2022-08-15
Fix i18n workflow bug when marking a translation as completed in translation database

roadlittledawn:
# Summary
When downloading and deserialization was successful, we update the translation database to mark a file as COMPLETED before we then clean up and delete all completed files and jobs from database.

in rare circumstances the same file URI could be IN_PROGRESS for the same locale at the same time in two different projects. if so, we only want to delete the record for the project at hand (human or machine). otherwise both records are deleted and one of the workflows will start to fail when it can't find the relevant record for a given file for a given locale.

# Related info
See jira for full details https://issues.newrelic.com/browse/NR-39576


LizBaker (approved):
glorious 🎉

Issue #8806, opened 2022-08-05
Issue: NRQL syntax, clauses, and functions


tanaka-takayoshi:
## How can we make our docs better?

Add boolean type condition syntax.

If the type of attribute is boolean, `=` operator doesn't work.

not working example:

```
SELECT  * FROM JavaScriptError WHERE firstErrorInSession = true
```

The comparison operation must be `is`.

working example:

```
SELECT  * FROM JavaScriptError WHERE firstErrorInSession is true
```

It seems `is [not] true` and `is [not] false` are acceptable.

https://docs.newrelic.com/docs/query-your-data/nrql-new-relic-query-language/get-started/nrql-syntax-clauses-functions#sel-where

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/query-your-data/nrql-new-relic-query-language/get-started/nrql-syntax-clauses-functions
* **OS**: Mac OS 10.15.7
* **Browser**: Chrome 103.0.0.0
* **Device**: Unknown

zuluecho9:
Thank you @tanaka-takayoshi. I'll change this and look for other similar errors.

zuluecho9:
@tanaka-takayoshi if you wanted to take a look at this fix and see if it makes sense: https://github.com/newrelic/docs-website/pull/8856

PR #8856, opened 2022-08-16
fix(nrql): adding booleans

zuluecho9:
Reported in https://github.com/newrelic/docs-website/issues/8806. Added details for booleans to NRQL operators section.


bradleycamacho (approved):
Nice simple change, go for it!

PR #8862, opened 2022-08-16
Fix link to Metric Streams NerdGraph mutation

maxlemieux:
The link to the NerdGraph information for linking AWS accounts to New Relic for Metric Streams was pointing to the section for API Polling integrations. This change fixes it to point to the Metric Streams version of the mutation.


rhetoric101 (approved):
@maxlemieux thanks for correcting this link!

PR #8863, opened 2022-08-16
Fix: Update OTLP table to reflect Infinite Tracing + HTTP support

jgals:
This PR updates the integrations table in our OpenTelemetry docs to reflect the fact that we now support Infinite Tracing + HTTP/1.1 (as of June 2022).


jgals:
👋 FYI @MattWhelan, is there any other info we should add here aside from updating this table?

rhetoric101 (approved):
@jgals thanks for making this update!

PR #8865, opened 2022-08-16
Fix(Go Agent): corrected names of functions, added multi-string usage for IgnoredPrefix

nr-swilloughby:
* Existing version of the documentation misspelled `IgnoredPrefix` as `IgnorePrefix`. This corrects that typo.
* The upcoming release adds new functionality by allowing multiple string arguments to the `WithIgnoredPrefix` and `ConfigCodeLevelMetricsIgnoredPrefix` functions instead of just a single one as before. This is now documented.
* The `Config` structure's `IgnoredPrefix` member is now deprecated in favor of the new `IgnoredPrefixes` (the former being a single `string` value, the latter being `[]string` to allow multiple prefixes to be specified.


nr-swilloughby:
These changes will apply to the Go Agent as of version 3.18.1, which we hope to release on 8/18.

homelessbirds (approved):
looks good, thank you!

PR #8868, opened 2022-08-17
Fix(Go Agent): Added notes about compatibility with legacy code

nr-swilloughby:
Release 3.18.0 introduced API changes in a backwards-compatible manner to prevent this from being a breaking change for users of the Go Agent. While this is the case for what we believe to be typical usage of the agent, there is a possible edge case where a change to customer code is necessary if they wrote their own custom code around the agent.

This PR adds a notice to the release notes for 3.18.0 to point this out to users so they are fully informed of the issue.


homelessbirds (approved):
Looks good! Thanks

PR #8874, opened 2022-08-17
feat(custom headers): Added custom header option

bpeckNR:
Added custom request headers to the advanced options for ping and simple browser monitors. This feature has been available in synthetics for a few months but was not added to the documentation when released.


homelessbirds (approved):
Looks good!

PR #8886, opened 2022-08-17
Clarify supported logging framework versions for in agent log forwarding

JcolemanNR:
This clarifies the requirements to use in agent automatic log forwarding


homelessbirds (approved):
Looks good!

PR #8888, opened 2022-08-17
fix(node agent): add compressed_content_encoding configuration setting

jordigh:
This has been added in the following PR:

https://github.com/newrelic/node-newrelic/pull/1326


homelessbirds (approved):
Looks good!

PR #8909, opened 2022-08-19
Fix mysql TLS setting

carlossscastro:
This PR is to fix a small discrepancy we have in our mysql settings and example docs

The correct syntax to enable TLS support is `ENABLE_TLS`
https://github.com/newrelic/nri-mysql/blob/master/mysql-config.yml.sample#L6


sigilioso (approved):
Thanks @carlossscastro for addressing this 👍

zuluecho9 (approved):

PR #8926, opened 2022-08-19
feat(synthetic IPs): Added details for additional IP ranges

bpeckNR:
- Adjusted wording of a few sections
- Made it clearer that both IP list and IP range lists should be accounted for
- Added details for additional IP ranges that will be added later in the year


akristen (approved):
Looks good! Thanks Brian!!

PR #8930, opened 2022-08-19
[NR-39298] AGP 7.2.1 breaks transform API, fixed in 7.2.2

ywang-nr:
* What problems does this PR solve? - We've identified one of the version of an essential library breaks our agent functionality


ywang-nr:
@cthomas-newrelic for review

zuluecho9 (approved):

PR #8933, opened 2022-08-19
Attribute requirements for External Services

lucidobservor:
## Give us some context

A customer opened an issue with us asking why they weren't seeing any data on their External Services page. We noticed that they weren't passing the correct attributes, and we further noticed that the docs did not contain information on what attributes were required. This PR aims to correct that.


zuluecho9:
@lucidobservor fyi, we'll need you to sign the docs site CLA. (you may have signed in past but we had a tech issue that erased those records, sorry).

zuluecho9 (approved):

PR #8947, opened 2022-08-22
change order

iazam6:
changing the order of whats most important to users and what is used most


bradleycamacho:
Hi @iazam6 could you please resign the CLA above? We had to reset it last week and everyone has to sign it again :)

PR #8951, opened 2022-08-22
Updating requirements information for Partner accounts with V2 user model

cchesser:
_What problems does this PR solve?_

We are working through the V2 user model requirements, and we are under the impression that New Relic supports V2 for Partner accounts. In this document, it appears they are not supported; however, that contradicts information we have received from New Relic (as we are a partner that has been notified in moving to V2).


zuluecho9:
thanks @cchesser!

zuluecho9 (approved):

PR #8952, opened 2022-08-22
Add OpenTelemetry to main logs in context page

rhetoric101:
Clarify that logs in context also works with OpenTelemetry.


paperclypse (dismissed):
Works for me!

PR #8964, opened 2022-08-23
chore: add cf parameters to aws metric stream

thezackm:
this PR adds details to the 7 parameters required by the CloudFormation template used to implement AWS CloudWatch Metric Streams.


akristen (approved):
Thanks for adding these details!

PR #8965, opened 2022-08-23
Create tile select component for java install docs

caylahamann:
## https://issues.newrelic.com/browse/NR-41516
- [x] Create tiles for dropdown 
- [x] Add property to config yaml (i.e. java.yml) that maps to icons
- [x] Source the icons in graphql
- [x] From graphql, query for the icons and then put them into the tiles 
- [x] Create tile selection (see design)


LizBaker (approved):
gawgeous

PR #8968, opened 2022-08-23
Store app info options in url state

caylahamann:
## https://issues.newrelic.com/browse/NR-37793
- [x] For each app info option, store in url state when selected
- [x] Loading page with app info options in url state should prepopulate / configure the page


LizBaker (dismissed):
✨

LizBaker (approved):
i approve

PR #8972, opened 2022-08-23
Sept 7 morning (prob): Add note about 60 days free support

zuluecho9:
NR-39735. Slated to go out Sept 7 morning, but we should confirm that first.


bradleycamacho (approved):

akristen:
Waiting to hear back from contacts to confirm when this can go out.

PR #8980, opened 2022-08-23
nav refactor

LizBaker:
We'll probably ultimately make a feature branch to merge this into but for now I wanted to get a PR going for review

This adds a script to build one giant nav from the smaller nav yaml files. It runs at the beginning of the build and creates a new nav file that gets sourced into gatsby's graphql and used in our nav plugin. I added it to the gitignore so folks don't worry about what's going on in there

I renamed some nav yaml files so we could programmatically reference them based on the file paths of the old "root" nav's children. we might want to do something else there but it works for now.

You'll also notice that the collapsers have redundant children, that's just because of the way we currently have the nav yamls set up, I figured there was no point in refactoring those until we know how we want things to look!

The localized nav seems to be just fine without any changes so I left things as they were in that department


scripts/createSingleNav.js
```diff
@@ -0,0 +1,47 @@
...
+      acc.push({ ...page });
+    }
+    return acc;
+  }, []);
+  const navToRuleThemAll = { ...rootNavYaml, pages: rootNavChildren };
+  const navToRuleThemAllYaml = yaml.safeDump(navToRuleThemAll, {
+    lineWidth: 99999,
+  });
+
+  fs.writeFileSync(`src/nav/generatedNav.yml`, navToRuleThemAllYaml);
```
roadlittledawn, line 44: i see the build preview had a 40m build time.

do you think that's due to building this file? if so, is there another way to build this file outside the build and just source it during build? adds some complexity, but might be worth it if it saves us ~10min of build time

src/layouts/MainLayout.js
```diff
@@ -90,14 +85,10 @@ const MainLayout = ({ data = {}, children, pageContext }) => {
               padding-bottom: ${contentPadding};
             `;
 
-            return nav?.id === rootNav.id ? (
+            return (
               <animated.div style={style} css={containerStyle}>
                 <RootNavigation nav={nav} />
```
roadlittledawn, line 90: Can we remove `Subnavigation` component entirely since it won't be used? Or were you waiting on doing that until we get better sense of rest of work involved?

scripts/createSingleNav.js
LizBaker: i've seen up to 47 min lately with develop. the script itself is really quick so i'd be surprised! but we should keep an eye on it

src/layouts/MainLayout.js
LizBaker: i was going to leave that in case but we probably could!

scripts/createSingleNav.js
roadlittledawn: oh god. really? that's...a long time.

scripts/createSingleNav.js
```diff
@@ -0,0 +1,47 @@
...
+          'utf8'
+        )
+      );
+    }
+    if (subNavChildren) {
+      acc.push({ ...page, pages: [subNavChildren] });
+    } else {
+      acc.push({ ...page });
+    }
+    return acc;
```
roadlittledawn, line 37: [nit] someone once told me when i used `reduce()` you could also write it without side effect of `.push()` if you're so inclined. something like:

```
return subNavChildren ? {...acc, { ...page, pages: [subNavChildren]} } : {...acc, { ...page }}
```

roadlittledawn (dismissed):

Issue #8989, opened 2022-08-24
Issue: Set up the Amazon CloudWatch Metric Streams integration


somensari:
The "Instant Observability" link under "Curated Content" is broken.

(https://docs.newrelic.com/docs/using-new-relic/welcome-new-relic/get-started/new-relic-quickstarts-overview/).

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/infrastructure/amazon-integrations/connect/aws-metric-stream-setup/
* **OS**: Mac OS 10.15.7
* **Browser**: Chrome 104.0.0.0
* **Device**: Unknown

rhetoric101:
@somensari thanks for pointing out this broken link! I wasn't able to find any curated dashboards for Amazon CloudWatch Metric Streams, so I am going to remove this section.

rhetoric101:
Closed by #8992

PR #8992, opened 2022-08-24
Remove section about curated dashboards

rhetoric101:
Since we had a broken link in the "Curated dashboards" section, we looked at the section more closely and realized that we don't actually have a quickstart with a dashboard for Metric Streams. So, we are removing this section.


rhetoric101:
This closes GitHub issue #8989.

homelessbirds (approved):

PR #9001, opened 2022-08-25
fix: missing escaped characters on manual install

vairoletto:
Fixed a broken command on the manual installation part on install-ecs-integration.mdx, added missing escaped characters


homelessbirds (approved):
looks good!

homelessbirds:
Thanks!

PR #9022, opened 2022-08-26
New override structure

caylahamann:


clarkmcadoo (approved):

roadlittledawn:
not sure if this going to be an issue or not, BUT

one weird quirk i noticed is if i have `deployment` set to `applicationServer` and `framework` set to `tomcat` say, i can't de-select `tomcat` to remove the framework option and see whatever content is available without a framework on a app server. step 3 then is still for docker unless i'm able to remove / de-select the framework

roadlittledawn (approved):

PR #9023, opened 2022-08-26
Collapser fixes for java install docs

caylahamann:
* removes scrolling issue 
* if the optionType is used more than once, it will collapse all of them (for example the framework selection and the top and bottom collapse at the same time)

### DIFF
--- a/gatsby/should-update-scroll.js
+++ b/gatsby/should-update-scroll.js
@@ -1,0 +1,16 @@
+const shouldUpdateScroll = ({
+  prevRouterProps,
+  routerProps: { location },
+  getSavedScrollPosition,
+}) => {
+  if (
+    location.pathname.includes('/install/') &&
+    location.pathname === prevRouterProps.location.pathname
+  ) {
+    return false;
+  }
+  const currentPosition = getSavedScrollPosition(location);
+  return currentPosition || [0, 0];
+};
+
+export default shouldUpdateScroll;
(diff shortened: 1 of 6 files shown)


gatsby/should-update-scroll.js
```diff
@@ -0,0 +1,16 @@
+const shouldUpdateScroll = ({
+  prevRouterProps,
+  routerProps: { location },
+  getSavedScrollPosition,
+}) => {
+  if (
+    location.pathname.includes('/install/') &&
```
roadlittledawn, line 7: would using `startsWith` make more sense here to avoid disabling scroll on any page url paths that contain `/install/` ?

src/components/TileSelect/TileOption.js
```diff
@@ -18,15 +18,15 @@ const TileOption = ({ displayName, value, logo, onChange, isSelected }) => {
         text-align: center;
         padding: 0.5rem;
         color: var(--system-text-primary-light);
-        border: 1px solid #3a444b;
+        border: ${isSelected
+          ? '1.25px solid var(--attention-notification-info)'
+          : ' 1px solid #3a444b'};
```
roadlittledawn, line 23: this outline is hard to see in dark mode. not a showstopper but may want to ask design for suggestion

s

roadlittledawn (approved):
Two minor suggestions but lookin' good! Nice find on the scroll stuff

gatsby/should-update-scroll.js
caylahamann: oooo sorry just saw this one, i think it would actually make sense to do that and can put that in a fix

PR #9053, opened 2022-08-30
updated image

iazam6:
I have updated the email settings image under the email column.


x8a (approved):
Thanks for the updated image!

PR #9054, opened 2022-08-30
Align Api-Key header example with error response

leslie-alldridge:
## Give us some context

Aligns the curl example with an error response people may see.

The error is:

```
Please provide an API key via the `Api-Key` header
```

So folks may be lead to believe the `API-Key` is incorrect in the example code, when in fact they need to create a `user` API key.


x8a (approved):
Thanks for the typo correction!

PR #9056, opened 2022-08-30
fix(node agent): Document the minimum version of Apollo Server that we support

jordigh:
* Closes https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/133


homelessbirds:
This looks good! thanks!

PR #9066, opened 2022-08-30
Page tools right sidebar with headings and contributing guidelines

caylahamann:
- [x] Page tools component is in install UI
- [x] h2 header links link to walkthrough component steps 
- ~Get rid of feedback part of component??~ - my opinion was to keep the create an issue just in case a user gets blocked and doesn't scroll to the end of the page, at least they have a pathway to create an issue 

https://issues.newrelic.com/browse/NR-44518

### DIFF
--- a/src/pages/install/{installConfig.agentName}.js
+++ b/src/pages/install/{installConfig.agentName}.js
@@ -3,6 +3,8 @@ import { graphql } from 'gatsby';
 import { css } from '@emotion/react';
 import {
   Walkthrough,
+  TableOfContents,
+  ContributingGuidelines,
   useQueryParams,
   Layout,
   useTessen,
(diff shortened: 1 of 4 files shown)


src/pages/install/{installConfig.agentName}.js
```diff
@@ -14,6 +16,14 @@ import AppInfoConfig from '../../components/AppInfoConfig';
 import AppInfoConfigOption from '../../components/AppInfoConfigOption';
 import InstallNextSteps from '../../components/InstallNextSteps';
 
+const slugify = (str) =>
+  str
+    .trim()
+    .toLowerCase()
+    .replace(/\s+/g, '-')
+    .replace(/-+/, '-')
+    .replace(/[^a-z0-9-]/g, '');
```
LizBaker, line 25: we do have this as an exportable function in src/scripts/utils if you wanna import instead

caylahamann: the import was being weird - I think bc of ESM vs CommonJs modules or something like that

LizBaker (approved):

PR #9070, opened 2022-08-30
NR-40035: Fix redirect build time warnings

bradleycamacho:
Work for NR-40035

Looks like #7891 added a redirect from a path, but didn't delete the underlying file. This PR deletes that files and should fix those warnings.

# REVIEWER 

- Build the entire site locally, and look at your terminal output to verify there are no errors that look like the following 

image


paperclypse (approved):
Works for me. Thanks for catching this!

PR #9107, opened 2022-09-01
Tessen fix

LizBaker:
tessen categories need to be sentence case, some events were silently failing

### DIFF
--- a/src/pages/install/{installConfig.agentName}.js
+++ b/src/pages/install/{installConfig.agentName}.js
@@ -32,6 +32,13 @@ const InstallPage = ({ data }) => {
 
   const tessen = useTessen();
 
+  const capitalize = (word) => {
+    let capitalizedWord = word.split('');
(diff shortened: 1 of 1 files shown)


src/pages/install/{installConfig.agentName}.js
```diff
@@ -55,7 +62,7 @@ const InstallPage = ({ data }) => {
   const handleAgentConfigChange = ({ name, value }) => {
     tessen.track({
       eventName: 'agentConfigFileUpdated',
-      category: `${name}AgentConfigFileUpdated`,
+      category: `${capitalize(name)}AgentConfigFileUpdated`,
```
caylahamann, line 65: it won't let me add a suggestion for line 17, but we also need to pull in location 

```
const InstallPage = ({ data, location }) => {
```

so that `location.pathname` is defined on line 53 and 68 and elsewhere

caylahamann (approved):

PR #9118, opened 2022-09-02
update tessen pt. 2

LizBaker:
This removes the value of the text entered and also only submits an event once per page render

### DIFF
--- a/src/pages/install/{installConfig.agentName}.js
+++ b/src/pages/install/{installConfig.agentName}.js
@@ -39,6 +39,7 @@ const InstallPage = ({ data, location }) => {
   const { queryParams, setQueryParam, deleteQueryParam } = useQueryParams();
   const [showGuided, setShowGuided] = useState(false);
   const [selectedIndex, setSelectedIndex] = useState(0);
+  const [agentConfigUpdate, setAgentConfigUpdate] = useState([]);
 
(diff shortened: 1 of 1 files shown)


src/pages/install/{installConfig.agentName}.js
```diff
@@ -69,15 +70,17 @@ const InstallPage = ({ data, location }) => {
...
-  const handleAgentConfigChange = ({ name, value }) => {
-    tessen.track({
-      eventName: 'agentConfigFileUpdated',
-      category: `${capitalize(name)}AgentConfigFileUpdated`,
-      key: name,
-      value,
-      path: location.pathname,
-      agentName,
-    });
+  const handleAgentConfigChange = ({ name }) => {
```
clarkmcadoo, line 73: If a user types in half of a name, then comes back to finish the name afterwards, would this fire the full name or the half name? probably nothing we could control...

LizBaker: oh! in this case name is the name of the input so "licenseKey" or "applicationName", not dynamic for the end user

clarkmcadoo: ahhh totally makes sense

clarkmcadoo (approved):

PR #9122, opened 2022-09-02
Clark/styles icons update

clarkmcadoo:


src/install/config/java.yaml
```diff
@@ -17,6 +17,7 @@ appInfo:
       - value: 'applicationServer'
         displayName: 'on an application server'
         recommendedGuided: true
+        icon: 'fe-thumbsup'
```
caylahamann, line 20:
```suggestion
        icon: 'fe-server'
```

just because we are likely going to use that

PR #9138, opened 2022-09-06
feat(agents): added blazor to net and nuxt to node

akristen:
Two additions to agent compatibility docs


akristen:
These integrations are currently in limbo, so am closing the PR as there is no release date yet

Issue #9150, opened 2022-09-07
Issue: Trace API general requirements and limits 


mrytower05:
## Broken Link

* The link in [THIS](https://docs.newrelic.com/docs/distributed-tracing/trace-api/trace-api-general-requirements-limits/#requirements) section where it says _"If you're using an integration that uses the Trace API (for example, [these integrations](https://docs.newrelic.com/docs/understand-dependencies/distributed-tracing/enable-configure/integrations-enable-distributed-tracing)), you must configure....."_ the link taking you to "these integrations" is broken. 

### Doc information (don't delete this section)

* https://docs.newrelic.com/docs/distributed-tracing/trace-api/trace-api-general-requirements-limits/
* **OS**: Mac OS 10.15.7
* **Browser**: Chrome 104.0.0.0
* **Device**: Unknown

rhetoric101:
@mrytower05 thanks for pointing out this broken link! We are fixing this in  PR #9153, so I am going to close this issue.

PR #9153, opened 2022-09-07
Update link to integrations powered by Trace API.

rhetoric101:
We had a link to an old page that listed integrations that used the Trace API. That page has been merged into our main setup page, and we only have one remaining integration that is powered with the Trace API (Kamon), so I just replaced the broken link with a link to that page.

This closes GitHub issue #9150.


akristen (approved):
Thanks for adding this fix @rhetoric101 !

PR #9164, opened 2022-09-08
Explain why multiple logins appear

rhetoric101:
This is a PR to clarify why some migrated users might still see multiple accounts when they go to log in.


akristen (approved):
Looks good! Thanks again. :)

PR #9165, opened 2022-09-08
Doc: Add documentation about new `SetName()` on the agent API

JcolemanNR:
## Give us some context

Version 10.1.0 of the .NET Agent (as yet unreleased) introduces is a new method to the `ISpan` API. These changes provide documentation about how to use this API.

Please check with our team before merging this as we would like to synchronize it with the Agent Release.


rhetoric101 (approved):
@JcolemanNR thanks for adding this information!

PR #9174, opened 2022-09-08
fix(Metrics): Make lower bound cardinality limit accurate

mcmellawatt:
The lower bound limit was not correct as it stands today.


mcmellawatt:
@pfraziernr sidekick that this was not correct and I'm not overlooking something here :).

rhetoric101 (approved):
@mcmellawatt thanks for making this update!

PR #9177, opened 2022-09-08
Add back manually deserialized files

LizBaker:
These are files we previously ignored to get the translation api and our DB back on track


LizBaker:
closing because some of these translations are in flight and smartling doesn't provide the previous version (making a ticket to revisit)

PR #9215, opened 2022-09-09
feat: create pasta recipe

bradleycamacho:
Please ignore

### DIFF
--- a/src/install/config/pasta.yaml
+++ b/src/install/config/pasta.yaml
@@ -1,4 +1,4 @@
-agentName: java
+agentName: pasta
 agentType: APM
 title: 'How to cook some great pasta'
 metaDescription: "Let's make Gordan Ramsey Proud"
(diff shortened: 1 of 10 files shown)


src/install/config/pasta.yaml
```diff
@@ -0,0 +1,36 @@
+agentName: java
```
caylahamann, line 1: just at a quick glance i assume that this is the issue! since there's already an agent with the name `java`

bradleycamacho: I believe I had it called `pasta` until I tried `java` to troubleshoot. But let me double check in case I'm wrong!

bradleycamacho: Yeah still 404ing locally w/out an error even with the new commit :(

caylahamann: so strange - i'm not having an issue building it locally 
Screen Shot 2022-09-09 at 7 39 49 PM

PR #9231, opened 2022-09-12
fix(accounts): add reason to add account

zuluecho9:
Adding additional reason about separating billing to add-account section


rhetoric101 (approved):
@zuluecho9 this looks good.

PR #9247, opened 2022-09-13
Fix configure-prometheus-openmetrics-integrations-large-kubernetes…

maxlemieux:
Replace smart quotes in NRQL with single quotes, other standardization to NRQL syntax

The existing versions with smart quotes do not run as-is on NR


x8a (approved):
Thanks!

PR #9250, opened 2022-09-13
Update firewall rule list for Pixie

ashmilhussain:
Add ```withpixie.ai:443``` to the Pixie firewall list in addition to ```work.withpixie.ai:443```. not adding this will result 'pl-cluster-secrets can not find' error

This was tested with AKS and Azure Firewall


x8a (approved):
Thank you!

PR #9252, opened 2022-09-13
chore(npm): added redirect

x8a:


x8a:
This redirect is not needed since the engineers had the wrong URL for the guided install.

PR #9292, opened 2022-09-14
feat(analytics): add optmizely snippet

roadlittledawn:
## Description

Adds optimizely snippet for marketing team to configure A/B tests and have them applied to site via scripts it injects.

Couldn't find any guidance on where script should be on page. They strongly suggest to not add `async` or `defer` unless absolutely needed (could cause flickering i guess).

Will try to work with marketing folks to try and test out before prod

## Related info
Closes https://issues.newrelic.com/browse/NR-44970

[How the Optimizely snippet works: Order of activation](https://support.optimizely.com/hc/en-us/articles/4410289779853)

### DIFF
--- a/gatsby/on-render-body.js
+++ b/gatsby/on-render-body.js
@@ -65,11 +65,16 @@ const onRenderBody = async ({ setPostBodyComponents }) => {
     />
   );
 
+  const optimizelySnippet = (
+    <script src="https://cdn.optimizely.com/public/7331003/s/web_blog_docs.js" />
+  );
+
   setPostBodyComponents([
     linkedInId,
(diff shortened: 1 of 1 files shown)


roadlittledawn:
Colin / Jason to try A/A test on preview build site so we can observe behavior for testing

roadlittledawn:
walked through with jason chen. page loads then applies treatment. for experiments they plan to run, he said it behavior should be sufficient to get reliable data.

bigger A/B tests that could change intent are typically done via fastly which routes it to one route or the other (so it's served up without weird flickering / changes happening in client)

clarkmcadoo (approved):

PR #9313, opened 2022-09-15
fix(php-agent):Update PHP compatibility specs

mfulb:
This PR contains changes to the compatibility specification page for the PHP agent requested by the PM.

These changes can go live when they have been approved, there is no need to sit on them.

Thanks!


homelessbirds (approved):

homelessbirds:
Thank you for this! Looks good.

PR #9340, opened 2022-09-16
Naming and wording fixes

keegoid-nr:
- set name to synthetics job manager
- remove use of "customer"
- fix spacing


akristen (approved):
Good catches @keegoid-nr !

PR #9354, opened 2022-09-19
Add support for JVMs page and metric/span toggle.

rhetoric101:
We have expanded the scope of the metric/span toggle so it now includes the JVMs page. This is for NR-43760 outside of the sprint process.


akristen (approved):
Looks good! My only recommendation would be to add a screenshot of some sort, if it's relevant. But if it doesn't add anything meaningful, then this is fine as is. :)

PR #9362, opened 2022-09-20
fix 8.0.0 changelog link on Ruby agent update page

fallwith:
Fix the v8.0.0 changelog link (by adding a leading slash to convert the relative link into an absolute one) on the Ruby agent update instructions page.


x8a (approved):
Thanks!

PR #9371, opened 2022-09-20
fix: wrap math in code

bradleycamacho:
Wraps equations in code


rhetoric101 (approved):
@bradleycamacho thanks for handling these math examples!

PR #9427, opened 2022-09-22
Update docs to reflect Python Agent's deprecation of Python 3.6

lrafeei:
Update New Relic docs to reflect Python Agent's deprecation of Python 3.6.

There are documents in various languages that I was not able to modify to reflect this (since I do not speak the languages), but the ones in English or with links to pages I was able to update.


zuluecho9:
@lrafeei thank you. we actually don't have to touch the other-language docs, as those are automatically updated. just for future reference.

zuluecho9 (approved):

PR #9438, opened 2022-09-22
Update scripted browser example to use async/await

keegoid-nr:
- Take advantage of features of our new runtime on Chrome 100+


x8a (approved):
Thanks a lot for the update!

PR #9447, opened 2022-09-23
Fix ibmmq readme and IDs of collapsers

paologallinaharbur:
Changing some ids and text


x8a (approved):
Thanks for these updates!

PR #9458, opened 2022-09-23
Add documentation about new environment variables in .NET Agent

JcolemanNR:
We are adding a few new environment variables to the .NET agent to control if and when data is sent during shutdown.


rhetoric101 (approved):
@JcolemanNR thanks for making these updates!

PR #9466, opened 2022-09-26
feat(runtime troubleshooting): Synthetics runtime troubleshooting

bpeckNR:
Added a troubleshooting doc to address common confusions and errors for synthetics runtime upgrades.


akristen (approved):
I've added this to the 'look over one more time' for next sprint. As it stands, though, the content is fine and needs to go out ASAP! Thanks @bpeckNR !

PR #9469, opened 2022-09-26
feat(request module): Request module compatibility details

bpeckNR:
Customers now attempting to use the request module in scripted API runtimes will receive our request like implementation of the got module instead.


akristen (approved):
Same comment as the last one~

PR #9480, opened 2022-09-26
Update .NET agent logs-in-context configuration instructions

nr-ahemsath:
## Give us some context

The instructions were incorrect for configuring the "local log decoration" scenario, for a couple of .NET logging libraries (log4net and Serilog), in the case of JSON-formatted output.  The original instructions were based on an incorrect assumption that the logging ingest API would handle having the `NR-LINKING` metadata string in a separate property from the message.

These new instructions demonstrate how users of log4net or Serilog who need to use the local decoration scenario and require JSON-formatted output can accomplish this with custom formatting classes.


nr-ahemsath:
Our team is collaborating on coming up with even better instructions so I'm going to close this PR and will open a new one when we're ready.

PR #9489, opened 2022-09-27
Add kafka python agent docs

hmstepanek:

## Give us some context


src/content/docs/apm/agents/python-agent/supported-features/python-message-queues.mdx
```diff
@@ -17,49 +18,21 @@ import rabbitmqTransactionTraceSummary0 from 'images/rabbitmq-transaction-trace-
...
-      </th>
-
-      <th>
-        **Publish a message (`Put` in UI)**
+        **Message Queue**
       </th>
 
       <th>
-        **Receive a message (`Take` in UI)**
+        **Python Agent Version Required**
```
hmstepanek, line 35: Good call on making this table.

src/content/docs/apm/agents/python-agent/supported-features/python-message-queues.mdx
```diff
@@ -71,21 +44,55 @@ Supported entry points for queue operations appear as `Put` (publish a message)
...
+* Time your app spends processing/consuming messages
+
+[APM](#view-queue) conveniently groups and reports operations that interact with queues. By analyzing this information, you can more easily identify bottlenecks and areas for performance improvement in your message passing architecture.
+
 ## View in New Relic UI [#view-queue]
 
-Queue operations appear on APM's [**Transactions** page](/docs/apm/applications-menu/monitoring/transactions-dashboard) for the selected app. The `Put` and `Take` metrics appear in the **Breakdown table** and are categorized as `MessageBroker` metrics. Here is an example:
+Queue operations appear on APM's [**Transactions** page](/docs/apm/applications-menu/monitoring/transactions-dashboard) for the selected app. 
+
+The `Put` and `Take` metrics appear in the **Breakdown table** and are categorized as `MessageBroker` metrics for RabbitMQ. Here is an example:
```
hmstepanek, line 95: I think we should clarify somewhere what `Put` and `Take` translate to in terms of kafka (`Produce` and `Consume`).

src/content/docs/apm/agents/python-agent/supported-features/python-message-queues.mdx
```diff
@@ -114,15 +121,3 @@ The [**Transaction trace summary** page](/docs/apm/transactions/transaction-trac
 <figcaption>
   **[one.newrelic.com](https://one.newrelic.com/) > APM > (select an app) > (select a transaction trace):** In this example, the selected transaction trace's **Summary** shows RabbitMQ operations in the colored bar chart. The **Slowest components** section also lists the RabbitMQ `Put` and `Take` results.
 </figcaption>
-
```
hmstepanek, line 117: Why did we remove this?

src/content/docs/apm/agents/python-agent/supported-features/python-message-queues.mdx
hmstepanek: It might be helpful to extend this table to show what monitoring features we support. For example, serialization, deserialization, consumer, producer, heartbeat, etc. I'm specifically thinking of the fact that we don't support deserialization on the consumer for kafka-python here and how we might call that out. Also how we might call out/advise the customer to setup alerts for heartbeat.

hmstepanek: I'm thinking it might also be helpful to provide some clarification around how we expect the customer to use the consumer for each library version because we do not support other usage models. Certainly I think this would come in useful for support if nothing else. I'm not sure if we generally go into this sort of detail like this in the docs but just a thought. Something like:

In kafka-python each iteration of the consumer is reported as an individual transaction:
```
for message in consumer:
   print(message)
```

In confluent-kafka, each call to consumer.poll is reported as an individual transaction:
```
while True:
    record = consumer.poll(0.5)
    if not record:
        break
```

We could also simplify/summarize the above examples to the following:
> When consuming messages via kafka, transactions are reported for each message as it is consumed by the consumer. Transactions will not be reported for bulk consumption operations where multiple messages are consumed at once.

hmstepanek (commented):
A couple requested changes/suggestions.

src/content/docs/apm/agents/python-agent/supported-features/python-message-queues.mdx
lrafeei: In terms of the specific examples, I think we're already expecting that the customer has written code that works for them (as in, the examples we should provide are things that integrate NR APIs and such).  That being said, good call on the summary and I'll add that as a callout.

src/content/docs/apm/agents/python-agent/supported-features/python-message-queues.mdx
lrafeei: This was an old feature in the APM that has been removed since the writing of the Pika RabbitMQ instrumentation, so I just took it out.

src/content/docs/apm/agents/python-agent/supported-features/python-message-queues.mdx
lrafeei: Since I am pointing to an example for RabbitMQ, I'll add framework agnostic clarification to the terms, but I will add a table before this section that will clarify what terms correspond with what brokers.

src/content/docs/apm/agents/python-agent/supported-features/python-message-queues.mdx
lrafeei: Also, adding a separate table for features supported.

PR #9500, opened 2022-09-27
docs: add lang identifiers to example code

brnhensley:
I also corrected the indentation which was wonky from cutting/pasting the doc together


x8a (approved):
Thanks for adding these!

PR #9522, opened 2022-09-28
Add custom attribute to identify type of page

roadlittledawn:
## Description

Adds custom attribute to Browser `PageAction` and `PageView` events to explicitly identify the type of page for more reliable querying.

Partiuclarly useful when querying data for new install UI pages generated via `src/pages/install/{installConfig.agentName}.js`

Also removes tessen tracking for when a step is made active because it's very noisy and likely not very useful

## Categories

Kinda just making up a nomenclature, but I thought it could be useful to categorize pages first as:
* Interactive
  * Pages that change based on user input
* Dynamic
  * Pages where content on the page updates automatically / dynamically based on where data is sourced
* Template
  * Page of templated content rendered via a template in `src/templates`

Then just try to match file names mostly so it's obvious what template or page generated it

### DIFF
--- a/src/pages/attribute-dictionary.js
+++ b/src/pages/attribute-dictionary.js
@@ -31,7 +31,7 @@ const AttributeDictionary = ({ data, pageContext, location }) => {
   if (typeof window !== 'undefined' && typeof newrelic === 'object') {
     window.newrelic.setCustomAttribute(
       'pageType',
-      'Interactive:PagesAttributeDictionary'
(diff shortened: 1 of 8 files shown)


roadlittledawn:
Liz gave verbal approval. I swear.

roadlittledawn:
one caveat doing it this way:

this attaches the attribute to the `PageAction` and `PageView` events that Tessen sends with our Browser agent. it doesn't send this attribute to segment and beyond

PR #9524, opened 2022-09-28
Replace CAT reference with DT reference in nav

rhetoric101:
We had an outdated URL and title in the API section of the Java instructions. I updated the title and the URL.

### DIFF
--- a/src/nav/apm.yml
+++ b/src/nav/apm.yml
@@ -280,8 +280,8 @@ pages:
                 path: /docs/apm/agents/java-agent/api-guides/java-agent-api-instrument-external-calls-messaging-datastore-web-frameworks
               - title: 'API: Annotating example app'
                 path: /docs/apm/agents/java-agent/api-guides/java-agent-api-custom-instrumentation-annotation-example-app
-              - title: 'API example: Datastore and CAT'
-                path: /docs/apm/agents/java-agent/api-guides/java-agent-api-instrumenting-example-app-external-datastore-calls-cat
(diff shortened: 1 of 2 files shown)


src/nav/apm.yml
```diff
@@ -280,8 +280,8 @@ pages:
                 path: /docs/apm/agents/java-agent/api-guides/java-agent-api-instrument-external-calls-messaging-datastore-web-frameworks
               - title: 'API: Annotating example app'
                 path: /docs/apm/agents/java-agent/api-guides/java-agent-api-custom-instrumentation-annotation-example-app
-              - title: 'API example: Datastore and CAT'
-                path: /docs/apm/agents/java-agent/api-guides/java-agent-api-instrumenting-example-app-external-datastore-calls-cat
+              - title: 'API: External datastore calls and distributed tracing'
+                path: /docs/apm/agents/java-agent/api-guides/java-agent-api-instrumenting-example-app-external-datastore-calls-dt
```
austin-schaefer, line 284: Does the filename need to change since the path changed here?

austin-schaefer (approved):

src/nav/apm.yml
rhetoric101: @austin-schaefer I not sure I understand your question. The paths remain the same, as far as I can see. What changed was the filename. Plus, I renamed the node name in the navigation file so it now mentions distributed tracing instead of cross application tracing. It seems to work in the preview, so I think we're good! I did make one additional fix where I replaced "DT" in the title with "distributed tracing."

PR #9525, opened 2022-09-28
Event harvest config callout in Python Agent

lrafeei:
This PR adds a callout tip in [this file](/Users/lrafeei/repo/docs-website/src/content/docs/apm/agents/python-agent/configuration/python-agent-configuration.mdx) to clarify that if there is an event data count of < 12, it will round down to 0.


rhetoric101 (approved):
@lrafeei these changes look good! I just made a few minor changes.

PR #9526, opened 2022-09-28
What's new: OTel metrics-based Transactions

rkbodenner:
A revision of Transactions for OTel services will be released soon.

PLEASE HOLD merging until the feature is released to all customers. I'll make a comment to that effect here when it's ready.


rkbodenner:
OK, this is ready to go! Ship when ready, please.

PR #9540, opened 2022-09-29
feat: infra-agent macos arm support

josemore:
## Give us some context

* Starting with infrastructure agent 1.30.0, macos is supported on Apple silicon too (M1, M2 processors).


x8a (approved):
Thanks a lot!

PR #9549, opened 2022-09-29
Update nav design pt. 1

LizBaker:
This includes a theme update and updates the general design but not the hierarchy of the nav items themselves

### DIFF
--- a/src/nav/root.yml
+++ b/src/nav/root.yml
@@ -1,6 +1,7 @@
 title: root
 path: /
 pages:
+  - title: Welcome
   - title: Welcome to New Relic
     path: /docs/new-relic-solutions/get-started/intro-new-relic
   - title: Install New Relic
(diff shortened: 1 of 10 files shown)


src/nav/root.yml
```diff
@@ -1,6 +1,7 @@
 title: root
 path: /
 pages:
+  - title: Welcome
```
clarkmcadoo, line 4: `nit` Do we also wanna add the "Full Stack Monitoring"? after the line break?

Screen Shot 2022-10-03 at 3 01 35 PM

LizBaker: I'm gonna save that part for the ticket where we refactor the nav items themselves. I just added `welcome` so we could see the styling 💅

clarkmcadoo (approved):

PR #9555, opened 2022-09-29
fix(add accounts): add reqs

zuluecho9:


rhetoric101 (approved):
@zuluecho9 This looks good! I added some missing periods. In a future edit, you might change the comment about not adding too many accounts like this: See the reasons for keeping your accounts to a minimum in [Organization and account structure](/docs/accounts/accounts-billing/account-structure/new-relic-account-structure).

PR #9556, opened 2022-09-29
Regex to update layouts on pages that exist

clarkmcadoo:
Had an issue where the 404 page template was being applied to pages that had "404" anywhere in the file name.  This will update the layout once again if the URL is one that exists in the "/docs/" page routing and is not getting a 404 from gatsby.  I thought this wouldn't work at first since the url doesn't get updated when you get a 404, but it seems like when the 404 page is displayed this doesn't update the layout.

https://issues.newrelic.com/browse/NEWRELIC-3231

BEFORE:
Screen Shot 2022-09-29 at 5 24 51 PM

AFTER:
Screen Shot 2022-09-29 at 5 35 21 PM

to test, go to the build and enter the slug: `/docs/release-notes/mobile-apps-release-notes/new-relic-ios-release-notes/new-relic-ios-4042`

Then, go to a page that doesn't exist to see the 404 page working correctly :)

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -368,6 +368,10 @@ exports.onCreatePage = ({ page, actions }) => {
     page.context.layout = 'basic';
   }
 
+  if (page.path.match(/404/) && page.path.match(/\/docs\//)) {
+    page.context.layout = 'default';
+  }
+
   if (page.path.includes('/install/')) {
     const pagePathArray = page.path.split('/');
     const lastItem = pagePathArray[pagePathArray.length - 1];


roadlittledawn (changes requested):
Can we dig into this tomorrow? Because I don't totally understand why this works. But also the theme has some [similar code in gatsby-node.js](https://github.com/newrelic/gatsby-theme-newrelic/blob/ee98235d61349accf826a4ba3c899c64c5dd72be/packages/gatsby-theme-newrelic/gatsby-node.js#L319-L320) which we may want to update as well. I also think we may want to specifically look for 404 page slug instead of matching on path

roadlittledawn:
fyi. i added this bit to both docs and theme's gatsby-node.js and it's workin' for both the relevant release note files with `404` in the path AND custom 404 page.

```js
const slug404 = ['/404.html', '/404-page', '/dev-404-page'];

if (slug404.includes(page.context.slug)) {
...
}
```

PR #9596, opened 2022-10-03
Verify install script WIP

LizBaker:
This adds a new script that can be run and will check the install config file for a given install page/agent. Currently it's only checking some basics but we can add to it later

### DIFF
--- a/scripts/verifyInstallPage.js
+++ b/scripts/verifyInstallPage.js
@@ -1,0 +1,140 @@
+/* eslint-disable no-console */
+const fs = require('fs');
+const yaml = require('js-yaml');
+const path = require('path');
+
+const verifySteps = (steps, installPage) => {
+  let issues = 0;
+  steps.forEach((step) => {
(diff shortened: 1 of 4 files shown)


scripts/verifyInstallPage.js
```diff
@@ -0,0 +1,140 @@
...
+        }
+      });
+    }
+  });
+  if (issues > 0) {
+    console.error(
+      '%s\x1b[31m%s\x1b[0m',
+      'please fix ',
+      issues,
+      'issue/s above \n'
```
clarkmcadoo, line 33: `huge nit` it cool if we change this to "issue(s)" just think it'd be easier to read maybe 🤷

clarkmcadoo (dismissed):

scripts/verifyInstallPage.js
LizBaker: ooh yeah i'm down

clarkmcadoo (approved):

PR #9601, opened 2022-10-04
Add EU info to Intro

iazam6:
Based on updates across NR AIOps docs, this paragraph was requested to be moved to this page by Shachar because it is currently on the "Getting Started with Incident Intelligence" page, which will be removed completely.


x8a (approved):
Thanks for adding this!

PR #9604, opened 2022-10-04
Navigation Changes

iazam6:
Based on updates across NR AIOps docs, these changes were requested by Shachar. The content on "Getting Started with Incident Intelligence" page has been moved to other pages because, this page should only have information on environments, which is also why the page was renamed. All other content was moved to other pages which also have pull requests. Finally, this page will now be under the Settings navigation under Applied Intelligence since it is only focused on setting up account environments. 

Tl;Dr Summary:
1) Content deleted on this page was moved to other doc pages which have their own pull requests
2) The page was renamed because it now only contains content on environments
3) Because the content is only on setting up environments, it was moved on the navigation menu to the column under settings


x8a (dismissed):
Thanks for these changes!

PR #9608, opened 2022-10-04
Update Image on Overview Page

iazam6:
Based on updates across NR AIOps docs, the overview page image was changed based on the request of Shachar since we also updated the overview page flow.


x8a (approved):
Looks great, thanks!

PR #9616, opened 2022-10-04
fix: add serialization handlers for the various Tabs Components

clarkmcadoo:


roadlittledawn (changes requested):
i believe we need to expose certain text for translation and not serializes the whole component

scripts/actions/utils/handlers.js
```diff
@@ -196,15 +196,17 @@ module.exports = {
...
     deserialize: deserializeComponent,
   },
   TabsPages: {
     serialize: serializeComponent,
     deserialize: deserializeComponent,
   },
   TabsPageItem: {
-    serialize: serializeComponent,
+    serialize: (h, node) =>
+      serializeComponent(h, node, { textAttributes: ['id'] }),
```
roadlittledawn, line 209: i don't think this is necessary since we don't need to translate id values

roadlittledawn (approved):
:shipit:

PR #9621, opened 2022-10-04
Add callout re: Sitecore log4net incompatibility

nr-ahemsath:
## Give us some context

Since releasing the in-agent log forwarding and local decoration features, the .NET agent has had several customers using Sitecore CMS wondering why their Sitecore logs weren't being forwarded (or decorated) by the agent.  Unfortunately Sitecore uses an older version of log4net with different assembly and class names that our log4net instrumentation does not match at the present time.  This PR updates the relevant doc to specifically warn users about Sitecore's log4net being unsupported.


homelessbirds (approved):

homelessbirds:
Thank you for this!

PR #9623, opened 2022-10-04
Add more checks to script

LizBaker:
Now verifies that step and override filePaths actually point to real files and that overrides always have a selectedOptions key


clarkmcadoo (approved):

clarkmcadoo:
this is great

PR #9626, opened 2022-10-05
Drop old MSSQL versions

kang-makes:
## Give us some context

While theoretically we can still support all those versions as the communication protocol has not changed and we should still be able to monitor them, we have lost the possibility to test them as there are no installers available for them.

Microsft does not support these versions so we are dropping support for them too.


x8a:
Hey, @kang-makes! Thanks so much for the PR!

Could you sign the CLA? We need all contributors to do so before we can merge their PRs. Thanks!

kang-makes:
Hi @x8a!

It is now ready to be merged. The CLA was signed but it seems there is an issue with the webhook as one of them is still showing that I had not signed it.

Let me know if there is something else I can do.

x8a:
It's all looking good, so I'll go ahead and merge it. Thanks a lot!

PR #9628, opened 2022-10-05
fix(synthetics apdex): Removed references for editing Apdex T

bpeckNR:
The synthetics UI and APIs no longer allow you to modify the Apdex T value used in the SLA report. I clarified the documentation and referred to the NRQL apdex function if an Apdex value is needed.


x8a (approved):
Thanks a lot for the update!

PR #9631, opened 2022-10-05
Use single config for gatsby-plugin-newrelic

LizBaker:
gatsby-plugin-newrelic deprecated multiple configs awhile back but we continued to use that syntax. this uses ternaries to toggle between the envs (per the plugin's suggestion) https://github.com/newrelic/gatsby-plugin-newrelic#installation


LizBaker:
re-triggering checks after github actions issue

clarkmcadoo (approved):

PR #9650, opened 2022-10-06
feat(style): Updated branching guide doc with info about the verify YAML script

paperclypse:
I added some info about our new YAML verify script and what it checks for.


zuluecho9 (approved):
removed var tags, as they don't work in backtick code styling, and we are removing them also generally

PR #9651, opened 2022-10-06
Link to supported Java frameworks

planteater:
Similar to what we do for Python, I've added a link to supported Java frameworks.


akristen (approved):
Thanks for this update! :)

PR #9653, opened 2022-10-06
New 'streaming export' ng doc

zuluecho9:
NR-55229


akristen (dismissed):
Looks good! Nothing sticks out -- am wondering if we should default to collapsers for NerdGraph in general, or if audience typically prefers it to be laid out like this.

zuluecho9:
> Looks good! Nothing sticks out -- am wondering if we should default to collapsers for NerdGraph in general, or if audience typically prefers it to be laid out like this.

my general preference is to not use collapsers, esp as I think coders like to skim quickly through docs. we've gotten complaint in possible that too much collapsers make things hard to parse. i avoid them mainly except when it's really a case of a set of options being mutually exclusive options

PR #9656, opened 2022-10-06
Update RestSharp supported versions info.

vuqtran88:
Update the compatibility matrix with RestSharp supported versions info.


rhetoric101 (approved):
@vuqtran88 thanks for making this update about the versions!

PR #9657, opened 2022-10-06
What's new: nerdgraph org user mgmt update

aalfano7:
Announcing new NerdGraph org and user management functionalities. Please backdate this to September 30th to match other related comms. I am a PMM, so this post does not require PMM review. Thank you!


rhetoric101 (approved):
@aalfano7 thanks for adding this. I changed the filename and tested this locally. It looks good.

PR #9658, opened 2022-10-06
More refactoring

LizBaker:
I wanted to make it more obvious how the logic in root.yml was working so it now expects:
- any paths without a slash to be a subnav yaml file in the nav folder
- if there's a slash it expects this is an actual page and links to it
- if there's no page at all it assumes it's a category. it also ignores whats-new and release notes since those have their own separate nav generation


LizBaker:
wrong branch 😭

PR #9668, opened 2022-10-07
fix(Metrics): Fixing time window in query

mcmellawatt:
We enforce a 50k per 30 second window cardinality limit, not per 24 hours.  Updating docs to reflect that.


homelessbirds (approved):

homelessbirds:
thank you!

PR #9678, opened 2022-10-07
Add syntax highlighting to NRQL docs

brnhensley:
## Give us some context

I hit most of the main nrql docs with language identifiers for sql, which looks good in most cases I've seen before, but some of these docs have more complicated queries so it might not come off as well.

I also fix a number of smaller typos and syntax issues.


brnhensley:
thanks @homelessbirds

PR #9728, opened 2022-10-11
Updates to Java branching doc (NR-56333)

paperclypse:
Making some tweaks and improvements.

### DIFF
--- a/src/install/java/license.mdx
+++ b/src/install/java/license.mdx
@@ -1,0 +1,10 @@
+---
+componentType: default
+headingText: Get your license key
+---
+
+In order to finish installing the APM agent, you need a valid New Relic license key.
+
+Go to the [API keys page](https://one.newrelic.com/api-keys) in the UI and click **Create a key**, and then create an ingest-license key type. Alternately, if you have an existing license key, click **...**, then click **Copy key ID**.
+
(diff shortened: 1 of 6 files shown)


src/install/java/license.mdx
```diff
@@ -0,0 +1,10 @@
+---
```
bradleycamacho, line 1: I'm not opposed either way, but why not follow the normal pattern of linking out to the license key doc here?

bradleycamacho (approved):
approved except one small comment

src/install/java/license.mdx
paperclypse: Good question! The main reason is that I think our license key doc sort of muddies the waters a bit. And also I'd rather someone just went and did the thing than read a doc about it. But if you think it's worth linking to, I'm happy to.

austin-schaefer: Generally links into the UI are a good conversion spot, so I think I'd prefer a direct link over a link to the (somewhat complicated) license key doc

LizBaker:
re-trigger gatsby

src/install/java/license.mdx
paperclypse: I've come up with a both/and solution here. We can always remove the link to the license key doc later if we want to.

PR #9733, opened 2022-10-12
Remove instructions to use NerdGraph to delete Browser applications

philweber:
`entityDelete` cannot delete entities of type BROWSER-APPLICATION: https://docs.newrelic.com/docs/apis/nerdgraph/examples/nerdgraph-entities-api-tutorial/#delete-entities


akristen (approved):
Thanks for the update!

PR #9742, opened 2022-10-12
fix: remove unnecessary hashes from redirect paths

roadlittledawn:
## Description

Removes hashes from redirect url paths because they're unnecessary. And could cause weird build issues (at least it did in netlify)


clarkmcadoo (approved):

akristen (approved):
Thanks for this!

PR #9776, opened 2022-10-13
docs: put code in backticks

brnhensley:
`tls.SLAB_BUFFER_SIZE` appears to be pretty old and no longer a thing. there is no mention of it in that link, and googling the phrase only brings up very old docs (Node 0.x) as well as this page on docs.NR and a 2014 stack overflow. I couldn't figure out what the updated version might be though.


zuluecho9:
thanks @brnhensley

zuluecho9 (approved):

PR #9785, opened 2022-10-13
Note semantic conventions, tweak page descriptions

rkbodenner:
Improved descriptions of what you can do with each UI page.

Included an overall statement on how we use semantic conventions to power the UIs, so users can understand why data might not be showing up as expected.


rhetoric101 (dismissed):
@rkbodenner these descriptions are a nice improvement! Thanks for adding them!

rhetoric101 (approved):
@rkbodenner good to have these better descriptions since this page gets some traffic!

PR #9786, opened 2022-10-13
Metrics/spans, semantic conventions, server/client distinction

rkbodenner:
The choice of presenting the 'server' role for a service (vs. 'client', etc.) is not always clear, so calling that out here with references to the OTel data model.

Adding a note on metrics along with a known limitation.


rhetoric101 (approved):
@rkbodenner thanks for adding this material!

PR #9790, opened 2022-10-13
Refer to global search, reorder "find your entity" parts

rkbodenner:
- Moved the sections on finding your entity next to each other and made one a sub-section of the other
- Added a tip callout for global search, with specific advice for OTel services
- Cleaned up language about the overall page being about both finding entities and filtering data
- Consistent formatting for numbered lists
- Fixed name of External services


rhetoric101 (approved):
@rkbodenner thanks for adding these details!!

PR #9797, opened 2022-10-14
chore: Fixed CoE left navigation, updated images with latest

khickey-newrelic:
Hey, docs hero - this is just an image update (to match what we're presenting a Futurestack) as well as a small capitalization correction.


akristen (approved):
These are nice images! Thanks for the update. :)

PR #9804, opened 2022-10-14
Add new queues to SyntheticsPrivateLocationStatus

keegoid-nr:
This allows one to see if monitors on a particular runtime are contributing to queue growth. This can happen if:

- Monitors are set to the new runtimes (Chrome 100+, Node 16+) without having a [Synthetics Job Manager](https://docs.newrelic.com/docs/synthetics/synthetic-monitoring/private-locations/install-job-manager/) to process those jobs for a private location.
- Monitors are set to the legacy runtimes (Chrome 72-, Node 10-) without having a [Containerized Private Minion](https://docs.newrelic.com/docs/synthetics/synthetic-monitoring/private-locations/install-containerized-private-minions-cpms/#start) to process those jobs for a private location.

The new queues are:

- `pingJobs`
- `jobManagerHeavyweightJobs`
- `minionHeavyweightJobs`

### DIFF
--- a/src/content/docs/synthetics/synthetic-monitoring/private-locations/monitor-private-locations.mdx
+++ b/src/content/docs/synthetics/synthetic-monitoring/private-locations/monitor-private-locations.mdx
@@ -48,7 +48,7 @@ The following Private Minion dashboard example JSON can be imported to your acco
           "description": null,
           "widgets": [
             {
-              "title": "pending checks by location",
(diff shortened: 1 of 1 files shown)


keegoid-nr:
Both the job manager and the minion can process ping jobs, so that's why there is only one `pingJobs` queue for both.

akristen (approved):
Thanks for adding all this content!

PR #9860, opened 2022-10-18
Add CircleCI integration

planteater:
1. Reorganized and renamed the “How to use CodeStream” section, as the current paradigm didn’t work anymore. Originally the titles were oriented around walking the user through a workflow, but as we add more features into CS that model is breaking down.
2. Added new CI/CD article to that section.
3. Added expander for Blame Hovers to the UI Overview article. Also, added CI/CD in bullet list in Sidebar section.


src/nav/codestream.yml
```diff
@@ -47,7 +47,7 @@ pages:
         path: /docs/codestream/troubleshooting/git-issues
       - title: GitHub organizations
         path: /docs/codestream/troubleshooting/github-org-repos
-      - title: GitLab Self-Managed version
+      - title: GitLab self-managed version
```
planteater, line 50: @zuluecho9 - "GitLab Self-Managed" is a product name and needs to be capitalized.

zuluecho9: i saw it lower case here and elsewhere: https://docs.gitlab.com/ee/subscriptions/self_managed/. are you seeing it capitalized elsewhere? @planteater

planteater: Looks like GL is inconsistent in their display, although that feels like a recent thing. I think we should stick with upper either way though, as we use that all throughout the product. And it's simply clearer that it's a product name.

zuluecho9: ok

zuluecho9 (approved):

zuluecho9:
sorry @planteater restarting the build, think it timed out. hopefully will go out in morning.

planteater:
@zuluecho9 - Seems like things are still stuck here?

planteater:
There some urgency in having this live tomorrow if at all possible. Thanks!

zuluecho9:
sorry @planteater, thought I merged that. merging now

PR #9864, opened 2022-10-18
NEWRELIC-4620 - Nodejs Agent: correct version for DT enabled by default

jmartin4563:
There was some confusion these docs caused with a customer, in [7.2.0](https://docs.newrelic.com/docs/release-notes/agent-release-notes/nodejs-release-notes/node-agent-7-2-0/) we updated the default config that customers can copy as an example to have DT enabled, but DT was not always enabled by default until [8.3.0](https://docs.newrelic.com/docs/release-notes/agent-release-notes/nodejs-release-notes/node-agent-8-3-0/)

This PR uses the correct version number for DT enabled by default.

Closes [NEWRELIC-4620](https://issues.newrelic.com/browse/NEWRELIC-4620)


zuluecho9:
thank you @jmartin4563

PR #9870, opened 2022-10-18
chore(the site): added /all-capabilities to the end of the one.nr.com…

akristen:
Updating the URL across the whole docs site! I checked the first ~100ish or so.


austin-schaefer (approved):
Spot-checked several dozen without finding any issues.

PR #9877, opened 2022-10-19
NEWRELIC-3025 Infra Agent: Added ntp metric

rubenruizdegauna:
Added new metric to the Infra Agent.
* config option
* metric description

### DIFF
--- a/src/content/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings.mdx
+++ b/src/content/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings.mdx
@@ -5040,6 +5040,268 @@ If you are having problems with proxy configuration, see [Proxy troubleshooting]
       </tbody>
     </table>
   </Collapser>
+
+  <Collapser
+      className="freq-link"
+      id="ntp-metrics"
+      title="ntp-metrics"
+    >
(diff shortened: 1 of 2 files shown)


src/content/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings.mdx
```diff
@@ -5040,6 +5040,271 @@ If you are having problems with proxy configuration, see [Proxy troubleshooting]
...
   </Collapser>
+
+  <Collapser
+      className="freq-link"
+      id="ntp-metrics"
+      title="ntp_metrics"
+    >
+     `ntp_metrics` is a key-value map usted to configure the NTP. When enabled and an NTP Hosts list is provided,
+     the agent will provide the host's NTP Offset in the System Sample. This value is requested to the provided NTP
+     hosts pool every `interval` minutes. Between intervals the last requested value will be reported.
```
josemore, line 5051:
```suggestion
     `ntp_metrics` is a key-value map used to configure the time offset metric. When enabled and an NTP Hosts list is provided,
     the agent will provide the host's `ntp offset` metric (in seconds). This value is checked against the provided NTP
     hosts pool every `interval` minutes (default: 15). Between intervals, the last known offset is reported.
     If the reported offset is greater than few seconds it can cause issues with alert conditions, NRQL queries and data gaps in dashboards.
```

PR #9887, opened 2022-10-19
Update opentelemetry-best-practices-metrics.mdx

jonahaapala:
## Give us some context

Clarify how OTLP summaries will be translated into NR summaries. We had a user wondering how to query percentiles on their summary metrics which is not supported.


homelessbirds:
Thanks!

PR #9912, opened 2022-10-20
chore(homepage): We have 500 quickstarts now!

austin-schaefer:
Tweaks homepage to say 500+ quickstarts instead of 470+


paperclypse (changes requested):
You should also update the left nav with this info too, where it says: "See our 470+" quickstarts @austin-schaefer

austin-schaefer:
> You should also update the left nav with this info too, where it says: "See our 470+" quickstarts @austin-schaefer

@paperclypse Good catch, fixed

paperclypse (approved):
Assuming the build works, approved!

PR #9933, opened 2022-10-21
NEWRELIC-4300 Infra Agent: Added new memory metrics

rubenruizdegauna:
Added new memory metrics to Infra Agent


x8a (approved):
Thanks for adding these!

PR #9945, opened 2022-10-21
chore(screenshots): Remove unused images

austin-schaefer:
Ran the `node remove-unused-images` script


paperclypse (approved):
Looks good! Basically, skimmed through the list looking for any images that matched our new naming syntax, but didn't spot any.

PR #9970, opened 2022-10-24
chore(images): Remove unused images

austin-schaefer:
using `yarn remove-unused-images`


paperclypse (approved):
It's weird to me that a lot of these images are in the new naming syntax. I would expect these to be used? But maybe you've already checked into that?

austin-schaefer:
> It's weird to me that a lot of these images are in the new naming syntax. I would expect these to be used? But maybe you've already checked into that?

@paperclypse Just checked and none of these images are used despite their seemingly new naming conventions. ¯\_(ツ)_/¯

PR #9974, opened 2022-10-24
Style guide revision for Usage page

bhecknr:
Added "backend" and "frontend." This was decided on word-nerds.


x8a (approved):
Thanks for adding this!

PR #9988, opened 2022-10-24
chore(apm): delete old install doc + add redirects

bradleycamacho:
Looks like the old java install doc was left in place after we launched the new branching install doc. I assume it was to not overload the new doc with redirects, but there are two issues:

- It's causing warnings in the terminal
- If you go to a URL that is included in the redirects for the original doc, the original doc loads for a few seconds before redirect to the new doc as there's two sets of redirects it has to parse (doc 1 redirects to doc 2 which redirects to doc 3)


paperclypse (approved):
@bradleycamacho mostly it was a failure of nerve :D But it's time. Thanks for noticing and taking action!

PR #9996, opened 2022-10-25
fix(style): small contraction pass

bradleycamacho:
Fixes a few small contraction issues I ran into while reading our style guide


akristen (approved):
It's lookin' gouda!

PR #10008, opened 2022-10-25
Node.js agent log forwarding API

jordigh:
We have merged a new API for forwarding logs, so let's document it more!

* https://github.com/newrelic/node-newrelic/pull/1380


zuluecho9:
thank you @jordigh

zuluecho9 (approved):

PR #10038, opened 2022-10-26
Revert single nav object

LizBaker:
We've got a lot of memory issues resulting from the recent single nav changes. This will revert back to sub navs but retain the styling updates


plugins/gatsby-source-nav/gatsby-node.js
```diff
@@ -116,7 +138,23 @@ const createWhatsNewNav = async ({ createNodeId, nodeModel }) => {
   });
 
   const posts = Array.from(entries);
-  const navItems = formatPosts(posts);
+
+  const currentYear = new Date().getFullYear();
+  const postsByYear = groupBy(posts, (post) => parseDate(post).getFullYear());
```
caylahamann, line 143: did we not order these by date before??

caylahamann (approved):
this all looks good to me!! i can't spot any issues from looking through

PR #10059, opened 2022-10-27
Update java requirements

planteater:
Update requirements to reflect that CLM is now on by default in Java agent 7.11.0 and higher.


akristen (approved):
Thanks for keeping the docs up to date!

PR #10075, opened 2022-10-27
Fix/#10051 kafka otel example

### LINKED ISSUE #10051: Issue: Collector for Confluent Cloud & Kafka monitoring

The problem:
* The documentation mentions both a Confluent Cloud API keys and a Confluent Resource API keys, but it is unclear where to use either in the config file because they are the same name in the example text. Namely `CONFLUENT_API_ID` and `CONFLUENT_API_SECRET`.
* The example `config.yaml` file does _not_ use the same names, but instead prefixes names with `INSERT_`.
* The example `config.yaml` has a config key at `receivers.kafkametrics.brokers`, but shows a single blank list entry. It is not clear if this list should be left blank or if actual Kafka broker endpoints should be entered.

Solution:
* Give the Confluent Cloud API key/secret and Confluent resource API key/secret different names.
* Use the matching names in the right places in the example `config.yaml`.
* Fix the broker list config to be clearer.

* https://docs.newrelic.com/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/collector/collector-configuration-examples/opentelemetry-collector-kafka-confluentcloud/
* **OS**: Unknown
* **Browser**: Unknown
* **Device**: Unknown

abeach-nr:
This fixes #10051.

The replace variables are renamed to match the rest of the document. Also updated missing path config.

### DIFF
--- a/src/content/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/collector/collector-configuration-examples/opentelemetry-collector-kafka-confluentcloud.mdx
+++ b/src/content/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/collector/collector-configuration-examples/opentelemetry-collector-kafka-confluentcloud.mdx
@@ -51,23 +51,26 @@ Create a new file called `config.yaml` from the example below.
 
 Replace the following keys in the file with your own values:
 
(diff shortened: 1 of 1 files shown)


akristen (approved):
Thanks for updating this! Sorry this had such a delay to get built / merged. :)

PR #10090, opened 2022-10-28
Update Python agent compatibility docs.

umaannamalai:
This PR updates the Python agent compatibility docs with Python version and agent support information.


akristen (approved):
Thanks for these updates!

PR #10094, opened 2022-10-28
fix(noticeError): dedupe anchor id

RebootJeff:
The `#promise-js` id attribute is used twice. This PR fixes the 2nd occurrence by renaming the id to a new, unique value so then it fixes the ability to directly link to the "Capturing custom attributes example" heading of the page.


homelessbirds:
@RebootJeff would you mind signing the CLA?

homelessbirds (approved):

PR #10120, opened 2022-11-01
Fix references to some UI elements

rkbodenner:
Refer to the new nav name for Explorer. Consistency in the links to further alerts docs, since they go to the same place.


homelessbirds (approved):

homelessbirds:
Thanks for catching this!

PR #10138, opened 2022-11-02
Add margins so entries in attribute-dictionary don't clip

ecchilds:
### Description

This Clipping could be seen in the Attribute Dictionary at widths less than 1240px:
image

This PR will fix this by changing the "position" field on Attribute headers to "initial" on screen widths less than 1240px.

### DIFF
--- a/src/pages/attribute-dictionary.js
+++ b/src/pages/attribute-dictionary.js
@@ -241,6 +241,9 @@ const EventDefinition = memo(
           data-swiftype-index="false"
           css={css`
             margin-bottom: 1rem;
+            @media (max-width: 1240px) {
+              margin-top: 4.1rem;
+            }
           `}
         >
           <span
--- a/src/pages/attribute-dictionary.js
+++ b/src/pages/attribute-dictionary.js
@@ -212,7 +212,7 @@ const EventDefinition = memo(
             }
 
             @media (max-width: 1240px) {
-              position: relative;
+              position: initial;
             }
           `}
         >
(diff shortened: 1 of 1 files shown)


src/pages/attribute-dictionary.js
```diff
@@ -241,6 +241,9 @@ const EventDefinition = memo(
           data-swiftype-index="false"
           css={css`
             margin-bottom: 1rem;
+            @media (max-width: 1240px) {
+              margin-top: 4.1rem;
+            }
```
clarkmcadoo, line 246:
```suggestion
```

I don't think we need this here, but rather need to update how the `h2` is being styled at the < 1240px width, [here](https://github.com/newrelic/docs-website/pull/10138/files#diff-acff009f7a9161a38ad8760ef5c8d75ae49e0974a4e76fcda0de178d14dcb693R214-R216) on line 215.

`position: relative` is updating this element to be relative to some other component which is creating the issue (I didn't track down which one).  Instead i just updated this to `position: initial` since I believe we're just trying to not make these headers sticky when we're at a sub 1240px width.

https://user-images.githubusercontent.com/26727669/199598756-ab615edf-0119-459f-8ac1-5739d2ae8ed7.mov

bradleycamacho:
Hi @ecchilds, thanks for the fix! Tagging our engineering team to approve and merge this in.

src/pages/attribute-dictionary.js
ecchilds: Ahh, I hadn't figured that out, thanks for the tip! New commit inbound.

clarkmcadoo (approved):

PR #10147, opened 2022-11-03
Clarify session/IP limits and move to better location

zuluecho9:
Liaison work involving: 
Move the sessions/IP-address limit to 'Factors affecting access' doc
Restructure that doc a little bit
Point to that limit from a couple other docs


paperclypse (approved):
Looks good to me!

PR #10193, opened 2022-11-07
Python Agent EOL Updates

TimPansino:
## Give us some context

* Updates Python Agent EOL page with new version releases, and removes old version.


akristen (approved):
Thanks for this update!

PR #10206, opened 2022-11-08
fix(navs): flattening some stuff

zuluecho9:
Attempt to make some nav stuff less nested, especially in the beginning areas of the navs


paperclypse (dismissed):
I left some CodeStream comments. Some minor quibbles. Otherwise, looks good! A definite improvement.

PR #10218, opened 2022-11-08
Update synth price and details

zuluecho9:
liaison work: update synth price/rules


rhetoric101 (dismissed):
@zuluecho9 Thanks for making these changes!

PR #10242, opened 2022-11-10
Updated "Logs API example" logtype value

csbertran:
Inconsistency of the description with the example of "Logs API".
Changed the value of logtype from "accesslog" to "nginx" as said in the description above. 
As you can see in this other part of the documentation, indeed the logtype value for NGINX parse rules is "nginx" and not "accesslog":  https://docs.newrelic.com/docs/logs/ui-data/built-log-parsing-rules#nginx


x8a:
Closing this PR as the change is introduced in PR #10241.

PR #10249, opened 2022-11-10
Fix(Go agent): corrected release note metadata

nr-swilloughby:
In the 3.20.0 release, the release note document had an incorrect metadata header. This corrects that error.


rhetoric101 (approved):
@nr-swilloughby thanks for making these updates to the release notes!

PR #10252, opened 2022-11-10
fix(node agent): remove hide_internals config option

jordigh:
This configuration option was removed in node release v9.6.0 and is no longer needed in the agent.


rhetoric101 (approved):
@jordigh thanks for removing this collapser!

PR #10263, opened 2022-11-10
fix(node agent): document new grpc.ignore_status_codes flag

jordigh:
It's very analogous to a similar flag for HTTP status codes, except nothing is ignored by default here.


rhetoric101 (approved):
@jordigh thanks for adding this section!

PR #10273, opened 2022-11-11
feat(dashboards): add new templ variables doc

zuluecho9:
NR-57876. 

Work done: 
New 'template variables' doc in dashboards category
Add to nav
Mention new doc in 'Manage dashboards' doc


src/content/docs/query-your-data/explore-query-data/dashboards/dashboard-template-variables.mdx
```diff
@@ -0,0 +1,214 @@
...
+
+And here's an example of the resulting widget, on the right, with the `country` dropdown to the left. 
+
+<img
+  title="Country template variable example"
+  alt="Country template variable example"
+  src={dashboardsCountryVariableExample}
+/>
+
+When you're done a) defining a template variable and b) adding a widget that references that variable, you can verify it's working as expected by choosing different options from the template variable bar and seeing if the widget changes based on your selection.
```
jeff-colucci, line 177: Do we need the a) and b) here? It seems to read better without them, unless we want to change them to 1) and 2) to more directly reference the preceding steps. My gut says it would read better without them though. Just a thought.

src/content/docs/query-your-data/explore-query-data/dashboards/dashboard-template-variables.mdx
```diff
@@ -0,0 +1,214 @@
...
+
+Here are some different types of template variable implementations. 
+
+<CollapserGroup>
+  <Collapser
+    id="multiple-variables"
+    title="Using multiple variables for one widget"
+  >
+
+You can define one template variable at a time, but you can use more than one template variable in a single widget.
```
jeff-colucci, line 202: This reads a little weirdly to me, I think I'm expecting it to be "but you can't" since it started with "you can". What do you think about replacing "but you can" with "and you can also"?

jeff-colucci (dismissed):
Just a commented on a pair of small things I noticed in my review. Approving for merge because I don't think they're major things, just considerations.

src/content/docs/query-your-data/explore-query-data/dashboards/dashboard-template-variables.mdx
zuluecho9: lol. yeah that's a bad one. whoops.

PR #10281, opened 2022-11-11
Echilds/caching single nav

ecchilds:
### Summary

This PR re-introduces the single nav sidebar that was previously reverted, but with these additional features:
- The nav is generated when nodes are being sourced, and is stored in Gatsby's GraphQL instance.
   - This is also done for the Style Guide and Agile Handbook navs, resulting in three nav nodes.
- The result of this generation is cached, so repeat builds will be faster and use less memory.
- The resolvers for the nav have thus been greatly simplified, simply returning the result of a single GraphQL query.
   - The results of these queries are also cached, so they only need to be performed once for each of the three navs.

### The stats

The previous version of the single nav ran at around 4.7 GB of heap usage during build, sometimes more. This version's heaviest build step is `onPostBuild`, which runs at around 3.8 GB of heap usage, peaking at 4.5 GB.

Screen Shot 2022-11-11 at 12 42 23 PM

### Still to be done

The memory usage on this, despite all the tricks, is still very high. @LizBaker and I have determined that the particular part of `onPostBuild` causing the biggest drain is that of the `gatsby-plugin-sitemap` plugin. This plugin is used by the contained `@newrelic/gatsby-theme-newrelic` module, which is imported in this repo. The best bet to reduce memory usage further will _likely_ be to fiddle with this plugin's settings, though it is still possible that something else is causing the memory spikes.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -230,7 +230,7 @@ module.exports = {
     {
       resolve: `gatsby-source-filesystem`,
       options: {
-        path: `./src/nav/`,
+        path: `./src/nav/generatedNav.yml`,
       },
     },
     {
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -236,6 +236,18 @@ module.exports = {
     {
       resolve: `gatsby-source-filesystem`,
       options: {
+        path: `./src/nav/style-guide.yml`,
+      },
+    },
+    {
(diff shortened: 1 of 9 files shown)


austin-schaefer:
Closing out all open 2022 PRs, feel free to re-open if there's active work here or this becomes active again!

PR #10292, opened 2022-11-13
.NET 7.0 updates

nr-ahemsath:
Update the .NET Core agent compatibility/support doc to include .NET 7.0, which GAed last week.


x8a (approved):
Thanks for adding this!

x8a (approved):

PR #10333, opened 2022-11-15
chore(fix): Replace MRI with CRuby

kaylareopelle:
## Give us some context

There are multiple Ruby language interpreters.
MRI stands for Matz's Ruby interpreter. Matz refers to the creator of Ruby, Yukihiro Matsumoto. Matz would prefer the community refer to this interpreter as CRuby. This is how it is named in the official Ruby language documentation.


akristen (approved):
Thanks!

PR #10338, opened 2022-11-15
Fix(Go Agent)correct pluralization of function names

nr-swilloughby:
Go agent release 3.20.1 corrects pluralization of some method function names. This updates the docs to match.


akristen (approved):
Looks good. Thanks!

PR #10372, opened 2022-11-17
feat(java branching): added ansible tile and shorter screenshot in intro

akristen:
- Ansible logo
- Is this image still too long?


src/install/config/java.yaml
```diff
@@ -72,6 +72,8 @@ appInfo:
       - value: 'aws'
         displayName: 'AWS Lambda'
         logo: 'awsLambda'
+      - value: 'ansible'
```
akristen, line 75: something is missing here*** AWS lambda immediately shortens and keeps the simple 2-step structure (a la VM). ask shawn about maneuvering this around

paperclypse: I just noticed this comment @akristen. Still need help with this?

akristen: No! We talked about it. I need to add value: ansible where the value: aws appears. :) Thanks for checking in!

paperclypse (approved):
Looks great!

PR #10340, opened 2022-11-15
Distributed tracing: Reverse planning sections and fix links (Publish on 11/18/2022)

rhetoric101:
I am not in the office, but could someone review this and get it published on 11/18/2022? We have already had some customers stumbling over these navigation links.

PR #10379, opened 2022-11-17
bug: correct infinte tracing location in NR1

brnhensley:
Infinite tracing was moved with the new UI and customers are getting lost. NR Edge is gone now, it's all done with _Infinite Tracing settings_, and that has moved to _All Capabilities_.

image


rhetoric101:
@brnhensley this seems to overlap with this PR: https://github.com/newrelic/docs-website/pull/10340/files. Could we close yours in favor of that one?

brnhensley:
Yes Rob, Do you know when that one will be merged? We've gotten two support tickets for this.

PR #10413, opened 2022-11-21
fix(node agent): properly document how to set up request parameters

jordigh:
At some point in the distant past (circa 2018) the way we capture request parameters changed, but our documentation just wasn't properly updated to reflect this. Let's finally fix this, since now, four years later, someone tried to make sense of it, failed, and told us about it.

Closes NEWRELIC-5477


homelessbirds (approved):
Looks good!

homelessbirds:
Thanks for this!

PR #10433, opened 2022-11-22
Remove data dictionary code

tabathadelane:
More work to see how we can reduce site build times. 

Local cold build time:
 - without data-dictionary: 20min
 - on develop: tbd


tabathadelane:
Closing this PR because removing this code did not significantly reduce build times.

PR #10447, opened 2022-11-23
Update configure-logs-context-go.mdx

niku:
As of go v1.19.3, logger doesn't have `Default` (struct), it has `Default()` (function).


iamemilio:
Thanks! Looks good to me.

zuluecho9 (approved):

PR #10454, opened 2022-11-23
Add Windows installation guide to Nginx integration (NR-68231)

kang-makes:
## Give us some context

A customer pointed us that we were missing the Windows installation procedure. This PR should fix the issue.


x8a (approved):
Thanks for adding this back!

PR #10465, opened 2022-11-23
chore(PHP troubleshooting): removed a file that is no longer relevant…

jeff-colucci:
… due to PHP no longer supporting MacOS as of Dec 2020

Info came in from the help documentation channel, verified that the content is removed and no longer referenced in the nav files.


paperclypse (approved):
Looks good!

PR #10499, opened 2022-11-29
chore: update list of modules

brnhensley:
The list of modules was missing a few listed in this plugin's github README, so I added them: https://github.com/newrelic/newrelic-node-apollo-server-plugin
\


rhetoric101 (approved):
@brnhensley thanks for adding these two modules!

PR #10532, opened 2022-11-30
fix(browser): rename translated files

bradleycamacho:
Fixes some translated files that were left un-renamed ;)


jeff-colucci (approved):
I can confirm that the files have been renamed so I will approve!

PR #10358, opened 2022-11-16
feat: remove all redirects with urls that have trailing slash, enable…

clarkmcadoo:
remove all redirects with urls that have trailing slash, enable dsg

PR #10534, opened 2022-11-30
Revert "feat: remove all redirects with urls that have trailing slash, enable…"

clarkmcadoo:
Reverts newrelic/docs-website#10358


tabathadelane (approved):
ABORT

PR #10554, opened 2022-12-01
Retain pathWithTrailingSlash and re-enable trailingSlash: always

LizBaker:


clarkmcadoo (approved):

roadlittledawn (approved):
- [X] Verified URL that's redirected correctly redirected on initial page load
- [X] Verified URL that's redirected redirects when clicking on internal link on page that has loaded already
- [X] Verified that there's always a trailing slash
- [X] Verified the above for jp pages as well
- verified behavior on non-MDX page like data dictionary

i noticed that anchor links in the On this page right rail have the hash after trailing slash like so
```
https://docswebsitedevelop-lizredirectfix73736.gatsbyjs.io/docs/infrastructure/infrastructure-monitoring/get-started/get-started-infrastructure-monitoring/#logs-context
```

on initial load i experienced they sometimes don't scroll to relevant section.

not a show stopper, but worth keeping eye on.

PR #10558, opened 2022-12-01
Add section on automated sources of tags

rkbodenner:
Describe the various ways tags can be automatically applied to entities. There is a bias toward service entities here, so any more detail on hosts or other entities might be nice. I don't think the doc has to be completely exhaustive, though. Would appreciate a review from @erabug @josemore @jlegoff 

Also make it a little more clear that tags go on entities, while attributes go on data.


akristen:
@rkbodenner Let us know when your review cycles are done and the docs team can merge this in!

rkbodenner:
No feedback from reviewers, let's merge when ready :)

akristen (approved):
Looks good! Going ahead to merge. :)

PR #10576, opened 2022-12-02
Add comment and email input to install feedback form

tabathadelane:
- remove the previous thumbs up/down system
- add comment and optional email boxes
- introduce some initial styling changes


src/components/InstallFeedback.js
```diff
@@ -4,222 +4,281 @@ import {
...
-      >
-        <h2
-          css={css`
-            color: ${feedbackStyles.headingTextColor};
-            margin-bottom: 0;
-          `}
-        >
-          {feedbackText}
-        </h2>
+              <div css={css``}>
```
LizBaker, line 239: got an empty css prop here
```suggestion
              <div>
```

LizBaker:
right now you can submit a comment with an empty text box and an email, is that expected behavior/do we want that?

tabathadelane:
> right now you can submit a comment with an empty text box and an email, is that expected behavior/do we want that?

oooh. i dont know how i missed that. ill look!

PR #10583, opened 2022-12-02
fix(apm): update key transaction pathing

bradleycamacho:
Corrects pathing for key transactions


rhetoric101 (approved):
@bradleycamacho these navigation changes look good to me! Thanks!

PR #10610, opened 2022-12-05
Install feedback recaptcha & feedback service request

tabathadelane:
add recaptcha and user-feedback submission and tessen submit event

this POST request does not contain a `rating` field which is currently required in the `user-feedback-service`. that will need to be changed before this goes live


src/components/InstallFeedback.js
```diff
@@ -229,8 +288,8 @@ const InstallFeedback = () => {
               </div>
               <div
                 css={css`
-                  margin-top: -1rem;
                   font-size: 0.75rem;
+                  margin-top: -2rem;
                   color: var(--secondary-text-color);
                   width: 50%;
                 `}
```
LizBaker, line 295: I know this is from the last PR really buuut i think it's worth removing this negative margin and moving this into a p tag that sits inside the div with the email. having the text in a p tag is better for accessibility and moving it to the email div lets us remove negative margin which can be a headache to deal with during refactors

tabathadelane: this was already a huge headache because of the `enter valid email` p tag that appears. i have no idea how these should be showing up/stacking in the ui. i didn't receive any design feedback on them. i'll try looking at it again. i agree that it should be a p tag.

LizBaker: as long as it's below the email validation i think it should basically look the same as it does now, just structurally it'll be different

LizBaker (approved):

PR #10637, opened 2022-12-07
docs: Add notification triggers docs to workflows

nzur-newrelic:

## Give us some context
We added new functionality to workflows called notification triggers.
To make it clearer for users I added an explanation about what is it and how it should be used.


jeff-colucci (approved):
Made some slight styling changes but other than that it looks good.

PR #10643, opened 2022-12-07
feat(APIs): added new using custom events section

jeff-colucci:
Awaiting feedback from other SMEs but I think what we have here covers the original ticket in an acceptable capacity


rhetoric101 (approved):
@jeff-colucci thanks for making these last-minute changes!

PR #10645, opened 2022-12-07
chore(pathing): Fix APM database path

austin-schaefer:
Fixes a couple stray paths.

@akristen : Can we do a check to see if this is a one-off or if there are more lurking old paths like this?


akristen:
I can do this thing

akristen:
Ok looks like there are a few more. I'll do a sweep for this error. To confirm, do all instances of > Explorer > need to change as well?

jeff-colucci (approved):
Looks good!

PR #10655, opened 2022-12-08
Add vale ini file and styles

LizBaker:
This adds the default microsoft styles as well as the folder/files needed to make adjustments to our accepted and rejected terms (caught by Vale.Spelling, Vale.Terms, and Vale.Avoid)

for review: 
please download the vale cli (brew install vale) and the vale vscode extension, then restart vscode and switch to this branch. you should be able to both run `vale src/content/docs/mlops/bring-your-own/mlops-byo.mdx` as well as see linting in the files themselves when you open them, and a list of issues in the problems section of vscode


clarkmcadoo (dismissed):

bradleycamacho:
Great job getting this implemented! Maybe this is all work for a future PR, but here are some thoughts:

1 - I can't get the `problems` tab to populate any errors, even if I save the file or make any changes. I only get content when I run the `vale` command.

2 - I think we'll have to remove the spelling rule or else VERY quickly start implementing our own spellings to it. It currently creates a ton of noise because it perceives `https` `<td>` `dataframe` and tons of other things as incorrect. One option is to keep it on and have writers slowly implement false-positives into the dictionary. I guess one person could also spend a day going through each spelling error and flagging the false-postives and implementing them in the dictionary haha

3 - Lots of spelling errors are coming from codeblocks and inline code. I know there's a way to get Vale to always ignore these, but that's where I hit a boulder when I was trying to implement this myself months ago. Godspeed 🖖🏼

clarkmcadoo (dismissed):

roadlittledawn (dismissed):
:shipit:

PR #10676, opened 2022-12-08
style(logs): Improve structure of "disable log"

austin-schaefer:
Partially addresses user feedback in NR-65567


jeff-colucci (approved):
Just saw a few small grammatical things but other than that looks good!

PR #10688, opened 2022-12-09
Update android crash reporting flag

kthai-nr:
Updated the default value for different Android agent crash reporting flags and wrote a description for `withCrashReportingEnabled()` to reflect current implementation.

### DIFF
--- a/src/content/docs/mobile-monitoring/new-relic-mobile-android/android-sdk-api/android-agent-configuration-feature-flags.mdx
+++ b/src/content/docs/mobile-monitoring/new-relic-mobile-android/android-sdk-api/android-agent-configuration-feature-flags.mdx
@@ -170,9 +170,43 @@ All settings, including the call to invoke the agent, are called in the `onCreat
 ## Crash and error reporting settings [#crash-error-reporting-settings]
 
 <CollapserGroup>
+
   <Collapser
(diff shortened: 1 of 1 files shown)


src/content/docs/mobile-monitoring/new-relic-mobile-android/android-sdk-api/android-agent-configuration-feature-flags.mdx
```diff
@@ -170,9 +170,43 @@ All settings, including the call to invoke the agent, are called in the `onCreat
 ## Crash and error reporting settings [#crash-error-reporting-settings]
 
 <CollapserGroup>
+
   <Collapser
     id="ff-withCrashReportingEnabled"
-    title={<>withCrashReportingEnabled<br/>FeatureFlag.CrashReportingEnabled</>}
```
cthomas-newrelic, line 175: @kthai-nr Careful here: `withCrashReportingEnabled` is not the same as using the feature flag. To disabled all crash reporting, the feature flag should be used. To enable "_just in time_" crash reporting, use `.withCrashReportingEnabled(true)` when starting the agent. We don't advise setting JIT but have left the option in case it comes up.

akristen (approved):
Looks good. Thanks for this!

Issue #10652, opened 2022-12-07
NRQL math page needs updates


thezackm:
context: [NRQL math](https://docs.newrelic.com/docs/query-your-data/nrql-new-relic-query-language/get-started/nrql-math-using-select/) page

 * In the `clamp` and `rounding` sections, we're showing an old Seldon chart image
 * Throughout this document; there are multiple functions that are not documented on the more verbose [NRQL clauses and functions](https://docs.newrelic.com/docs/query-your-data/nrql-new-relic-query-language/get-started/nrql-syntax-clauses-functions/) page. I think it would make sense to ensure the functions are all documented in a single page and pages like this are treated more like "example / best-practice" type of documentation. It's confusing to try and piece together a NRQL query using functions from multiple pages.

jeff-colucci:
Hi @thezackm, thanks for your feedback! I agree about the disjointed state of some of the docs. I'm currently revamping the NRQL docs and I plan on fixing both the syntax and math pages once I get to that phase. As far as the screenshots, I can make the changes to update those easily. If you can provide screenshots that you think should be there I can get it done and merged in the same day, otherwise I will put it on my list of things to do on a future sprint and get it done there.

thezackm:
@jeff-colucci - thanks for the info; here's a few shots using the current UI in both light and dark; hope that helps!

```sql
FROM SystemSample SELECT
average(cpuPercent) AS 'raw',
round(average(cpuPercent), .01) AS 'rounded',
floor(average(cpuPercent)) AS 'floor',
ceiling(average(cpuPercent)) AS 'ceiling'
TIMESERIES
```

![round-floor-ceil-dark](https://user-images.githubusercontent.com/30831476/206735994-0fe15412-0360-473f-b097-1e8cb1db9424.jpg)
![round-floor-ceil-light](https://user-images.githubusercontent.com/30831476/206736007-c86b6bf0-4ca6-4a8f-8891-3fb32e88083e.jpg)


```sql
FROM SystemSample SELECT
average(cpuPercent) AS 'raw',
clamp_min(clamp_max(average(cpuPercent), 98), 96) AS 'clamped'
TIMESERIES
```

![clamp-light](https://user-images.githubusercontent.com/30831476/206736012-390512e6-d4a1-4f19-9594-f59226151d62.jpg)
![clamp-dark](https://user-images.githubusercontent.com/30831476/206736018-5438bcf7-772a-4e2c-a77b-38c7683f5446.jpg)

[...]

PR #10690, opened 2022-12-09
fix(NRQL): fixed screenshots

jeff-colucci:
Marking work done from [this ticket](https://github.com/newrelic/docs-website/issues/10652)


thezackm (changes requested):
@jeff-colucci - thanks so much! one quick note: 

above the clamp image there's pre-existing text:
```
The following chart shows the result of clamping both min and max to keep the value in the range 70-90.
```

The new query is this:
```sql
FROM SystemSample SELECT
average(cpuPercent) AS 'raw',
clamp_min(clamp_max(average(cpuPercent), 98), 96) AS 'clamped'
TIMESERIES
```

so the text should now read: 
```
The following chart shows the result of clamping both min and max to keep the value in the range 96-98.
```

---

I'd argue that we should also add a sample query there, but with the upcoming changes you've already planned it's low priority.

jeff-colucci:
@thezackm Thanks for the suggestion and good eye! Is it as straightforward as adding the query you put in your comment to the documentation? I can make the change to say "the following query and chart" instead of just "the following chart" and then list the query as well.

thezackm:
@jeff-colucci - yes I think so; just a code block for the NRQL - it's pretty generic and is a good example of how to nest the clamp functions together

jeff-colucci:
@thezackm Very good, I've updated the doc to have the example in it and tweaked the wording around it as well. If you could take a look at it and let me know that it came out correctly I think I can get a TW review and merge it in today. As an aside, if you have any other ideas for examples of NRQL throughout the documentation, either new examples or refining some older ones we have in place, please reach out at any time and I will get them changed. NRQL examples are something I'm kind of constantly looking at so any help there would be very appreciated.

thezackm (approved):
🚀

jeff-colucci:
Thanks @rhetoric101 and good call, I'll make a note and probably have a ticket to do that next sprint

PR #10695, opened 2022-12-09
fix(synthetics-usage): Clarify usage timestamps

bpeckNR:
Clarifies which timestamp is used to calculate synthetics usage details.


akristen (dismissed):
Thanks for adding this Brian!

akristen (approved):

PR #10708, opened 2022-12-12
Resolve sharp packages to single version

tabathadelane:
Explicitly resolving `sharp` and `gatsby-plugin-sharp` to dedupe them in the `yarn.lock`

### DIFF
--- a/package.json
+++ b/package.json
@@ -151,7 +151,12 @@
     "remark-mdx": "2.0.0-next.8",
     "ini": "^1.3.6",
     "yargs": "^13.3.2",
-    "yargs-parser": "^18.1.1"
+    "yargs-parser": "^18.1.1",
+    "gatsby-plugin-sharp": "4.19.0",
+    "sharp": "0.30.3"
+
+
+
   },
   "keywords": [
     "gatsby"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -8747,31 +8747,7 @@ gatsby-plugin-robots-txt@^1.6.8:
...
-    potrace "^2.1.8"
-    probe-image-size "^6.0.0"
-    progress "^2.0.3"
-    semver "^7.3.5"
-    sharp "^0.29.3"
-    svgo "1.3.2"
-    uuid "3.4.0"
-
-gatsby-plugin-sharp@~4.19.0:
+gatsby-plugin-sharp@4.19.0, gatsby-plugin-sharp@4.4.0, gatsby-plugin-sharp@~4.19.0:
```
clarkmcadoo, line 8750: i think these are just labels in the lock file, but this looks weird, i imagine it just needs to be `gatsby-plugin-sharp@4.19.0:`

package.json
```diff
@@ -151,7 +151,12 @@
     "remark-mdx": "2.0.0-next.8",
     "ini": "^1.3.6",
     "yargs": "^13.3.2",
-    "yargs-parser": "^18.1.1"
+    "yargs-parser": "^18.1.1",
+    "gatsby-plugin-sharp": "4.19.0",
+    "sharp": "0.30.3"
+
```
clarkmcadoo, line 157: just some extra lines we don't need

PR #10727, opened 2022-12-14
fix: render icon suffix for external links

jb-newrelic:
This PR fixes markdown links pointed at external domains which have not been rendering with the icon suffix.

## Give us some context

* What problems does this PR solve?
This PR fixes markdown links pointed at external domains which have not been rendering with the icon suffix.

* Add any context that will help us review your changes such as testing notes,
  links to related docs, screenshots, etc.
  
Screen Shot 2022-12-14 at 12 08 10 PM

  
* If your issue relates to an existing GitHub issue, please link to it.
https://issues.newrelic.com/browse/NR-35724

PR #10729, opened 2022-12-14
Revert "fix: render icon suffix for external links"

jeff-colucci:
Reverts newrelic/docs-website#10727


jeff-colucci:
Ok cool, if that one gets done before the 12:00pm merge I can close this one without any issue. I can also delay the 12:00 publish a bit if necessary to give you a bit more time. So sorry for all the confusion and the inconvenience.

PR #10741, opened 2022-12-14
chore: fix code examples

brnhensley:
The code examples had typos and the `<var>`s were showing as plain text.
image

The `$` cmd prompt is added when you make format the code as `shell` or `bash`, so it was there twice:
image


jeff-colucci (approved):
Looks good, good eye!

PR #10746, opened 2022-12-14
updates for Ruby agent v8.14.0 (9am 12/15 publish)

fallwith:
content related to New Relic Ruby agent v8.14.0, which offers a new configuration parameter


jeff-colucci (approved):
Had to make one change to the added CollapserGroup entry but the docs should build correctly now. Everything else looks good!

PR #10748, opened 2022-12-15
Add back changes

LizBaker:
We had some weird merge conflict issues in dev/main and completely dropped changes on the file, this adds em back


jeff-colucci (approved):
I'll be attempting to do this one as it's own release after the 9am release, fingers crossed it goes through ok.

PR #10768, opened 2022-12-15
UI writing style - update - rps

mmfred:
We didn't have an entry for requests per second in the style guide. This updates that.


jeff-colucci (approved):
Looks good!

jeff-colucci:
Bypassing the CLA check as this is coming from an internal contributor

PR #10783, opened 2022-12-16
Try lessBabel option in gatsby-plugin-mdx

LizBaker:
This was suggested as possibly helpful for our issue with builds not failing on mdx errors when they should


roadlittledawn (approved):
:shipit:

PR #10785, opened 2022-12-16
chore: Document broken config

kaylareopelle:
## Give us some context

Ruby agent 8.14.0 introduced `defer_rails_initialization` to toggle some code for a bugfix.
This fix is run during the initialization process, before configs are applied so using the config has no value at this time.

https://github.com/newrelic/newrelic-ruby-agent/issues/662 
https://github.com/newrelic/newrelic-ruby-agent/pull/1702

cc @homelessbirds


homelessbirds:
Looks good :)

PR #10591, opened 2022-12-02
fix(PHP agent): Clarify newrelic_record_datastore_segment string rest…

mfulb:
This change make it clear that a special character ('/') should not be used with certain parameters for the `newrelic_record_datastore_segment()` PHP agent API call.

PR #10793, opened 2022-12-19
fix(PHP agent): Clarify newrelic_record_datastore_segment string restrictions

mfulb:
This change make it clear that a special character ('/') should not be used with certain parameters for the `newrelic_record_datastore_segment()` PHP agent API call.


mfulb:
This is [PR 10591](https://github.com/newrelic/docs-website/pull/10591) redone as that one seemed to be failing in odd ways so we thought we'd try recreating the PR.

jeff-colucci (approved):
I've checked the anchor links in the preview build so we don't have a repeat of the last problem and the new content looks good!

PR #10798, opened 2022-12-19
docs: fix broken header

jonathonlacher:
## Give us some context

Fixes broken header.

image


akristen (approved):
Thanks for catching this and giving it an update!

PR #10810, opened 2022-12-19
Conditionally add install steps

LizBaker:
this adds a boolean to overrides called addStep. setting it to true will cause the step to not render unless the options defined in the override are selected


sunnyzanchi (approved):
👍 💯

PR #10825, opened 2022-12-20
chore: minor code cleanup, added missing config option for fedramp

brnhensley:
I fixed a bunch of small things like:
- indentation being inconsistent (3 spaces for example). 
- spaces between `key = "something"` in XML which isn't usually done
- added `ini` as a code type for env vars. It looks nice where I've seen it elsewhere, so I figured I'd add it here while I was at it.


brnhensley:
Let me know what the issue is with this PR and I'll fix it. not sure what the vale linter is asking for

a-sassman (approved):

PR #10845, opened 2022-12-22
fix: CAT is deprecated

brnhensley:
I moved the section for CAT below DT and 8T, we want customers to use those instead but the deprecated option was presented first in the doc. I also added a note about that deprecation, and updated the default value for the CAT setting to `false`.

I also added an 8T link in the navigation links.

## Give us some context


rhetoric101 (approved):
@brnhensley thanks for moving cross application tracing lower in the order and clarifying the EOL.

PR #10858, opened 2022-12-28
fix(style guide): update var tag guidance

zuluecho9:
This was work to: 

* Update style guide with better guidance on customer-specific value placeholders, which previously we said to use `<var>` tags for, but now we have to update because those tags disable code syntax highlighting. 
* I left in room to use `<var>` tags if we really want to and are okay with there not being syntax highlighting, but main guidance is to avoid var tags almost entirely.  

Questions: 
* Is saying to include 'INSERT' in placeholder text too much? Is it enough to just say 'YOUR_X...'? Or would it be more helpful to say 'INSERT_YOUR_X...' ? On one hand, getting rid of `<var>` tag makes it so that the placeholders are harder to recognize, so having 'INSERT' would help make things more clear in that regard. On other hand, it would take up more room.

### DIFF
--- a/src/content/docs/style-guide/formatting/code-examples.mdx
+++ b/src/content/docs/style-guide/formatting/code-examples.mdx
@@ -173,30 +173,24 @@ To format one or more lines of code, insert three backticks ( ```` ``` ```` ) ab
   </Collapser>
 </CollapserGroup>
 
-## Customer-specific values [#var-tags]
+## Customer-specific values [#value-placeholders]
 
-Traditionally, we used `<var>` tags to highlight customer-specific values in code snippets. For example: we'd use `<var>YOUR_ACCOUNT_ID</var>` to draw a user's attention to a place they'll need to input their account ID.
+Traditionally, we've used `<var>` tags for customer-specific value placeholders in code snippets. For example: we'd use `<var>YOUR_ACCOUNT_ID</var>` to draw a user's attention to a place they'd need to input their account ID.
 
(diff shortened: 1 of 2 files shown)


roadlittledawn (dismissed):
I like it

src/content/docs/style-guide/formatting/code-examples.mdx
```diff
@@ -167,85 +167,52 @@ To format one or more lines of code, insert three backticks ( ```` ``` ```` ) ab
...
-  * [Instrumentation procedure](/docs/serverless-function-monitoring/aws-lambda-monitoring/get-started/enable-new-relic-monitoring-aws-lambda#java) (at bottom of the Java section): the var tag wouldn't be necessary because it's an example, not part of a procedure, and the main goal is seeing the general structure. Also, because it's an example of app code, the concept of user-specific values doesn't have much meaning, because the entire code will vary dependent on how the customer has written their code. (If anything, this would be a potential for using [`<var>` format](#highlighting) to emphasize the New Relic functions.)
-* Response/output code: If the code is meant to show an expected return, and is not related to a procedure, then var tags shouldn't be used. Here's [a doc section](/docs/apis/synthetics-rest-api/monitor-examples/manage-synthetics-monitors-rest-api#get-all-monitors) (the returned JSON) where var tags are not needed.
+* Use `INSERT` to make it more clear what customers are to do. 
+* Address the reader directly. For example, using `INSERT_YOUR_ACCOUNT_ID` and not `INSERT_ACCOUNT_ID`).
```
jeff-colucci, line 187: Should the ) be at the end of line 187?

jeff-colucci (dismissed):

PR #10859, opened 2022-12-28
fix(infra-agent): support dates

josemore:
## Give us some context

* Suggesting more details on the EOL policy for the infra agent (feel free to rewrite). The policy is: releases after July-2022 (starting with 1.28.0) are supported for one year, older releases (before July 2022) have a two-year support cycle.


akristen (approved):
Looks good!

PR #10860, opened 2022-12-28
fix(infra-agent): version param deprecated

josemore:
## Give us some context

* The version param has been deprecated long time in the ansible repo for the infra-agent. We are removing this from doc to avoid confusion until there's an alternative param.


akristen:
Adjusted the table, but otherwise looks good! Thanks for updating this.

akristen (approved):
Thanks!

PR #10864, opened 2022-12-28
Update curl and printf commands

keegoid-nr:
- curl should use -fsSL options to retry on failure, silent, show errors, and redirect if needed
- Ubuntu 22.04 needs to use `gpg` instead of `apt-key` to store the key in a different location
- `echo` is more user friendly than `printf` since it adds a line break


akristen (approved):
Thanks for these updates!

keegoid-nr:
Tested new cURL command with `gpg` in Ubuntu: 16.04, 18.04, 20.04, 20.10, 21.04, and 22.04

It works in all versions.

PR #10878, opened 2022-12-30
fix typo

kb-newrelic:
Even though english is not my first language, I am pretty sure this is a typo.


bradleycamacho (approved):
Haha great catch! Thanks for the edit :)

PR #10886, opened 2023-01-03
Updates to Debian

keegoid-nr:
- remove Debian 8 and 9 because they are EOL
- update curl command to use gpg due to deprecated apt-key


nbaenam:
Hi @keegoid-nr! Happy New Year! Thank you for the update. Please, let me know when this modification must be published. Thanks!

nbaenam (approved):

PR #10890, opened 2023-01-03
chore(Docs website): Replacing var with nothing

nbaenam:
Removing <var> and </var> from the entire docs-website except from these files: */i18n/*,*.json,*i18n*


austin-schaefer (approved):
Spot-checked about 50 files and all look like good substitutions to me!

austin-schaefer (dismissed):

x8a (approved):

x8a (approved):

PR #10900, opened 2023-01-04
Updating style guide with placeholder/highlight-code guidance

zuluecho9:
Work done: 
* Added info about highlighting multiple lines of code
* Updated guidance on var tags (don't use) and updated style guide on placeholders (removed 'INSERT', updated with better guidance in general


src/content/docs/style-guide/formatting/code-examples.mdx
```diff
@@ -173,36 +173,43 @@ To format one or more lines of code, insert three backticks ( ```` ``` ```` ) ab
...
+  ```js lineHighlight=1,3-5
+  const coolThing = {
+    key: 'HI',
+    tada: 'hello'
+  }
+  // this is an important object
+  ```
 
-That said, `<var>` tags can still be used if you think it's important to highlight something in a code snippet and you're okay losing syntax highlighting. For tips on using var tags, see [Var tags](#var-tags).
+The first line reads: `js lineHighlight=1,3-5`, which tells it to use JavaScript syntax highlighting, and tells it to highlight the first line, and the third through fifth lines. Highlighting lines requires the simultaneous use of syntax highlighting: you must indicate a language if you want to highlight lines.
```
rhetoric101, line 188: @zuluecho9 readers can't see the code fence because the parser simple processes this and displays the output. How about including a screenshot of the entire snippet that shows the enclosing backticks and code fence?

src/content/docs/style-guide/formatting/code-examples.mdx
```diff
@@ -173,36 +173,43 @@ To format one or more lines of code, insert three backticks ( ```` ``` ```` ) ab
...
-* If you explicitly say that a customer must replace some values, you can break some of the usual rules. For example, if you preface a code snippet with an instruction that they'll have to replace some of the values, you can use more brief placeholder text, like `YOUR_ACCOUNT_ID` or `YOUR_USER_KEY`. 
-* **Don't** use placeholders for either of these situations: 
-  * Example code or files: if you're showing an example config file or another example that's simply meant to give a rough sense of what something will look like, and there is no direct procedure involved, you don't have to worry about placeholders and can use example values. 
-  * Results of API calls: if you're including the response of a call, you also don't have to use placeholders. 
+* Address the reader directly using `YOUR`. For example, use `YOUR_ACCOUNT_ID` and not `ACCOUNT_ID`.
+* Use `NR` when you think it may not be clear if a component is something of ours or a third party's. For example, when writing about connecting AWS to New Relic, you may want to specific that an account or API key is New Relic's. For example: `YOUR_NR_ACCOUNT_ID`.
```
rhetoric101, line 198: @zuluecho9 I think there's a small typo and that it should be, "...you may want to specify...." Also, I've not see any examples where we're too verbose with placeholders. We're doing our readers a favor by not using abbreviations.

src/content/docs/style-guide/formatting/code-examples.mdx
```diff
@@ -173,36 +173,43 @@ To format one or more lines of code, insert three backticks ( ```` ``` ```` ) ab
...
+* Don't combine them with other punctuation to indicate variables (such as wrapping the text in angle brackets or curly braces).
```
rhetoric101, line 199: @zuluecho9 this may be too subtle to communicate here, but the main convention we're bucking here is the use of angle brackets <like this>. We have a bunch of examples in our code that are frankly confusing. For example, we present the syntax for a bash variable `${APPID}` and then proceed to tell users to replace the whole variable with the literal value for the application ID. Instead, we should tell users they can do either approach: replace the entire variable or simple export APPID in the shell.

src/content/docs/style-guide/formatting/code-examples.mdx
```diff
@@ -173,36 +173,43 @@ To format one or more lines of code, insert three backticks ( ```` ``` ```` ) ab
   </Collapser>
 </CollapserGroup>
 
-## Customer-specific values [#value-placeholders]
+## Highlighting lines of code [#highlight-code]
```
rhetoric101, line 176: @zuluecho9 somewhere in this section, you might insert a callout that explains that we can't highlight anything less that full lines--no highlighting of individual words or groups or words.

rhetoric101 (approved):
@zuluecho9 I inserted a few comments.

PR #10911, opened 2023-01-04
Adding and updating templates

dneth:
This adds templates for PagerDuty, OpsGenie, VictorOps and xMatters.  It also updates and corrects several of the existing mapping and webhook template entries.  Finally, it updates the table to be more consumable since many of the notes weren't helpful or descriptive and that column took room away from the expressions.


homelessbirds (approved):
Looks good!

PR #10916, opened 2023-01-05
fix: update option from bar to area chart for units

thezackm:
small fix to update the scope of the "units customization" to include area charts, not bar charts (which are not time series and don't have this option).

### DIFF
--- a/src/content/docs/query-your-data/explore-query-data/use-charts/use-your-charts.mdx
+++ b/src/content/docs/query-your-data/explore-query-data/use-charts/use-your-charts.mdx
@@ -172,7 +172,7 @@ Read the sections below to learn more about the customization features. You can 
     id="dash-units"
     title="Units customization"
 >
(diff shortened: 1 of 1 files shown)


zuluecho9:
thanks @thezackm

zuluecho9 (approved):

miguelsancheznewrelic:
@thezackm , thanks!
I would also correct "and stacked charts", which they mean "stacked bar charts".
At the end of the day, our area chart and stacked bar charts are both stacked.

miguelsancheznewrelic:
> @thezackm , thanks! I would also correct "and stacked charts", which they mean "stacked bar charts". At the end of the day, our area chart and stacked bar charts are both stacked.

cc @zuluecho9

thezackm:
thanks @miguelsancheznewrelic - added PR #10950 to address this

cc: @zuluecho9

PR #10917, opened 2023-01-05
OM section: move out web perf and page load docs into Best Practices

zuluecho9:


akristen (approved):
Looks good! Nav was structured correctly in the preview. :)

PR #10921, opened 2023-01-05
Update MacOS infra agent filepath

amandaguthrie:
The MacOS configuration file location differs by chipset. On a Mac with a Silicon processor, /usr/local/etc/.. doesn't exist.

Added chipset clarification to label of current location and the location for Macs with Silicon processors.


zuluecho9:
thank you @amandaguthrie !

PR #10926, opened 2023-01-05
OM section: improve intro section

zuluecho9:
Work done: 

* Added mention of implementation guide, clarifying that beginners should go there
* Added introductory context for each of the four value driver sections
* Took out link to style guide 'OM guide template', as I'm not sure we want to encourage customers to do that (and don't think they ever will do it) and replaced that with note about giving ideas/comments using feedback button


jeff-colucci (approved):
Looks great!

PR #10962, opened 2023-01-09
switch from `mdxAST` to `tableOfContents`

sunnyzanchi:
this changes the GraphQL query for the docPage template to use the `tableOfContents` field instead of `mdxAST`. `tableOfContents` is a less expensive field to access since it doesn't require the AST to be generated for every MDX page, but it does take more string parsing to get usable data

### DIFF
--- a/src/templates/docPage.js
+++ b/src/templates/docPage.js
@@ -2,6 +2,7 @@ import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/react';
 import { graphql } from 'gatsby';
+import { takeWhile } from 'lodash';
 import PageTitle from '../components/PageTitle';
 import MDXContainer from '../components/MDXContainer';
 import {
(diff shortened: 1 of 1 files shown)


a-sassman (approved):

src/templates/docPage.js
```diff
@@ -14,14 +15,30 @@ import {
 import MachineTranslationCallout from '../components/MachineTranslationCallout';
 import SEO from '../components/SEO';
 import GithubSlugger from 'github-slugger';
-import { parseHeading } from '../../plugins/gatsby-remark-custom-heading-ids/utils/heading';
```
clarkmcadoo, line 17: is this getting used anywhere else or can we delete it from the repo?

sunnyzanchi: looks like it's still used in [TableOfContents.js](https://github.com/newrelic/docs-website/blob/develop/src/components/TableOfContents.js#L9) and [serialize-mdx.js](https://github.com/newrelic/docs-website/blob/develop/scripts/actions/serialize-mdx.js#L12)

PR #10972, opened 2023-01-10
Service maps now supported for OTel services

rkbodenner:
Service map is now in production for OTel services, so we have parity now with APM agents.


akristen (approved):
Lookin' good!

PR #10979, opened 2023-01-11
fix(alerts): removed two outdated docs and added relevant information…

homelessbirds:
Removing two outdated docs and then adding information from deleted docs to the alerts and applied intelligence overview page.


jeff-colucci (approved):
Looks awesome!

nbaenam (approved):

PR #10992, opened 2023-01-11
browser.nav updates

ghost:
Task: to reorganize the browser.nav (table of contents) so the doc locations make sense and the most popular pages (based on Google Analytics) are easy to find.


src/nav/browser.yml
```diff
@@ -1,66 +1,116 @@
 title: Browser monitoring
 path: /docs/browser
 pages:
-  - title: Introduction to browser monitoring
+  - title: Intro to browser monitoring
```
a-sassman, line 4: Thoughts on "intro" vs. "introduction" vs. "overview"?

src/nav/browser.yml
```diff
@@ -1,66 +1,98 @@
...
     path: /docs/browser/browser-monitoring/getting-started/introduction-browser-monitoring
-  - title: Compatibility and requirements
-    path: /docs/browser/new-relic-browser/getting-started/compatibility-requirements-browser-monitoring
-  - title: Browser agent EOL policy
-    path: /docs/browser/browser-monitoring/getting-started/browser-agent-eol-policy
-  - title: Browser agent release notes
-    path: /docs/browser/new-relic-browser/getting-started/browser-agent-release-notes
-  - title: Guides
+  - title: Single page app (SPA) monitoring
```
a-sassman, line 6: @paperclypse It's not within scope, but I'm planning on merging all these SPA docs into the primary browser docs. For now though, I think they're going to be weirdly out of place.

src/nav/browser.yml
```diff
@@ -1,66 +1,98 @@
...
+        path: /docs/browser/new-relic-browser/browser-pro-features/spa-monitoring
+      - title: SPA data collection
+        path: /docs/browser/single-page-app-monitoring/use-spa-data/spa-data-collection
+      - title: Query SPA data
+        path: /docs/query-your-data/nrql-new-relic-query-language/nrql-query-tutorials/browserspa-nrql-query-examples
+      - title: SPA APIs
+        path: /docs/browser/new-relic-browser/browser-apis/using-browser-apis#spa-api
+  - title: Best practices guide
+    path: /docs/browser/browser-monitoring/guides/browser-monitoring-best-practices-guide
+  - title: Page load timing
```
a-sassman, line 26: @paperclypse I feel compelled to move these docs higher up because the "Page load timing process" page is a popular page. But I don't love how clunky the whole page tree is, especially being above the installation docs.

src/nav/browser.yml
```diff
@@ -1,66 +1,98 @@
...
         path: /docs/browser/browser-monitoring/installation/install-browser-monitoring-agent
       - title: Update the browser agent
         path: /docs/browser/new-relic-browser/installation/update-browser-agent
-      - title: Disable browser monitoring
-        path: /docs/browser/new-relic-browser/installation/disable-browser-monitoring
-      - title: Delete apps from New Relic
-        path: /docs/browser/new-relic-browser/installation/delete-apps-new-relic-browser
+      - title: Compatibility and requirements
+        path: /docs/browser/new-relic-browser/getting-started/compatibility-requirements-browser-monitoring
   - title: Configuration
```
a-sassman, line 50: @paperclypse Thoughts on this title? I tried other variations, and they were too long or didn't match well with the other tree titles :(

src/nav/browser.yml
```diff
@@ -1,66 +1,98 @@
...
-      - title: Alerts in browser
-        path: /docs/browser/new-relic-browser/configuration/view-browser-apps-alert-information
-  - title: "Single page app (SPA) monitoring"
+      - title: Disable browser monitoring
+        path: /docs/browser/new-relic-browser/installation/disable-browser-monitoring
+      - title: Rename browser apps
+        path: /docs/browser/new-relic-browser/configuration/rename-browser-apps
+      - title: Delete apps from New Relic
+        path: /docs/browser/new-relic-browser/installation/delete-apps-new-relic-browser
+  - title: Browser monitoring UI
```
a-sassman, line 72: @paperclypse One idea I have for these docs is to rename the parent title to "Browser monitoring features" or "Monitor browser performance" and transition the child pages from UI overview docs into topic-based articles for each browser data type. For example, "Ajax page" would be "Ajax requests", "JS Errors page" would be "JavaScript errors", etc., and the page would cover how to monitor that aspect of a browser app, the business impact of poor impact, how to interpret Ajax data in the UI, any customizations/troubleshooting they need to know, etc. From an SEO side, it could be impactful when someone searches "monitor javascript errors" and they come across our doc that covers all things JS error monitoring. But this would be larger work. In the meantime, what do you think the parent title be?

src/nav/browser.yml
paperclypse: @a-sassman "Intro" is definitely better for the nav, simply because "Introduction" is so many characters. In that way, too, "Intro" has an advantage over "Overview". I'd probably keep it as "Introduction" in the actual, header title.

I do feel like there's a subtle difference between an introduction and an overview. I don't think I've thought too long about the difference. It's more of an I-know-it-when-I-see-it-thing. :) I guess I think of an overview as an attempt to wholly describe what a thing can do in as brief a space as possible, while an intro doesn't need to have that kind of comprehensiveness.

This doc has qualities of both an intro and an overview, but it feels a bit more like an intro...

src/nav/browser.yml
paperclypse: @a-sassman Works for me! I think it's fine in the short term.

src/nav/browser.yml
paperclypse: @a-sassman Is there a way to change the nav title "Single page app (SPA) monitoring" to something more action oriented? "Monitor SPA pages" or something? I don't know how SPA is typically referred to...

src/nav/browser.yml
paperclypse: @a-sassman I think it's fine. It makes it pretty clear what the doc is all about.

src/nav/browser.yml
paperclypse: @a-sassman I do prefer more active titles, like "Monitor browser performance" or even "Monitor your browser page performance". I'd look for opportunities to do that wherever you can.

I think your impulse to reorganize these docs into more task-related categories is a positive one and I wholeheartedly approve!

One thing to keep in mind is that where a doc is in the nav doesn't affect SEO at all. In a search result, the full doc title is returned and not the nav title.

paperclypse (dismissed):
A definite improvement over the current nav. I think the only thing that stands out as a bit odd is the title and location of "Single page app (SPA) monitoring".

src/nav/browser.yml
a-sassman: I'm going to change it to "Monitor single page apps (SPA)". Good suggestion!

paperclypse (approved):
Looks good to me!

PR #11081, opened 2023-01-18
chore: add syntax highlighting to terminal cmds

brnhensley:
No functional changes, just add language identifiers for syntax highlighting the terminal commands

### DIFF
--- a/src/content/docs/infrastructure/install-infrastructure-agent/linux-installation/tarball-manual-install-infrastructure-agent-linux.mdx
+++ b/src/content/docs/infrastructure/install-infrastructure-agent/linux-installation/tarball-manual-install-infrastructure-agent-linux.mdx
@@ -56,6 +56,7 @@ Before installation, check the [compatibility and requirements](/docs/infrastruc
       </td>
     </tr>
   </tbody>
+</table>
 
 To install the agent:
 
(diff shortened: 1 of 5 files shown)


brnhensley:
Looks like the linter is unhappy about the use of 'cannot' and 'will not' in part of the files I didn't change.

akristen:
> Looks like the linter is unhappy about the use of 'cannot' and 'will not' in part of the files I didn't change.

Yes, that shouldn't be an issue. I closed/reopened the PR to see if the final check went through... Not sure what would cause the build to fail, but I've reached out to Eng just in case too.

brnhensley:
Oh interesting. It built fine before my last commit, I made a change based on the preview build. Hopefully it was just a fluke.

src/content/docs/infrastructure/install-infrastructure-agent/linux-installation/tarball-manual-install-infrastructure-agent-linux.mdx
```diff
@@ -56,7 +56,6 @@ Before installation, check the [compatibility and requirements](/docs/infrastruc
       </td>
     </tr>
   </tbody>
-</table>
```
roadlittledawn, line 59: looks like this may be causing the build failure

brnhensley: good catch!

akristen (approved):
Looks good! Thanks for fixing that `</table>` tag

PR #11090, opened 2023-01-18
fix(PAPI): clarify limitations for adding accounts

zuluecho9:
Work done: clarify that for new user model, they can't use PAPI account-creation but must use nerdgraph


jeff-colucci (approved):
Looks good, there are a few contraction points that Vale has picked up but other than that I don't see anything to change.

PR #11115, opened 2023-01-19
Update supported version for .NET MySQL compatibility

JcolemanNR:
Update supported version for .NET MySQL compatibility


bradleycamacho (approved):

bradleycamacho:
Hi @JcolemanNR, thanks for the edit! Looks good, just want to confirm we want to be replacing the old version instead of adding the new one alongside it as a verified compatible version.

PR #11123, opened 2023-01-19
Add step components

LizBaker:
This pr
* updates the theme giving access to new step components
* adds resolvers for serializing and deserializing (both tested and working as expected)
* adds an isTutorial frontmatter key
* adds a tutorial meta tag on pages with isTutorial: true
* filters pages with isTutorial frontmatter from json generation

### DIFF
--- a/src/templates/docPage.js
+++ b/src/templates/docPage.js
@@ -64,23 +64,19 @@ const BasicDoc = ({ data, location, pageContext }) => {
   const {
     title,
     metaDescription,
-    type,
+    tags,
     translationType,
     dataSource,
     isTutorial,
   } = frontmatter;
 
-  let { tags } = frontmatter;
+  let { type } = frontmatter;
 
   if (typeof window !== 'undefined' && typeof newrelic === 'object') {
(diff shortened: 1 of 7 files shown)


clarkmcadoo (approved):

src/templates/docPage.js
```diff
@@ -65,15 +65,23 @@ const BasicDoc = ({ data, location, pageContext }) => {
     title,
     metaDescription,
     type,
-    tags,
     translationType,
     dataSource,
+    isTutorial,
   } = frontmatter;
 
+  let { tags } = frontmatter;
```
tabathadelane, line 73: How is this component setting the metatag document type to tell DLS that this is a tutorial doc? Does it get the same info from tags?

https://github.com/newrelic/docs-website/blob/develop/src/components/SEO.js#L70

LizBaker: whoops, this is just adding a tag, i'll switch it to be a type 🙊

tabathadelane (approved):

clarkmcadoo (approved):

PR #11134, opened 2023-01-20
feat(logging): Add info about restricted attribute name appId behaviour

danybmx:
* What problem do this PR solve?
There are some restrictions to query `appId` attribute on logs, so this PR adds a line of info under attributes that has a restriction.

### DIFF
--- a/src/content/docs/logs/log-api/introduction-log-api.mdx
+++ b/src/content/docs/logs/log-api/introduction-log-api.mdx
@@ -475,7 +475,7 @@ Restrictions on logs sent to the Log API:
 Some specific attributes have additional restrictions:
 
 * `accountId`: This is a reserved attribute name. If it is included, it will be dropped during ingest.
-* `appId`: Must be an integer. If you set a string or other type, the data is ingested but becomes unqueryable.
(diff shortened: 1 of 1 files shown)


jsubirat (approved):

src/content/docs/logs/log-api/introduction-log-api.mdx
```diff
@@ -475,6 +475,7 @@ Restrictions on logs sent to the Log API:
 Some specific attributes have additional restrictions:
 
 * `accountId`: This is a reserved attribute name. If it is included, it will be dropped during ingest.
+* `appId`: Must be an integer. If you set a string or other type, the data is ingested but becomes unqueryable.
```
jeff-colucci, line 478: What is "other type" here? I think that could do with just a touch of clarification, but other than that looks good!

danybmx: Ops, other data type hehe, added a commit, thanks @jeff-colucci!

jeff-colucci (approved):
Looks good! Just a slight style tweak

PR #10975, opened 2023-01-10
docs: updating OTLP metrics support for FedRAMP

tclayton-newr:
Please follow [conventional commit standards](https://www.conventionalcommits.org/en/v1.0.0/)
in your commit messages and pull request title.

## Give us some context

* What problems does this PR solve?
  * This updates our documentation to be more accurate regarding OTLP Metrics and our support for FedRAMP accounts. 
* Add any context that will help us review your changes such as testing notes,
  links to related docs, screenshots, etc.
  * N/A
* If your issue relates to an existing GitHub issue, please link to it.
  * N/A

PR #11140, opened 2023-01-20
fix(OTel): remove port 443 example from HTTP

tclayton-newr:
## Give us some context

  * This removes an example URL that won't work for FedRAMP customers. 
  * I overlooked this change from https://github.com/newrelic/docs-website/pull/10975 and am fixing it here. 


rhetoric101 (approved):
@tclayton-newr thanks for removing that 443 listing.

PR #11155, opened 2023-01-23
fix: change `eventType()` to `entityType()`

JimHagan:
Seems like `entityType` no longer works but `eventType()` does.

working example:

```
FROM Mobile, MobileRequest, MobileRequestError, MobileSession, MobileHandleException, MobileCrash select rate(bytecountestimate() / 10e8, 1 day) as 'GB Ingest' facet appName, eventType() limit 100 TIMESERIES since 1 day ago
```


rhetoric101 (approved):
@JimHagan thanks for making these changes!

PR #11169, opened 2023-01-23
Remove img lazy load to fix scroll

LizBaker:
the scroll was scrolling to the point on the page where the header would be if all of the images had a height of 0. i think this was happening because the scroll height was getting calculated before the images had loaded due to lazy loading. turning that off seems to fix the issue completely 

example page:

old: https://docs.newrelic.com/docs/codestream/how-use-codestream/performance-monitoring/#investigating-errors

new: https://docswebsitedevelop-lizturnofflazyload.gatsbyjs.io/docs/codestream/how-use-codestream/performance-monitoring/#investigating-errors

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -409,6 +409,7 @@ module.exports = {
         allPageHeaders: [
           'Referrer-Policy: no-referrer-when-downgrade',
           'Content-Security-Policy: frame-ancestors *.newrelic.com',
+          'Cache-Control: no-cache',
         ],
       },
     },
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -409,7 +409,6 @@ module.exports = {
         allPageHeaders: [
           'Referrer-Policy: no-referrer-when-downgrade',
           'Content-Security-Policy: frame-ancestors *.newrelic.com',
-          'Cache-Control: no-cache',
         ],
       },
     },
--- a/src/components/MDXContainer.js
+++ b/src/components/MDXContainer.js
@@ -58,7 +58,6 @@ const defaultComponents = {
                   margin: '0 0.25rem',
                 }
           }
-          loading="lazy"
         />
       </Lightbox>
     ),


sunnyzanchi (commented):
the fix works for me! could you add more context to the commit message about why lazy loading causes this? if i found this commit a year from now i don't think i'd understand how this has anything to do with scrolling

tabathadelane (dismissed):
If I remove the trailing slash, it works about 50% of the time. But when it works, it works well.

LizBaker:
> the fix works for me! could you add more context to the commit message about why lazy loading causes this? if i found this commit a year from now i don't think i'd understand how this has anything to do with scrolling

i'm gonna add some explanation to the PR message which we can definitely track down but it feels like a bit much to fit in a commit message

clarkmcadoo (dismissed):

tabathadelane (approved):

sunnyzanchi:
> i'm gonna add some explanation to the PR message which we can definitely track down but it feels like a bit much to fit in a commit message

hmm i would disagree, i think the first paragraph of the PR description is the perfect amount of detail for a commit message. it describes exactly what's happening and why, and why the change fixes it. i actually think it's pretty tough to put _too much_ in a commit message. i used to feel the same way about not needing much in the commit if the PR explained things, but after needing to track down so many PRs from less-than-detailed commits, i think it's so much nicer for our future selves to be able to look at a commit locally and understand it without going through any extra steps

LizBaker:
> > i'm gonna add some explanation to the PR message which we can definitely track down but it feels like a bit much to fit in a commit message
> 
> hmm i would disagree, i think the first paragraph of the PR description is the perfect amount of detail for a commit message. it describes exactly what's happening and why, and why the change fixes it. i actually think it's pretty tough to put _too much_ in a commit message. i used to feel the same way about not needing much in the commit if the PR explained things, but after needing to track down so many PRs from less-than-detailed commits, i think it's so much nicer for our future selves to be able to look at a commit locally and understand it without going through any extra steps

gotcha, i'm gonna leave this as is for now (especially when it's a single commit PR) but we can definitely discuss adopting that as a team. at the moment it's not something i find super necessary

PR #11171, opened 2023-01-23
fix: InstallFeedback overflowing on mobile

sunnyzanchi:


roadlittledawn (approved):
:shipit:

clarkmcadoo:
Screen Shot 2023-01-24 at 5 16 26 PM

its much better, but still getting some overlap

sunnyzanchi:
@clarkmcadoo good catch. i added some more breakpoints to make it look better at smaller sizes, i was just able to squeeze down to 375px

Screen Shot 2023-01-24 at 6 56 24 PM

i think the real reason this stuff is so squished though is that the page has two `main` tags, both with left/right padding, so mobile is smaller than it should be. we should probably make another ticket for that

Kapture 2023-01-24 at 18 52 44

clarkmcadoo (approved):

PR #11186, opened 2023-01-24
Releasing .NET Agent 10.6.0

chynesNR:
Adds 10.6.0 to EOL list, removes one older version that has reached EOL.


homelessbirds:
Looks good!

homelessbirds (approved):

PR #11213, opened 2023-01-25
style(Sonic): Optimize AI (81701)

jeff-colucci:
- [x] Creating pathways into product UI that align with user intent will enable this doc to generate signups.

- [x] Rework the first fold so there is a clearer user story and journeys into the product. Screenshots? Videos? Links?


src/content/docs/alerts-applied-intelligence/overview.mdx
```diff
@@ -31,33 +31,79 @@ redirects:
...
+* **[Overview](/docs/alerts-applied-intelligence/new-relic-alerts/get-started/alerts-ai-overview-page)**: Find your most recent open, closed, and muted incidents.
+* **Issues & activity**: Review your most recent [issues](/docs/alerts-applied-intelligence/new-relic-alerts/get-started/alerts-ai-overview-page/#issues), [incidents](/docs/alerts-applied-intelligence/new-relic-alerts/get-started/alerts-ai-overview-page/#incidents), and [anomalies](/docs/alerts-applied-intelligence/new-relic-alerts/get-started/alerts-ai-overview-page/#anomalies).
+  </Collapser>
+  <Collapser
+    className="freq-link"
+    id="Detect"
+    title="Detect"
+  >
+Define problems in your system you want to be notified about with alert conditions.
+
```
jeff-colucci, line 65: I feel like we are missing two links here to better align it with the UI screenshot above, that is Anomaly detection and Alert coverage gaps.

src/content/docs/alerts-applied-intelligence/overview.mdx
```diff
@@ -31,33 +31,79 @@ redirects:
...
 
-Alerts allow you to define the conditions in your system you want to be notified about, letting you focus on the things you care about most. Applied intelligence is a machine-learning engine that notifies your team about potential problems by identifying anomalous behavior, correlating any relevant incidents, alerting your team of active issues, and providing a root cause analysis.
+New Relic's alerts and applied intelligence were designed for busy DevOps teams to end response fatigue: you can detect and diagnose incidents and prioritize them for immediate action. If you haven't already, If you haven't already, [sign up for your free account](https://newrelic.com/signup) to [start receiving alerts](https://one.newrelic.com/alerts-ai) about your data right away!
```
austin-schaefer, line 36: I wonder if there's an intermediary step we can add here (or even a 1,2,3 procedure). There's really a step between "signup" and "start alerting."

What about something that instead acknowledges the intervening realities and perhaps builds confidence in the whole process?

e.g.

1. Create an account, if you don't have one already [not sure we need this step]
2. Install one or more agents and integrations [link to product IO]
3. [alerting stuff]

jeff-colucci: Agreed, I've been giving this some thought since our meeting. Going to try to cobble something together in a rework of this and see if I can make a better getting started path.

PR #11229, opened 2023-01-25
Create stepper skeleton

LizBaker:
This is a WIP for an upcoming component. For the first round of review we can ignore css, mostly we want to review the structure of the step mdx and its translation process


src/homepageSteps/step1.mdx
```diff
@@ -0,0 +1,12 @@
+---
+homePage: true
```
tabathadelane, line 2: What is this formatter for? I can't find it referenced anywhere else.

tabathadelane (dismissed):
My only question would be about the `homePage` frontmatter being used to describe this. Is this just an early way to differentiate between 'Stepper' folders if in the future there are multiple?

LizBaker:
> My only question would be about the `homePage` frontmatter being used to describe this. Is this just an early way to differentiate between 'Stepper' folders if in the future there are multiple?

@tabathadelane that gets referenced [here](https://github.com/newrelic/docs-website/pull/11229/files#diff-26ca35c3be6d07c84bb77f1e63a1b19487279d5b9c9a9392179b3a35647e1254R224) so we can query/filter for just the step mdx

scripts/actions/fetch-and-deserialize.js
```diff
@@ -130,23 +129,25 @@ const fetchTranslatedFilesZip = (locale) => {
...
+      .map((entry) => {
+        if (entry.entryName.endsWith('.mdx')) {
+          return {
+            path: entry.entryName,
+            html: zip.readAsText(entry, 'utf8'),
+          };
+        }
+        return null;
+      })
+      .filter(Boolean);
```
sunnyzanchi, line 150:
```suggestion
      .filter((entry) => entry.entryName.endsWith('.mdx'))
      .map((entry) => ({
        path: entry.entryName,
        html: zip.readAsText(entry, 'utf8'),
      }))
```

sunnyzanchi (dismissed):
left one suggestion but otherwise looks good to me

scripts/actions/fetch-and-deserialize.js
LizBaker: ooooh nice

LizBaker:
cache clear build: [build-12d43ac8-c4c2-4de4-ac93-8cb0d7c28f3a.gatsbyjs.io](https://build-12d43ac8-c4c2-4de4-ac93-8cb0d7c28f3a.gatsbyjs.io/)

tabathadelane:
> > My only question would be about the `homePage` frontmatter being used to describe this. Is this just an early way to differentiate between 'Stepper' folders if in the future there are multiple?
> 
> @tabathadelane that gets referenced [here](https://github.com/newrelic/docs-website/pull/11229/files#diff-26ca35c3be6d07c84bb77f1e63a1b19487279d5b9c9a9392179b3a35647e1254R224) so we can query/filter for just the step mdx

I was just curious how that might carry over into future uses like maybe within a tutorial-like doc, but I think I am seeing that is a way to query the individual uses, so this would be fine.

tabathadelane (approved):

PR #11245, opened 2023-01-26
chore(images): Transparent icon backgrounds

austin-schaefer:
These look awkward/ugly in the various "tile install" things across the site


bradleycamacho (approved):

homelessbirds (approved):
Looks good to me! I can't see an example of how it will look live on the site but the individual images look great.

PR #11263, opened 2023-01-26
Remove unwanted link in Node.js agent to distributed tracing introduction

rhetoric101:
The Node.js agent navigation tree had a link to the introduction to distributed tracing page under "Extend your instrumentation," but that meant that the left nav would light up for both Node.js and for the introduction to distributed tracing. Since we already have a distributed tracing topic under "Install and configure," I suggest we remove this left-navigation link.


jeff-colucci (approved):
Looks removed to me!

jeff-colucci (approved):

PR #11282, opened 2023-01-27
fix(alerts ): fixing broken nav links

homelessbirds:
I deleted two deprecated docs in [this PR](https://github.com/newrelic/docs-website/pull/10979/files) but neglected to delete them from the nav. Fixing here.


paperclypse (approved):
Looks good to me! 

I'm assuming these are the correct two docs and that the Gatsby build will be good.

PR #11308, opened 2023-01-31
Update cpuPercent.md

tanaka-takayoshi:
## Give us some context

Add a clarification that CPU percent may be greater than 100% in the case of multi-core CPU, similar to ProcessSample.


nbaenam (approved):

nbaenam:
Hi @tanaka-takayoshi! Thank you for the update! All is good!

PR #11316, opened 2023-01-31
Update advanced-config.mdx

Mesverrum:
## Give us some context
Corrected an incorrect default value


jeff-colucci (approved):
Seems like a simple change so I'm approving. Thanks for catching it!

PR #11319, opened 2023-01-31
style(Sonic): Initial pass on optimizing tutorial

jeff-colucci:


jeff-colucci:
Gatsby works and the changes are small, so approving for time constraints

PR #11323, opened 2023-01-31
.NET Agent: Add 2.19.0 to list of verified compatible MongoDB.Driver versions

nr-ahemsath:
Add 2.19.0 to the list of verified compatible MongoDB.Driver versions for the .NET Agent.


nbaenam (approved):
Hey @nr-ahemsath, thanks your the update! It looks good to me!

PR #11332, opened 2023-02-01
fix(stream data): various improvements

zuluecho9:
Add 'Azure' to data streaming in Data Plus doc. 
Clarify the 'streaming data' doc in APM is for APM, and made some improvements related to 'supported agent version' wording.


rhetoric101 (approved):
@zuluecho9 thanks for adding that Azure information and moving the supported versions to a different spot.

PR #11337, opened 2023-02-01
fix: Updates Security API docs to reflect internal API

colfortis:
## Give us some context

Updating format of findings for Security API to match internal API format


homelessbirds (approved):

homelessbirds:
Looks good! Thanks!

PR #11339, opened 2023-02-01
feat: add `LicenseKey` serializer & deserializer

sunnyzanchi:


clarkmcadoo (approved):

tabathadelane (approved):

tabathadelane:
Should we wait until we update the theme, so we can also bump the version here, or do we just want to do that separately?

sunnyzanchi:
> Should we wait to merge until we update the theme, so we can also bump the version here, or do we just want to do that separately?

i'll go ahead an merge this, we can always bump the theme later

PR #11343, opened 2023-02-01
FIX(Go agent) Release notes for 3.20.3 metadata correction

nr-swilloughby:
The previous 3.20.3 release notes doc had incorrect metadata. This corrects that so the correct release note shows as current.


homelessbirds (approved):

homelessbirds:
Looks good! thanks!

PR #11348, opened 2023-02-02
feat(Dashboards): Added the <LIKE> and <RLIKE> example

nbaenam:
Added a new example for `<LIKE>` and `<RLIKE>`.

This is the ticket: [NR-83531](https://issues.newrelic.com/browse/NR-83531)


x8a (approved):
Looking good!

PR #11366, opened 2023-02-03
chore(synthetics): updated module versions

bpeckNR:
Updated a few module versions in the Node 16.10 and Chrome 100 runtimes.


x8a (approved):
Thanks for the update!

PR #11367, opened 2023-02-03
[FIX] CSP recommendations in Browser Compatibility Document

metal-messiah:
## Give us some context
This PR updates the CSP section of the browser agent compatibility requirements doc, as recommended in https://issues.newrelic.com/browse/NR-77599 .  It was missing a route for EU customers.  But beyond this, the list of domains was getting long and could be included as `*.nr-data.net` if desired.  All specific subdomains should still be configurable if the `*` does not meet the customer's needs.

This PR also fixes an error in our "supported versions" in the same document.


rhetoric101 (approved):
@metal-messiah, thanks for updating this details!

PR #11376, opened 2023-02-03
fix scripted browser best practices example

keegoid-nr:
- update script to match recent platform changes
- add missing awaits
- add a second page
- improve comments and script log output
- fix `err` object which should have been named `exception`


rhetoric101 (approved):
@keegoid-nr thanks for updating this script!

PR #11377, opened 2023-02-03
Updated CLM implementation

planteater:
Implementation of code-level metrics has changed slightly, and we're now showing just average duration and error rate. Updated copy and screenshots to reflect this. Also eliminated some references to "golden". Since we're now showing just two of the four classic golden metrics, didn't seem to make sense to keep using that term.


rhetoric101 (approved):
@planteater thanks for adding these updated images and for clarifying the text.

PR #11391, opened 2023-02-06
PUBLISH DATE: WAIT TO PUBLISH ON 2/13/23 Updates related to CLM on by default

mfulb:
The PHP agent team is working on releasing the agent with code level metrics (CLM) on by default.
These changes reflect that the CLM feature is now the default instead of having to be manually enabled.
I tried to find all places this is mentioned but it is possible there are others so it would be good to let those responsible for the CLM sections of the site aware of this change in the agent.


mfulb:
This document is linked with https://github.com/newrelic/docs-website/pull/11444 - we will notify when ready to merge.

mfulb:
We have completed the release so this can be merged.

homelessbirds (approved):

PR #11394, opened 2023-02-06
fix(PHP agent): Add Codestream/CLM to compat page

mfulb:
Updates to the PHP agent compatibility page clarifying Codestream/CLM support as well as recommended PHP versions.


akristen (approved):
Looks good! Thanks for this update :)

PR #11402, opened 2023-02-06
fix(alerts): improving create your first alert

homelessbirds:
Applying the new steps component to "Create your first alert."


paperclypse (approved):
Looks good!

PR #11409, opened 2023-02-06
chore: add license key component instances inside collapsers

tabathadelane:
This reinstates the instances(32) of the use of the new LicenseKey component that was removed from the initial PR due to a UI bug. 

This was the [commit](https://github.com/newrelic/docs-website/pull/11368/commits/ceb5d532703d1aa84d73477fc278f1937ebb3401) that removed them if a reference is needed to double check.

We are waiting for a theme repo update before merging to confirm it is working as expected.


akristen (approved):
SO COOL!

PR #11436, opened 2023-02-07
Changes 'tip' to cover third use-case

mrdlesniak:
In the 'Update your synthetic monitors', the current tip implies that not including the runtime will downgrade a monitor. However, you have to instead include runtime and set the attributes to empty strings ("") to downgrade the monitor.

## Give us some context

'Update your synthetic monitors' section does not cover the use-case for downgrading monitors

Was told this by Engineering, verified in my test account


x8a (approved):
Thanks!

PR #11463, opened 2023-02-09
fix(Distributed tracing): Removed broken link to forum

paperclypse:
The link goes to a forum post that's locked. NR-87387


nbaenam (approved):
Looks good to me!

PR #11464, opened 2023-02-09
Create doc banner

LizBaker:
This adds a dismiss-able banner to specific docs based on frontmatter. When dismissed a value is set in local storage unique to the doc title. You can test this out on [`/docs/mlops/get-started/intro-mlops/`](https://docswebsitedevelop-lizdocbanner.gatsbyjs.io/docs/mlops/get-started/intro-mlops/) and [`/docs/mlops/bring-your-own/mlops-byo/`](https://docswebsitedevelop-lizdocbanner.gatsbyjs.io/docs/mlops/bring-your-own/mlops-byo/) which have temporary banners
fyi if you are logged into production it shouldn't show up at all 👀 

TODO
- [ ] delete temp banner frontmatter
- [x] design tweaks
screenshot


src/templates/docPage.js
```diff
@@ -34,6 +37,8 @@ const splitTOCTitle = (title = '') => {
   return [titleText.join(' '), slug];
 };
 
+const bannerHeight = '78px';
```
clarkmcadoo, line 40: can we `SCREAMING_SNAKE_CASE` this?

src/templates/docPage.js
```diff
@@ -78,6 +84,23 @@ const BasicDoc = ({ data, location, pageContext }) => {
   if (isTutorial) {
     type = 'tutorial';
   }
+  const useBannerDismissed = createLocalStorageStateHook(
+    `docBannerDismissed-${title}`
+  );
+
+  const [bannerDismissed, setBannerDismissed] = useBannerDismissed(null);
+  const [bannerVisible, setBannerVisible] = useState(bannerDismissed !== true);
```
tabathadelane, line 92: Do we check anywhere if they are logged in?

LizBaker: 😂 🙈 ohhoho not yet we don't

LizBaker: fixed now!

src/templates/docPage.js
```diff
@@ -78,6 +84,33 @@ const BasicDoc = ({ data, location, pageContext }) => {
...
     type = 'tutorial';
   }
+  const useBannerDismissed = createLocalStorageStateHook(
+    `docBannerDismissed-${title}`
+  );
+
+  const { loggedIn } = useLoggedIn();
+
+  const [bannerDismissed, setBannerDismissed] = useBannerDismissed(null);
+  const [bannerVisible, setBannerVisible] = useState(bannerDismissed !== true);
```
sunnyzanchi, line 94: i think we can remove `bannerVisible` from state. since it can't be set independently (it wouldn't make sense for `bannerVisible` to be anything other than false if `bannerDismissed` or `loggedIn` are true), it can just be a derived value

```suggestion
  const bannerVisible = !loggednIn && !bannerDismissed
```

src/components/DocPageBanner.js
```diff
@@ -0,0 +1,82 @@
...
+        }
+      `}
+    >
+      <p>{text}</p>
+      <Button
+        css={css`
+          height: 40px;
+        `}
+        as={Link}
+        to={url ?? 'https://newrelic.com/signup'}
```
sunnyzanchi, line 60: what do you think about moving this to a default arg on line 7 in the destructured props? i think it's easier to see a default where something is defined instead of where it's used

tabathadelane (dismissed):
✨

sunnyzanchi (changes requested):
looks great! just left a couple comments

src/components/DocPageBanner.js
LizBaker: because the cta and url props are always getting passed in (just sometimes as null) the usual default syntax doesn't work and the null will override it. i could make new vars at the top though so it's easier to see

src/templates/docPage.js
LizBaker: hoohooo so true

sunnyzanchi (approved):
✨ ✅ ✨

sunnyzanchi (approved):

PR #11488, opened 2023-02-10
chore(PHP agent): Remove Kohana from supported frameworks

mfulb:
Support for the Kohana framework was removed in PHP agent release 9.17 and this fix removes it from the compatibility page.


rhetoric101 (approved):
@mfulb thanks for removing this entry in the table!

PR #11490, opened 2023-02-10
chore(PHP agent): Remove kohana framework from config docs

mfulb:
Removing reference to unsupported Kohana framework for PHP agent.


rhetoric101 (approved):
@mfulb thanks for catching this other reference to kohana!

PR #11503, opened 2023-02-13
Update browser-app-settings-page.mdx

ghost:
Updated instructions to match the UI. This just needs a Hero review.

### DIFF
--- a/src/content/docs/browser/new-relic-browser/configuration/browser-app-settings-page.mdx
+++ b/src/content/docs/browser/new-relic-browser/configuration/browser-app-settings-page.mdx
@@ -7,7 +7,7 @@ redirects:
 
 There are several configuration options that regulate what data the browser agent reports. This is a factor when [enabling browser monitoring](/docs/browser/browser-monitoring/installation/install-browser-monitoring-agent), and you can also change these configurations afterward.
 
-To get to the app settings page:
+To get to the **Application settings** page:
 
 1. Go to **[one.newrelic.com](https://one.newrelic.com/all-capabilities) > Browser**.
 2. Select an app.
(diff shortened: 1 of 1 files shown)


src/content/docs/browser/new-relic-browser/configuration/browser-app-settings-page.mdx
```diff
@@ -11,6 +11,6 @@ To get to the app settings page:
 
 1. Go to **[one.newrelic.com](https://one.newrelic.com/all-capabilities) > Browser**.
 2. Select an app.
-3. On the left navigation, click **App settings**.
+3. On the left navigation, click **Application settings**.
 
-If you don't have access to this, it may be due to permissions.
+If you cannot access this page, it may because you don't have admin-level permissions.
```
a-sassman, line 16:
```suggestion
If you cannot access this page, it may because you don't have admin-level permissions.
```

src/content/docs/browser/new-relic-browser/configuration/browser-app-settings-page.mdx
```diff
@@ -7,10 +7,10 @@ redirects:
...
-To get to the app settings page:
+To get to the **Application settings** page:
 
 1. Go to **[one.newrelic.com](https://one.newrelic.com/all-capabilities) > Browser**.
 2. Select an app.
-3. On the left navigation, click **App settings**.
+3. On the left navigation, click **Application settings**.
 
-If you don't have access to this, it may be due to permissions.
+If you cannot access this page, it may because you don't have admin-level permissions.
```
zuluecho9, line 16: I think we could actually just remove this. We have a lot of things like this that are gated by various capabilities (in this case, this is related to the 'application settings' capability here: https://docs.newrelic.com/docs/accounts/accounts-billing/new-relic-one-user-management/user-capabilities/#browser) but in practice I think it's fine to not document all these things, because a) customers will generally assume if they can't see something it's due to permissions, and b) NR customers have admins dedicated to deciding what their users get access to and for that they'd review the user mgmt docs and make those decisions.

zuluecho9 (dismissed):

src/content/docs/browser/new-relic-browser/configuration/browser-app-settings-page.mdx
a-sassman, line 16:
```suggestion
```

PR #11511, opened 2023-02-13
Move doc tile component

LizBaker:
This moves the doc tile component out of the index page and adds an optional wrapper for formatting.
passing `numbered` to the wrapper will auto number the tiles, but you can also pass a number to individual tiles (not used in the wrapper) via a `number` prop. there is also a label prop which takes an object with text and a color.

tested serialization and deserialization without issue

### screenshots
Screen Shot 2023-02-13 at 1 54 29 PM
Screen Shot 2023-02-13 at 1 54 35 PM
Screen Shot 2023-02-13 at 1 58 45 PM
Screen Shot 2023-02-13 at 1 58 51 PM

### DIFF
--- a/src/components/DocTile.js
+++ b/src/components/DocTile.js
@@ -118,6 +118,10 @@ DocTile.propTypes = {
   label: PropTypes.array,
   date: PropTypes.string,
   instrumentation: PropTypes.object,
+  children: PropTypes.node,
+  path: PropTypes.string,
+  number: PropTypes.number,
+  className: PropTypes.string,
 };
 
 export const DocTiles = ({ children, numbered = false }) => (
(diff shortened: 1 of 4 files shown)


src/components/DocTile.js
```diff
@@ -0,0 +1,164 @@
...
+            color: var(--primary-text-color);
+            margin-left: auto;
+          `}
+        />
+      </div>
+    </div>
+  </SurfaceLink>
+);
+
+DocTile.propTypes = {
```
sunnyzanchi, line 136: i think these need to be updated

sunnyzanchi (changes requested):
looks good! left one comment, and i saw this minor visual bug at <1050px

Screen Shot 2023-02-13 at 5 34 13 PM

LizBaker:
@sunnyzanchi oof there were some silly media queries in there that i don't think we really suffer from removing (the tiles are a bit tighter on the homepage at smaller screen sizes) so i dumped em

sunnyzanchi (approved):

clarkmcadoo (approved):

PR #11523, opened 2023-02-14
chore: lots of minor improvements

brnhensley:
This was supposed to be a small fix that got out of hand. I made a lot small improvements like
- add missing punctuation
-  wrapping code in backticks
- add syntax highlighting to one codeblock
- tried to make our style consistent. In yaml you don't need to use `""` for most strings unless they contain special chars like `\` or `,`
    - so I removed `"` where they aren't needed (since we went back and forth) and added them where they would be needed (like where lists containing `,`)


zuluecho9:
thank you @brnhensley. god bless

zuluecho9 (approved):

PR #11529, opened 2023-02-14
fix(Alerts): Tried to fix some of the pathing in this doc

paperclypse:
Tried to quickly address the issue raised in NR-88762, but I think there are more than just pathing issues in this doc.

@nbaenam I think this PR addresses some minor issues with this doc, but I think the View alert history doc is almost entirely referring to UI that doesn't exist. We should maybe just delete this doc? What do you think?

### DIFF
--- a/src/content/docs/apm/applications-menu/events/view-alert-history.mdx
+++ b/src/content/docs/apm/applications-menu/events/view-alert-history.mdx
@@ -16,13 +16,13 @@ redirects:
   - /docs/apm/applications-menu/events/alerts-page-legacy
 ---
 
-APM's **Alerts** page shows a table of alerts over the past two weeks that you can drill down into detailed information, including throughput, error rate percentage, and history. You can view alert history for a specific application or for all applications.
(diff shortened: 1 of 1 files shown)


nbaenam:
> Tried to quickly address the issue raised in NR-88762, but I think there are more than just pathing issues in this doc.
> 
> @nbaenam I think this PR addresses some minor issues with this doc, but I think the View alert history doc is almost entirely referring to UI that doesn't exist. We should maybe just delete this doc? What do you think?

@paperclypse, yes, I think we should delete this page because it doesn't add new info or explain different things. Besides, I can't see in the UI the categories mentioned at the end of the doc, maybe these categories were in the UI in the past?

We already have the info on the [Alerts & AI overview page](https://docs.newrelic.com/docs/alerts-applied-intelligence/new-relic-alerts/get-started/alerts-ai-overview-page/) page.

nbaenam (approved):

paperclypse:
@nbaenam I'm going to close this PR. Would you please delete the View alert history doc and add a redirect to the doc that has the duplicate info? Thanks!

PR #11539, opened 2023-02-15
Deduping few lines

gani-gcp-zm:
"When the pool size limit is reached, the New Relic Android agent will begin sampling events, discarding some old and some new events, until the pool of events are transmitted with the next harvest cycle." 

This line is repeated. Removed one entry.


nbaenam (approved):
Good catch, thank you!

PR #11541, opened 2023-02-15
PHP CLM tweaks

planteater:
Mention min required version of PHP, and add missing hyphen.


akristen (approved):
Thanks for adding this! Looks good :)

PR #11545, opened 2023-02-15
Update azure-extensions-infrastructure.mdx

kupsand:
Fixed I vs. i for infrastructure. I changed this doc to support any supported OS vs. just Linux.


akristen (approved):
Thanks for the updates!

PR #11558, opened 2023-02-15
feat: create QuickstartChooser component

sunnyzanchi:
the background gradient isn't exactly the same as the designs — the gradient in Figma uses a feature not in CSS, so i approximated as best i could. it's definitely more green though. mobile responsiveness is okay, i went for "good enough" instead of perfect

you can add `<QuickstartChooser />` to the homepage, though it will have gutters on either side as it's with the rest of the content under `<main>`. i figure the new page we use this on won't have that so this should fill the entire width

## screenshot

Screen Shot 2023-02-15 at 6 36 03 PM


tabathadelane (changes requested):
Can we add a variant style or second component version of this that like the one at the bottom of the page?

sunnyzanchi:
> Can we add a variant style or second component version of this like the one at the bottom of the page on the figma mockup?

whoops. @tabathadelane just added
Screen Shot 2023-02-16 at 11 18 23 AM

tabathadelane (dismissed):

clarkmcadoo (approved):

PR #11563, opened 2023-02-16
docs: remove unsupported agent version v1198

coreyarnold:


nbaenam (approved):

PR #11593, opened 2023-02-16
fix: Correct URL for listing location failure conditions

relic-js:
This doc had previously displayed the wrong endpoint for listing location failure conditions. I've added the correct one (instead of passing `policy_id` as a parameter, you include it in the URL).


a-sassman:
@relic-js Can you please sign the SLA agreement? Thank you!

a-sassman (approved):

PR #11597, opened 2023-02-16
fix(sonic): Weird license key component

austin-schaefer:
It renders really strangely https://docs.newrelic.com/docs/apis/intro-apis/new-relic-api-keys/#overview-keys


x8a (approved):
I agree, I don't think it works well in the table.

PR #11610, opened 2023-02-17
docs(guided-install): instructions update

amolero-nr:
This is a general readability update to our `guided install` instructions, including addition of the Windows command.


rhetoric101 (approved):
@amolero-nr thanks for adding this Windows information and making those other helpful edits!

PR #11616, opened 2023-02-17
Add MSMQ support to compatibility doc

nr-ahemsath:
Update the .NET Agent compatibility and support document (for .NET Framework only) to document our instrumentation support for MSMQ.  No specific version information is called out because MSMQ client support is built in to .NET Framework.


rhetoric101 (approved):
@nr-ahemsath, thanks for adding the row about MSMQ.

PR #11627, opened 2023-02-17
Sunny/add inline signup

sunnyzanchi:
for the life of me i couldn't get this component to work if i was importing it from the theme, but having it here seems to work

### DIFF
--- a/src/components/MDXContainer.js
+++ b/src/components/MDXContainer.js
@@ -16,6 +16,7 @@ import { DocTile, DocTiles } from './DocTile';
 import LandingPageHero from './LandingPageHero';
 import LandingPageTile from './LandingPageTile';
 import LandingPageTileGrid from './LandingPageTileGrid';
+import InlineSignup from './InlineSignup';
 import InstallFeedback from './InstallFeedback';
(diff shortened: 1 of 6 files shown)


LizBaker (dismissed):

src/components/MDXContainer.js
```diff
@@ -16,6 +16,7 @@ import { DocTile, DocTiles } from './DocTile';
 import LandingPageHero from './LandingPageHero';
 import LandingPageTile from './LandingPageTile';
 import LandingPageTileGrid from './LandingPageTileGrid';
+import InlineSignup from './InlineSignup';
```
clarkmcadoo, line 19: this need to be `WrappedInlineSignup`?

sunnyzanchi: `WrappedInlineSignup` is the default export in InlineSignup.js so this should be fine. i don't think the build should actually be failing, but i'm going to cut another theme release to remove the InlineSignup component there. once i bump the theme here again i think that should fix the build failure

LizBaker (approved):

PR #11631, opened 2023-02-18
docs: update field mappings for Workflows

emetcalf9:

Updating the recommended field mappings to match the mappings used during automatic migrations


nbaenam (approved):
Thank you for the update! It looks good to me!

PR #11636, opened 2023-02-20
fix(Java agent): Removed the ansible page

nbaenam:
Because the [new-relic-java-agent-ansible-role](https://github.com/newrelic/newrelic-java-agent-ansible-role) repository is not going to be supported any longer, these are the modifications:

- Removed the _Install New Relic Java agent with Ansible_ page in the English, Japanese, and Korean versions. 

- Removed the _Ansible_ option in the _Step 1. Tell us about your Java app_ .

This is the ticket: [NR-87379](https://issues.newrelic.com/browse/NR-87379).


x8a (approved):
Looks good to me!

PR #11639, opened 2023-02-21
Adding new env variable

meda123:
Adding the NEW_RELIC_METADATA_SERVICE_VERSION=XXXXXXXX variable. DO NOT publish until I send a confirmation to the help doc channel.

## Give us some context


src/content/docs/apm/agents/net-agent/configuration/net-agent-configuration.mdx
```diff
@@ -239,6 +239,7 @@ NEW_RELIC_UTILIZATION_DETECT_DOCKER=<var>true|false</var>
 NEW_RELIC_UTILIZATION_DETECT_KUBERNETES=<var>true|false</var>
 NEW_RELIC_FORCE_NEW_TRANSACTION_ON_NEW_THREAD=<var>true|false</var>
 NEW_RELIC_CODE_LEVEL_METRICS_ENABLED=<var>true|false</var>
+NEW_RELIC_METADATA_SERVICE_VERSION=XXXXXXXX
```
obarney, line 242: XXXX was used to indicate private information.   It looks like the pattern here is to put an example like "1.2.3" or "the version of your software".
Frankly, I think the java page organizes this information a lot better and this section should be changed to align with that.  Short of this happening I think we need to add a section below this block that explains what NEW_RELIC_METADATA_SERVICE_VERSION is for using the same words we did in the Java agent docs.

obarney (changes requested):
Frankly, I think the java page organizes this information a lot better and this section should be changed to align with that.  Short of this happening I think we need to add a section below this block that explains what NEW_RELIC_METADATA_SERVICE_VERSION is for using the same words we did in the Java agent docs.

PR #11661, opened 2023-02-21
chore: formatting code

brnhensley:
I removed the indentation in the lists of env vars, I think that was just a cut/paste thing, it didn't serve a purpose. I also added language identifiers for syntax highlighting.


rhetoric101 (approved):
@brnhensley thanks for adding this code formatting.

PR #11674, opened 2023-02-22
Fix explanation about hiding calls based on customer feedback

rhetoric101:
See NR-87120 for details.


nbaenam (dismissed):
It looks good to me except for the screenshot. I think that it could be better without the navigation bar.

PR #11673, opened 2023-02-22
fix: + incompatible versions to ServiceStack.Redis

brnhensley:
Please follow [conventional commit standards](https://www.conventionalcommits.org/en/v1.0.0/)
in your commit messages and pull request title.

## Give us some context

* What problems does this PR solve?
* Add any context that will help us review your changes such as testing notes,
  links to related docs, screenshots, etc.
* If your issue relates to an existing GitHub issue, please link to it.

PR #11699, opened 2023-02-22
.NET Agent: correct compatibility doc for ServiceStack.Redis

nr-ahemsath:
A PR (https://github.com/newrelic/docs-website/pull/11673) was done yesterday to update the .NET Agent's compatibility and support docs regarding known compatible vs. incompatible versions of our instrumentation for `ServiceStack.Redis`; however, the data was accidentally added to the table row for `StackExchange.Redis` which is a different library.  This PR corrects the error.


brnhensley:
Thanks for spotting my mistake Alex.

ghost (approved):

PR #11734, opened 2023-02-23
Update documentation to make note of resource attribute decoration on spans

toddwest:

This PR updates our OpenTelemetry best practices for tracing to clarify how resource attributes are handled. This was done as a result of at least one customer being surprised when they compared the size of tracing data in NRDB compared to the payload being sent in.


homelessbirds (approved):
Looks good! I just made a few changes.

PR #11748, opened 2023-02-24
fix: link to host file should be HTTPS

pnvnd:
Issues with clicking on the the link to download the host file.  Browser appears to download and delete immediately, or take a long time to verify download.  Using HTTPS instead of HTTP, the file downloads immediately, as expected.


zuluecho9:
thanks @pnvnd

zuluecho9 (approved):

PR #11753, opened 2023-02-24
fix: the ESD needs an entityGuid to be returned

keegoid-nr:
In order for the Entity Status Daemon (ESD) to find the associated Synthetic monitor, the ESD needs an entityGuid to be returned as part of the data gathered by the alert condition query. Moving the identifying attribute out of the `filter()` function will accomplish this and allow the monitor's status to turn red or green rather than remain grey in Synthetics.

## Give us some context


keegoid-nr:
Requesting review from @relic-js. I don't think I have permission to add reviewers? At least I can't find a way.

relic-js:
As it's currently written in the docs, the `filter()` example will **not** associate the alert condition's signal with an entity. A signal can only be assigned to an entity based on the conditions in the body of the query, not on any conditions within the `SELECT` statement. Adjusting the query so that an individual monitor is defined in the main `WHERE` clause will mean that, for any aggregation window where the monitor performs a check, the signal will be tied to that entity.

This is just to say that your edit looks good. I do have a couple suggestions:

* We might consider changing the `filter()` clause to read `WHERE result = 'FAILED'`. In that situation, a window with a successful result would return a `0`, allowing the condition's threshold to resolve on its own. This also puts it more in line with the earlier example.
* Speaking of that earlier example, it actually has a typo — `WHERE result = 'FAILURE'`. This might be a good opportunity to change that to `FAILED`.

zuluecho9 (approved):

PR #11771, opened 2023-02-27
chore(Java): Updating eol document

meiao:
## Give us some context
Updates EOL table, removing unsupported versions and adding latest release.


rhetoric101 (approved):
@meiao thanks for updating this EOL table!

PR #11779, opened 2023-02-27
feat(Sonic): Updating homepage tile (88204)

jeff-colucci:
Changing vuln management tile to browser monitoring


paperclypse (approved):
Looks good! (I'm assuming the Gatsby build will work)

Issue #11786, opened 2023-02-28
Broken link : Compatibility and requirements for the Java agent


qryuu:
## How can we make our docs better?
[Built-in instrumentation](https://docs.newrelic.com/docs/apm/agents/java-agent/getting-started/compatibility-requirements-java-agent/#auto-instrumented) >[ App/Web servers](https://docs.newrelic.com/docs/apm/agents/java-agent/getting-started/compatibility-requirements-java-agent/#auto-instrumented)

The following two links included in the above section do not exist

[WebSphere 8.5.x to 9.x](https://docs.newrelic.com/docs/agents/java-agent/additional-installation/websphere-application-server/)
WebSphere Liberty Profile 8.5 to latest
[WildFly 8.0.0.Final to 21.x](https://docs.newrelic.com/docs/agents/java-agent/additional-installation/wildfly-11-installation-java/)

PR #11791, opened 2023-02-28
chore(Java agent): Fixed broken links

nbaenam:
Fixed broken links.

See [11786](https://github.com/newrelic/docs-website/issues/11786)


rhetoric101 (approved):
@nbaenam thanks for removing these links. I think these docs were combined into the new branching format.

PR #11792, opened 2023-02-28
feat(vm): added more api details

maxa101:


src/content/docs/vulnerability-management/integrations/trivy.mdx
```diff
@@ -23,6 +22,8 @@ curl --location --request POST 'https://security-api.newrelic.com/security/v1?re
 --header 'X-Scanner: Trivy' \
 -d @nameOfSarifFile.json
  ```
+ 
+Replace <RepositoryUrl> with your integration repository (optional).
```
x8a, line 26: Do we really need this? I updated the example above to show users how to add their GH repository.

x8a (dismissed):

src/content/docs/vulnerability-management/integrations/trivy.mdx
maxa101: I think this caused some confusion as well since the field isn't documented here https://docs.newrelic.com/docs/vulnerability-management/integrations/security-data-api/ (which might be a better place)

x8a: I see! Then we can add it in the security-data doc and then have the example here.

x8a:
Closing this PR, since all edits will be introduced with #11912.

PR #11819, opened 2023-03-01
Add metadata to Java agent release notes

jasonjkeller:
This PR brings the Java agent release notes inline with the Agent Release Notes spec by adding required metadata.


nbaenam (approved):
Thank you for the update! I've added some fixes, other than that it looks good to me!

PR #11821, opened 2023-03-01
Hide left & right navs on mdx pages with frontmatter

tabathadelane:
if `hideNavs: true` frontmatter is added to a doc, the page layout changes. The left nav will collapse, and the right rail will be removed. 

Some docs pages are in experiments as JS, so there is also an added check for those pathnames.

For preview purposes, there is a temporary frontmatter added to 
 - [docs/new-relic-solutions/get-started/quick-launch-guide/](https://docswebsitedevelop-tabathahidebothnavs.gatsbyjs.io/docs/new-relic-solutions/get-started/quick-launch-guide/)
 - [docs/apm/new-relic-apm/getting-started/introduction-apm/](https://docswebsitedevelop-tabathahidebothnavs.gatsbyjs.io/introduction-apm/) (JS page)
 - [docs/apm/new-relic-apm/getting-started/apm-agent-data-security/](https://docswebsitedevelop-tabathahidebothnavs.gatsbyjs.io/docs/apm/new-relic-apm/getting-started/apm-agent-data-security/)

TODO:
 - [x] remove the frontmatter from mdx pages

### DIFF
--- a/src/layouts/MainLayout.js
+++ b/src/layouts/MainLayout.js
@@ -45,11 +45,14 @@ const MainLayout = ({ children, pageContext, sidebarOpen = true }) => {
   /*
    * [VSU] some docs pages are being designed as JS for faster experimenting
    * and will never have the frontmatter property
+   * Using regex for check to account for paths with and without trailing slash
    */
-  const introDocsAsJS = ['/introduction-apm'];
-  if (introDocsAsJS.includes(location.pathname) && !hideNavs) {
-    hideNavs = true;
-  }
+  const docsAsJS = ['/introduction-apm'];
+  docsAsJS.forEach((docUrl) => {
+    if (location.pathname.match(docUrl) && !hideNavs) {
+      hideNavs = true;
+    }
+  });
 
   useEffect(() => {
     setIsMobileNavOpen(false);
--- a/src/layouts/MainLayout.js
+++ b/src/layouts/MainLayout.js
@@ -24,7 +24,8 @@ import { MainLayoutContext } from '../components/MainLayoutContext';
 const MainLayout = ({ children, pageContext, sidebarOpen = true }) => {
   const { loggedIn } = useLoggedIn();
   const { sidebarWidth, contentPadding } = useLayout();
-  const { locale, slug, hideNavs } = pageContext;
+  const { locale, slug } = pageContext;
+  let { hideNavs } = pageContext;
   const location = useLocation();
   const [isMobileNavOpen, setIsMobileNavOpen] = useState(false);
   const [searchTerm, setSearchTerm] = useState('');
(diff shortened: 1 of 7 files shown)


src/layouts/MainLayout.js
```diff
@@ -41,9 +42,21 @@ const MainLayout = ({ children, pageContext, sidebarOpen = true }) => {
...
+   * [VSU] some docs pages are being designed as JS for faster experimenting
+   * and will never have the frontmatter property
+   * Using regex for check to account for paths with and without trailing slash
+   */
+  const docsAsJS = ['/introduction-apm'];
+  docsAsJS.forEach((docUrl) => {
+    if (location.pathname.match(docUrl) && !hideNavs) {
+      hideNavs = true;
+    }
+  });
```
sunnyzanchi, line 55: how about this?
```suggestion
  const isJSDoc = docsAsJS.some((docUrl) => docUrl.test(location.pathname));
  hideNavs ||= isJSDoc
```

`String#match` would work here, but it returns either an array of matches or `null`, whereas `RegExp#test` returns a boolean, which feels more appropriate here

src/layouts/MainLayout.js
```diff
@@ -41,9 +42,21 @@ const MainLayout = ({ children, pageContext, sidebarOpen = true }) => {
     }
   };
 
+  /*
+   * [VSU] some docs pages are being designed as JS for faster experimenting
+   * and will never have the frontmatter property
+   * Using regex for check to account for paths with and without trailing slash
+   */
+  const docsAsJS = ['/introduction-apm'];
```
sunnyzanchi, line 50: did you mean this?
```suggestion
  const docsAsJS = [/introduction-apm/];
```
`String#match` will automatically try to convert its argument to a regex, so the string `'/introduction-apm'` is converted to `/\/introduction-apm/`

src/layouts/MainLayout.js
```diff
@@ -41,9 +42,21 @@ const MainLayout = ({ children, pageContext, sidebarOpen = true }) => {
...
+  docsAsJS.forEach((docUrl) => {
+    if (location.pathname.match(docUrl) && !hideNavs) {
+      hideNavs = true;
+    }
+  });
+
   useEffect(() => {
     setIsMobileNavOpen(false);
-    setSidebar(sidebarOpen);
+    setSidebar(hideNavs ? !hideNavs : sidebarOpen);
```
sunnyzanchi, line 59: nit: these are technically equivalent but i think it'd be more clear to just use `false`

```suggestion
    setSidebar(hideNavs ? false : sidebarOpen);
```

sunnyzanchi (changes requested):
looks good, just left a couple comments

src/layouts/MainLayout.js
tabathadelane: Classic overthinking it.

src/layouts/MainLayout.js
tabathadelane: Okay. I was wondering about these Regex nuances, but it was behaving as expected, so I just left it to the will of the PR gods. I'll make an update.

sunnyzanchi (approved):
✅

LizBaker (dismissed):

tabathadelane:
I'm going to remove the frontmatter from testing which will dismiss the reviews. I'll just do an override once the checks finish. 👍

sunnyzanchi (approved):

PR #11824, opened 2023-03-01
fix(PHP agent): Correct minimum glibc version supported

mfulb:
The PHP team would like to correct the minimum GLIBC version supported for the PHP agent.  We have determined the value we used was too new as the operating systems we say we support use older versions than 2.35.  The new minimum version is from Centos 7 which is a supported OS and uses the oldest GLIBC of any of the ones we support.


rhetoric101 (approved):
@mfulb thanks for expanding the range of glibc support.

PR #11833, opened 2023-03-01
fix(sonic): inline signup to intro doc

bradleycamacho:
Adds the inline signup form to the Get started w/ NR doc


jeff-colucci (dismissed):
Looks good

PR #11834, opened 2023-03-01
feat(sonic): apm popover sweep

bradleycamacho:
This adds the APM popover component to a selection of docs. This is only a portion of instances where we could place the component, we will revisit later based on how these perform.

Component:
`<InlinePopup type="apm" />`


akristen (commented):
Looks like the inline is only adding an awkward indent, but not displaying the popover

PR #11835, opened 2023-03-01
Remove unsupported Python agent versions from EOL table.

umaannamalai:
The Python agent EOL table included several outdated agent versions. This PR removes those unsupported agent versions.


rhetoric101 (approved):
@umaannamalai thanks for keeping this table up to date!

PR #11836, opened 2023-03-01
Display steps on mount

LizBaker:
To get around some hydration issues we previously only rendered the page on mount, but this broke our site indexing. I'm experimenting with wrapping only the steps with this logic


tabathadelane (approved):
I'm not sure how to know if swiftype is happy, but otherwise lgtm

clarkmcadoo (approved):

PR #11878, opened 2023-03-03
Update tracing-gap-aws-x-ray-integration.md

aalfano7:
Updating per request from TSS: NR and AWS traces cannot be blended together into a single end-to-end trace


homelessbirds (approved):

homelessbirds:
Looks good! Thanks!

PR #11879, opened 2023-03-03
create popovers.json for different popover types

sunnyzanchi:
this JSON will populate the soon-to-be-added InlinePopover component. the top-level keys are the different `type`s that can be used


tabathadelane (approved):
✨

PR #11881, opened 2023-03-03
fix: Update choose-your-aggregation-method.mdx

relic-js:
Our docs provide a clear, concise example for the particular scenario in which you'd want to use the cadence aggregation method. However, we immediately follow it up with a contradictory statement:

> If you're currently using cadence, choose one of the other aggregation methods.

This edit softens the statement to note that we only advise against cadence outside of the case mentioned above it. Before going any further, though, I'd like @HenryTech to give this approach a review from his standpoint.


HenryTech:
I approve this change -- thanks for thinking of this @relic-js !

nbaenam (approved):
Thank you for the update! It looks good to me!

PR #11884, opened 2023-03-04
fix: Update new-relic-api-keys.mdx

relic-js:
The section on user keys is ambiguous enough to potentially mislead a customer into thinking that the REST API supports cross-account user key authentication. However, our REST API endpoints only recognize a user key as being associated with its root account.

This edit attempts to clarify that point.


x8a (approved):
Thanks for the edits!

PR #11898, opened 2023-03-06
fix(sonic): Make tabs non-headers

bradleycamacho:
image

The tabs here were accidentally set as `### H3s`, leading to strange rendering and a poor viewing experience. This removes the header styling from them.


jeff-colucci (approved):
They look removed to me

PR #11902, opened 2023-03-06
chore: Add correct class prefix to class method

kaylareopelle:
The documented strategy for calling the method is wrong.


homelessbirds:
thanks for fixing this @kaylareopelle!

PR #11910, opened 2023-03-07
Add a new setting to RabbitMQ

kang-makes:
We added a new parameter to RabbitMQ so the timeout can be configurable: https://github.com/newrelic/nri-rabbitmq/pull/99

We would like to have it available in the docs.


x8a (approved):
Thanks

PR #11917, opened 2023-03-07
feat: create `InlinePopover` component

sunnyzanchi:
also sets up the `TwoButton` layout (currently the only one), and makes it easy to add more

## testing
- pick your favorite MDX file
- add `<InlinePopover type="apm" />`, or one of the other `type`s in `popovers.json`
- check that page and make sure the popover renders correctly and Tessen events are sent


src/components/InlinePopover/layouts/TwoButton.js
```diff
@@ -0,0 +1,130 @@
...
+  primaryButton,
+  primaryButtonUrl,
+  secondaryButton,
+  secondaryButtonUrl,
+  type,
+  ...props
+}) => {
+  const capsType = capitalize(type);
+  const instrumentation = (eventName) => ({
+    category: `${capsType}Popover`,
```
tabathadelane, line 26: Nit: In `InlinePopover.js` we are sending `${type}InlinePopover` Is there a reason these should be different, or is it just an oversight?

sunnyzanchi: ah good catch, no they should be the same

sunnyzanchi: changed this in `InlinePopover.js` to just `${type}Popover`

clarkmcadoo (approved):

tabathadelane:
I love it 🤩 
Screen Shot 2023-03-08 at 9 53 56 AM

It doesn't love being at the edge of the screen, but i dont know if there is anything we can do about that
Screen Shot 2023-03-08 at 9 57 18 AM

sunnyzanchi:
> It doesn't love being at the edge of the screen when there are no navs, but i dont know if there is anything we can do about that

hmm, i copied over the original logic from the LicenseKey popover which is supposed to avoid that scenario by offsetting the popover left/right. let me double check

PR #11931, opened 2023-03-07
sonic: get started synthetic monitor tutorial

akristen:
- Added one sentence value statement to each monitor
- Added new tutorial section
- Moved in line sign up form to the beginning
- Deleted reference doc from What's next?
- Moved CTAs to body of text from fig captions
- Updated intro copy to lead with 'no installation' needed

Is the intro concise and snappy, and does it capture that:

- Synthetics helps with proactive monitoring
- Synthetics does not need an installation
- Synthetics is easy to use (Without saying 'it's easy')


src/content/docs/synthetics/synthetic-monitoring/getting-started/get-started-synthetic-monitoring.mdx
```diff
@@ -58,32 +65,119 @@ Grab the URL of a page you want to test and drop it in the **URL** field.
...
 </SideBySide>
 
 </Step>
 
 </Steps>
 
-## Create a scripted browser monitor [#script]
+## Test site behavior with scripted monitoring [#script]
+
+Scripted browser monitors are a flexible monitor type that let you test website behavior. Scripting lets you deploy monitors that respond to your web apps like a customer might, choosing different paths across your site depending on conditions you've scripted.
```
jeff-colucci, line 174: "Scripted browser monitors are a flexible monitor type" reads a little weird to me. Maybe just cut out the "are a flexible monitor type" and just go straight to what SBMs actually do, more direct value prop.

jeff-colucci (approved):
I only saw one small thing, I'm going to go ahead and approve for publishing. Overall looks great to me! I feel like the added value props really add a lot to the meat of the doc without making it a burden to get through.

PR #11934, opened 2023-03-08
fix(alerts and applied intelligence ): removing outdated doc

homelessbirds:
This is an outdated doc that should be removed entirely.


nbaenam (dismissed):
I've added a redirect ;-)

nbaenam (approved):

PR #11945, opened 2023-03-08
feat(sonic): Iterate on APM copy

bradleycamacho:


x8a (approved):
Just some minor comments about punctuation. I think we should be persistent and decide whether we want to have full stops 🙌🏻.

x8a (approved):
LGTM

PR #11959, opened 2023-03-08
Tabatha/mini search

tabathadelane:
Adds an animated search input that only shows when on the newUserView
Also updates the text in the toggle / its popup


sunnyzanchi (approved):
looks good! there's a WIP commit in here so i'd do a squash merge when you merge this in

LizBaker (approved):

PR #11960, opened 2023-03-08
fix(style-guide): Use en dashes as described

dvanoni:
## What problem does this PR solve?

Our style guide docs state that we should use en dashes for ranges and spans, but the docs themselves don't actually use en dashes in the examples. This PR fixes that. 🙂


homelessbirds (approved):

homelessbirds:
Good catch! Thanks!

PR #11962, opened 2023-03-09
fix: Revise query example in NRQL conditions doc

relic-js:
Fixed incorrect NRQL syntax in a query example (`FROM` should not follow `WHERE`).


x8a (approved):
Thanks for the fix!

PR #12006, opened 2023-03-10
feat: add function to generate EOL date

jmartin4563:
Created a new function called `getEOLDate()` that will programatically generate an EOL date based on a piece of software's release date and our EOL policy, which is described
[here](https://docs.newrelic.com/docs/licenses/end-of-life/notification-changes-new-relic-saas-features-distributed-software/#impact).

Function accepts two inputs: the first is the release date of the piece of software. The function assumes this input is a string in yyyy-MM-dd format. Second input is an override EOL date. If passed, the function will favor this EOL date over the generated one.

The intent of this function is to allow this repo to programatically generate the various EOL pages it supports, to ease the release process burden on NR developers (updating HTML programatically is hard and humans can forget to do manual steps).

### DIFF
--- a/src/utils/__tests__/getEOLDate.test.js
+++ b/src/utils/__tests__/getEOLDate.test.js
@@ -1,0 +1,40 @@
+const getEOLDate = require('../getEOLDate');
+
+describe('eolDate function', () => {
+    it('should throw an error if dateString is missing', () => {
+      expect(() => getEOLDate()).toThrow('releaseDate must be a valid date in yyyy-MM-dd format')
+    })
+  
+    it('should throw an error if dateString is malformed', () => {
+      expect(() => getEOLDate('October 1, 2019')).toThrow('releaseDate must be a valid date in yyyy-MM-dd format')
+    })
+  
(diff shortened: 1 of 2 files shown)


clarkmcadoo:
hey @jmartin4563! looks like these tests are failing locally when I run that file, all just by 1 day.  I'm not sure why they passed in the workflow, I'm checking that out now.

jmartin4563:
@clarkmcadoo I'm seeing the same thing locally, I think it's something to do with timezones locally vs in CI 🤦🏻‍♀️

clarkmcadoo:
ohhhh great call

jmartin4563:
@clarkmcadoo latest commit has a fix, I updated to add the timezone offset to prevent the midnight barrier issue that was causing local failure, but there's another way to fix this that would require a new dep `date-fns-tz`, curious to get your thoughts on if you'd prefer that over the fix I committed.

clarkmcadoo (approved):

sunnyzanchi (approved):
looks great!

PR #12010, opened 2023-03-10
feat(sonic): updating NR Lookout

akristen:
Was in the area, saw that this experience was a little confusing. Spruced up content:

- Deep linked to NR Lookout
- Spruced up intro 
- Deleted callout
- Updated pathing


jeff-colucci (approved):
Looks good!

PR #12017, opened 2023-03-10
feat(synth): Support custom job timeouts on Job Manager

bpeckNR:
- Removed the limitation for custom job timeouts from the job manager transition guide
- Added config details for CHECK_TIMEOUT (Docker) and globals.checkTimeout (K8s) to the job manager configuration guide


rhetoric101 (approved):
@bpeckNR thanks for adding these two entries and removing that transition guide bullet.

PR #12041, opened 2023-03-13
Node.js Agent v9.12.0 Release Notes 

mrickard:
Node.js Agent v9.12.0 Release Notes 

Version 9.12.0 is likely to be released on March 14, 2023.


x8a (approved):
Thanks for adding these!

PR #12070, opened 2023-03-14
fix: add serialization handlers for InlinePopover

tabathadelane:
update serializers for LicenseKey component becoming InlinePopver

### DIFF
--- a/scripts/actions/utils/handlers.js
+++ b/scripts/actions/utils/handlers.js
@@ -123,12 +123,12 @@ module.exports = {
     deserialize: deserializeComponent,
     serialize: serializeComponent,
   },
-  LicenseKey: {
+  InlinePopover: {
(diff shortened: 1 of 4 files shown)


sunnyzanchi (dismissed):

scripts/actions/utils/handlers.js
```diff
@@ -123,20 +123,20 @@ module.exports = {
...
     serialize: (h, node) =>
       h(
         node,
         'span',
         {
           'data-type': 'component',
-          'data-component': 'LicenseKey',
+          'data-component': 'InlinePopover',
         },
         [u('text', 'account license key')]
```
clarkmcadoo, line 141: we should probably update some of these, i'm not sure what `u` does though

clarkmcadoo (approved):

PR #12083, opened 2023-03-15
Remove Spring Boot support designation

tbradellis:
Spring Boot is not a library in and of itself.  It is a collection of Spring modules.  Our instrumentation targets certain libraries/modules and we provide instrumentation for a range of versions for some individual libraries.  It is misleading to say we support Spring Boot X, because we do not provide instrumentation for every library in Spring Boot.   Instead, we should list the libraries for which we provide instrumentation, along with the range of versions.  Removing this Spring Boot line achieves that.

## Give us some context


nbaenam (approved):
Thank you for the update!

PR #12088, opened 2023-03-15
fix(user model): clarify how u know which on

zuluecho9:
Liaison work: clarify the best way to know which user model you're on. 

And some assorted clean-up/simplification i noticed as I did that.


x8a (approved):
LGTM!

PR #12112, opened 2023-03-16
chore(apm): fix typo in ruby-agent

tnir:
Fix a typo in https://docs.newrelic.com/docs/apm/agents/ruby-agent/installation/ruby-agent-heroku/#configuring


nbaenam (approved):
Thank you for your update! I've added minor modifications.

PR #12154, opened 2023-03-17
chore(Ruby agent): Mention Rails credentials

kaylareopelle:
## Give us some context:

Completes customer issue: https://issues.newrelic.com/browse/NR-87816


kaylareopelle:
@homelessbirds  - I wonder if tips like the one I'm adding to this doc might be helpful in other docs? 

My hunch is very few people visit the Rails plugin documentation, as installing the agent using a gem is much more common than using a plugin.

akristen (approved):
Looks good!

PR #12156, opened 2023-03-17
new workflows feature

iazam6:
In this doc update, I've added information about a new feature under Workflows called Workflows Log. This request is from Matan, the PM leading Workflows.


akristen (approved):
Looks good! Made some copy changes :)

PR #12165, opened 2023-03-20
Remove open preview callout nri-winservice

gsanchezgavier:
Removing the Open Preview callout as the integration [has been promoted to GA](https://github.com/newrelic/infrastructure-agent/releases/tag/1.39.1).


nbaenam (approved):
Thank you for the update!!

PR #12166, opened 2023-03-20
infra_agent: fix manual installation arch

rubenruizdegauna:
Currently, the architecture in the manual installation one-liner is set to `amd64` . This PR replaces it with the `ARCH` variable.


nbaenam (approved):
Thank you for the update!!

PR #12211, opened 2023-03-21
Homepage tweaks

LizBaker:
This fixes the footer bug we had and also makes the inline signup banner small/expandable on tablet width
I also adjusted the mobile header because it had some weird inconsistencies in light/dark


sunnyzanchi (approved):

tabathadelane (approved):
✨

PR #12213, opened 2023-03-21
Update advanced-config.mdx

Mesverrum:
Removing one of the example configs because it leads users toward a bad pattern.  They should not be using multiple files for the devices lists.


nbaenam (approved):

PR #12214, opened 2023-03-21
add labeler

sunnyzanchi:
adds the Labeler workflow to automatically add the 'eng' label to PRs where JS files have changed


tabathadelane (approved):

.github/workflows/label.yml
```diff
@@ -0,0 +1,22 @@
+# This workflow will triage pull requests and apply a label based on the
+# paths that are modified in the pull request.
+#
+# To use this workflow, you will need to set up a .github/labeler.yml
+# file with configuration.  For more information, see:
+# https://github.com/actions/labeler
+
+name: Labeler
```
roadlittledawn, line 8: i _think_ this is the name that will show up in the Checks pane on a PR page right?

is it possible to name this with something that's a bit more explicit, esp for tech writers who may wonder what that check is? maybe something like `Add eng PR label` or slightly more generic `Add PR labels`?

roadlittledawn (changes requested):
:ship

.github/workflows/label.yml
sunnyzanchi: yep that's right. and good idea, i'll update it

roadlittledawn (approved):

PR #12219, opened 2023-03-22
NR-99637 Fix errors in the documentation

vetusbs:

## Give us some context

It was reported that one of the request of the https://docs.newrelic.com/docs/apis/nerdgraph/examples/nerdgraph-entities-api-tutorial/ contains an invalid request body.

The request body was

```
{
        actor {
          entitySearch(query: "domain='APP' and alertable is false") {
            results( {
              entities {
                name
              }
            }
          }
        }
      }
```
This is wrong because there is a `(` after the word `results` and this is a syntax error. Also, the domain APP does not exist, so I'm changed this to APM.
I've also changed it to `alertable is TRUE` to make this request return results.


naxhh (approved):

rhetoric101:
@vetusbs it looks like you still need to sign our license agreement so we can use your content. Could you sign this again even if you've signed this in the past?

PR #12226, opened 2023-03-22
fix/ct nerdgraph typo

thezackm:
fixes 2 problems with a current page for Change Tracking and adds a clarification on another page:

 * `user` is not a New Relic field, it is an optional field; this puts the field in the right section
 * `timestamp` needs to be epoch ms; this clarifies that and adds a callout about a limitation for timestamps to be within 24 hours of the current time
 * NRDB events are immutable and cannot be deleted; this adds a callout to that point since we are receiving a lot of questions about this in the internal slack channels


rhetoric101 (approved):
@thezackm thanks for adding these clarifications about change tracking!

PR #12235, opened 2023-03-22
Adapt all mobile hybrid docs to one template

ghost:
The hybrid docs should all be based off a singular template because the process to install the agent is the same for every agent. This PR captures unified hybrid mobile app docs modeled after the newest Flutter agent doc.


ally-sassman:
Closed until I have time for this in the future.

PR #12240, opened 2023-03-22
fix(manage-accounts-nerdgraph): document regionCode argument

kdef-nr:
From support, the allowed values for `regionCode` in the `accountManagementCreateAccount` mutation are missing.

- add mutation example with `regionCode`
- note the valid values are `us01` and `eu01`
- remove note about only one region being allowed (that's not always true)


rhetoric101 (approved):
@kdef-nr thanks for adding this example!

PR #12294, opened 2023-03-24
Instrument nav collapse clicks and various searches

LizBaker:
This adds instrumentation to the nav being opened or closed and also adds more specific categories to the various searches we have (sidebar, persona view, default view)

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -7,6 +7,7 @@ import {
   Link,
   Icon,
   SearchInput,
+  useTessen,
   useTranslation,
   useInstrumentedHandler,
 } from '@newrelic/gatsby-theme-newrelic';
@@ -18,6 +19,7 @@ import HomepageSlabs from '../components/HomepageSlabs';
 import useMediaQuery from '../hooks/useMediaQuery';
 
 const HomePage = ({ data }) => {
+  const tessen = useTessen();
   const {
     site: { layout },
     allMarkdownRemark: { edges: whatsNewPosts },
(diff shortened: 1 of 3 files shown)


src/pages/index.js
```diff
@@ -67,7 +69,13 @@ const HomePage = ({ data }) => {
         isIconClickable
         alignIcon={SearchInput.ICON_ALIGNMENT.RIGHT}
         onChange={(e) => setSearchTerm(e.target.value)}
-        onSubmit={() => navigate(`?q=${searchTerm || ''}`)}
+        onSubmit={() => {
+          tessen.track({
+            eventName: 'defaultViewSearchTriggered',
+            category: 'HomepageSearch',
```
tabathadelane, line 75: I think these got flipped. The other search events look like this:
```suggestion
            eventName: 'homepageSearchEntered',
            category: 'DefautViewSearch',
```

LizBaker: whoops, i decided halfway through that i liked the category to describe if it was default, persona, or sidebar. but the category could be the same for every one of them. maybe i'll ask austin if there's a preference here

LizBaker: actually i think i'm just gonna make that change and have it all be one search category with varied events so we can narrow a select down by category and then facet on event

clarkmcadoo (approved):

tabathadelane (approved):

sunnyzanchi (approved):

PR #12303, opened 2023-03-24
Change docs to support OpenShift v4.12

vuqtran88:
- Add OpenShift v4.12 to the support compatibility matrix.
- Update the "Setup mTLS for ECTD with OpenShift" with working steps and improve the existing one with examples..
- Add additional service account setting for control plane.


akristen (approved):
Looks good! Thanks

PR #12305, opened 2023-03-25
Add missing parenthesis in conf example

philweber:


x8a (approved):
Thanks for catching this!

PR #12316, opened 2023-03-27
feat(sonic): Create service level/benchmark doc

bradleycamacho:
A draft version of the benchmark/SLO doc


nbaenam (commented):
I think the tags `<Steps>` and `<Step>` and their closing tags must be reviewed. Some of those tags are missing.

PR #12358, opened 2023-03-29
Create branching doc for Kubernetes installation

rhetoric101:


rhetoric101:
I am closing this because the Kubernetes changed the focus of the page so it just talks about the guided installation. This was a big improvement, so we no longer need to try a branching approach.

PR #12360, opened 2023-03-29
docs: add agent version metadata for Node.js supported releases NR-93129

coreyarnold:
adds frontmatter for supported releases


nbaenam (approved):
Thank you for the update!! I've added some minor modifications.

PR #12363, opened 2023-03-29
March 31 - Azure native, add private offers

zuluecho9:
We're adding this feature on March 31, can publish morning of.


rhetoric101 (approved):
@zuluecho9 thanks for adding this!

PR #12375, opened 2023-03-29
add nav animations

sunnyzanchi:
this PR adds a slide in/out animation to the nav sidebar, and a cute lil mini animation to the collapser. it also adds transition animations to the nav items when they open/close

the way the slide animation works is:
- nav is open. page layout is normal, containing div is `display: grid;` with two columns, one for the sidebar `aside` and one for `main`
- when clicking the collapser, the page width increases to `calc(100% + var(--sidebar-width))`. there's a new top-level containing div with `overflow: clip` to prevent horizontal scrolling. the transition starts, the entire containing div translates to the left by an amount equal to `--sidebar-width`, to slide the navbar out, while translating `PageTools` to the right to keep it in the same position and translating Layout.Main to the left slightly to adjust for the new padding
- the inverse happens when opening the nav from its collapsed state

currently, this works perfectly at widths <1600px. above that, things jump a little bit because `main` hits its max-width, `var(--site-max-width)` and it becomes centered in its grid track. this is probably fixable, i just haven't figured out how yet. it might be that it's too dynamic to try doing this with more or less hardcoded numbers like i'm doing, in which case we'd need to consider something like [react-flip-toolkit](https://github.com/aholachek/react-flip-toolkit)

## clip

[nav animation.webm](https://user-images.githubusercontent.com/14365008/228979133-412f53ee-4a55-4c08-a2bb-4130adac11bb.webm)

## todo
- [x] fix collapser not fixed on homepage


package.json
```diff
@@ -52,7 +52,8 @@
     "react-dom": "^17.0.0",
     "react-helmet": "^6.1.0",
     "react-scroll": "^1.8.7",
-    "react-spring": "9.1.2",
```
sunnyzanchi, line 55: i initially upgraded `react-spring` bc newer versions support interpolating CSS vars. unfortunately, this only works if the var is declared on `:root`, so this didn't end up mattering. doesn't hurt to upgrade though

src/templates/docPage.js
```diff
@@ -179,28 +182,61 @@ const BasicDoc = ({ data, location, pageContext }) => {
...
         </LoggedInProvider>
         {!hideNavs && (
-          <Layout.PageTools
-            css={css`
-              @media screen and (max-width: 1240px) {
-                margin-top: 1rem;
-                position: static;
-              }
-            `}
+          // TODO pass nodeRef to avoid `findDOMNode` usage
```
sunnyzanchi, line 185: this todo isn't _strictly_ necessary, it's just recommended. passing a `ref` through to PageTools (which is in the theme) would just be 😩 so i'm inclined to say it's ok for now

tabathadelane:
The icon is backwards when closed on the homepage. Maybe because we aren't animating it here?

https://user-images.githubusercontent.com/47728020/229203958-c327b945-a83f-4a54-9796-c394b3154cb7.mov

tabathadelane:
Also, the ticket AC asks for some simple open/close animations on the category collapsers within the nav. Is that still a reasonable ask for this ticket?

sunnyzanchi:
had to cache clear and manually restart, here's the build https://build-477239d4-1b1c-41e3-95ca-2733e4b58ebf.gatsbyjs.io/

clarkmcadoo (approved):

LizBaker (approved):

PR #12428, opened 2023-03-31
adjust language before <InlineSignup /> component

erichsauer:
[https://issues.newrelic.com/browse/NR-99726](https://issues.newrelic.com/browse/NR-99726)

fixes several instances where installation steps are missing/unclear if the `<InlineSignup />` component is not rendering (ie when a user is already logged into a NR account)

removes a stray `]` from `src/content/docs/mobile-monitoring/new-relic-mobile/get-started/introduction-mobile-monitoring.mdx`


x8a (approved):
Thanks for the changes!

PR #12448, opened 2023-04-04
chore(Java): updating compatibility with Oracle R2DBC

meiao:
## Give us some context

Updating compatibility of the Java agent with Oracle R2DBC.


x8a (approved):
Thanks for the update!

PR #12457, opened 2023-04-04
fix: add times to `releaseDate`s to fix sort order

sunnyzanchi:


sunnyzanchi:
these dates can't include times, it causes the build to error

PR #12044, opened 2023-03-13
add persona homepage

sunnyzanchi:
this is a big PR, but most of it has already been reviewed. you should just need to review my last six commits

everything looks good from 760px on up, but the left sidebar kind of throws a wrench in things — things look not so great with it open at the smallest widths

PR #12468, opened 2023-04-04
Bring back video homepage layout

tabathadelane:
Questions:
* Should we also remove the [useMediaQuery hook](https://github.com/newrelic/docs-website/pull/12044/files?diff=split&w=0#diff-92800dca3c43d80c17a292f3b418f395e6382a5f729e5db5204450beabf66e15) that was introduced with the Persona Homepage? Or is this useful enough to keep around?
* Do we still want to keep the `forceLightMode` and `animated` options on [DocTile](https://github.com/newrelic/docs-website/pull/12044/files?diff=split&w=0#diff-b35fec885d56b0f64f82169c3441af9d4c2f564b5bd82d388db8a18dae663ca5) that we added for the Persona Homepage


sunnyzanchi (approved):
looks good! i think it's okay to keep the `light` DocTile variant and the `useMediaQuery` hook around, i can imagine them being useful outside of this context

src/components/ToggleView/ToggleSelector.js
```diff
@@ -19,6 +19,7 @@ export const TOGGLE_VIEWS = {
 const mobileBreakpoint = '450px';
 
 const ToggleSelector = ({ className, showTooltip }) => {
+  showTooltip = true;
```
LizBaker, line 22: should this be here? this means it pops up even if you're already on the alternate view

tabathadelane: nope 🤡

LizBaker (approved):

PR #12471, opened 2023-04-04
fix(.net install): more human readaable names

zuluecho9:


rhetoric101 (approved):
@zuluecho9 these filename changes look good!

PR #12487, opened 2023-04-05
chore: add language identifiers for syntax highlighting

brnhensley:
I also wrapped the rest of the command lines in backticks and made the API key/Account ID stand-ins consistent (removed the `<>` from one code block)

I'm not positive if docs.newrelic recognizes `powershell` code blocks, but we can add it if it's not yet recognized.


brnhensley:
It does work to use `powershell` as a lang identifier

image

rhetoric101 (approved):
@brnhensley yes, we already support powershell highlighting. This looks good!

PR #12494, opened 2023-04-06
fix(billing ingest): new data plus table

zuluecho9:
Liaison work: updating data plus info


rhetoric101 (approved):
@zuluecho9 I worked with PMM to finalize this today.

PR #12506, opened 2023-04-06
fix(Metric normalization): Updated old screenshot 

paperclypse:
based on customer feedback (NR-104845)


paperclypse:
This PR is too old. I need to close it and start over.

PR #12507, opened 2023-04-06
style(Solutions): Moved some text to make things slightly less confusing

paperclypse:
It didn't make sense to have "Otherwise, let's get you started:" as part of the side by side content.

Also, funny things happen to the video when my browser is too wide (it shrinks it down pretty small it bumps it up to the top right)
Screen Shot 2023-04-06 at 1 50 13 PM

This is based on customer feedback NR-104922


homelessbirds (approved):
Makes sense to me!

PR #12522, opened 2023-04-07
fix(Sonic): fixing tile paths using lab tile structure

jeff-colucci:


homelessbirds (dismissed):
It looks like it's just the links on the bottom? They work/look good to me :)

austin-schaefer (approved):

PR #12562, opened 2023-04-11
[Synthetics Job Manager] adds release notes for sjm and node-browser-runtime

kondracek-nr:
## Give us some context

Adds release notes for release-133 of Synthetics Job Manager and 1.1.108 of Synthetics-Node-Browser-Runtime.


rhetoric101 (approved):
@kondracek-nr thanks for adding these two release notes!

PR #12574, opened 2023-04-12
Update whats-new-03-27-querying-custom-infra-attributes-and-tags.md

aalfano7:
Removed duplicate text.


nbaenam (approved):

PR #12576, opened 2023-04-12
Improve MongoDB Atlas requirements

carlossscastro:
Clarify the Atlas users still need to use a hosted environment to run the agent_integration


nbaenam (approved):
Thank you for the update!! I've added a couple of modifications.

PR #12589, opened 2023-04-12
feat: Document GUI Configuration Option (Pathpoint)

JimHagan:
#### The current docs reference only the JSON file method of configuring pathpoint.  As of May 2022, Pathpoint has support GUI based editing of stages, steps, and touchpoints.  This could be much more detailed but it should be adequate in the short term.  There is a risk to interested users thinking they can only configure Pathpoint with a JSON file.
image


homelessbirds:
I just made a few changes! Looks good. Thanks @JimHagan

homelessbirds (approved):

PR #12592, opened 2023-04-12
Update scripted browser example

keegoid-nr:
- Target a docs page which doesn't change as frequently as the main page.
- Highlight steps to get started with New Relic.


nbaenam (approved):

PR #12619, opened 2023-04-14
Tabatha/quickstart chooser bugs

tabathadelane:
Just some cleanup in the APM intro doc.  A content update caused the text to overlap the absolute positioned svg arrow in the top QuickstartChooser, and the secondary layout wasn't' centering on large widths.


clarkmcadoo:
Screen Shot 2023-04-17 at 9 33 14 AM

we should probably look at making these smaller on mobile, but that's outside the scope of the bug work here.

clarkmcadoo (approved):

PR #12638, opened 2023-04-17
fix(QueryBuilder): Update PromQL link

unamaria:
Current link seems to be broken.


nbaenam (approved):

PR #12640, opened 2023-04-17
docs(logging): Fix sample grok parsing rule

danybmx:
## Give us some context

It fixes a GROK pattern example that is wrapped in double quotes, if you copy/paste that without removing them, it won't work. So removing them.

N/A

N/A


jeff-colucci (approved):
Looks good to me, assuming Gatsby build success we can push this out this morning!

PR #12652, opened 2023-04-17
include jpgs in webp conversion script

sunnyzanchi:
also:

- adds some error handling
- only checks for MD/JS in `src/` (previously running this with the `--global` flag would go through node_modules 😩)
- fixes a bug with the markdown references not being updated when running with the `--global` flag


tabathadelane (approved):
✨LGTM✨

clarkmcadoo (approved):

PR #12653, opened 2023-04-17
MobileBreadcrumb is now a New Relic event

ywang-nr:
* What problems does this PR solve? - Update docs, we have moved mobilebreadcrumb from custom events to New Relic events.


jeff-colucci (approved):
Looks good!

PR #12656, opened 2023-04-17
fix(Synthetics): fixing broken links

jeff-colucci:
fixing some broken links via 97471


paperclypse (approved):
Links all look good, thanks!

PR #12679, opened 2023-04-18
Persist nav open/close state

tabathadelane:


src/utils/navState.js
```diff
@@ -0,0 +1,15 @@
+export const isNavClosed = () => {
+  if (window.localStorage) {
```
sunnyzanchi, line 2: might be worth it to check if `typeof window !== undefined` here before accessing localStorage

sunnyzanchi (approved):
tested it out, looks good!

LizBaker (dismissed):
🎉

sunnyzanchi (approved):

PR #12682, opened 2023-04-19
fix cluster documentation config for rabbitmq integration

gsanchezgavier:
It clarifies how to configure the integration in a cluster environment.


nbaenam (approved):
hank you for the update!! It looks good to me!

PR #12683, opened 2023-04-19
migrated NPM lab from developer site

mehreentahir16:
This is a copy of the [Network Performance Monitoring Lab](https://developer.newrelic.com/collect-data/network-performance-monitoring) from the developer site, as part of the migration.


x8a (approved):
Looks great! Thanks!

PR #12684, opened 2023-04-19
rabbitmq queue size limitation (NR-107429)

gsanchezgavier:
The rabbitmq integration has a non documented limitation regarding the maximum amount of queues that can be collected. this PR documents that


nbaenam (approved):

PR #12697, opened 2023-04-19
Updated broken link ref.

ammartens:
Description:

This pull request fixes a broken link in the Start Using New Relic homepage. The original link was pointing to a page that no longer exists, so I updated it to the new URL. The new link directs users to the correct page with the relevant information.

Changes Made:

Updated the broken link to the new URL
Tested the link to ensure it is working as expected

Testing:

Clicked on the updated link to verify it directs to the correct page.


rhetoric101 (approved):
@ammartens thanks for correcting this link!

PR #12700, opened 2023-04-19
Fixup docs

hmstepanek:
## Clarify a couple common problem areas in the Python docs
* Add additional info and clarify causes of active application warning.
* Add in note about not being able to monitor the main process of celery.
* Clarify under what circumstance startup_timeout should be used.

### DIFF
--- a/src/content/docs/apm/agents/python-agent/troubleshooting/activate-application-warning-python.mdx
+++ b/src/content/docs/apm/agents/python-agent/troubleshooting/activate-application-warning-python.mdx
@@ -12,6 +12,9 @@ redirects:
 ---
 
 ## Problem
+<Callout variant="important">
+  Note this is only an issue if no data is being reported.
+</Callout>
 
 You are using the [latest Python agent](/docs/release-notes/agent-release-notes/python-release-notes), and you see a log entry with a warning message such as:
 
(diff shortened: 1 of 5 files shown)


src/content/docs/apm/agents/python-agent/troubleshooting/activate-application-warning-python.mdx
```diff
@@ -75,19 +78,20 @@ To solve this problem with your app:
...
+    import newrelic.agent
+    
+    # This will cause the agent to activate whenever custom_event is imported
+    app = newrelic.agent.application()
+    
+    def custom_event():
+        newrelic.agent.record_custom_event('CustomEvent', {}, application=app)
+    ```
+
+1. This issue can also occur when monitoring services like celery where a main parent process launches worker processes. It occurs when an agent application instance is created on the main process prior to forking the worker processes. Since the agent is launched in the parent process, no data is collected in the child process.
```
lrafeei, line 97: This number should be "2." instead of "1."  Or maybe we can use bullet points?  
< ul>
< li>first thing< /li>
< li>second thing< /li>
< /ul>

src/install/python/python-agent-non-web-apps.mdx
```diff
@@ -146,7 +146,7 @@ Due to the limitation on what Django management commands can be monitored, you n
 instrumentation.scripts.django_admin = syncdb sqlflush
 ```
 
-By default, the startup timeout is automatically specified to be 10.0 seconds when monitoring the Django management commands. If you need to override the startup timeout, you can set the `instrumentation.background_task.startup_timeout` setting within the same `import-hook:django` configuration section.
+By default, the startup timeout is automatically specified to be 10.0 seconds when monitoring the Django management commands. This means New Relic will block the command for up to 10s giving New Relic a chance to start up so it can monitor the command, otherwise it would not be able to start in time and would not be able to instrument the command. If you need to override the startup timeout, you can set the `instrumentation.background_task.startup_timeout` setting within the same `import-hook:django` configuration section.
```
lrafeei, line 149: "10 seconds" instead of "10s"

lrafeei, line 149: Replace the comma after "command" with a semicolon or just make it its own sentence

lrafeei (approved):
I'll pre-approve this, but here are some small changes that I think might make it easier to read

src/content/docs/apm/agents/python-agent/troubleshooting/activate-application-warning-python.mdx
nbaenam: Added number 2.

nbaenam (approved):
Thank you for the update!! Added some modifications

PR #12706, opened 2023-04-20
chore: Remove Process sample netio metrics

Mesverrum:
## Give us some context
Two attributes show in the processSample dictionary, but they do not show up in product at all and per dev conversation they should have been deleted long ago.
Internal Slack reference https://newrelic.slack.com/archives/CK58517B6/p1588722673340200


nbaenam (approved):
Thank you for the update!!

PR #12739, opened 2023-04-21
Add ingest guidance

rhetoric101:
Inform users about the typical impact of turning on change tracking.


nbaenam (approved):
I'd try to avoid the passive sentence and I'd write this: The Deployment namespace stores the change tracking activity.

PR #12750, opened 2023-04-24
Add error boundary

LizBaker:
This adds an error boundary component that can wrap our templates and js pages should there be a js error. it'll render a message in the UI and sent a browser event that we can use for alerts.

from [react's docs](https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary) "There is currently no way to write an error boundary as a function component". it's a little inconvenient, but since you can't use hooks in class components i'm passing the tessen call through

example event:
Screen Shot 2023-04-24 at 4 05 47 PM

UI:
Screen Shot 2023-04-24 at 3 44 46 PM

Screen Shot 2023-04-24 at 3 44 55 PM

### DIFF
--- a/src/components/ErrorBoundary.js
+++ b/src/components/ErrorBoundary.js
@@ -1,0 +1,28 @@
+/* eslint-disable react/sort-comp */
+import React from 'react';
+import ErrorMessage from './ErrorMessage';
+
+class ErrorBoundary extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { hasError: false };
+  }
+
+  static getDerivedStateFromError() {
+    return { hasError: true };
+  }
+
+  componentDidCatch() {
+    this.props.callTessen();
+  }
+
+  render() {
+    if (this.state.hasError) {
+      return <ErrorMessage />;
+    }
+
+    return this.props.children;
+  }
+}
+
+export default ErrorBoundary;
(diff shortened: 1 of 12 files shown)


tabathadelane (approved):
LGTM

clarkmcadoo (dismissed):

clarkmcadoo:
Do we want to include this on the homepage?

clarkmcadoo (dismissed):

src/components/ErrorBoundary.js
```diff
@@ -0,0 +1,28 @@
...
+
+  render() {
+    if (this.state.hasError) {
+      return <ErrorMessage />;
+    }
+
+    return this.props.children;
+  }
+}
+
```
sunnyzanchi, line 27: rather than have each call site pass a callback that calls Tessen, i think it might be easier if we export a wrapper component that passes in Tessen so the class component can call it itself. something like

```jsx
const ErrorBoundaryFC = ({ children, eventName }) => {
  const tessen = useTessen()

  return <ErrorBoundary eventName={eventName} tessen={tessen}>{children}</ErrorBoundary>
}

export default ErrorBoundaryFC

// and then in the class component

 componentDidCatch() {
    this.props.tessen.track({
      category: 'PageErrored',
      eventName: this.props.eventName
    });
  }

```

then when we use it we'd only have to do
```jsx
<ErrorBoundary eventName="attributeDictionary">...</ErrorBoundary>
```

sunnyzanchi (changes requested):

src/components/ErrorBoundary.js
LizBaker: ooh yeah this is nicer

sunnyzanchi (approved):
awesome, looks great!

PR #12794, opened 2023-04-26
Update EOL doc for 10.10.0 release

nr-ahemsath:
Update .NET EOL policy doc for 10.10.0 release.


nr-ahemsath:
Is there something holding this PR up?

PR #12835, opened 2023-04-28
fix(synthetics): add multiple clarifications

bpeckNR:
- Added clarifications that AMD64 / x86_64 processors are required for CPM and Job Manager installs
- Removed a duplicate and incorrectly named timestamp attribute from SyntheticRequest, adding SyntheticRequest to the event types of the timestamp attribute used for SyntheticCheck
- Clarified when sslCertificateExpiration attributes are created
- Fixed typos in the name of sslCertificateExpirationDaysRemaining
- Clarified the description of sslCertficiateExpiration related attributes


jeff-colucci (approved):
Looks good!

PR #12843, opened 2023-04-28
PUBLICATION: MAY 1; Add ruby agent 9.2.1 to agent eol page for upcoming release

tannalynn:
We're releasing 9.2.1 on monday, so this PR adds this versions dates to the EOL table for the ruby agent


rhetoric101 (approved):
@tannalynn thanks for updating the EOL doc!

PR #12861, opened 2023-05-01
chore: add secondary sort for releases with same day

tabathadelane:

## Give us some context


sunnyzanchi (approved):
there is an edge case here — the slug is a string so it's sorting in lexographic order, which means a case like this

Screen Shot 2023-05-01 at 2 17 12 PM
comes out like 
Screen Shot 2023-05-01 at 2 17 23 PM

but this is only disambiguating when two releases are on the same date, and i doubt that `v*.*.1` and `v*.*.10` will release the same day. so i think this is fine

LizBaker (approved):

PR #12874, opened 2023-05-02
[draft] Configs from custom command and http headers

josemore:
## Give us some context
In preparation for the next Infrastructure Agent release, we are introducing the following features:
* Ability to define custom http headers to use in all client requests to the backend.
* Ability to integrate a custom command to load configurations.

### DIFF
--- a/src/content/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings.mdx
+++ b/src/content/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings.mdx
@@ -6032,6 +6032,74 @@ If you are having problems with proxy configuration, see [Proxy troubleshooting]
       </tbody>
     </table>
   </Collapser>
+</Collapser>
+</CollapserGroup>
+
+## HTTP Client
+
+<CollapserGroup>
+  <Collapser
+    className="freq-link"
+    id="http-client-settings"
+    title="http_client_settings"
+  >
+  Configure the HTTP client used for backend communication.
+
+  <Collapser
+    className="freq-link"
+    id="http-headers"
+    title="http_headers"
+  >
+    Custom HTTP headers to include in all requests to the New Relic backend.
+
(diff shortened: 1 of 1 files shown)


src/content/docs/infrastructure/host-integrations/installation/secrets-management.mdx
```diff
@@ -743,3 +743,67 @@ variables:
...
+
+It can then be used in any configuration file using the `${myVariable.key}` notation.
+In the following example, a `my-custom-auth` command is executed with a `domain` parameter then loading the resulting `token` as an agent configuration: 
+
+```
+variables:
+  myToken:
+    command:
+      path: '/path/to/my-custom-auth-json'
+      args: ["domain", "myDomain"]
```
rubenruizdegauna, line 780: It's a bit confusing to me the `args` part as they look to me as "named parameters" instead of possitional ones. wdyt if we change it to something like:
 ```
args: ["--domain", "myDomain", "--other_param", "otherValue"]
```
or
 ```
args: ["arg1", "arg2"]
```
?

src/content/docs/infrastructure/host-integrations/installation/secrets-management.mdx
```diff
@@ -743,3 +743,67 @@ variables:
...
+# my-custom-auth output example: {"data":{"token":"XXXXYYYYZZZZ"}, "ttl": "24h"}
+# the `token` result can be included as ${myToken.token} in any configuration.
+  
+http:
+  headers: 
+    "Proxy-Authorization": ${myToken.token}
+```
+
+### String (text/plain)
+If the command output is a plain string, then it can be referenced in the configuration using the ${myVar} notation.
```
tabathadelane, line 792: this variable seems to be breaking the build

```suggestion
If the command output is a plain string, then it can be referenced in the configuration using the `${myVar}` notation.
```

x8a (approved):
Looks great! Thanks for the content!

PR #12879, opened 2023-05-02
fix: unify UI name of Issues & activity page

ghost:
Found inconsistencies with Issues & activity UI page. This just needs a quick hero review.


nbaenam (approved):
Good catch!

PR #12897, opened 2023-05-03
Update data-ingest-billing.mdx

anikoo-nr:
### Problem
The unit for counting the number of data points is not consistent 

### Solution
Convert all data point counting units to B.


zuluecho9:
@anikoo-nr i enquired about this and our marketing team uses 1T alongside 1B so this might be something you could talk with them about if you think it's not good. I'm going to close this for now but do let us know if you think there's more to talk about here. thank you!

PR #12903, opened 2023-05-03
TEST PR: DO NOT MERGE!

telehodown:
This is a test PR. DO NOT MERGE!

### DIFF
--- a/src/content/docs/errors-inbox/errors-inbox.mdx
+++ b/src/content/docs/errors-inbox/errors-inbox.mdx
@@ -89,8 +89,8 @@ While our managed rules are able to provide automatic error grouping based on a 
 
 For example,  consider these two error messages:
 
-* "Parsing Error: you put 124 and that is not allowed"
-* "Parsing Error: you put 456 and that is not allowed"
+* `Parsing Error: you put 124 and that is not allowed`
+* `Parsing Error: you put 456 and that is not allowed`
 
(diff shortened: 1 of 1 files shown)


src/content/docs/errors-inbox/errors-inbox.mdx
```diff
@@ -89,8 +89,8 @@ While our managed rules are able to provide automatic error grouping based on a
 
 For example,  consider these two error messages:
 
-* "Parsing Error: you put 124 and that is not allowed"
-* "Parsing Error: you put 456 and that is not allowed"
+* `Parsing Error: you put 124 and that is not allowed`
+* `Parsing Error: you put 456 and that is not allowed`
 
 These error messages look similar but these error occurrences are less likely to be grouped properly due to numbers that are shorter than 4 digits (124). If there is additional information (e.g. SQL statements, ids, etc) you would like to include in the error message, it can be sent via custom attributes.
```
nwhite222, line 95: @telehodown I looked at your fixes for lines 92 and 93 - looks good! I also suggested some small copy edits to line 95.
CURRENT:
These error messages look similar but these error occurrences are less likely to be grouped properly due to numbers that are shorter than 4 digits (124). If there is additional information (e.g. SQL statements, ids, etc) you would like to include in the error message, it can be sent via custom attributes.  
EDITED:
These error messages look similar but they are less likely to be grouped properly due to numbers that are shorter than four digits. If there is additional information you would like to include (for example, SQL statements, ids, etc) in the error message, it can be sent via custom attributes.

nwhite222 (commented):
@telehodown Thanks for your contribution! I inserted a couple of potentially helpful comments below.

rhetoric101:
This was a test PR. I'm finally closing it.

PR #12908, opened 2023-05-03
fix: Make cardinality() doc more clear and consistent

mcmellawatt:
`attribute` is confusing here when it really should be `metric_name`.  Also including all args in the top level example like the other functions in this list instead of a separate nested example.


bradleycamacho (approved):

zuluecho9:
thanks @mcmellawatt

PR #12938, opened 2023-05-05
fix: change page to path so link works

rudouglas:

## Give us some context

* The link for `New Relic's impact on your app's performance` was dead because the parameter in the `browser.yml` Nav file was `page` instead of `path`


nbaenam (approved):

PR #12949, opened 2023-05-05
fix: typo

keegoid-nr:
It should be called `ExecuteChangeSet`
https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_ExecuteChangeSet.html


jeff-colucci (approved):
Looks good!

jeff-colucci (approved):
Looks good!

PR #12986, opened 2023-05-09
feat: Adds Security to list of data ingest sources

colfortis:
* Adds absent Security section in "Understand and manage data ingest" page
* [NR-108868](https://issues.newrelic.com/browse/NR-108868)


x8a (approved):
Thanks!

PR #13009, opened 2023-05-11
infra-agent release notes 1.41.0 patch

josemore:
## Give us some context

* Added better explanation for features included in release 1.41.0 of the infra agent.


x8a (approved):
Thanks!

PR #13022, opened 2023-05-11
release notes action/job

sunnyzanchi:
just a simple job to run the upload script. i also added an `--upload` option to the script. omitting it means the script won't attempt to upload to S3 but just outputs to stdout

i used console.error, which outputs to stderr, for info messages so we can do useful stuff with the actual output


.github/workflows/agent-release-notes.yml
```diff
@@ -0,0 +1,17 @@
+name: Generate Agent Release Notes
+
+on:
+  push:
+    branches:
+      - master
```
tabathadelane, line 6: i think this should be `main`

tabathadelane (changes requested):

tabathadelane (approved):

PR #13039, opened 2023-05-12
Removed doc about deprecated feature/product

akristen:
* Received Jira feedback that the original doc had a dead link to Insights licenses
* Verified this is a deprecated product. Could not find a replacement page for Insights licenses
* Unclear if we're allowed/able to remove documentation about product. This PR deletes the doc entirely and removes the doc from our nav
* Need to confirm with @austin-schaefer that this is an OK doc to delete from our site.


akristen:
Been notified that we can't delete this doc yet!

PR #13041, opened 2023-05-12
fix: updated screenshots

pnvnd:
Updated screenshots with the correct entity names


akristen:
Thanks for this PR! 

Waiting to hear from eng about why our webp conversion script isn't working to convert these images for you. We're currently switching from png to webp files, so will need to make that change before we merge this PR!

akristen:
Confirmed with eng that we can merge these with pngs without ruining anything! Thanks

PR #13062, opened 2023-05-15
EKS Fargate: Add logging options

rhetoric101:
Add logging options since New Relic logging isn't yet supported on Fargate Nodes.


x8a (approved):
Looks good!

PR #13088, opened 2023-05-17
fix: remove out of context sign-up note

josemore:
## Give us some context

* There's a sign-up note that seems to be out of context in the Infra Agent upgrade documentation. Removed.


nbaenam (approved):

PR #13111, opened 2023-05-18
fix(Whats-new): Fixed title

nbaenam:
Fixed title: [PR - 13106](https://github.com/newrelic/docs-website/pull/13106)


x8a (dismissed):
Good catch! I think the problem was the symbol that was at the end of the string, and I don't think we want `&` in the docs.

nbaenam:
I merge it without a TW review because this PR must be published asap. I only changed the title and added the file extension

PR #13123, opened 2023-05-18
Update net-configure-logs-context-all.mdx

FlorentATo:
The use of _relative_ path leads to https://docs.newrelic.com/docs/logs/logs-context/net-configure-logs-context-all/docs/apm/agents/net-agent/attributes/enable-disable-attributes-net/#attruls/, which is wrong.


jeff-colucci (approved):
Good catch! Looks good to me

FlorentATo:
Thanks @jeff-colucci :) The issue appears to be much broader and affects several pages. I'll open another PR.

PR #13145, opened 2023-05-19
Update .NET Performance Metrics documentation to point to new UI page

nr-ahemsath:
We finally have a real APM UI page for .NET CLR (VM) performance metrics.  This PR updates the documentation about these metrics to point users to the new UI.


rhetoric101:
Hey docs hero, @nr-ahemsath is passing the work for this to their colleagues next week. Please keep tabs on this. The gist is that this doc about .NET performance metrics has been out in the wild for a while, but the location of where users find these metrics is changing to the Dotnet VMs page. See the Slack thread about this in help-documentation.

x8a (approved):
Thanks a bunch for the edits!

PR #13163, opened 2023-05-22
OpenTelemetry: Make "Collector" upper case when it follows "OpenTelemetry."

rhetoric101:
To conform with the OpenTelemetry spec, we needed to make "Collector" upper case when it follows "OpenTelemetry" because this is the full name. When collector appears by itself, we can use lower case.


bradleycamacho (approved):

rhetoric101:
@bradleycamacho thanks for the review!

PR #13176, opened 2023-05-23
style(agents): fixed typo

x8a:


x8a:
I thought it'd be better to break the sentence in two. What do you think?

nbaenam (approved):
Yes, good idea!!

PR #13180, opened 2023-05-23
docs: add details about .NET AddCustomAttributes parameters

brnhensley:
I'm not sure if this will look good (the table in the table), but the important details about these API methods is in a separate doc (https://docs.newrelic.com/docs/apm/agents/net-agent/attributes/custom-attributes-net/)

I'm not sure why, but I feel this info should be front in center, nobody (including support) knows to look for these details in another section of the .NET docs


brnhensley:
fixed conflict

nbaenam:
The build has been triggered by a [manual rebuild](https://build-72a8ea44-29d1-48aa-a601-271f49e85a50.gatsbyjs.io/)

PR #13193, opened 2023-05-24
fix homepage tooltip cut off, add usePosition hook

sunnyzanchi:
this won't update the Tooltip's position if the page is resized, but since it only shows briefly i think this is okay


tabathadelane (dismissed):
✨

clarkmcadoo:
Is there a good page to QA this? @sunnyzanchi or @tabathadelane

tabathadelane:
> Is there a good page to QA this? @sunnyzanchi or @tabathadelane

the homepage if you load it after selecting the default view

clarkmcadoo (dismissed):

clarkmcadoo (approved):

PR #13195, opened 2023-05-24
Update snmp-profiles.mdx

Mesverrum:
## Give us some context
Product direction has shifted to where customers should expect to make changes to profiles via the self service method instead of the previous issue based process.


akristen (approved):
Looks good!

PR #13205, opened 2023-05-25
Fix code example rendering

bradleycamacho:
Fixes a very annoying and hard to fix bug with some example code in the following doc: https://docswebsitedevelop-bradleycamachopatch17.gatsbyjs.io/docs/journey-demand/autoscale-your-infra/


akristen (approved):
THANK YOU!

PR #13207, opened 2023-05-25
Update compatibility-requirements-python-agent.mdx

sjyothi54:
### Description:
Updated the text from "Aiohttp" to "AIOHTTP"  as per the official suggestion provided on the AIOHTTP website.


x8a (approved):
They use both `AIOHTTP` and `Aiohttp`, but I'm ok with using the former to refer to the framework name.

PR #11389, opened 2023-02-06
docs: Change curl command destination

kenzoarima:
Please follow [conventional commit standards](https://www.conventionalcommits.org/en/v1.0.0/)
in your commit messages and pull request title.

## Give us some context

* Running the curl command with `https://api.newrelic.com/graphiql` does not work.
* Changing from `https://api.newrelic.com/graphiql` to `https://api.newrelic.com/graphql` works. Verified on my own terminal.

PR #11176, opened 2023-01-24
fix(Apis): updated urls

nbaenam:
Updated these URLs:

- From: https://api.newrelic.com/graphql to https://api.newrelic.com/graphiql
- From: https://api.eu.newrelic.com/graphql to https://api.eu.newrelic.com/graphiql

PR #13243, opened 2023-05-26
Fix NG endpoint typo

stellanor:
Hi! I'm a New Relic engineer on Unified API. 

PR #11176 was an incorrect change. `/graphql` is the API endpoint and `/graphiql` is the API explorer user interface. 

I see that #11389 partially reverted it, but this endpoint was still wrong. We had a customer report the issue. Feel free to reach out to me internally (epoley on Slack) if needed to discuss this further.


rhetoric101 (approved):
@stellanor sorry for this mistake! Thanks for correcting these URLs!

PR #13252, opened 2023-05-29
Update magento-business-insights.mdx

sjyothi54:
Ticket URL: https://issues.newrelic.com/browse/NR-121265
Description: The restart step is inside step 8 of 9.
This should be new step.


nbaenam (approved):
Thank you for the update! I've added a minor modification.

PR #13270, opened 2023-05-30
Clarifying workflow incident_id behavior

dneth:
The doc suggests the incident_id will be available far more often than it actually will.  This change will help to inform consumers of the field what they can expect post-migration.


rhetoric101:
This was stalled, and I think it was related to an issue we found in develop earlier today. I just updated with develop.

rhetoric101 (approved):
@dneth thanks for adding this clarification. I only made a minor edit.

dneth:
Thanks @rhetoric101!

PR #13332, opened 2023-06-02
Remove Helm 2 section, as it is no longer supported

jsubirat:
Removed the Helm 2 section, since Helm 2 is not supported anymore.


nbaenam (approved):

PR #13340, opened 2023-06-02
Remove Exception on AWS Metric Streams

shredtechular:
The AWS Metric Streams is not listed on the Services Not In Scope on this page (https://docs.newrelic.com/docs/security/security-privacy/compliance/certificates-standards-regulations/fedramp-moderate/#not-scope) anymore, so I think we should remove it on the FEDRAMP Compliant Endpoint page too.  This PR removes that "EXCEPTION" verbiage to be consistent across the documentation.


akristen (approved):
Looks good! (For record keeping, this removal is part of this change: https://github.com/newrelic/docs-website/pull/13343)

PR #13355, opened 2023-06-05
NR-122412 added a collapser for NRFeatureFlag_GestureInstrumentation

mbruin-NR:
There currently isn't any mention of gesture instrumentation in the docs website, this pr adds that and mentions that it is deprecated and may cause instability. 
https://issues.newrelic.com/browse/NR-122412


nbaenam (approved):
Thank you for the update!! It looks good to me! I've removed the modifications in the Japanese and Korean files because there is a team in charge of those modifications.

PR #13358, opened 2023-06-05
fix: Remove #config-format

nwhite222:
Removed a nonexistent anchor tag link in the 'Update older integration configuration' section of the page. Jira: NR-87663


paperclypse (approved):
Looks good. Thanks! (Sorry, didn't notice this review request right away)

PR #13424, opened 2023-06-07
Add user journey buttons

LizBaker:
This component is used by adding props for the previous and next steps like so: 

```
<UserJourneyControls
  nextStep={{path: "/", title: "Next Step", body: "do this third"}}
  previousStep={{path: "/", title: "Previous step", body: "do this first"}}
/>
```
I added a custom serializer/deserializer so we can get the props translated

https://github.com/newrelic/docs-website/assets/39655074/d5c90068-84a7-4edd-ac58-1088fdc19523

### DIFF
--- a/scripts/actions/utils/serialization-helpers.js
+++ b/scripts/actions/utils/serialization-helpers.js
@@ -14,6 +14,7 @@ const removeParagraphs = () => (tree) => {
   });
 };
 
+// this converts the string version of props to a json string so we can use JSON.parse on it
 const createJsonStr = (str) =>
   str
     .replace(/(\w+:)|(\w+ :)/g, function (matchedStr) {
--- a/scripts/actions/utils/serialization-helpers.js
+++ b/scripts/actions/utils/serialization-helpers.js
@@ -14,6 +14,13 @@ const removeParagraphs = () => (tree) => {
   });
 };
 
+const createJsonStr = (str) =>
+  str
+    .replace(/(\w+:)|(\w+ :)/g, function (matchedStr) {
+      return `"${matchedStr.substring(0, matchedStr.length - 1)}":`;
+    })
+    .replace(/'/g, '"');
+
 const attributeProcessor = unified()
   .use(toMDAST)
   .use(remarkMdx)
@@ -109,8 +116,10 @@ const stripNulls = (obj) =>
   Object.fromEntries(Object.entries(obj).filter(([, value]) => value != null));
 
 module.exports = {
+  createJsonStr,
   serializeComponent,
   serializeProps,
   serializeTextProp,
   serializeJSValue,
+  serializeAttributeValue,
 };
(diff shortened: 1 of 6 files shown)


src/components/UserJourneyControls.js
```diff
@@ -0,0 +1,142 @@
...
+          <animated.div style={leftButton}>
+            <Icon
+              name="fe-arrow-left"
+              size="3rem"
+              css={css`
+                color: var(--system-text-primary);
+                margin-right: 2rem;
+                .doc-tiles-labs & {
+                  display: none;
+                }
```
sunnyzanchi, line 50: is this needed since this is separate from the DocTiles and doesn't have variants? same on line 94

scripts/actions/utils/serialization-helpers.js
```diff
@@ -14,6 +14,13 @@ const removeParagraphs = () => (tree) => {
   });
 };
 
+const createJsonStr = (str) =>
```
sunnyzanchi, line 17: could you add a doc comment explaining what this does/is used for? i think it turns a string from something like `{ something: 'hey' }` into `{ "something": "hey" }`? but it'd be good to be clear about it and why it's needed. is it because MDX doesn't automatically recognize JS objects in props?

tabathadelane:
for some reason, my buttons are rendering like this with the right arrow aligned with the text instead of the right edge of the box

Screenshot 2023-06-08 at 10 13 58 AM

src/components/UserJourneyControls.js
```diff
@@ -0,0 +1,142 @@
...
+        }}
+        className={className}
+        css={css`
+          border: solid 2px var(--brand-button-primary-accent);
+          &:hover {
+            color: var(--primary-text-color);
+            box-shadow: -2px -2px var(--brand-button-primary-accent),
+              2px -2px var(--brand-button-primary-accent),
+              -2px 2px var(--brand-button-primary-accent),
+              2px 2px var(--brand-button-primary-accent);
```
sunnyzanchi, line 77: currently this creates the thicker border on hover which looks nice, but it also removes the existing box shadow which looks a little off

```suggestion
            box-shadow: -2px -2px var(--brand-button-primary-accent),
              2px -2px var(--brand-button-primary-accent),
              -2px 2px var(--brand-button-primary-accent),
              2px 2px var(--brand-button-primary-accent),
              var(--shadow-3);
```

sunnyzanchi (changes requested):
looks great! just left a few comments

scripts/actions/utils/serialization-helpers.js
LizBaker: ah yeah, the serializer has access to the props as a string so "{key: 'value'}" but to convert it to an object via JSON.parse it needs to be a json string

src/components/UserJourneyControls.js
LizBaker: eep i didn't even notice

sunnyzanchi (approved):
brilliant ✨

tabathadelane:
I'm not sure how much we are prioritizing small screens, but the buttons stop being responsive < 450px. It doesn't stretch the layout though.
Screenshot 2023-06-09 at 12 06 28 PM

clarkmcadoo (approved):

tabathadelane (approved):

PR #13438, opened 2023-06-08
Update java agent config to document FedRAMP ingest endpoints

jasonjkeller:


rhetoric101 (approved):
@jasonjkeller thanks for adding this FedRamp information!

PR #13439, opened 2023-06-08
.NET Agent: update supported versions of instrumented libraries

nr-ahemsath:
The .NET agent team has successfully tested some newer versions of libraries we instrument and want to update our compatibility and support documentation accordingly.


chynesNR (approved):

rhetoric101 (approved):
@nr-ahemsath thanks for updating these versions!

PR #13444, opened 2023-06-08
fix(amazon sqs monitoring): capitalization

nwhite222:
NR-97779 - Fixes a missed capitalization in the "ApproximateNumberOfMessagesVisible" query


src/content/docs/infrastructure/amazon-integrations/aws-integrations-list/amazon-sqs-monitoring-fix
```diff
@@ -94,7 +94,7 @@ This integration collects the following Amazon SQS metrics. For full metric desc
 
     <tr>
       <td>
-        `approximateNumberOfMessagesVisible`
```
zuluecho9, line 97: why is the first letter capitalized; that isn't our usual standard and none of the other ones in here follow that pattern

zuluecho9 (approved):

PR #13445, opened 2023-06-08
docs: Remove unused core dependencies from iOS app

jvanmetre:
Update licenses used by the iOS Mobile App.


rhetoric101 (approved):
@jvanmetre thanks for making these license updates!

PR #13453, opened 2023-06-09
chore: explain how to disable container monitoring (ContainerSample)

josemore:
## Give us some context

* Adding a note explaining how to disable this capability since it's a FAQ in our support channels.


nbaenam (approved):

PR #13462, opened 2023-06-09
feat: Create comment component for markdown pages

tabathadelane:
in place of comments, this is a component that writers can use to leave important notes about content

```
<CONTRIBUTOR_NOTE>
  This is a note to future authors about the MDX content.

  It does not render in the UI
</CONTRIBUTOR_NOTE>
```


src/components/MDXContainer.js
```diff
@@ -90,6 +91,7 @@ const defaultComponents = {
   ButtonLink: (props) => <MDXButton as={Link} {...props} />,
   ButtonGroup: MDXButtonGroup,
   DoNotTranslate: ({ children }) => <>{children}</>,
+  ContributorNote,
```
sunnyzanchi, line 94: i think we may want to declare it like this

```suggestion
  CONTRIBUTOR_NOTE: ContributorNote,
```

src/components/ContributorNote.js
```diff
@@ -0,0 +1,20 @@
...
+ */
+const ContributorNote = ({ children }) => <NoteStyles>{children}</NoteStyles>;
+
+ContributorNote.propTypes = {
+  children: PropTypes.node,
+};
+
+const NoteStyles = styled.div`
+  display: none;
+`;
```
sunnyzanchi, line 18: i think we can just make this an empty component and it won't output anything, but it can still be used in MDX

```suggestion
const ContributorNote = () => null;
```

sunnyzanchi (changes requested):

sunnyzanchi (approved):

LizBaker (approved):

PR #13469, opened 2023-06-09
[chore] Update upgrading-kubernetes.mdx

maxlemieux:
Standardize release name to match guided install, as `newrelic-bundle`

The release name must be used when referring to a Helm chart in the CLI. If the release name doesn't match the initial installation, it could cause confusion.

Other docs seem to also use `newrelic-bundle`, agreeing with the guided install.


nbaenam (approved):

PR #13487, opened 2023-06-13
feat(NRQL): Updated the facet order by clause

nbaenam:
Updated the `facet...order by` clause to specify the `ASC` and `DESC` modifiers can't be used.

Request added in the help-documentation channel.


akristen (approved):
Looks good!

PR #13494, opened 2023-06-13
fix: Update incident engagement KPI queries

JimHagan:
* Fixed the incident engagement queries for AQM.  They were still using the custom AQM incident event.
* Changed wording in the migration section

Need to get reviews form the OMA team.

@adwetzelnr @khickey-newrelic @AlecIsaacson


AlecIsaacson (approved):
Changes look ok

JimHagan:
@akristen We have an SME `okay` so we can merge when ready.

nbaenam (approved):
Thank you for the update! I've added some minor modifications.

PR #13503, opened 2023-06-14
Note about configs for azure vm extension

josemore:
## Give us some context

* Expanding the tip section to explain that customers can optionally customize the agent default settings via environment variables.


nbaenam (approved):
Thank you for the update! I've added some minor modifications.

PR #13532, opened 2023-06-15
Update nested-aggregation-make-ordered-computations-single-query.mdx

kinseypdx:
The first codeblock shows syntax with `integer units` as a way to indicate that it should be a number and unit of time. 

The following code blocks after are examples... therefore should use an actual integer and unit. One of these is already doing this correctly with Transaction, but this CPU one needs to also have a real integer and unit.

## Give us some context


nbaenam (approved):

PR #13593, opened 2023-06-20
Update java-agent-attributes.mdx

jasonjkeller:
Clarify how Java agent attribute system works with request headers.


akristen (approved):
Good clarifying updates. Thanks for the contribution!

PR #13597, opened 2023-06-20
Update high traffic images

nwhite222:
Updated images in high traffic DT sections, mostly to reflect preference for dark mode. NR-119134


src/images/scatter-plot-hover.webp
x8a, line 1: I think we're not supposed to remove images from the repo. The thing is that the translated files still have reference to the old image, so until the translation script runs to make the imgs updates, the builds will fail. I'd add the image back.

nwhite222: Added it back. Thanks!

x8a:
The new images look great! 💃🏻

I have a question about the images inside the collapsers though. I see they're pretty outdated and in light mode, was the plan to only update the ones outside the collapsers?

x8a (commented):
The `distributedtracingFragmentedTrace` screenshot is still in light mode. Do you think we can retake it as well?

x8a (approved):
> The `distributedtracingFragmentedTrace` screenshot is still in light mode. Do you think we can retake it as well?

Natasha confirmed that she'll come back to this screenshot in a future PR, since she couldn't retake it easily.

PR #13602, opened 2023-06-20
chore: adding notes on CIDR sizing recommendations

thezackm:
adding some notes to better explain issues with discovery job timeouts using the network monitoring agent.


x8a (approved):
Thanks for the edits, as always!

PR #13639, opened 2023-06-22
fix(Sonic): alerts iterations (121952, 127899)

jeff-colucci:


jeff-colucci:
Build succeeded on manual rebuild [here](https://www.gatsbyjs.com/dashboard/34e708d2-68a1-48ca-b66f-d2604890b061/sites/a0b77314-cdea-4790-8fde-7d3457d590ce/pull-requests/9da0fe49-6505-411f-be7c-c4282acdf6c2/builds)

src/content/docs/tutorial-create-alerts/create-an-alert.mdx
```diff
@@ -13,37 +17,148 @@ import alertsThresholds from 'images/alerts_screenshot-crop_set-thresholds.webp'
...
+<SideBySide>
+    <Side>
+Starting from a chart is one of the easiest ways to set up a New Relic alert condition. Instead of creating a NRQL query from scratch, charts let you use pre-existing NRQL queries to begin the process. 
+    
+In this example, we'll set up an alert condition using an error response chart. Follow the steps below with any type of chart you like to begin.
+    </Side>
+    <Side>
+    <img
+    title="Create chart alert"
+    alt="A screenshot displaying where to create an alert condition from a chart in New Relic"
```
akristen, line 62: the way these images are stacked sets up the user to think that the right screenshot (clicking the `...` to get the expanded menu) corresponds with the left text, but the first screenshot goes with step 1 of the succeeding paragraph. I think this would be clearer if you described the steps then put the text in, and maybe removed the side by side.

akristen: Alternatively, moving step 1 of the stepped instructions to be step one in the first part of the component. something to that effect

src/content/docs/tutorial-create-alerts/create-an-alert.mdx
```diff
@@ -13,37 +17,148 @@ import alertsThresholds from 'images/alerts_screenshot-crop_set-thresholds.webp'
...
     src={alertsSignal}
-/>
+    />
+    </Side>
+    <Side>
+1. Go to any dashboard. Then, find a chart you want to use, click the "**...**" menu, and select **Create alert condition**.
+
+2. Select the signal, or stream of data, you want to use to create the alert condition. To keep things simple, **Build a query** is the default option for conditions created this way.
+
+3. Modify the NRQL query to match what you want you to monitor, if necessary. For now, we recommend leaving the query as is, but once you have a better handle on alerts and NRQL queries, you can change the query to monitor nearly any signal.
```
akristen, line 80: this is good information. do you think there's space to describe what the NRQL query translates into in English terms?

src/content/docs/tutorial-create-alerts/create-an-alert.mdx
```diff
@@ -13,37 +17,148 @@ import alertsThresholds from 'images/alerts_screenshot-crop_set-thresholds.webp'
...
-    <Step>
+
+<Callout variant="tip">
+NRQL stands for New Relic's Query Language—a standardized query method that powers all our charts, and which you can use to analyze and query your data. If this is your first time working with NRQL, you may want to check out our [NRQL docs](/docs/query-your-data/nrql-new-relic-query-language/get-started/introduction-nrql-new-relics-query-language/) to learn more about NRQL, including [how it works](/docs/query-your-data/nrql-new-relic-query-language/get-started/introduction-nrql-how-nrql-works/) and an [introductory tutorial](/docs/query-your-data/nrql-new-relic-query-language/get-started/introduction-nrql-tutorial/).
+</Callout>
+        </TabsPageItem>
+        <TabsPageItem id="2">
+<SideBySide>
+    <Side>
+One of the easiest ways to get started is to use our **Golden signal or metric** option. When selected, this capability gives you a guided creation path for making a condition. All you have to do is select the options you want and let us do the rest!
```
akristen, line 114: Could we make this the first tab that users immediately see? If it's the easiest way, then we should probably default it so they don't miss the hack

src/content/docs/tutorial-create-alerts/create-an-alert.mdx
```diff
@@ -13,37 +17,148 @@ import alertsThresholds from 'images/alerts_screenshot-crop_set-thresholds.webp'
...
+<SideBySide>
+    <Side>
+<img
+    title="golden metric entities"
+    alt="A screenshot displaying the entities selection screen for golden metrics alerts"
+    src={alertsGoldenEntities}
+/>
+    </Side>
+    <Side>
+1. From **one.newrelic.com**, select [**Alerts & AI**](https://one.newrelic.com/alerts-ai), then [**Alert conditions (policies)**](https://one.newrelic.com/alerts-ai/alerts-classic/policies).
```
akristen, line 135: IMO i would move steps 1-3 to the first chunk of text left of the screenshot. then i would make step 4 in the subsequent text section a new step 1. LMK if you have questions/this isn't clear

akristen: Alternatively, you can make steps 1-3 into more exposition style description. That will fill the empty text in this first text description and eliminate some awkwardness from the next side by side where the text is longer than the image itself

akristen (approved):
Good addition! Left some non urgent comments -- the tabs work and the images are appearing in the deploy. i don't have any content feedback on the sentence level, but made some suggestions about info architecture. If you apply those changes, LMK and I can approve again!

src/content/docs/tutorial-create-alerts/create-an-alert.mdx
jeff-colucci: Hmm, good point. I thought about how best to fix it, I think that moving step 1 up to the left side is the simplest solution and also the most direct in terms of keeping the information to it's implied paired screenshot. It also fills out the white space nicely. See what you think.

src/content/docs/tutorial-create-alerts/create-an-alert.mdx
jeff-colucci: I moved the callout to earlier in the procedure to provide more context for NRQL and present opportunities for learning about it to before the reader commits to the steps in the procedure. Thanks for the feedback!

src/content/docs/tutorial-create-alerts/create-an-alert.mdx
jeff-colucci: I spoke about this with the SME, we came to the conclusion that charts are the most common way that people make alerts, so we settled on this order. I'd say let's test it out and see and in a couple of months when the UI gets it's revamp we can revisit and make any necessary changes since we'll be making changes to the page anyway.

src/content/docs/tutorial-create-alerts/create-an-alert.mdx
jeff-colucci: Good call, I changed the formatting of the side by side for this section and it should be more aesthetically pleasing and take up less room. Thanks very much!

akristen (approved):
I think this looks good! I'm going back/forth on the back to back images (even though I recommended that change). I think visually it's still a bit dense, but from a customer POV it's probably more helpful and clearer what's going on and how they should follow those steps.

PR #13648, opened 2023-06-22
chore: add language identifiers to heroku docs

brnhensley:
I also fixed some formatting and put code/filenames in backticks instead of BOLD, which is more consistent with the rest of our docs


nbaenam (approved):

PR #13684, opened 2023-06-26
chore: added SLES note with alternative command

josemore:
## Give us some context

* Added a note explaining the alternative way to import GPG key on SLES to solve some compatibility issues in older versions.


nbaenam (approved):
Thanks for adding this! It looks good to me!

PR #13685, opened 2023-06-26
Correct typo in if() function example

crshanks:
Fixes a typo in the if() statement nrql example: [Use with AND and OR](https://docs.newrelic.com/docs/query-your-data/nrql-new-relic-query-language/get-started/nrql-syntax-clauses-functions/#func-if).

Replace non-standard apostrophes: ’ with '.  This allows the query to run without error.


nbaenam (approved):

PR #13687, opened 2023-06-26
feat(Dictionary): Added NrAiIssue

nbaenam:
Because of a request in the help-documentation channel, I've added `NrAiIssue` in the dictionary.


src/data-dictionary/events/NrAiIssue/activaeTime.md
```diff
@@ -0,0 +1,8 @@
+---
+name: activaeTime
```
zuluecho9, line 2:
```suggestion
name: activateTime
```

src/data-dictionary/events/NrAiIssue/closeCause.md
```diff
@@ -0,0 +1,42 @@
+---
+name: closeCause
+type: attribute
+events:
+  - NrAiIssue
+---
+
+If applicable, what caused the incident to close.
```
zuluecho9, line 8: we tend to avoid having very long attribute definitions like this. I recommend we create a doc for this info and not have it in the attribute.

src/data-dictionary/events/NrAiIssue/correletionRuleIds.md
```diff
@@ -0,0 +1,9 @@
+---
+name: correletionRuleIds
```
zuluecho9, line 2:
```suggestion
name: correlationRuleIds
```

src/data-dictionary/events/NrAiIssue/correletionRuleNames.md
```diff
@@ -0,0 +1,8 @@
+---
+name: correletionRuleNames
```
zuluecho9, line 2:
```suggestion
name: correlationRuleNames
```

src/data-dictionary/events/NrAiIssue/lastModifiedTime.md
```diff
@@ -0,0 +1,9 @@
+---
+name: lastModifiedTime
+type: attribute
+units: seconds (s)
```
zuluecho9, line 4: is this really a unit of 'seconds'? If it's a timestamp it wouldn't be that format. e.g., createTime isn't 'seconds'

src/data-dictionary/events/NrAiIssue/sources.md
```diff
@@ -0,0 +1,8 @@
+---
+name: sources
+type: attribute
+events:
+  - NrAiIssue
+---
+
+If applicable, the source of the issue.
```
zuluecho9, line 8: i think as a general point, we can avoid/erase 'if applicable' or 'present when...' types of notes. Reason being: many of our attributes are only present when x/y/z things are true, so it can go without saying that they may or may not be present. This is just a general tip and I think we talk about this and other aspects in the data dictionary style guide.

src/data-dictionary/events/NrAiIncident/closeCause.md
```diff
@@ -5,21 +5,28 @@ events:
   - NrAiIncident
 ---
 
-If applicable, what caused the incident to close.
-
 Available values: 
+
```
zuluecho9, line 9: left similar note on an event: we want to avoid these long attribute definitions: this seems like something we should have in the docs and not here.

src/data-dictionary/events/NrAiIssue/sources.md
nbaenam: ok, there are several "If applicable"

src/data-dictionary/events/NrAiIssue/lastModifiedTime.md
nbaenam: It's milliseconds ;-)

src/data-dictionary/events/NrAiIssue/closeCause.md
nbaenam: OK, I added like that because I saw it in other attributes. I've moved the info to the [Incident event attributes](https://docs.newrelic.com/docs/alerts-applied-intelligence/new-relic-alerts/advanced-alerts/understand-technical-concepts/violation-event-attributes/)

nbaenam:
Triggered by a manual build: https://build-17dbfef4-71c1-4f72-834b-14b78de89c6b.gatsbyjs.io/

PR #13708, opened 2023-06-26
Clark/attribute dictionary json action

clarkmcadoo:
This creates the workflow and corresponding script to update the docs site with an up to date version of the attribute dictionary that will be held in GHE


.github/workflows/update-attribute-dictionary-json.yml
```diff
@@ -0,0 +1,58 @@
+name: Update Attribute Dictionary JSON
+
+on:
+  workflow_dispatch:
+  schedule:
+    # 12pm every day
+    - cron: '0 12 * * *'
```
LizBaker, line 7: i think this is UTC so it's technically 5am pdt i think? not that the time even matters but we might wanna update the comment

LizBaker (approved):

LizBaker (approved):

PR #13733, opened 2023-06-28
fix(AWS): Fixed incorrect AWS steps

paperclypse:
This is why we don't love documenting external systems.

Related to NR-138896


akristen (approved):
Change itself is good!

PR #13742, opened 2023-06-29
fix(tags): clarify cloud/host reporting

zuluecho9:
Attempt to make tags doc more helpful about how we report tags for hosts, including cloud environments. This first attempt is very rough and likely wrong, as it's just attempt by @zuluecho9 to clarify it a bit more based on his understanding. Comments/feedback/additions very much appreciated.


nbaenam (approved):
LGTM!

PR #13749, opened 2023-06-29
.NET Agent: add exception for Alpine Linux under ARM64 support statement

nr-ahemsath:
We've discovered the the ARM64 version of our .NET agent does not work on Alpine Linux.  This PR updates our documentation with that information.


nbaenam (approved):
Thank you for the update! I've added some minor TW edits

PR #13753, opened 2023-06-29
fix(workloads): matching UI

zuluecho9:
NR-139000

Work done: 
Attempted to get docs to match the workloads UI. 
Replaced some screenshots
Deleted some stuff about 'Header'; seemed to be talking about some rather obvious basic stuff
Moved a thing about adjusting the 'workloads view' to the 'workloads UI' as it seemed like something we should promote more. 
Improved unclear or too-much-detail wording as I saw it.


akristen (approved):
I made some changes to the initial Workloads doc in terms of structure. I wanted it to look simpler, so added bullet points and collapsers and eliminated some H3s and H2s.

PR #13763, opened 2023-06-30

bhecknr:
Entry under "Name" was listed as "New Relic Query Language"; it should be "New Relic query language" (lowercase "query language").


rhetoric101 (approved):
@bhecknr thanks for the update!

PR #13802, opened 2023-07-11
New attribute dictionary page/data

LizBaker:
This removes all the event/attribute definitions (now living in our service in GHE) and adds a workflow to bring in updates from our service. these updates are stored as json and read by our original (updated) data dictionary page


sunnyzanchi (approved):
🤩

PR #13803, opened 2023-07-11
Update browser-javascript-injection-causes-problems-page.mdx

heatherzurek:
Correcting page path to disable browser monitoring from incorrect path. More info below.

## Give us some context

On the Browser "JavaScript injection causes problems with a page" doc page, step 1 is to disable browser monitoring, but when you click the link, it goes to Browser settings: UI options for Apdex and geography, when it should go to Disable browser monitoring. I corrected this link.


rhetoric101 (approved):
@heatherzurek thanks for correcting this link!

PR #13808, opened 2023-07-12
Add new metrics to nri-postgresql

gsanchezgavier:
Add new metrics to the nri-postgresql integration. 

The new version has been released so the PR is ready to merge after review/approval.


nbaenam (approved):
Thanks for adding this! It looks good to me!

PR #13838, opened 2023-07-13
Clark/attribute dictionary fix

clarkmcadoo:
- updates API url to hopefully match our secret key stored in GH
- update the json update script to have better error handling
- add eng label to automatically generated PR
- remove unused script


.github/workflows/update-attribute-dictionary-json.yml
```diff
@@ -56,3 +56,5 @@ jobs:
           base: develop
           delete-branch: true
           team-reviewers: docs-engineering
+          lables: |
+            eng
```
roadlittledawn, line 60: i think we want content folks to review this no?

roadlittledawn: i don't think we need to be involved unless there's a failure. similar to i18n PRs

roadlittledawn (changes requested):
can add `content` label and maybe ping tech docs to let them know

roadlittledawn (approved):
LGTM. i wonder if we may want to eventually change this to hit prod nerdgraph endpoint to ensure the site is only getting content that is live in prod.

PR #13844, opened 2023-07-13
fix(attribute-dictionary): apply alpha sort to attribute tables

roadlittledawn:
## Description

Sort each event's table by attribute name in alph order.

## Screenshot(s)
**Before**

**After**

### DIFF
--- a/src/pages/attribute-dictionary.js
+++ b/src/pages/attribute-dictionary.js
@@ -189,15 +189,35 @@ const EventDefinition = memo(
     let filteredAttributes = [];
 
     if (searchedAttribute) {
-      filteredAttributes = event.attributes.filter(({ name }) =>
-        name.toLowerCase().includes(searchedAttribute.toLowerCase())
-      );
+      filteredAttributes = event.attributes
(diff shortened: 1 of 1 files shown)


src/pages/attribute-dictionary.js
```diff
@@ -189,15 +189,35 @@ const EventDefinition = memo(
...
 
     if (searchedAttribute) {
-      filteredAttributes = event.attributes.filter(({ name }) =>
-        name.toLowerCase().includes(searchedAttribute.toLowerCase())
-      );
+      filteredAttributes = event.attributes
+        .filter(({ name }) =>
+          name.toLowerCase().includes(searchedAttribute.toLowerCase())
+        )
+        .sort((a, b) => {
```
clarkmcadoo, line 196: Let's abstract this to a callback function just to keep this more DRY

roadlittledawn: 🤦 good call

pushed it up

clarkmcadoo (approved):
🔥

PR #13853, opened 2023-07-14
[NR-67942] Added note about ARM instances

josemore:
## Give us some context

* Addressing customer feedback when installing on ARM.


nbaenam (approved):

PR #13864, opened 2023-07-14
Improvements to synth cat

zuluecho9:
NR-125568

Work done: 
Moved some things around in the nav 
Moved runtime docs together
Added another mention of the 'three strike' behavior
Assorted edits about things that were a bit confusing to me.


src/nav/synthetics.yml
```diff
@@ -117,10 +119,10 @@ pages:
         path: /docs/synthetics/synthetic-monitoring/pages/synthetic-monitoring-troubleshoot-downtime
       - title: SLA report
         path: /docs/synthetics/synthetic-monitoring/pages/synthetic-monitoring-aggregate-monitor-metrics
-  - title: Synthetics API
+  - title: API
```
akristen, line 122: I would clarify which API this is, and maybe even break it out of its own category. Feels overkill to have a drop down for one doc

src/content/docs/synthetics/synthetic-monitoring/administration/synthetics-api.mdx
```diff
@@ -0,0 +1,18 @@
...
+  - Synthetic monitoring
+  - Administration
+metaDescription: 'For New Relic synthetic monitoring, how to use APIs to manage monitors.'
+redirects:
+---
+
+We have two options for managing your New Relic synthetic monitors via API: 
+
+* NerdGraph (recommended): this is the primary and recommended way. See [our NerdGraph synthetics tutorial](/docs/apis/nerdgraph/examples/nerdgraph-synthetics-tutorial).
+* REST API: Some customers prefer to use our older REST API. See [Synthetics REST API](/docs/apis/synthetics-rest-api/monitor-examples/manage-synthetics-monitors-rest-api)
```
akristen, line 14: I would x2 check if it's worth calling this out. If we do want it listed, might want to include a note that we recommend NerdGraph because we might not support REST... but confirm with SME

src/nav/synthetics.yml
zuluecho9: it was like that before, and mainly serves to point people to docs elsewhere. i wasn't going to try to improve that with this jira

PR #13871, opened 2023-07-17
fix(Dashboards): Updating icon and parameter in the TV mode section

nbaenam:
Updated icon and parameter in the TV mode section


nwhite222 (approved):
Looks good!

PR #13878, opened 2023-07-17
Remove all code related to embed pages

LizBaker:
Now that the one official use of the /embed pages is being EOLed (docs library service) there's no reason for us to keep the pages around 🎉


sunnyzanchi (approved):
you love to see it

PR #13879, opened 2023-07-17
fix: CORECLR_ENABLE_PROFILING env var for Core apps

brnhensley:
I added some syntax highlighting to other docs while I was at it


jeff-colucci (approved):
Looks good!

PR #13883, opened 2023-07-17

bhecknr:
Updating use of "New Relic Explorer" name, which will no longer be used in docs or marketing materials.


jeff-colucci (approved):
Looks good

PR #13890, opened 2023-07-17
whats-new: NRUG Summer Series

aalfano7:
This is a WN post to promote our NR User Group Meetup Summer Series. It is an experiment to see if this is an effective place to advertise Event information. Please post on 7/19. The time of day is not important for this post, but if we are sharing any product launch announcements on 7/19, please post it before the launch information so that the launch information is ranked higher/more recently in Whats New (in product). We want to make sure that this announcement does not cannibalize views of product launch info.

It is written by Anna Alfano, a PMM - so it does not require PMM review. Thank you!


nbaenam (approved):

PR #13906, opened 2023-07-19
fix: SLES signature validation improvements

josemore:
## Give us some context

* infrastructure agent 1.44.0 has a fix for known issue, workaround is no longer needed.


nbaenam (approved):
Thank you for the update! I've added a couple of TW edits

PR #13952, opened 2023-07-21
Update nodejs-agent-configuration.mdx

tytyhibye:
per the Node Agent repo, browser monitoring is set to false by default.

Please follow [conventional commit standards](https://www.conventionalcommits.org/en/v1.0.0/)
in your commit messages and pull request title.

## Give us some context

* What problems does this PR solve?
* Add any context that will help us review your changes such as testing notes,
  links to related docs, screenshots, etc.
* If your issue relates to an existing GitHub issue, please link to it.

PR #13955, opened 2023-07-21
fix: browser monitoring is off by default

brnhensley:
goes along with Tyler's fix here: https://github.com/newrelic/docs-website/pull/13952


akristen (approved):
Good additional context! Thanks

PR #13958, opened 2023-07-21
feat(style): add label syntax + style guide structure edits

bradleycamacho:
Adds the syntax for left nav labels and rearranges some of the style guide nav. The style guide needs larger edits, but I think the nav edits make it a bit easier to find content and improve the titles by making them closer to the actual content of the doc.


paperclypse (approved):
Seems like a solid improvement

PR #13985, opened 2023-07-25
feat:Add service characterization instructions for browser

khickey-newrelic:
## Give us some context

This commit does two things: 
1. Adds documentation that helps a user understand whether or not their browser application(s) is/are instrumented correctly. 
2. Changes some of the academic language in service characterization to something that is simpler and easier to read.

It includes images in png formal instead of webp format - let me know if this is an issue and I'll correct it.


jeff-colucci (approved):
I made a few changes but overall, looks great!

PR #13993, opened 2023-07-25
Update update-or-disable-policies-conditions.mdx

relic-js:
This update simply clarifies that copying a condition does not copy any tags that have been added to the original condition.


bradleycamacho:
Hi @relic-js, would you mind shooting me a quick ping on slack when you have a second to confirm a few things?

PR #13996, opened 2023-07-26
Added missing recipes on guided install page

kidk:
Adds a couple of recipes and updates the text to match the new recipes. It used to be only on host integrations but now also contains some agents (infra, APM)


NSSPKrishna (approved):

nbaenam (approved):

PR #14002, opened 2023-07-26
add fedramp docs for k8s

bpschmitt:
## Give us some context

Adding docs for enabling FedRAMP endpoints during K8s integration installation.


jeff-colucci (approved):
Made a few small grammar and style changes, but everything else looks great!

PR #14016, opened 2023-07-27
chore(Java): rename release notes

meiao:
## Give us some context
The release notes for Java went out with the wrong filename.
Fixing it.


rhetoric101 (approved):
@meiao thanks for fixing this filename! I also switched 6.5.3 to 653.

PR #14035, opened 2023-07-27
add: config option for 8T span resevoir

brnhensley:
Added info about an undocumented configuration option to set the amount of infinite tracing spans the agent will hold in memory before dropping them. [Other agents suggest setting this to 100k](https://docs.newrelic.com/docs/distributed-tracing/troubleshooting/missing-trace-data/#spans), so that's the number I used in this doc.

- https://github.com/newrelic/node-newrelic/blob/9be213963938ba9bbfac251ba594952a964631bc/lib/config/default.js#L1103-L1124

I'd like to include a maximum value and the time period for these queues (it might be 60 seconds or 5 seconds (12 times a minute), different agents send on different periods).

Maximum value seems to not be a regular number, but a payload size so I'm not sure I can add one. I'm requesting that the node team look this over to confirm my changes.

### DIFF
--- a/src/content/docs/apm/agents/net-agent/install-guides/install-net-agent-using-nuget.mdx
+++ b/src/content/docs/apm/agents/net-agent/install-guides/install-net-agent-using-nuget.mdx
@@ -37,7 +37,7 @@ Here's an example of using NuGet via Visual Studio to install the .NET agent:
        id="nuget-framework"
        title="Environment variables for monitoring .NET Framework applications"
      >
-       ```
+       ```ini
        COR_ENABLE_PROFILING=1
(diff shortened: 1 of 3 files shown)


rhetoric101:
@brnhensley let us know when you get confirmation from the Node team.

nbaenam:
Hi @brnhensley, I'm going to move moved this PR to our draft category. When you're ready to publish, reach out to the Docs team and we'll get it merged!

brnhensley:
@rhetoric101 node team told me more about how this works and I've updated my note. This is ready for review from your team now.

brnhensley:
I added an unrelated commit here since the PR is still open, to avoid making more for small things

bradleycamacho (approved):

PR #14047, opened 2023-07-28
use `eval` to parse MDX prop object string

sunnyzanchi:
## testing

- `yarn serialize src/content/docs/distributed-tracing/ui-data/understand-use-distributed-tracing-ui.mdx -o test`
- `cat test` should look like some HTML
- `yarn deserialize test`


roadlittledawn (approved):
:shipit:

PR #14096, opened 2023-08-02
Update ruby-custom-instrumentation.mdx

Catsuko:
Trying to run this code snippet did not work for me so i've updated the code snippet to include the proper namespacing. This is consistent with the other code snippets here.


nbaenam (approved):
Hi @Catsuko, thanks for bringing this to our attention! I've checked out with our Ruby team experts (ruby-agent channel) and added the right code.

PR #14124, opened 2023-08-04
Remove manual proxy configuration step

RyanKadri:
This step is no longer needed as the in-product wizard will properly populate the proxy settings into the necessary file


akristen (approved):
Thanks for this update!

PR #14163, opened 2023-08-08
Add note on GraphQL usage limits

mattbaker:
NerdGraph will be enforcing concurrent request limits soon, this documentation is a reference for customers and support in the event they see a 429 response from the GraphQL API.

@vosechu, a review would be welcome if you have a moment to spare.


vosechu (approved):
This looks magnificent!

mattbaker:
@vosechu @homelessbirds — I realized the URL was going to be odd since it would reference the REST API, so I've split the pages.

If I did something wrong in terms of directory structure or naming please let me know! I'll test out a preview build when it's available.

jmiraNR:
@homelessbirds,  @mattbaker and I chatted, we'll go through human translation for this page so the frontmatter is good from my perspective. Thx

zuluecho9:
Just noticed this. We will want to mention this in our main limits doc. https://docs.newrelic.com/docs/data-apis/manage-data/view-system-limits/

mattbaker:
Does that need to block this merge @zuluecho9?

zuluecho9 (dismissed):

PR #14174, opened 2023-08-09
all headers in pagetools

tabathadelane:
- use mdxAST for headers in PageTools instead of graphql's tableOfContents to allow us to grab deeper nested headings
- remove unused TableOfContents component that seemed to only be stitching on an mdxAST fragment to the docsPage query. I just added mdxAST to the query itself. Functionality doesn't seem to be affected by this change.
- added a temp mdx page with headers inside a step component for demo purposes to get feedback from docs teams


austin-schaefer:
@tabathadelane : Is this still a PR we're actively working? If not can we close and reopen the related Jira for the backlog?

PR #14177, opened 2023-08-10
feat(Alerts): Added the other metrics option

nbaenam:
Updated the [Create an alert condition using New Relic](https://docs.newrelic.com/docs/tutorial-create-alerts/create-an-alert/) page to add the new option "Other metrics".


rhetoric101 (approved):
Nati, I left-aligned the first image in step 7 and indented it so it aligns with the step. I also indented the follow-up line and image for step 7 so they are all indented under step 7.

See if this look OK.

PR #14190, opened 2023-08-10
.NET Agent API usage guide: update links for DT APIs to non-obsolete versions

nr-ahemsath:
## Give us some context

This PR updates some links in the .NET Agent's public API usage guide to point to current (non-deprecated) APIs.


jeff-colucci (approved):
Looks good

PR #14242, opened 2023-08-14
fix: remove plain text escape chars `\`

brnhensley:
when `wrapping text in backticks` escape chars like `\` show up as plain text.

I also add lang identifiers to codeblock. `ini` works well on github, not sure how well it will work on docs-website

image


jeff-colucci (approved):
Looks great as always, thank you!

PR #14249, opened 2023-08-14
fix(Hero): fixing some bits and bobs

jeff-colucci:


jeff-colucci:
Publishing after I checked the links myself

PR #14252, opened 2023-08-14
Update .NET browser monitoring info

brnhensley:
It's often missed that automatic browser monitoring only works for .NET framework


jeff-colucci (approved):
Seems good to me!

PR #14282, opened 2023-08-16
Updated supported runtimes list for lambdas.

umaannamalai:
This PR updates the compatibility docs for lambda with supported runtimes for Node and Python.


rhetoric101 (approved):
@umaannamalai thanks for making this update!

PR #14302, opened 2023-08-18
fix: details about identifying infra agent versions in the customer environment

josemore:
## Give us some context

* Adding more details to explain how to check the infra agent versions running in the customer environment and how to identify agents running older versions.


zuluecho9:
thanks @josemore !

zuluecho9 (approved):

PR #14307, opened 2023-08-18
fix(user mgmt): detail about user records

zuluecho9:
Add detail about how unique users are determined by user ID.


paperclypse (approved):
Looks good!

bradleycamacho (approved):

PR #14323, opened 2023-08-21
More detail on user record tracking/counting

zuluecho9:
Clarifying how users are tracked/counted in a new section, and pointing to that section from a couple spots


jeff-colucci (approved):
Looks good to me!

PR #14380, opened 2023-08-25
1st draft browser agent version pinning tutorial

ajtran303:

## Give us some context

* This is documentation for a current product-facing NG API that we are preparing for customer-facing release.


rhetoric101:
I switched this to a draft where we'll keep this until it's ready.

src/content/docs/apis/nerdgraph/examples/browser-monitoring-config-nerdgraph.mdx
```diff
@@ -302,4 +302,119 @@ Here's the response:
...
+  }
+}
+
+```
+
+Note that the return fields for this mutation will contain the updated `pinnedVersion` value, as well as the updated `jsLoaderScript`.
+
+Changing the `loader` value to `NONE` will remove any pinned semantic version. 
+
+### Step 4. Decode the JavaScript Snippet (if using Copy and Paste)
```
rhetoric101, line 408: Maybe step 4 should be "Insert snippet into you application"? In the details, you could tell users that if they used the copy/paste routine, they'll need to decode the JavaScript snippet into HTML before they insert it. 

Still, I'm not really clear about how this works. The mutation generates JavaScript, so doesn't the JavaScript remain the same, but they just insert it between <script> tags?

src/content/docs/apis/nerdgraph/examples/browser-monitoring-config-nerdgraph.mdx
```diff
@@ -302,4 +302,119 @@ Here's the response:
...
+
+The `jsLoaderScript` value is a JSON encoded string. The string **MUST** be decoded into a valid HTML `<script>` element before it can be used in your application. 
+
+### Last step. Re-deploy your application or restart your agent
```
rhetoric101, line 412:
```suggestion
### Step 5. Re-deploy your application or restart your agent
```

src/content/docs/apis/nerdgraph/examples/browser-monitoring-config-nerdgraph.mdx
```diff
@@ -302,4 +302,119 @@ Here's the response:
...
+
+For applications using the copy/paste method, replace the existing JavaScript snippet with the new, decoded `<script>`. Then re-deploy your application.
+
+For applications that have browser monitoring via a backend-installed agent (such as Ruby, Python, etc.), the agent will need to be restarted in order to fetch the new values. 
+
+In either case, the next time your application serves a browser page, it should be using the pinned agent version to report data to your New Relic One account.
```
rhetoric101, line 418:
```suggestion
In either case, the next time your application serves a browser page, it should be using the pinned agent version to report data to your New Relic account. 
```

src/content/docs/apis/nerdgraph/examples/browser-monitoring-config-nerdgraph.mdx
```diff
@@ -302,4 +302,119 @@ Here's the response:
...
+
+### Last step. Re-deploy your application or restart your agent
+
+For applications using the copy/paste method, replace the existing JavaScript snippet with the new, decoded `<script>`. Then re-deploy your application.
+
+For applications that have browser monitoring via a backend-installed agent (such as Ruby, Python, etc.), the agent will need to be restarted in order to fetch the new values. 
+
+In either case, the next time your application serves a browser page, it should be using the pinned agent version to report data to your New Relic One account. 
+
+Remember that you can check the pinned version and loader type by using the query in Step 1.
```
rhetoric101, line 420:
```suggestion
Remember that you can check the pinned version and loader type by using the query in the first step.
```

rhetoric101 (approved):
@nbaenam and @ajtran303 this looks great! I just inserted some changes you might consider.

src/content/docs/apis/nerdgraph/examples/browser-monitoring-config-nerdgraph.mdx
ajtran303: > Maybe step 4 should be "Insert snippet into you application"?

I'm okay with this change.

> The mutation generates JavaScript, so doesn't the JavaScript remain the same, but they just insert it between <script> tags?

The JavaScript snippet would change if the customer changes the pinned version.

ajtran303: Maybe, step 4 could just be combined into Step 5

src/content/docs/apis/nerdgraph/examples/browser-monitoring-config-nerdgraph.mdx
```diff
@@ -302,4 +302,119 @@ Here's the response:
...
+- **Testability**: Easily test new versions in isolated environments before deciding to update.
+- **Stability**: Minimize unexpected disruptions and maintain consistent application behavior.
+- **Efficiency**: Reduce deploy time and mitigate deployment difficulties.
+
+Here is an overview of how to use the Browser Agent Version Pinning API:
+
+1. Check agent type and which version is currently pinned, if any
+1. Reference the available versions of the agent
+1. Select and pin the desired version
+1. Redeploy or restart your service to use the pinned version of the agent
```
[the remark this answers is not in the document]
ajtran303: > Do we need to add the decoding step that appears down in the main steps?

I think, that in the case of step 4 and 5 being combined as I suggested in the other comment, it would be okay to not include the decoding step in the overview.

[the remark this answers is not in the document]
nbaenam: Removed step 4. Decode the JavaScript snippet

src/content/docs/apis/nerdgraph/examples/browser-monitoring-config-nerdgraph.mdx
ajtran303: Oh I understand what you are asking.

`jsLoaderScript` is actually HTML `<script>` tag that contains the JavaScript agent.

An example response:

```json
{
  "data": {
    "actor": {
      "entity": {
        "browserProperties": {
          "jsLoaderScript": "<script type=\"text/javascript\">\n;window.NREUM|| ... p\n</script>\n"
....
```

> Would users just copy the jsLoaderScriptJavaScript from the mutation and insert it into their existing <script> section of their apps? 

No, users would replace their existing HTML script element.

> What's involved in "decoding"? 

The GraphQL mutation result returns a JSON-encoded string. I.e., quotation marks are escaped and there are newline literals....:

```
<script type=\"text/javascript\">\n;window.NREUM|| ... p\n</script>\n"
```

So to have valid JavaScript, a user would have to JSON-decode it to escape the slashes and interpret the newlines.

In my prior experience, I paste the entire string into a tool called "Boop" and use the "JavaScript Eval" functionality. It runs the input as JavaScript code, and the output is text that is valid `<script>` and JavaScript inside.

Theoretically, a developer could use a different method to do it, as long as the main constraints are 1. removing slash mark escaping from quotation marks and 2. removing newline literals.

> Is it more than just nesting the JavaScript inside the <script> tags?

They will not need to nest JavaScript code in HTML <script> tags because the API provides both of those.

> Also, where would they put it in their existing <script> code? 

They would remove their existing script and replace it with this one. The point of the version pinning API is so that they can "freeze" their browser agent from getting updates. Also if they keep their existing script and add this one, I'm not sure how that would work. "It's not supported," I suppose!

> At the end?

This script would be pasted into the <head> tags, where the old one went. By following the directions in the copy/paste documentation they will be ok. 👍 

> If users are not using the copy/paste routine, what else would they do?

For users who enable browser monitoring through an APM agent like Python or Ruby, they would have to restart their agent for the new changes to take effect. From then on, any new web pages served will be automatically instrumented with their desired browser agent with the pinned version they specified.

nbaenam: Thank you for the explanation, @ajtran303!

PR #14392, opened 2023-08-28
chore(php): add laravel 9 and 10 to supported frameworks

hahuja2:
Update the PHP Agent Compatibility page by adding Laravel 9 and 10 as supported frameworks.


akristen (approved):
Thanks for the update!

PR #14402, opened 2023-08-29
Fix wording in Azure log forwarding integrations docs

jsubirat:
It fixes the wording for the Microsoft Azure log forwarding integrations. These integrations were referring to an old UI component that has changed its naming from "Add more data sources" to "Add data". Also, the instructions were pointing to the "top right of the page", and this has now been physically moved to the left sidebar.


zuluecho9 (approved):

zuluecho9:
thanks @jsubirat !

PR #14406, opened 2023-08-29
update: explain what labels do, remove more `<var>`s

brnhensley:
Labels were not clearly explained, so added some info about what they do and don't do. Internal source: https://source.datanerd.us/agents/agent-specs/blob/b657613819369c60be766556ea4a62cb82fd8597/Labels.md

I also removed more `<var>`s which we don't use anymore. I left them in code samples that shows multiple values, like `true|false` or a series of possible options

For examples that did show `true|false` before I kept the non-default value. So if the default of `X=true`, in the example of configuring `X` I made it 
```ini
X=false
```

## Give us some context


brnhensley:
I found a mistake in this change about Tags, let me do more research before this goes to review

brnhensley:
I made some mistakes here and will close this at least for now

PR #14457, opened 2023-09-01
Add webdriver test to github actions

tabathadelane:
webdriver seems to be working in [gh actions](https://github.com/newrelic/docs-website/actions/workflows/webdriver.yml)

I had to modify the synthetics scripts versions slightly, but tried to leave comments where relevant.

I have this setup to run on PRs to main so we can always use the develop build url. Otherwise, i think we would have to pull in the github api and slugify branch names to try and match develop PR build urls. And then we would have try and setup the action to only run after the gatbsy build is complete - which would never run if we have to do a cache clear build. 

TODO:

~~remove 'develop' branch and 'workflow_dispatch' from action yaml. Just added these for testing purposes~~

~~fix collapser failing on webdriver-desktop (only in gh action)
fix doctile failing on webdriver-mobile~~


clarkmcadoo (approved):

sunnyzanchi (approved):
🌟

PR #14473, opened 2023-09-05
fix: update the query for suggested success sli

NilVentosa:
## Give us some context

We updated the suggested query for bad events in the success SLI for APM entities. This is just an update to reflect that.


rhetoric101 (approved):
@NilVentosa thanks for making this improvement to the query!

PR #14485, opened 2023-09-05
fix(nav): fix some entries

zuluecho9:


rhetoric101 (approved):
@zuluecho9 thanks for making these fixes!

PR #14495, opened 2023-09-06
Update addcustomattributes-python-agent-api.mdx

vicenterocha:
Hey, I believe the metric key shouldn't be a string, but whatever value the variable `server_ip` has.

Akin to the example [here](https://docs.newrelic.com/docs/apm/agents/python-agent/python-agent-api/addcustomattribute-python-agent-api)


homelessbirds:
Thank you!

PR #14515, opened 2023-09-07
Add note about Grok multiline

bmcfeely:
## Give us some context

* I had a customer issue around this problem, thought it would be good to tell all customers about


akristen (approved):
Looks good! Thanks for the addition

PR #14523, opened 2023-09-07
chore(java): removing unimportant item from release notes

meiao:
## Give us some context

This specific item relates to a bad release of a library. Since older versions of the agent are compatible with the same versions, this item should be removed to avoid confusion.


akristen (approved):
Looks good, thanks!

akristen (approved):
Thanks!

PR #14560, opened 2023-09-12
Update kubernetes-components.mdx

reese-lee:
Adds:
* link to `newrelic-infrastructure` chart
* blurb about `nri-bundle`
* table of the components that come by default when you install the integration using `nri-bundle`
* table of optional components you can install using `nri-bundle`

Renames page title to more accurately reflect its contents.


nbaenam (approved):
Thanks for adding this! I've added a commit to fix the tables. Once it builds, I'll merge it. Thanks!

nbaenam (approved):

PR #14611, opened 2023-09-18
[fix] OTel host.id is disabled by default

josemore:
## Give us some context

* The OTel community changed the host.id attribute to be disabled by default (v0.82.0 or greater). Adding it back to be enabled in the recommended config for host monitoring.


nbaenam (approved):

PR #14623, opened 2023-09-18
create validate-release-note-json gh action

tabathadelane:
- runs on PRs to develop
- verifies that every agent has release notes
- verifies our expected fields exist

Related: [synthetics monitor](https://staging.onenr.io/0kLwGb4bAQ6) that checks the status of the docs link associated with each agents most recent release note to help us foresee potential JSON/Gatsby builds being out of sync.

### DIFF
--- a/scripts/releaseNotes.mjs
+++ b/scripts/releaseNotes.mjs
@@ -12,6 +12,7 @@ import { Command } from 'commander';
 
 import getAgentName from '../src/utils/getAgentName.js';
 import getEOLDate from '../src/utils/getEOLDate.js';
+import { Agent } from 'http';
 
 const program = new Command();
 program
@@ -108,10 +109,11 @@ const releaseNotes = (
 console.error('📦 release notes JSON generated');
 
 const validateReleaseNotesAgents = (json) => {
+  // this set excludes 'sdk', 'node' and '.net' from the one above
   const JSON_AGENTS = new Set([
     'android',
     'browser',
-    'dontnet',
+    'dotnet',
     'go',
     'infrastructure',
     'ios',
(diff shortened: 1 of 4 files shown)


scripts/releaseNotes.mjs
```diff
@@ -105,6 +107,52 @@ const releaseNotes = (
...
+  json.forEach((note) =>
+    requiredData.forEach((key) => {
+      if (!note[key]) {
+        console.error(note);
+        console.error(`😵 Missing ${key} data`);
+        process.exitCode(1);
+      }
+    })
+  );
+  console.error(`✨ Release notes JSON validated`);
```
sunnyzanchi, line 153: i think we should put a conditional here with a different message if validation fails. rn i think this kinda of reads like the notes are valid whether they pass or not

scripts/releaseNotes.mjs
```diff
@@ -105,6 +107,52 @@ const releaseNotes = (
 );
 console.error('📦 release notes JSON generated');
 
+const validateReleaseNotesAgents = (json) => {
```
sunnyzanchi, line 110: maybe a nit but i think we should name this arg `releaseNotes`. `json` might imply that it's already stringified

```suggestion
const validateReleaseNotesAgents = (releaseNotes) => {
```

sunnyzanchi (changes requested):

scripts/releaseNotes.mjs
tabathadelane: Do you mean for eng readability? or for the Actions console? This line of code actually doesn't run if it [fails](https://github.com/newrelic/docs-website/actions/runs/6228948584/job/16906611920).

tabathadelane: updated! I actually like [this](https://github.com/newrelic/docs-website/actions/runs/6241138043/job/16942723937) better

sunnyzanchi (approved):

PR #14646, opened 2023-09-20
feat(Alerts): Added content in the "Set up a Slack destination" section

nbaenam:
Added content at the end of "Set up a Slack destination" on the [Alerts and applied intelligence notification integrations](https://docs.newrelic.com/docs/alerts-applied-intelligence/notifications/notification-integrations/#slack) page.

### DIFF
--- a/src/content/docs/alerts-applied-intelligence/notifications/notification-integrations.mdx
+++ b/src/content/docs/alerts-applied-intelligence/notifications/notification-integrations.mdx
@@ -696,7 +696,10 @@ If successful, there'll be a link to view the incident in JIRA.
       src={accountsSlackAuthenticationCompleted}
     />
 
-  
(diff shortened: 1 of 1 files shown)


src/content/docs/alerts-applied-intelligence/notifications/notification-integrations.mdx
```diff
@@ -696,7 +696,10 @@ If successful, there'll be a link to view the incident in JIRA.
       src={accountsSlackAuthenticationCompleted}
     />
 
-  
+The Slack integration isn't tied to a specific user. If the user who originally authenticates with Slack leaves the organization, or their account is deleted, the Slack destination will continue to function. When you authenticate with Slack, our platform goes to Slack and asks for an OAuth token. This token expires at some point, and it's refreshed with the token from the original transaction.
```
homelessbirds, line 699:
```suggestion
Our Slack integration isn't tied to a specific user. If the user who originally authenticates with Slack leaves the organization, or their account is deleted, the Slack destination will continue to function. When you authenticate with Slack, our platform goes to Slack and asks for an OAuth token. This token expires at some point, and it's refreshed with the token from the original transaction.
```

homelessbirds (approved):

PR #14660, opened 2023-09-21
chore(entity): fix callout fatigue

bradleycamacho:
This doc had too many callouts. Did a quick pass to reduce callout fatigue.


jeff-colucci (approved):
Looks good!

PR #14705, opened 2023-09-26
docs: Minor streaming export compression correction

ecchilds:
## Give us some context

* Jira ticket: NR-164732
* Currently, the documentation for export compression says that users will save on "egress COGS" while using it. This is incorrect terminology, they will actually save on ingress costs. This PR corrects the mistake.


nbaenam (approved):

PR #14733, opened 2023-09-29
fix(Logs): Removed a confused text about the limit

nbaenam:
Removed this text `The limit is not a fixed value; it scales up or down proportionally to the volume of data stored daily by the account and the environment size that is subsequently allocated to support that customer.` from the [Parsing log data](https://docs.newrelic.com/docs/logs/ui-data/parsing/#limits) page.

Requested in the #help-documentation channel by Alexander Mosher.


jeff-colucci (approved):
Looks good!

Issue #14676, opened 2023-09-22
OTel collector for Prometheus docs are inaccurate


thezackm:
[docs link here](https://docs.newrelic.com/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/collector/collector-configuration-examples/opentelemetry-collector-prometheus/#install-opentelemetry-collector)

In our install docs, we have step 3 listed as 2 parts:
 * install the collector
 * install the contrib distribution to include the prometheus and otlp receivers

>Install [OpenTelemetry Collector Contrib Distro](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-contrib) or other distribution including, at least, the following components:
> * [Prometheus receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver)
> * [OTLP exporter](https://pkg.go.dev/go.opentelemetry.io/collector/exporter/otlpexporter)

This is confusing as the link to the contrib distro doesn't have any installation instructions. Additionally, the vanilla install of the OTel collector has both of these components installed by default so the entire process is unnecessary. 

results from running `./otelcol components`:

```yaml
buildinfo:
    command: otelcol
    description: OpenTelemetry Collector
    version: 0.85.0
receivers:
    - name: otlp
      stability:
        logs: Beta
        metrics: Stable
        traces: Stable
    ...
    - name: prometheus
      stability:
        logs: Undefined
        metrics: Beta
        traces: Undefined
    ...
```

---

suggested resolution: remove the instructions to install the contrib distro as unneeded.

akristen:
Hi @thezackm -- just to clarify, **removing** the contrib distro step is what we need to do to update the Prometheus install doc. There's nothing to add or draft, just remove. Would you be willing to review this PR when it's ready? :)

thezackm:
@akristen - confirmed. 

I can review but would recommend checking with the PM as well to make sure everything is okay with them. 
CC: @mangulonr

mangulonr:
Thanks for the suggestion @thezackm 
I will come back to you tomorrow with an answer.

akristen:
Addressing this issue in this PR: https://github.com/newrelic/docs-website/pull/14739/

PR #14739, opened 2023-09-29
Update prometheus install docs

akristen:
See #14676

### DIFF
--- a/src/content/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/collector/collector-configuration-examples/opentelemetry-collector-prometheus.mdx
+++ b/src/content/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/collector/collector-configuration-examples/opentelemetry-collector-prometheus.mdx
@@ -23,12 +23,11 @@ Complete the steps below to collect Prometheus metrics.
 If you haven't already:
 
 * [Sign up for a New Relic account](https://newrelic.com/signup).
-* Ensure that the Prometheus metrics endpoint is accessible to the OpenTelemetry collector as the collector uses the Prometheus receiver component. 
-    * Prometheus metrics endpoint could be enabled by default or may need some additional configuration to enable it. 
-    * Some software may require a [third-party Prometheus exporter](https://prometheus.io/docs/instrumenting/exporters/#third-party-exporters) to expose metrics.
-* Finally, install the OpenTelemetry collector following the [Opentelemetry docs](https://opentelemetry.io/docs/collector/getting-started/). 
-
-In some cases but not all, you may use an open source project that requires you to install the [OpenTelemetry contrib distro](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-contrib). For example, if you use the Core distribution, the contrib distro contains additional components that the collector needs. Regardless of your distribution, you must have the [OTLP exporter](https://pkg.go.dev/go.opentelemetry.io/collector/exporter/otlpexporter) and the [Prometheus receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver) components installed. 
+* In your target app, make sure you're exposing the Prometheus metrics endpoint so it is accessible to the Prometheus receiver component in the OpenTelemetry collector.
+  * The Prometheus metrics endpoint could be enabled by default or you may need do do additional configuration to enable it.
+  * Some software may require a [third-party Prometheus exporter](https://prometheus.io/docs/instrumenting/exporters/#third-party-exporters) to expose metrics.
+* Install the OpenTelemetry collector by following the steps in the [OpenTelemetry docs](https://opentelemetry.io/docs/collector/getting-started/). In some cases but not all, you may use an open source project that requires you to install the OpenTelemetry contrib distro. For example, if you use the Core distribution, the contrib distro contains additional components that the collector needs. 
+* Regardless of your distribution, you must have the [OTLP exporter](https://pkg.go.dev/go.opentelemetry.io/collector/exporter/otlpexporter) and the [Prometheus receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver) components installed.
 
     </Step>
     <Step>
--- a/src/content/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/collector/collector-configuration-examples/opentelemetry-collector-prometheus.mdx
+++ b/src/content/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/collector/collector-configuration-examples/opentelemetry-collector-prometheus.mdx
@@ -26,8 +26,8 @@ If you haven't already:
 * In your target app, make sure you're exposing the Prometheus metrics endpoint so it is accessible to the Prometheus receiver component in the OpenTelemetry collector.
   * The Prometheus metrics endpoint could be enabled by default or you may need do do additional configuration to enable it.
   * Some software may require a [third-party Prometheus exporter](https://prometheus.io/docs/instrumenting/exporters/#third-party-exporters) to expose metrics.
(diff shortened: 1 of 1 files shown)


src/content/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/collector/collector-configuration-examples/opentelemetry-collector-prometheus.mdx
```diff
@@ -15,29 +15,25 @@ You can collect metrics from any [software that exposes Prometheus metrics](http
...
-
-## Step 3: Install Opentelemetry collector [#install-opentelemetry-collector]
-
-If not done yet, download and install the Opentelemetry collector following [Opentelemetry docs](https://opentelemetry.io/docs/collector/getting-started/).
+<Steps>
+    <Step>
+    ## Install Opentelemetry collector
 
-Install [OpenTelemetry Collector Contrib Distro](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-contrib) or other distribution including, at least, the following components:
+If not done yet, download and install the Opentelemetry collector following [Opentelemetry docs](https://opentelemetry.io/docs/collector/getting-started/).
```
akristen, line 30: @mangulonr This is the step that @thezackm called out as needing some updates. This step on its own isn't very helpful, either:

* This step is about installing the OTel Collector, but then links to Prometheus receiver and OTLP exporter. Can you explain how these two bullet points relate to the install process in OTel collector? Are they folded into the install process? Why are they separate bullet points? What are users supposed to do with these links -- just having them linked out doesn't tell us much
* I removed the Collector Contrib distro. Should I add anything about this repo that could be helpful for users? Zack mentioned that this is included out of the box as part of the vanilla install, but if someone needed to do a special configuration, would this repo be relevant?

akristen: Something else that confuses me (and might confuse our users) is how installing the OTel collector relates to the Prometheus collector. If I'm understanding correctly:

- The Prometheus collector integrates with our existing OTel Collector
- Users get the Prometheus collector by configuring the OTel Collector
- So step #1 of getting the Prometheus collector is really step 4, "Configure OTel Collector."

The prereqs would then read:

- Sign up for an account
- Install the OTel collector
- Ensure that the OTel collector has "something" enabled so that it can access the Prometheus metrics endpoints.

Does this sound correct?

mangulonr: @akristen 
Thanks for taking the initiative here.

We pretended to state that our customers could use any  OTel distribution that includes the components mentioned (Prometheus receiver and OTLP exporter). Customer can also create their own OTel distribution

In both links, you can check in which OTel distributions are those components included.

We used to recommend installing the contrib  distro because it not only contains components tied to open source projects ([like Core distribution](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol)) but also some vendor supported ones ([link](https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/distributions/otelcol-contrib/README.md)).

For example, in the case of other collector examples we have in our docs website like Docker and Redis we can't recommend the Core distro because it requires a vendor-specific receiver.

Maybe we can keep the recommendation of installing distro at least with the other examples but including the link to the install instructions you're proposing.

Prometheus receiver is just a component that can be included in a collector distribution it is not a collector for itself.

Also I'm checking that the sign-up widget is not displaying any content.

src/content/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/collector/collector-configuration-examples/opentelemetry-collector-prometheus.mdx
```diff
@@ -15,29 +15,23 @@ You can collect metrics from any [software that exposes Prometheus metrics](http
 
 Complete the steps below to collect Prometheus metrics.
 
-## Step 1: Sign up for your free account if you haven't already [#signup]
+<Steps>
+    <Step>
+    ## Ensure you have the prerequisites
```
akristen, line 20: Alternative title maybe? "Get set up" "Start here" -- not sure, but 'ensure you have the prereqs' seems a bit silly

src/content/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/collector/collector-configuration-examples/opentelemetry-collector-prometheus.mdx
```diff
@@ -15,29 +16,22 @@ You can collect metrics from any [software that exposes Prometheus metrics](http
...
 
-In both cases, the Prometheus metrics endpoint needs to be accesible from the OpenTelemetry Collector.
+If you haven't already:
 
-## Step 3: Install Opentelemetry collector [#install-opentelemetry-collector]
+* [Sign up for a New Relic account](https://newrelic.com/signup).
+* In your target app, make sure you're exposing the Prometheus metrics endpoint so it is accessible to the Prometheus receiver component in the OpenTelemetry collector.
+  * The Prometheus metrics endpoint could be enabled by default or you may need do do additional configuration to enable it.
+  * Some software may require a [third-party Prometheus exporter](https://prometheus.io/docs/instrumenting/exporters/#third-party-exporters) to expose metrics.
+* Install the OpenTelemetry collector by following the steps in the [OpenTelemetry docs](https://opentelemetry.io/docs/collector/getting-started/). In some cases but not all, you may use an open source project that requires you to install the OpenTelemetry contrib distro. For example, if you use the Core distribution, the contrib distro contains additional components that the collector needs.
```
paologallinaharbur, line 29: I think that the sentence "In some cases but not all, you may use an open source project that requires you to install the OpenTelemetry contrib distro." could be misunderstood.

Basically, the each otel distro can have different components. They should make sure that their particular distro has those components. On the other hand, if not available they cannot be installed, since they should be part of the binary.

I would rephrase it to something like "You need either to have or to install a distro with those components"

src/content/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/collector/collector-configuration-examples/opentelemetry-collector-prometheus.mdx
```diff
@@ -15,29 +16,22 @@ You can collect metrics from any [software that exposes Prometheus metrics](http
...
+* In your target app, make sure you're exposing the Prometheus metrics endpoint so it is accessible to the Prometheus receiver component in the OpenTelemetry collector.
+  * The Prometheus metrics endpoint could be enabled by default or you may need do do additional configuration to enable it.
+  * Some software may require a [third-party Prometheus exporter](https://prometheus.io/docs/instrumenting/exporters/#third-party-exporters) to expose metrics.
+* Install the OpenTelemetry collector by following the steps in the [OpenTelemetry docs](https://opentelemetry.io/docs/collector/getting-started/). In some cases but not all, you may use an open source project that requires you to install the OpenTelemetry contrib distro. For example, if you use the Core distribution, the contrib distro contains additional components that the collector needs. 
+* Regardless of your distribution, you must have the [OTLP exporter](https://pkg.go.dev/go.opentelemetry.io/collector/exporter/otlpexporter) and the [Prometheus receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver) components installed.
```
paologallinaharbur, line 30: this point is related to the previous one, but it is confusing as it is now.

Moreover, I would not use the word "installed"

src/content/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/collector/collector-configuration-examples/opentelemetry-collector-prometheus.mdx
akristen: Ah, ok, this makes a lot of sense. I'll update the language.

homelessbirds (approved):

PR #14750, opened 2023-10-02
Fix broken query

tcgbrett:
looks like a copy + paste issue. you can't order by a field that isn't queried in your select statement. i tested in my env and the query returns nil


bradleycamacho:
Hey @tcgbrett, thanks for the edit! I'll take a look and approve, but could you sign the CLA so I can get this merged in? You just have to click the yellow button in the comment above. Thanks!

PR #14760, opened 2023-10-03
Update issues-with-istio.mdx

xqi-nr:

## Give us some context

use a better word for the trouble shooting issue doc. mitigate means reduce but not resolved. So it is not a proper word in the context


rhetoric101 (approved):
@xqi-nr thanks for improving this wording!

PR #14784, opened 2023-10-05
Lookups api

akristen:
- Fixed H2s in sentence case
- Did a casing pass
- Added some structure to the beginning of the doc
- Put most content in collapsers to keep the doc more in line? Not sure how I feel though

### DIFF
--- a/src/content/docs/apis/lookups-service-api/lookups-service-api.mdx
+++ b/src/content/docs/apis/lookups-service-api/lookups-service-api.mdx
@@ -12,7 +12,7 @@ redirects:
 
 Use the NRQL Lookups API to create and manage [lookup tables](/docs/logs/ui-data/lookup-tables-ui/#overview).
 
-## Before you start
+## Prerequisites [#prereq]
 
 The NRQL Lookups API is a REST API that allows you to manage lookup tables programmatically. As another option you can also [manage lookup tables through our UI](/docs/logs/ui-data/lookup-tables-ui/#lookup-ui).
 
@@ -174,7 +174,7 @@ The variables required in the above NRQL Lookups API endpoints are defined below
     </tbody>
 </table>
 
-## Authentication
+## Authentication [#auth]
 
 Your <InlinePopover type="userKey" /> serves to authenticate your request to the NRQL Lookups API and needs to be passed as an HTTP header.
 
@@ -204,10 +204,9 @@ Your <InlinePopover type="userKey" /> serves to authenticate your request to the
   </tbody>
 </table>
 
+## Create and update a table [#create-and-update]
 
-## Create/Update a Table [#create-and-update]
-
-### HTTP Endpoints [#create-and-update-http-endpoints]
+### HTTP endpoints [#create-and-update-http-endpoints]
 
 #### Create
 `POST /v1/accounts/YOUR_ACCOUNT_ID/TABLE_NAME`
@@ -219,7 +218,7 @@ Used to upload a new table. The table cannot already exist. If it does, this cal
 
 Used to replace an existing table. If the table does not exist this call will result in a `404 Not Found` response.
 
-#### Request Query Parameters
+#### Request query parameters
 
 <table>
   <thead>
@@ -263,7 +262,7 @@ Used to replace an existing table. If the table does not exist this call will re
   </tbody>
 </table>
 
-### HTTP Headers [#create-http-headers]
+### HTTP headers [#create-http-headers]
 
 When creating your HTTP headers, use these guidelines:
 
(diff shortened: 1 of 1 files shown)


homelessbirds:
So I think your instinct to make these into tabs is a good one. I like the collapsers, but I think grouping them into horizontal tabs would also help. Like a heading that says something like "Manage tables" and then the `List`, `Delete`, `Create`, etc sections can be below, Also, it will make the lone `Error messaging` at the bottom look less weird and out of place. 

You could make those changes as a fast follow though!

src/content/docs/apis/lookups-service-api/lookups-service-api.mdx
```diff
@@ -10,31 +10,36 @@ redirects:
   - /docs/apis/lookup-service-api/lookup-service-api
 ---
 
-Use the NRQL Lookups API to create and manage [lookup tables](/docs/logs/ui-data/lookup-tables-ui/#overview).
+If you have custom data that's not reported to New Relic out of the box, you likely use [lookup tables](/docs/logs/ui-data/lookup-tables-ui/#overview) to query certain kinds of telemetry data. With our NRQL lookups API, you can programmatically create, update, or delete your lookup tables. This is a good option for users who want a more efficient way to query and handle their custom data.  
 
-## Before you start
-
-The NRQL Lookups API is a REST API that allows you to manage lookup tables programmatically. As another option you can also [manage lookup tables through our UI](/docs/logs/ui-data/lookup-tables-ui/#lookup-ui).
+The NRQL lookups API is a REST API that allows you to manage lookup tables programmatically. If you don't want to work with REST APIs, you can [manage lookup tables via our UI](/docs/logs/ui-data/lookup-tables-ui/#lookup-ui).
```
ccobb-nr, line 15: I'm not sure about the "If you don't want to work with the REST APIs" line. I think the original "As another option" is better here.  Or maybe something like "For simple, non-programatic Lookup Table maintenance, you can manage..".

At the very least, remove the "s" on "APIs".

src/content/docs/apis/lookups-service-api/lookups-service-api.mdx
```diff
@@ -10,31 +10,36 @@ redirects:
...
 
-[European Union](/docs/using-new-relic/welcome-new-relic/get-started/introduction-eu-region-data-center) (EU) endpoint:
+The European Union (EU) endpoint is:
 
 ```
 https://nrql-lookup.service.eu.newrelic.com
 ```
 
-### Endpoints
+### Choose an endpoint
```
ccobb-nr, line 35: "Choose an" doesn't seem quite right.  This section lays out all the different actions that can be done through the API. This just needs a simple header.

But, I also don't really like "Endpoints".  Maybe "REST API Methods and Paths"?

src/content/docs/apis/lookups-service-api/lookups-service-api.mdx
```diff
@@ -10,31 +10,36 @@ redirects:
...
+
+    Add an endpoint to the base URL depending on the action you want to take. There are two variables you'll need to update because they're specific to your account:
```
ccobb-nr, line 37: In the context of this sentence, "choose" actually makes more sense.  But endpoint is still a bit confusing. 
I think this should say "Choose an HTTP method and path based on the action you want to take. The path will be appended to the the base URL." And make 'base URL' link to the base URL section above.

In the table that follows, I'd create 3 columns:
* Action  - same as the current first column (the current "Method" column)
* HTTP Method - The first part of the current "Endpoint" column (POST, PUT, GET, DELETE)
* Path - The second part word of the current "Endpoint" column
* Description - Same as the current "Description"

src/content/docs/apis/lookups-service-api/lookups-service-api.mdx
```diff
@@ -10,31 +10,36 @@ redirects:
...
+
+    * `TABLE_NAME` is a string that names a stored table. Table names have a maximum character amount of 255 and can be a combination of alphanumeric characters, underscores, and colons.
+    * `YOUR_ACCOUNT_ID` is a number that the table belongs to.
```
ccobb-nr, line 40: Maybe flip these, putting "YOUR_ACCOUNT_ID" first, since it comes first in every path.

ccobb-nr:
I know you weren't sure about collapsing things, and I highly recommend **_NOT_** making all the collapse changes.  That just hide all the key information of this doc, and create a lot of extra work for the user.  (Leave the existing ones in the example sections.)

Collapsing things makes sense on pages like the [NRQL Reference](https://docs.newrelic.com/docs/query-your-data/nrql-new-relic-query-language/get-started/nrql-syntax-clauses-functions/) page, where your listing a grammar and functions of a whole language, but less sense for an REST API definition like this that defines basic CRUD functions.

Also, note that the [Log API](https://docs.newrelic.com/docs/logs/log-api/introduction-log-api/) page does not collapse most things.

PR #14805, opened 2023-10-09
feat(apm): clarified action for containerized environment with example

akristen:
Clarified some copy and added an example


rhetoric101 (approved):
@akristen this looks good to me!

PR #14810, opened 2023-10-09
clang/code cleanup

roadlittledawn:
## Description

Cleans up the repo some. Namely:

✅ remove old migration code

✅ remove codemods not being used?
- move utils to more relevant dir

✅ remove unused deps and dev deps (merged from another branch)

✅ remove unused imports (merged from another branch)

✅ remove unused components

✅ remove unnecessary local plugins

✅ remove unused github action workflows (validate datasource)

## Related
Closes https://issues.newrelic.com/browse/NR-160998


sunnyzanchi (commented):
what are the `dataSource` changes about? i'm not familiar with that property in frontmatter outside of the data-dictionary

roadlittledawn:
> what are the `dataSource` changes about? i'm not familiar with that property in frontmatter outside of the data-dictionary

this was originally going to be used to connect display content via docs library service in guided install in NR UI. but that project took another direction and these were never actually used. also the docs library service is now defunct, so they're completely unused

tabathadelane (dismissed):
Wasn't able to spot any issues caused by these deletions.

sunnyzanchi (changes requested):
there are still some places that are still querying `dataSource`. i got an error on the homepage from `docPage.js` but i think there are others

roadlittledawn:
> there are still some places that are still querying `dataSource`. i got an error on the homepage from `docPage.js` but i think there are others

nice catch! just removed dataSource from docPage and SEO. forgot we added those values as meta tags for swiftype in order for docs library service to filter for them.

sunnyzanchi (approved):
✨🧹💨 ✨

PR #14824, opened 2023-10-10
feat(Sonic): NRQL revamp

jeff-colucci:


jeff-colucci:
Merging with prior review

PR #14832, opened 2023-10-11
OpenTelemetry: Add tabs to Java tutorial

rhetoric101:
This doc has three tutorials and one of them is very long. The goal is to make this easier for readers to parse.


jeff-colucci (approved):
Vale has caught a few contractions and instances of passive voice, but I don't think they're gating so I'm going to approve and leave their fate up to you! Looks great!

PR #14838, opened 2023-10-11
Change phrasing of the browser proxy doc

ghost:
@cwli24 requested to change the language because the previous version made it sound like we do support proxy servers.


jeff-colucci (approved):
Looks good!

PR #14722, opened 2023-09-28
chore(k8s): create skeleton nav for future work

bradleycamacho:
Contains the skeleton of the new k8s nav.

PR #14840, opened 2023-10-11
[chore] update subchart name reference for newrelic-prometheus-agent

maxlemieux:
The name given for the New Relic Prometheus Agent Helm chart in the Kubernetes integration Helm docs currently is `prometheus-agent`, but the actual chart name is `newrelic-prometheus-agent` (ref: https://github.com/newrelic/newrelic-prometheus-configurator/blob/main/charts/newrelic-prometheus-agent/Chart.yaml#L2)

There is a risk of confusion at different stages of deployment, due to this mismatch.

This PR is to update this reference in docs, to match the actual subchart name.


bradleycamacho:
Hi @maxlemieux, thanks for the edit! We're actually about to roll out an update to the k8s install docs that combines the all the install docs into one. Instead of merging this PR in, I just updated the content in the new PR: https://github.com/newrelic/docs-website/pull/14722/commits/7a0fab488c2a1f790e2fe072dd0dc9350fc2daa0. You'll see this live in the next day or two - thanks!

PR #14847, opened 2023-10-12
Create Troubleshooting.mdx

jvadera:
Changing Troubleshooting location

## Give us some context


akristen (changes requested):
Hi team! @jvadera Thanks for submitting this PR. It looks like you broke out the Kafka config troubleshooting section into its own doc. This isn't a bad idea since the original Kafka config doc is pretty long and unwieldy. However, the troubleshooting section doesn't have a ton of content there. 

I have some follow up questions:

- Is there additional troubleshooting content that needs to be added to this doc? I can definitely see the utility of breaking this out if there are other troubleshooting use cases we can document here. To that end, I'm also open to hearing your POV about making it a separate doc as well. :) 

- If there isn't additional content, one option is to rework the original Configuration doc so the troubleshooting section isn't buried under a lot of tables and content. Want to hear your point of view first though!

PR #14862, opened 2023-10-13
fix: Fix Arguments to Python error_group_callback

TimPansino:
## Give us some context

* Fixes documentation around Python agent's error_group_callback to reflect the real arguments accepted.


rhetoric101 (approved):
@TimPansino thanks for updating these arguments!

PR #14883, opened 2023-10-17
feat: Added infra agent status endpoint settings

josemore:
## Give us some context

* The infrastructure agent provides an optional setting to enable a status / health check point.


jeff-colucci (approved):
A few style changes but looks good overall!

PR #14886, opened 2023-10-17
Fix left nav tutorials

akristen:
fixing left nav behavior for tutorial category


jeff-colucci (approved):
I clicked all the things and all the things did the things that the things were supposed to do, so I'm approving!

PR #14888, opened 2023-10-17
fix: Node agent compatibility typo

daffinito:
## Give us some context

Line 34, "The following are proposed time ranges. The actual release date may vary.", didn't belong there. It looked like a copy/paste error. The line is repeated further down on line 61 and looks to make sense in that location.


jeff-colucci (approved):
Looks good!

PR #14898, opened 2023-10-17
consolidate frontmatter dependencies, improve errors

sunnyzanchi:
this PR adds a new module,`scripts/utils/frontmatter.js`, to replace `@github-docs/frontmatter` and `front-matter` so we don't have multiple frontmatter dependencies that do basically the same thing with slightly different interfaces. our repo was split 50/50 between these two libraries, which are both unmaintained. this module is a small wrapper around a fork of `gray-matter`

this PR also improves error handling/messaging in the case of malformed frontmatter

### before
Screenshot 2023-10-17 at 4 41 29 PM

### after
Screenshot 2023-10-17 at 4 40 51 PM


scripts/utils/frontmatter.js
```diff
@@ -0,0 +1,84 @@
...
+  try {
+    const { content, data } = grayMatter(mdString);
+
+    result = {
+      attributes: data,
+      body: content,
+      error: null,
+    };
+  } catch (error) {
+    // make this common error message a little easier to understand
```
clarkmcadoo, line 67: left over comment, but i think the default error here is fine, you've got most of the error handling heavy lifting done w/in the various files

clarkmcadoo (approved):

clarkmcadoo (approved):

scripts/utils/frontmatter.js
sunnyzanchi: this was actually originally from `@github-docs/frontmatter`. i think they were just saying that this was to make the rather inscrutable 'can not read a block mapping entry' error make more sense 🤷 i went ahead and removed it

LizBaker (approved):
lgtm!

PR #14908, opened 2023-10-18
feat: Add support for boolean values in setCustomAttribute api

metal-messiah:
## Give us some context
This PR updates the browser agent setCustomAttribute() API docs to indicate support of allowing boolean-type data as an accepted value.


akristen (approved):
Thanks for updating this!

PR #14947, opened 2023-10-23
Update how-new-relic-distributed-tracing-works.mdx - added Trace Limi…

beatface:
…ts section

Added section to inform customers that we have internal tracing pipeline limits that do get increased periodically.


jeff-colucci (approved):
Looks good to me!

PR #14950, opened 2023-10-23
[fix] Update setup-prometheus-agent.mdx

maxlemieux:
The Prometheus Agent docs have `annotation` in many places where it should read `annotations` in the configuration.  This PR updates the examples to a working configuration.

Tested and confirmed that filtering requires the configuration `annotations` and will not work with `annotation`.

Please also see the Prometheus agent [chart values file](https://github.com/newrelic/newrelic-prometheus-configurator/blob/main/charts/newrelic-prometheus-agent/values.yaml) which documents the `annotations` usage.


jeff-colucci (approved):
Looks good to me!

PR #14959, opened 2023-10-24
chore: format output lines in bash codeblock

brnhensley:
The `[output]` in the bash codeblock should format those lines as output, without the `$` which indicates input
How it looks now:
image

what my change will do:
image


akristen (approved):
This is a good change that will definitely help our users! Thanks for the update + good eye :)

PR #14963, opened 2023-10-24
fix: groupID to groupId

brnhensley:
The doc went back and forth from `groupId` to `groupID`. I'm not positive if `groupID` is a problem, but I made it consistent with the way I know works.


rhetoric101 (approved):
@brnhensley thanks for catching this capitalization issue!

PR #14964, opened 2023-10-24
chore: update code examples

brnhensley:
I formated the toml config how [vector does in their docs](https://vector.dev/docs/reference/configuration/), and added the `toml` language identifier -  but that would likely need to be added to our prisma config to work like github shows


rhetoric101 (approved):
@brnhensley thanks for adding this formatting!

PR #14966, opened 2023-10-24
fix: downgrade autolink-headers plugin

tabathadelane:
- changing gatsby-remark-autolink-headers version resolves bug
- tried upgrading to 5.25.0 and it did not resolve.
- installed latest v4 and it resolves
- there is a new major version available on node 18 we could revisit

`/docs/query-your-data/explore-query-data/dashboards/manage-your-dashboard/#filter-legend`
 should scroll to **Filter using the chart legend** and not **Export and share your data**

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "gatsby-plugin-meta-redirect": "^1.1.1",
     "gatsby-plugin-react-helmet": "^5.4.0",
     "gatsby-plugin-sharp": "4.25.0",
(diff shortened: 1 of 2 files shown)


clarkmcadoo:
I think we're still stuck, the "Export and Share your data" header is getting the `filter-logs` hash
Screenshot 2023-10-24 at 2 45 56 PM

tabathadelane:
> /docs/query-your-data/explore-query-data/dashboards/manage-your-dashboard/#filter-legend

whyyy

PR #14967, opened 2023-10-24
Update send-prometheus-metric-data-new-relic.mdx

wildcardlinux:
Confirmed our prometheus agent default scrape value is 30s not 15s. The global default within prometheus documentation is 1m. Updated the recommendations sections to reflect the correct values.

Per prometheus documentation, the global scrape_interval is set to 1 minute by default. https://prometheus.io/docs/prometheus/latest/configuration/configuration/#configuration-file

Our prometheus agent is set to 30 seconds by default. I double checked with Engineering. https://github.com/newrelic/newrelic-prometheus-configurator/blob/7dcb9392349cef3d2a44aad5a0371e9719ea12ef/charts/newrelic-prometheus-agent/values.yaml#L134

https://newrelic.slack.com/archives/C04S2FT1LU9/p1698164053929219?thread_ts=1698162082.068379&cid=C04S2FT1LU9

### DIFF
--- a/src/content/docs/infrastructure/prometheus-integrations/get-started/send-prometheus-metric-data-new-relic.mdx
+++ b/src/content/docs/infrastructure/prometheus-integrations/get-started/send-prometheus-metric-data-new-relic.mdx
@@ -58,7 +58,7 @@ Benefits:
 
 
 Recommendations:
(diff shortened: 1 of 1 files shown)


rhetoric101 (approved):
@wildcardlinux thanks for correcting the default value!

PR #14968, opened 2023-10-24
fix: NR-172682 position() returns null if not found

mlobiondo-newrelic:


src/content/docs/query-your-data/nrql-new-relic-query-language/get-started/nrql-syntax-clauses-functions.mdx
```diff
@@ -2923,10 +2923,7 @@ Note: `aparse()` is case-insensitive.
 
   Returns:
     * The 0-based index of the starting character of the substr within str
-    * If the referenced occurrence of substr is not found, -1 is returned
-    * Return null if str or substr is null
-    * Searches are case-sensitive
-
+    * Null, if str is null, substr is null, or the referenced occurrence of substr is not found
```
akristen, line 2926: Hi @mlobiondo-newrelic. Thanks for getting this updated. This line isn't clear to me. Is it saying that:

When the `position()` function is used, it will return null if you query a string or substr that is null? 

Can you unspool what's meant by "referenced occurence of substr is not found"? What's the difference between a substr being null/not found and a referenced occurrence of a substr being not found?

mlobiondo-newrelic: Hi @akristen,
* str is null example: SELECT position(banana, 'a') FROM PageView (there's no attribute named "banana" in PageView, so it will be null)
* substr is null example: SELECT position('a string', banana) FROM PageView (same, no attribute named "banana")
* Referenced occurrence not found:
** SELECT position('banana', 'q') FROM <whatever> (there's no "q" in "banana"), or
** SELECT position('banana', 'a', 5) FROM <whatever> (there are not 5 occurrences of "a" in "banana")

akristen (approved):
Looks good!

PR #14971, opened 2023-10-24
Clark/docs by last date edited

clarkmcadoo:
run `node scripts/docsByLastDateEdited.mjs` to test.  Should create a CSV in the root directory with all file names and last edited date


scripts/docsAndLastDateEdited.mjs
```diff
@@ -0,0 +1,28 @@
...
+import { stringify } from 'csv-stringify';
+
+const allDocs = await glob('src/content/docs/**/*.mdx', {
+  ignore: '**/index.mdx',
+});
+
+const allDocsAndDates = allDocs.map((doc) => {
+  // mtime is the last modified time of a particular file
+  // https://nodejs.org/api/fs.html#statsmtime
+  const modifiedDate = statSync(doc).mtime;
```
roadlittledawn, line 14: i think we need to use git data and not the file system. using the file system will only tell you last time it was modified on machine's file system and not edit history via git commits

roadlittledawn (changes requested):
a couple quick notes:
- can you take that new eng PR template for a spin with this?
- can we add doc on this somewheres?
- i think you'll need to use git history and not file system modified date

clarkmcadoo:
moved to https://github.com/newrelic/docs-website/pull/15009 to use the PR checklist

PR #14982, opened 2023-10-25
Add env var for building languages individually

LizBaker:
### ✨ Summary
This adds logic around a BUILD_LANG env var. if you don't specify it all languages will be included in the build. if you do specify it it will ignore all other languages (except english). if you specify english (or any string that isn't an i18n language) it will only build english. it does build english no matter what but this ensures we get english backup pages. it still cuts build time down by 60% and could be acceptable for now without us having to refactor to keep the backup logic. I added BUILD_LANG=en to the start script to keep it in line with what it was doing (not building any other languages) and with no BUILD_LANG in the build script it should behave as normal in the preview

### 🔮 Engineering checklist

- [x] The build preview has been checked for expected functionality on desktop browsers


sunnyzanchi (changes requested):
i'm not sure this is working as expected. i ran
```
BUILD_LANG=jp
```
and i can go to localhost:8000/jp to see the Japanese homepage and the nav is in Japanese. but trying to navigate to any other page shows the English version, even though the URL still has /jp and the nav is in Japanese

LizBaker:
> i'm not sure this is working as expected. i ran
> 
> ```
> BUILD_LANG=jp
> ```
> 
> and i can go to localhost:8000/jp to see the Japanese homepage and the nav is in Japanese. but trying to navigate to any other page shows the English version, even though the URL still has /jp and the nav is in Japanese

this is just how the site behaves when built in english (things like the nav are sourced separately from the mdx pages). you'd need to remove or edit the BUILD_LANG=en in the start script because that's overriding yours. it's in the start script to speed up local dev by not building any languages other than english

clarkmcadoo (approved):

clarkmcadoo:
worked for me!

sunnyzanchi:
> this is just how the site behaves when built in english (things like the nav are sourced separately from the mdx pages). you'd need to remove or edit the BUILD_LANG=en in the start script because that's overriding yours. it's in the start script to speed up local dev by not building any languages other than english

🤦 my brain was not working on Friday. i just tried with `BUILD_LANG=jp yarn develop` and it works as expected

PR #15005, opened 2023-10-27
Update_log_drop_rules_docs

marcelschlapfer:
Updated screenshots for new Drop rule creation UI that includes partition selector.
Also I added a CAUTION about using Partitions and Drop rules.  Drop rules are partition specific, so making changes to partitions can cause drop rules to no longer match intended logs.  This could lead to logs not being dropped as exepected/intended after partition changes.

## Give us some context


akristen (approved):
Looks good to me!

PR #15056, opened 2023-11-02
Update java-agent-configuration-config-file.mdx

jasonjkeller:
Remove use_private_ssl config option as it no longer exists in the Java agent.


rhetoric101 (approved):
@jasonjkeller thanks for keeping the config file up to date!

PR #15059, opened 2023-11-02
fix(infra): remove invalid metric from metrics list of AWS DynamoDB Integration

mitchang:
Remove invalid one from metric list of AWS DynamoDB API Polling.

'userErrors' metric is region/account specific, it is not included dynamoDbTable provider.
The metric should be in dynamoDbRegion, which is already listed.

AWS Docs: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/metrics-dimensions.html#UserErrors


rhetoric101 (approved):
@mitchang thanks for removing this row!

PR #15060, opened 2023-11-02
nri-jmx: fix downloader path

rubenruizdegauna:
This PR fixes the path to nri-jmx downloader (currently points to an old version)


rhetoric101 (approved):
@rubenruizdegauna thanks for updating the file name! Don't forget that you don't need to change the Japanese or Korean files because we have an automated job that will do that for us.

PR #15069, opened 2023-11-02
feat(PHP): Adds log context data documentation

mfulb:
This PR contains updates documentation for the PHP agent for the added support of log context attributes for the Monolog library.

I would like this PR reviewed but please do not merge until I let you know.


akristen (approved):
Approving so docs can pair with the release, but am making a second PR with docs style edits

akristen (approved):

PR #15094, opened 2023-11-06
docs: .NET Agent multiple app names via environment variable

tippmar-nr:
## Give us some context

Updates the .NET Agent docs to include an example of setting multiple app names via an environment variable.


jaffinito (approved):

rhetoric101 (approved):
@tippmar-nr thanks for adding this example!

PR #15097, opened 2023-11-06
chore: Update get-started-kubecost.mdx to not preview status

brettj-newrelic:
Notes that kubecost integration with NR is currently in preview.

## Give us some context
Kubecost integration for NR is not thoroughly tested; this makes that clear to readers.


rhetoric101 (approved):
@brettj-newrelic thanks for clarifying that this feature is in preview.

PR #15098, opened 2023-11-06
fix: correct support of gauge types in the metric api

thezackm:
[gauge](https://docs.newrelic.com/docs/data-apis/ingest-apis/metric-api/report-metrics-metric-api/#json-payload-keys) is the default metric type used by the Metric API and should definitely be supported.


rhetoric101 (approved):
@thezackm thanks for catching that.

PR #15139, opened 2023-11-09
.NET agent 10.19.1 release notes

nr-ahemsath:
Release notes for .NET Agent 10.19.1


bradleycamacho:
hi @nr-ahemsath, thanks for the contribution! Looks like the build was failing because you included some unsupported front matter. I just removed those and will get this merged in the morning. Thanks!

bradleycamacho:
Hi @nr-ahemsath , sorry I was entirely wrong! The build was erroring out because of the front matter, and when I referenced other .NET release notes I didn't see any of those additional front matter sections. I must have only clicked on files published before we started using those.

I think the issue was actually just the formatting of the release date. I just pushed up a fix (hopefully).

nr-ahemsath:
@bradleycamacho Anything I can do to help get this merged?

PR #15170, opened 2023-11-10
PHP agent: Clarify multiple call behavior

rhetoric101:
This is to address customer feedback in NR-79484.


jeff-colucci (approved):
Looks good to me!

PR #15189, opened 2023-11-14
fix: broken link on netflow setup doc

thezackm:
small fix for a broken link to a 3rd party site


akristen (approved):
Link works! Thanks for fixing this

PR #15202, opened 2023-11-14
Install java guided install

akristen:
Some <10 minute fixes from burndown


jeff-colucci (approved):
There's a vale callout for haven't instead of have not, but other than that looks good to me!

PR #15227, opened 2023-11-15
chore(k8s): Fix title case in left nav

austin-schaefer:


bradleycamacho (approved):

bradleycamacho:
Good catch, I'll do a quick sweep of the k8s docs as a whole for title casing tomorrow since it seems to be something I miss!

PR #15234, opened 2023-11-16
feat(apm): remove errors tutorial from apm category

bradleycamacho:
Removes the errors tutorial from the install section of the APM category. We noticed this in a feedback burndown session and agreed this isn't the right place for a link to live.


jeff-colucci (approved):
Seems removed to me!

PR #15239, opened 2023-11-16
AWS Metric naming revamp

morecar:
This brings AWS metric naming section up to speed with its Azure counterpart.


jeff-colucci (approved):
Minor style tweak, otherwise looks great!

PR #15249, opened 2023-11-16
docs: update docs to reflect REST V1 eol

tracyatrf:
The REST V1 API should be considered EOL. These endpoints are being turned off for API access. Users are encouraged to move to NerdGraph.


jeff-colucci (approved):
Looks good to me!

jeff-colucci (approved):
Approving

PR #15259, opened 2023-11-17
chore(accounts): Tweak ID to remove dated term

austin-schaefer:
https://docs.newrelic.com/docs/style-guide/word-choice/usage-dictionary/#black-white


akristen (approved):
Good update! Always happy when we move to more inclusive language. :)

PR #15278, opened 2023-11-20
Chore/meraki config sample

thezackm:
adding a full config snippet for Meraki API polling and updating some of the `$VAR` placeholders to use the `$YOUR*` styling.


jeff-colucci (approved):
Looks great!

PR #15283, opened 2023-11-20
chore(Java): updating Mule compatibility

meiao:
Updating the compatibility of the Java agent with Mule server.


jeff-colucci (approved):
Looks good!

PR #15289, opened 2023-11-21
Document EU PrivateLink Endpoints

KevinCorcoran:
This change introduces a `<Tabs>` component and adds the EU endpoints to the documentation on configuring New Relic to use AWS PrivateLink.


jeff-colucci (approved):
Had to fix the tabs but other than that looks good!

PR #15294, opened 2023-11-21
Fix grey block formatting

hannahramadan:
Fix grey block formatting, [example](https://docs.newrelic.com/docs/apm/agents/ruby-agent/getting-started/ruby-agent-requirements-supported-frameworks/#databases). The style and rowSpans seem to be the outliers where colors are messed up, so I'm hoping the removal will fix this issue.


jeff-colucci:
@hannahramadan looks like the issue was caused by there being no entry for the problem columns, I think I've pushed up something to fix all of them. Only time (and Gatsby) will tell.

jeff-colucci (approved):

hannahramadan:
Thank you @jeff-colucci!

PR #15313, opened 2023-11-24
chore: added missing infra-agent status endpoint example

josemore:
## Give us some context

* Added a missing example showing how to query the infra-agent status endpoint.


akristen (approved):
Looks good!

PR #15326, opened 2023-11-27
fix: Update drop rules doc to specify that JOIN and subqueries cannot be used

ecchilds:
Joins and subqueries are not allowed in drop rules, but presently our docs do not reflect that. This PR adds those restrictions.

Jira: NR-172145


akristen (approved):
Looks good, thanks for the update!

PR #15343, opened 2023-11-28
Resolve image import and MDX errors

tabathadelane:
Clean up from a minor bug that allowed an image to be imported anywhere in the MDX content, and it will render with that const as an img src in a file without needing to be imported there. This has probably happened by cut+pasting content to a new doc and accidentally leaving the import statements behind. 

This doesn't effect functionality, but is throwing errors in our verify-mdx script, so we want to remove this noise. Now that verify-mdx is a required github check on PR's this shouldn't continue to happen as these would get flagged in the review process.


sunnyzanchi (dismissed):

sunnyzanchi (approved):

tabathadelane:
[build successfully here](https://www.gatsbyjs.com/dashboard/34e708d2-68a1-48ca-b66f-d2604890b061/sites/a0b77314-cdea-4790-8fde-7d3457d590ce/pull-requests/30e37eed-45d4-4803-9313-4a4abadfd31a/builds)

clarkmcadoo (approved):

PR #15375, opened 2023-11-30
Update definition of upstream and downstream

joeljames:

## Give us some context

* Users operate under different definitions of upstream and downstream, so we try not to use those terms in the UI or docs. This definition will match the new tooltip we are adding to the UI.


bradleycamacho:
hi @joeljames! Thanks for the edit! Could you sign the CLA by clicking the yellow button above so I can review and merge this? Thanks!

bradleycamacho (approved):

PR #15377, opened 2023-12-01
fix(api keys): Clarify original key behavior/name

austin-schaefer:


src/content/docs/apis/intro-apis/new-relic-api-keys.mdx
```diff
@@ -222,7 +222,7 @@ Here's how to rotate each of our four most common API keys:
 
       You can find license keys for your account in the API keys UI at [**one.newrelic.com/api-keys**](https://one.newrelic.com/api-keys) (or [**one.eu.newrelic.com/api-keys**](https://one.eu.newrelic.com/api-keys) for our EU datacenter). You'll only see keys you have permission to use, unless you have [All product admin permissions](/docs/accounts/accounts-billing/new-relic-one-user-management/user-management-concepts/#standard-roles) or a custom role that enables you to see those keys.
 
-      By default, all New Relic accounts are created with an organization-level license key. You can't delete this key from the API keys UI, but you can contact New Relic support to remove your key.
+      By default, all New Relic accounts are created with an original account license key, which is shared by the entire organization. You can't delete this key from the API keys UI, but you can contact New Relic support to rotate your key.
```
rhetoric101, line 225: @austin-schaefer for this collapser, you might consider making it clearer that there are two types of keys. I was initially confused because the transition from "original account license key" to "user-created license key" was not noticeable. How about something like this inside the collapser?

```
**Original account license key**

By default, all New Relic accounts are created with an original account browser key, which is shared by the entire organization. You can't delete this key from the API keys UI, but you can contact New Relic support to rotate your key.

**User-created license keys**

For license keys that your users create, here's how to rotate them: 
...
```

src/content/docs/apis/intro-apis/new-relic-api-keys.mdx
```diff
@@ -279,7 +279,7 @@ Here's how to rotate each of our four most common API keys:
 
       You can find browser keys for your account in the API keys UI at [**one.newrelic.com/api-keys**](https://one.newrelic.com/api-keys) (or [**one.eu.newrelic.com/api-keys**](https://one.eu.newrelic.com/api-keys) for our EU datacenter). You'll only see keys you have permission to use, unless you have [All product admin permissions](/docs/accounts/accounts-billing/new-relic-one-user-management/user-management-concepts/#standard-roles) or a custom role that enables you to see those keys.
 
-      By default, all New Relic accounts are created with an organization-level browser key. You can't delete this key from the API keys UI, but you can contact New Relic support to remove your key.
+      By default, all New Relic accounts are created with an original account browser key, which is shared by the entire organization. You can't delete this key from the API keys UI, but you can contact New Relic support to rotate your key.
```
rhetoric101, line 282: @austin-schaefer similar to the account license key collapser, you might consider making it clearer that there are two types of keys. How about something like this inside the collapser?

```
**Original account browser key**

By default, all New Relic accounts are created with an original account browser key, which is shared by the entire organization. You can't delete this key from the API keys UI, but you can contact New Relic support to rotate your key.

**User-created browser keys**

For browser keys that your users create, here's how to rotate them: 
...
```

rhetoric101 (approved):
@austin-schaefer I made a few small suggestions.

PR #15410, opened 2023-12-05
[Do not publish, release TBA] logs_timestamp_support_docs_page

marcelschlapfer:
DO NOT MERGE TO PRODUCTION DOCS.  JUST WANT THIS TO BE IN DEVELOP SO IT CAN BE REVIEWED AND EDITED BEFORE FINAL PUSH TO PROD.

Creating a docs page for how we support timestamps in NR Logs

## Give us some context


src/content/docs/logs/ui-data/timestamp-support.mdx
```diff
@@ -0,0 +1,200 @@
...
+
+There are many, many different timestamp formats as you can see in some of these example logs below.  There is no single standard for the format of a timestamp.  They may or may not contain certain fields like the date, ms precision, or the timezone. 
+
+```
+time="2023-11-16T14:50:44Z" level=info msg="checking pixie-operator.v0.1.4"
```
akristen, line 18: @marcelschlapfer Do we need to keep this example? I'm reworking the structure and I think the content works better without this particular example, but I won't remove it if your team thinks it's necessary for the integrity of the doc.

src/content/docs/logs/ui-data/timestamp-support.mdx
```diff
@@ -0,0 +1,200 @@
...
+  - Logs
+  - Timestamp
+  - Timestamps
+  - Forwarding Logs
+metaDescription: "How New Relic handles timestamps in logs."
+---
+
+## Introduction [#introduction]
+
+One of the most important attributes in a log event is the **timestamp**.  It is the exact moment in time that tells when the event happened and is critical for ordering events in chronological order as well as for searching logs in specific timeframes.  The most basic question all logs answer is **what happened, and when** it happened.  In most cases, the timestamp will appear somewhere in the beginning of the log event, but this is not always the case.  In some cases, it may appear much later in the log.  In logs that are in a JSON format they will usually contain some attribute that identifies the timestamp, like `timestamp` or `log_timestamp`, `time`, etc.  Some logs may not contain timestamps at all, and in these cases New Relic will always assign a timestamp automatically to the event at the time it is ingested into the system.  New Relic also indexes the timestamp of each log, since they are crucial for efficient querying.
```
akristen, line 13: For this sentence: Some logs may not contain timestamps at all, and in these cases New Relic will always assign a timestamp automatically to the event at the time it is ingested into the system.

Isn't this the default behavior anyway? Do we need to call this out if New Relic is always going to add the ingest time of the timestamp unless there's an override set up?

marcelschlapfer: It is the default behavior, I'm fine with removing this if you don't think it's necessary.  Was just trying to point out the default behavior somewhere in the intro even though I do call it out in the very beginning of the next section.

src/content/docs/logs/ui-data/timestamp-support.mdx
marcelschlapfer: You can remove it.  I just added several examples from logs we had in our test platform.  The point can still be made with 4 examples vs 5.

src/content/docs/logs/ui-data/timestamp-support.mdx
```diff
@@ -0,0 +1,200 @@
...
+{"bytes" : "10009528", "cacheStatus" : "1", "cliIP" : "172.16.103.13", "timestamp" : "1700146596212", "tlsOverheadTimeMSec" : "2"}
+```
+
+## How timestamps work for logs [#how_timestamps_work]
+
+**By default**, New Relic assigns the value of the log `timestamp` attribute (a reserved keyword) at the exact time when the log is ingested, i.e. ingest time.
+
+There are two options to override this default behavior and assign the `timestamp` value based on a `timestamp` contained in the log.
+1. If we receive a [JSON formatted log](#timestamps_in_json_logs) containing a `timestamp` attribute in one of our supported time formats, we will override the ingest timestamp value with the value of the timestamp from the json attribute.
+2. You can use [parsing rules](#timestamp_parsing_rules) at ingest to override the ingest `timestamp` value to the timestamp value that is contained in the log.  Using a parsing rule you can extract out our supported time formats with simple Grok expressions, or by using the `datetime` Grok type and a `pattern` to define the custom timestamp format in your log.
```
akristen, line 37: @marcelschlapfer what does Grok type and Grok expressions mean in this context?

marcelschlapfer: A example of a Grok expression would be `%{NUMBER:timestamp:long}`, the grok type in this case is `long`.
Another example in the context of timestamps would be the expression `%{DATA:timestamp:datetime;MM/dd/yyyy HH:mm:ss.SSS}`, where the grok type is `datetime`.

src/content/docs/logs/ui-data/timestamp-support.mdx
```diff
@@ -0,0 +1,200 @@
...
+```
+"timestamp": 1701445586607,
+"message":"This is my sample JSON log message with timestamp attribute in ISO8601 format"
+```
+
+## Setting the log timestamp using Parsing rules [#timestamp_parsing_rules]
+
+You can use **parsing rules** at ingest to override the ingest `timestamp` value to the timestamp value that is contained in the log.  Using a parsing rule you can extract out our supported time formats with simple Grok expressions, or by using the `datetime` Grok type and `pattern` to define the custom timestamp format in your log. 
+
+- The attribute name extracted by the expression must be `timestamp` and its data type string or long. If it is not provided, it defaults to string.  Note that for epoch timestamps with microsecond or nanosecond precision, you must include the `long` datatype even though timestamp precision will be truncated to milliseconds.
```
akristen, line 123: what does "and its type string or long" mean in this context?

akristen: maybe it's missing a verb. 'and its data type must be string or long'

src/content/docs/logs/ui-data/timestamp-support.mdx
```diff
@@ -0,0 +1,200 @@
...
+```
+
+We will override the ingest `timestamp` with the ISO8601 `timestamp` (converted to epoch) contained in the log, and the result will look like this:
+
+```
+"timestamp": 1701445586607,
+"message":"This is my sample JSON log message with timestamp attribute in ISO8601 format"
+```
+
+## Setting the log timestamp using Parsing rules [#timestamp_parsing_rules]
```
akristen, line 119: I'm a little lost at this section. I'm not sure what Grok expressions are and how they relate to parsing rules. This section also feels like it could be broken out into two, too. One is more reference material with some examples of how to use Grok expressions (?) and the second is to set up parsing rules specific to timestamps. What is the connective tissue here? @marcelschlapfer

src/content/docs/logs/ui-data/timestamp-support.mdx
```diff
@@ -0,0 +1,325 @@
+---
+title: Understanding and using timestamps in logs
```
akristen, line 2: Since this is reference material, it should be OK to follow the ui-data naming conventions. I think 'Timestamp formats in logs' is good enough

akristen (approved):
Looks good!

PR #15411, opened 2023-12-05
Update punctuation.mdx

bhecknr:
After discussion with Robert Friesel, added additional info in "Dashes and hyphens" section.


jeff-colucci (approved):
Looks good to me

PR #15425, opened 2023-12-06
feat(Sonic): remove stale ymls

jeff-colucci:
removing unneeded yml files


jeff-colucci:
Checked the left nav to make sure the page links work. Merging with approval from sprint review meeting.

PR #15464, opened 2023-12-11
infra-agent macos version support

josemore:
## Give us some context

* Updated supported version of the infrastructure agent on macOS, which is aligned with Apple & Brew support on the most recent 3 major versions.


jeff-colucci (approved):
Looks good!

PR #15475, opened 2023-12-12
fix: nvidiagpu integration default event name

josemore:
## Give us some context

* Updating integration docs based on feedback, default event name is different. 
* Added link to example source.


akristen (approved):
Gave the content a light refresh. Looks good to merge!

PR #15480, opened 2023-12-12
chore(HordeIPs): Update post

bpeckNR:
Clarified the what's new post describing what action will be taken with legacy Horde IPs on January 15, 2024.


akristen (approved):
looks good!

PR #15504, opened 2023-12-13
Fix codeblock indent

akristen:
Un-indented code block to get block to render correctly without image and callout


jeff-colucci (approved):
Looks good to me!

PR #15543, opened 2023-12-15
The output is named `newrelic`

luckslovez:
I am not sure why [this change](https://github.com/newrelic/docs-website/commit/7bd25c0996076b0517b2448cd19087a55f72de49) made it, but the plugin is actually named `newrelic`.


luckslovez:
I think there was a confusion here, since `nrlogs` is the name of the output plugin maintained by fluentbit https://docs.fluentbit.io/manual/pipeline/outputs/new-relic

rhetoric101:
After speaking with @luckslovez, it sounds like our customer who suggested the original change to `nrlogs` was using Fluent Bit's own plugin--not the New Relic-made plugin from here: https://github.com/newrelic/newrelic-fluent-bit-output

We will accept this change back to `newrelic`.

rhetoric101 (approved):
@luckslovez thanks for catching our mistake on this!

PR #15548, opened 2023-12-15
chore(monitorIDMapping): Added example query

bpeckNR:
Added a query example to retrieve a monitor entity GUID using the monitor ID


rhetoric101 (approved):
@bpeckNR thanks for adding this example!

PR #15553, opened 2023-12-15
APIs: Remove link to outdated video.

rhetoric101:
Now that the NerdGraph experience has moved into the New Relic UI and has some other minor internal layout changes, I'm removing the link to the outdated video.


paperclypse (approved):
Looks good

PR #15567, opened 2023-12-18
Document UTF-16LE handling

maxlemieux:
Add section to provide directions for UTF-16LE -> UTF-8 transform.

In some Windows environments, the Microsoft SQL Server error logs are written to disk in UTF-16LE encoding. Sending them directly to New Relic can result in formatting and display issues for the data stored to NRDB, since New Relic's API requires UTF-8 encoding ([reference](https://docs.newrelic.com/docs/data-apis/custom-data/custom-events/data-requirements-limits-custom-event-data/#general)).

Configuration is required in order to send any UTF-16 logs as UTF-8. Fluentd supports the required transformation between encodings ([reference](https://docs.fluentd.org/input/tail#encoding-from_encoding)).


jeff-colucci (approved):
Just some minor style changes but it looks great!

PR #15568, opened 2023-12-18
Browser: Remove js errors doc

rhetoric101:
Since the JS Errors doc under browser was migrated to errors inbox, remove the JS Errors doc and insert a redirect for it into the errors inbox grouping doc for browser. Also, insert a dummy doc link in place of the old JS Errors nav link (change it to Errors (errors inbox)).


homelessbirds (approved):
Works like a charm :)

PR #15615, opened 2023-12-21
fix(Hero): fixing job manager release note submissions

jeff-colucci:
Attempting to consolidate job manager and fix the non-updating overview list


akristen (approved):
Looks good!

PR #15636, opened 2023-12-22
fix(new relic solutions): removed slack collaboration

homelessbirds:


jeff-colucci (approved):
Looks great! Verified the nav works in the build, good to go!

PR #15639, opened 2023-12-22
(chore)Alerts: update image positioning

jeff-colucci:
Changed the positioning of the image in the doc to be less confusing


jeff-colucci:
Since this was just moving code from one place to another and there were no other changes, I verified that it worked in the build so I am publishing without review.

PR #15651, opened 2023-12-23
fix(mobile): add callout about platform-specific methods

ghost:
Added a callout for methods missing hybrid-specific instructions.


ally-sassman:
@ndesai-newrelic , we received customer feedback specifically on the setAppVersion page that it's missing a React Native tab, so I want to add a callout to the top of the 3 pages that only have Android/iOS methods. Can you review my language and confirm if it's accurate?

paperclypse (approved):
Looks good!

PR #15652, opened 2023-12-23
fix(api): update the UI instructions for adding a new api key

ghost:
Addresses this customer feedback: NR-186872


jeff-colucci (approved):
Looks good!

PR #15663, opened 2023-12-27
fix(alerts): update old screenshot and UI references

ghost:
This PR addresses customer feedback about the UI being outdated.


jeff-colucci (approved):
There is some passive voice but I don't think it's gating so I will approve!

PR #15665, opened 2023-12-27
Update header anchor position and add collapser header copy button

LizBaker:
This moves the header anchor link to the right of headers (to match the positioning in collapser headers) and also adds copy functionality to collapser headers via the theme update

### DIFF
--- a/plugins/gatsby-remark-autolink-headers-custom/index.js
+++ b/plugins/gatsby-remark-autolink-headers-custom/index.js
@@ -20,8 +20,7 @@ module.exports = (
     className = `anchor`,
     maintainCase = false,
     removeAccents = false,
-    enableCustomId = false,
-    isIconAfterHeader = false,
+    enableCustomId = true,
     elements = null,
   }
 ) => {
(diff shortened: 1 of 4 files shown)


plugins/gatsby-remark-autolink-headers-custom/index.js
```diff
@@ -64,10 +63,15 @@ module.exports = (
     if (icon !== false) {
       patch(data.hProperties, `style`, `position:relative;`);
       const label = id.split(`-`).join(` `);
-      const method = isIconAfterHeader ? `push` : `unshift`;
+      const method = `push`;
       node.children[method]({
```
sunnyzanchi, line 67: kind of a nit but if the method isn't being chosen dynamically anymore we should just call it
```suggestion
      node.children.push({
```

sunnyzanchi (changes requested):

sunnyzanchi (approved):

clarkmcadoo (approved):

PR #15669, opened 2023-12-27
fix(synth): redo a previous error changing gigabyte abbreviation

ghost:
Reverting a recent change. The original doc says GiB, which is different than a gigabyte (GB).


paperclypse (approved):
I had no idea GiB was a thing....

jeff-colucci (approved):
Looks good!

PR #15687, opened 2023-12-29
fix(PHP): Add Amazon Linux 2023

rhetoric101:
Add agent support for Amazon Linux 2023


bradleycamacho (approved):

rhetoric101:
I am waiting for SME confirmation about these two entries.

rhetoric101:
It turns out that Amazon Linux 2023 with PHP isn't supported (except with a tarball), so we're not going to publish this right now.

PR #15702, opened 2024-01-02
chore(NRQL): changing query builder info in docs

jeff-colucci:
Docs feedback: number 211423


akristen (dismissed):

akristen (approved):
Looks good! i saw one of the docs had query from anywhere bolded and i'm not sure that's within style guides. that said i have no idea what our bolding UI convention is and just know that i'm usually wrong when i bold anything LOL. but i think this can go out as is for sure!

PR #15721, opened 2024-01-03
fix(mobile): misc post-release fixes to Unity doc

ghost:
Two issues:
- The current screenshot doesn't match what the instructions say, so I'm working with a SME to get a screenshot that shows a sample app key pasted in the `appToken` field.
- Duplicate of "new relic" reference in Step 3


src/content/docs/mobile-monitoring/new-relic-mobile-unity/monitor-your-unity-application.mdx
```diff
@@ -89,13 +89,6 @@ To enable New Relic to authenticate your Unity app's data, follow these steps:
     2. Go to **Tools > New Relic > Getting Started** to open the New Relic editor window.
```
ally-sassman, line 89:
```suggestion
    2. Go to **Tools > New Relic > Getting Started** to open the editor window.
```

src/content/docs/mobile-monitoring/new-relic-mobile-unity/monitor-your-unity-application.mdx
```diff
@@ -89,13 +89,6 @@ To enable New Relic to authenticate your Unity app's data, follow these steps:
...
     3. Paste your copied application token(s) into the field labeled `appToken = ""`. If your app is deployed to both iOS and Android, you must provide separate tokens for each platform.
```
ally-sassman, line 91:
```suggestion

<img
  title="Unity AppToken Setup"
  alt="Add App Token in Unity Apps"
  src={unityAgentAppTokenSetup}
  style={{ maxWidth: '70%' }}
/>
```

src/content/docs/mobile-monitoring/new-relic-mobile-unity/monitor-your-unity-application.mdx
paperclypse, line 1: A few thoughts, all super minor:
* It might be helpful to have a screenshot of the guided installation UI. That might draw more people in. I worried that otherwise it might be a bit easy to skim past it.
* I'm torn about review requirements being its own step... Also, is there a way to summarize the info in "Android-native" and "iOS-native" requirements links? Maybe there's a way to abstract it to something like: We support the the SDKs and frameworks from the last X years. For more info, see our detailed Android-native requirements.
* I'd put the Unity package manager screenshot between the step title and the step text.
* Is there a way to rewrite "Add application token(s)" so we're not using the (s)?
* Look for places to simplify your wording. For example, "retrieve" could be "get", "utilizing" could be "use", etc
* I worry that Step 4 of 5 is a bit confusing. "If you're using Unity 2019 or later, complete the following steps" doesn't tell me what to do if I'm using something older and that info is sort of buried at the end of sub-step 2.
* "If you are utilizing an older version of Unity Studio, you can incorporate a lower version of the classpath as a dependency which supports Gradle version 7.0 and lower." this sentence is a bit of a tangle. Worth tightening up for  clarity, I think.

Otherwise, I think this looks good!

paperclypse: I like the specific use cases in the intro.
If we can, I think it would be nice to provide a little more context for why someone would want to choose between the guided installation and the manual installation, but I think that's probably outside the scope of this PR. (It's relevant to many docs)

homelessbirds (approved):

homelessbirds (approved):

PR #15738, opened 2024-01-05
fix(nri-mongodb): specify Atlas as special as of MongoDB metric attribute

sigilioso:

## Give us some context

* Specify that the `cl_role` attribute is empty for Atlas instances. Discovered by @kang-makes  while giving support in a GTSE.


sigilioso:
We keep it as a draft because we may include more details, depending on an on-going investigation cc/ @gsanchezgavier

sigilioso:
> We keep it as a draft because we may include more details, depending on an on-going investigation cc/ @gsanchezgavier

It was con concluded and there is no need of more details for now. So this is ready for review.

akristen (approved):

PR #15741, opened 2024-01-05
Nginx refresh

bradleycamacho:
Reviewer: 

Please take a look at the NGINX branching doc this implements. Please click around and take a look at all the options possible. I highly recommend having the PR open in one window and the preview in another to ensure you can track the content and that you're seeing as much of the content as possible.

Ignore the MongoDB content within the PR. 

A few notes:

* The intro is temporary. I'm working with SMEs to create a more OHI-specific intro
* The screenshots are for the wrong OHI and will be updated by the time this goes live
* Please take a look at a micro level (copyedits, small scructure edits, etc.) and also macro level (general flow, ordering of steps, size of steps, how skimmable the doc is, etc.)

Please reach out to me if you have any questions!

### DIFF
--- a/src/install/nginx/linux/install-tarball.mdx
+++ b/src/install/nginx/linux/install-tarball.mdx
@@ -1,0 +1,32 @@
+---
+componentType: default
+headingText: Download using tarball
+---
+
+    <Callout variant="tip">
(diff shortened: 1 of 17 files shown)


src/install/nginx/linux/install-tarball.mdx
```diff
@@ -0,0 +1,32 @@
+---
+componentType: default
```
rhetoric101, line 2: @bradleycamacho there seems to be a bug in the tarball logic. When I select this, I also see the generic section "Download the NGINX integration" which isn't necessary.

rhetoric101 (commented):
@bradleycamacho I took a spin through the NGINX docs. I hope I didn't accidentally add any comments for the MongoDB topics since those docs are in this PR.

PR #15755, opened 2024-01-08
fix: update theme to 9.2.0

clarkmcadoo:
This PR bumps the theme to bring in newly styled steps and a collapser anchor link fix. 
It also includes a change to verify-mdx and some mdx files themselves to prevent non <Step> children inside of <Steps> components (this causes a bug with the step numbering)


scripts/verify_mdx.js
```diff
@@ -21,12 +28,53 @@ const progressBar = new cliProgress.SingleBar(
...
+  return { hasNonStepChild, nodeInfo };
+};
+
 const readFile = async (filePath) => {
   let failed = false;
 
   const mdxText = fs.readFileSync(filePath, 'utf8');
   try {
     const jsx = mdx.sync(mdxText);
+    const mdxAst = createAST(filePath);
```
sunnyzanchi, line 68: i didn't notice this when we looked at it before, but i think we should pass `mdxText` to `createAST` instead of `filePath` so we don't read the file from disk twice

sunnyzanchi (changes requested):

sunnyzanchi (approved):

sunnyzanchi (approved):

PR #15860, opened 2024-01-18
chore: update netflow docs

thezackm:
multiple updates to cover new options enabled in kentik/ktranslate#647


jeff-colucci (approved):
Looks great!

PR #15346, opened 2023-11-28
[Do not merge until mid-Jan] Document blob, encode, decode in NRQL functions

maddie-boisvert:
Please follow [conventional commit standards](https://www.conventionalcommits.org/en/v1.0.0/)
in your commit messages and pull request title.

## Give us some context
This documents the `blob`, `encode`, and `decode` functions.

* What problems does this PR solve?
* Add any context that will help us review your changes such as testing notes,
  links to related docs, screenshots, etc.
* If your issue relates to an existing GitHub issue, please link to it.

PR #15861, opened 2024-01-18
feat(NRQL): add blob, decode, encode

jeff-colucci:
Added blob, encode, and decode to the syntax doc

Fixing branch conflicts from https://github.com/newrelic/docs-website/pull/15346

@maddie-boisvert


jeff-colucci:
Content has been reviewed in the original PR so it doesn't require a review here

PR #15866, opened 2024-01-18
Sunset Tessen

tabathadelane:


gatsby/on-route-update.js
```diff
@@ -1,10 +1,10 @@
 const onRouteUpdate = ({ location, prevLocation }) => {
   if (
-    window.Tessen &&
+    window.newrelic &&
     location.hash &&
     location.pathname !== prevLocation?.pathname
   ) {
-    window.Tessen.track({
+    window.newrelic.track({
```
clarkmcadoo, line 7: Does this need to be Segment or Browser Agent? If its browser agent, it could either import the `addPageAction` function or use the global but with `.addPageAction` vs `.track`

src/content/docs/apm/agents/c-sdk/install-configure/install-c-sdk-compile-link-your-code.mdx
```diff
@@ -26,7 +26,7 @@ Our C SDK auto-instruments your code so you can start monitoring applications. Y
 If you don't have one already, [create a New Relic account](https://newrelic.com/signup). It's free, forever.
 
 <ButtonLink
-  data-tessen="stitchedPathLinkClick"
+  data-nr-browser-agent="stitchedPathLinkClick"
```
clarkmcadoo, line 29: What if we just got rid of all of these?

gatsby/on-route-update.js
tabathadelane: we won't be using segment in these sites outside of setting the config options. It should be browser agent

src/content/docs/apm/agents/c-sdk/install-configure/install-c-sdk-compile-link-your-code.mdx
clarkmcadoo: these buttons become `MDXButton` which calls `Button` in the theme and this is not one of the listed prop types. Also, I can't find `data-tessen` in any files that aren't mdx so i really don't think we need this

LizBaker: yeah i didn't find any events associated with them either, i don't think they're hooked up to anything anymore. but.. do we want them to be? maybe they should be getting tracked?

tabathadelane: `stitchedLinkPath` is still actively being logged. Is there coming from somewhere else?

tabathadelane: I guess it's coming from the theme and the SignupLink. I don't am not familiar with what the point of this value is. if it's not hooked up, it does't seem like it's being used anymore.

gatsby/on-route-update.js
clarkmcadoo: lets use the browser agent method `addCustomAttribute()`!

tabathadelane: This has been added and we can see `hashRequest` values in browser events now.

clarkmcadoo (approved):

PR #15876, opened 2024-01-19
chore(k8s): Fix bullet indents

austin-schaefer:


jeff-colucci (approved):
Checked the build, everything seems good to me!

PR #15879, opened 2024-01-19
fix(user-permission): Update to include security capability

cpoggioli-newrelic:
## Give us some context

* A new capability has been added to manage access to new relic security products. We want to update the docs to include the 'vulnerabilities' capability that is already added.

Capability mentioned
Screenshot 2024-01-19 at 2 07 31 PM


jeff-colucci (approved):
Looks great!

PR #15888, opened 2024-01-19
refactor verify-mdx: add validators and tests

sunnyzanchi:
this PR refactors verify-mdx and breaks it up to organize it better. validators are now in their own file and are run in sequence which will make it easier to add new ones. in addition to Liz's `Steps` validator, i added a `Tabs` validator. i separated the reporting/cli functionality into its own file, so verify-mdx now just exports functions which return errors. finally, i added some tests


austin-schaefer:
Hey @sunnyzanchi , is this PR still valid?

gmanandhar-nr:
Hi @WriteMayur , the same comment added for [this PR](https://github.com/newrelic/docs-website/pull/19201#issuecomment-2716683542) would be valid here as well. Let's keep it open for now, and will revisit in future.

PR #15893, opened 2024-01-21
fix(Java): removing compatibility list from JVM section

jeff-colucci:
Removing compatibility list from JVM section to prevent customer confusion on supported Java versions.


meiao (approved):

paperclypse (approved):
It's removed! Are you worried at all about people not having this info, @jeff-colucci? If not, feel free to merge it in

jeff-colucci:
@paperclypse I'm always worried when we remove anything! But I consulted with the Java agent team and we determined this was the best way to alleviate confusion while we wait for some other changes to be made.

PR #15898, opened 2024-01-22
feat(Sonic): Added logs monitoring key popover

nbaenam:
This PR adds the mobile monitoring popover component.

This is the ticket: NR-90652

This is the component:
`<InlinePopup type="logs" />`


jeff-colucci (approved):
Looks great! Thanks for this!

PR #15899, opened 2024-01-22
Vmware OHI refresh

bradleycamacho:


nbaenam (commented):
Hi @bradleycamacho. The build failed and I can't see the result. There are changes in the java and mongodb files. Maybe it would be better to close this one and open a new PR with only the changes of the specific OHI. Also, this PR is to add the vSphere one, right? and the PR is called VMware, which is confusing, I think....

bradleycamacho:
Moving to #17012

PR #15953, opened 2024-01-25
Update overview.mdx

homelessbirds:
Fixing dead link


bradleycamacho (changes requested):
Looks like this is just the intro doc, not the overview doc. I believe the correct doc is https://docs.newrelic.com/docs/alerts-applied-intelligence/new-relic-alerts/get-started/alerts-ai-overview-page/

Instead of a redirect, I think the yaml file just needs to point to the correct file.

akristen (approved):

PR #15967, opened 2024-01-26
Changes per release 15.1

planteater:
- Update code-level metrics article to reflect new drill-down model for anomalies and reorder the article.
- Update error investigation article to include support for Browser errors.
- Update performance monitoring article to clarify which features are available for which service types.
- Clarify that admins can only create blame maps for users who have signed in to CodeStream.
- Update instructions for generating log files to remove reference to codestream@newrelic.com.


planteater:
Thanks for the updates @bradleycamacho !

PR #15997, opened 2024-01-30
.NET agent: update verified compatible versions of Elastic.Clients.ElasticSearch

nr-ahemsath:
The .NET agent team has successfully tested the latest version of a library the agent instruments, Elastic.Client.ElasticSearch v8.12.0.  This PR adds that version to the list of versions we support.

Additionally, I found that some incorrect versions had somehow made it into the list for one of the two compatibility docs; this fixes it.


rhetoric101 (approved):
@nr-ahemsath thanks for correcting these version numbers!

PR #16000, opened 2024-01-30
feat(OHI): base draft for Redis rewrite

bradleycamacho:
Still needs a deeper edit (wording, specifics, etc) and needs the "What's New" file edited for Redis and Redis metrics


src/install/redis/default-requirements.mdx
```diff
@@ -0,0 +1,7 @@
+---
+headingText: Check the compatibility and requirements
+componentType: default
+---
+
+- Our integration is compatible with Redis 4.0 or higher, Percona Server, and Redis Atlas M10 or higher.
+- To monitor Redis Atlas, you'll need a hosted environment to install the infrastructure agent and integration and remotely connect to the Atlas instance.
```
akristen, line 7: Let's make these two sentences, and maybe frontload the Atlas instance.  Something like:

"An Atlas instance in a hosted environment. The infrastructure agent and Redis integration connect to this banana through the apple."

Whatever the order of operations is here

akristen: Other info that may or may not bee necessary is maybe specifying the kind of hosted environment. Are there exclusions? if so, what is supported, etc., and if there aren't any, maybe just specifying that "in any kind of hosted environment."

src/install/redis/install-infra.mdx
```diff
@@ -0,0 +1,6 @@
+---
+componentType: default
+headingText: Install the infrastructure agent
+---
+
+To use the MySQL integration, you need to also [install the infrastructure agent](/docs/infrastructure/install-infrastructure-agent/get-started/install-infrastructure-agent-new-relic/) on the same host. The infrastructure agent monitors the host itself, while the integration you'll install in the next step extends your monitoring with Redis-specific data such as database and instance metrics.
```
akristen, line 6:
```suggestion
To use the Redis integration, you need to also [install the infrastructure agent](/docs/infrastructure/install-infrastructure-agent/get-started/install-infrastructure-agent-new-relic/) on the same host. The infrastructure agent monitors the host itself, while the integration you'll install in the next step extends your monitoring with Redis-specific data such as database and instance metrics.
```

src/install/redis/linux/install-apt.mdx
```diff
@@ -0,0 +1,16 @@
...
+headingText: Download using APT
+---
+
+1. From the command line, run these commands:
+
+```shell
+sudo apt-get update
+```
+
+2. Run the following command:
```
akristen, line 12:
```suggestion
1. Run the following command:
```

src/install/redis/host/percona.mdx
```diff
@@ -0,0 +1,31 @@
+---
+componentType: default
+headingText: Prepare Redis or Percona
```
akristen, line 3: I wonder if the "or Percona" aspect of this integration should be introduced in your intro. I'm not sure how these relate (does Percona fall under Redis? Are they parallel 'things' that are mutually exclusive?) Depending on that, maybe the title needs rework too

src/install/redis/host/percona.mdx
```diff
@@ -0,0 +1,31 @@
...
+---
+
+<Callout variant="tip">
```
akristen, line 6: calling out so you don't forget to remove, if you need to remove this call out

src/install/redis/host/percona.mdx
```diff
@@ -0,0 +1,31 @@
...
+    Complete these steps on the mongos to be monitored. If you want to collect mongod-level metrics,
+    such as host or replica set statistics, create the role
+    and user on each mongod as well.
+</Callout>
+
+1. In the Redis shell. switch to the admin database:
+
+```shell
```
akristen, line 14: indent code blocks

src/install/redis/host/atlas.mdx
```diff
@@ -0,0 +1,13 @@
+---
+componentType: default
+headingText: Prepare Atlas 
+---
+
+
+Using the Atlas CLI run the following command to create a new user, and assign `clusterMonitor` and `readAnyDatabase` roles to the user (adjust the `username` and `password` as required):
+
+```
```
akristen, line 9: indent

src/install/redis/host/atlas.mdx
```diff
@@ -0,0 +1,13 @@
+---
+componentType: default
+headingText: Prepare Atlas 
+---
+
+
```
akristen, line 6: We should split this up into two separate ideas, and consider front loading the action first rather than front loading the tool you're using to do the action. for example:

> Create a new user using the Atlas CLI. You'll assign `clusterMonitor` and `readanyDatabase` roles to this user. Be sure to adjust the `username` and `password` as needed:

src/install/redis/host/atlas.mdx
```diff
@@ -0,0 +1,13 @@
...
+Using the Atlas CLI run the following command to create a new user, and assign `clusterMonitor` and `readAnyDatabase` roles to the user (adjust the `username` and `password` as required):
+
+```
+atlas dbuser create -u username -p password --role clusterMonitor,readAnyDatabase
+```
+
+Alternatively, you can use the Redis Atlas UI to create your user and assign the same specific privileges. For more, information check the [Redis Atlas documentation](https://www.Redis.com/docs/atlas/security-add-Redis-users/#add-database-users).
```
akristen, line 13: I wonder if we should add this to the top in a more narrative way. I'm imagining someone who would've preferred this method (for whatever reason) but didn't realize that this was an option until after running the CLI. 

Maybe you can preface the step by saying users need to create users and assign roles to accomplish X. They can do this using the Atlas CLI tool, or by navigating away from this install doc to work in the Redis Atlas UI.

I'm sure this is outside of scope, but are there other steps they can do from the UI? Maybe if we send them away, we should either specify that they can accomplish the integration thru the Redis UI, or maybe let them know why they need to return to the doc (to finish doing X thing to accomplish their Y goal)

src/install/redis/default-configuration.mdx
```diff
@@ -0,0 +1,15 @@
+---
+componentType: default
+headingText: Configure the Redis integration 
+---
+
+Edit the `Redis3-config.yml` file to configure your integration as needed. The following example config fills in all the required options and reports all metrics the integration pulls:
+
+```yml
```
akristen, line 8: indent

src/install/redis/find-data.mdx
```diff
@@ -0,0 +1,22 @@
...
+componentType: default
+headingText: Find and use data
+---
+
+import infrastructureInstallMysqlDashboard from 'images/infrastructure_screenshot-crop_install-mysql-dashboard.webp'
+
+1. Go to **[one.newrelic.com](https://one.newrelic.com) > Add data** and type **Redis**.
+2. Under **Dashboards**, click **Redis overview dashboard**.
+<img
+  title="Install the Redis dashboards"
```
akristen, line 11: call out to update screenshot (im sure u won't forget but just in case)

akristen (commented):
Content looks good. I left a couple comments about framing and made some call outs around things that are prob on your to do, but you may want a reminder for to keep track of the tasks.

- Don't forget to indent code snippets and add the cute little coloring to 'em
- Some of the ordered lists are rendering as 1. 1. 1., or they're actually ordered (but not using the 1. 1. 1. markdown stuff. Just keep an eye out for that as you i terate
- The build is breaking because of the missing What's next component (which you have in your description as a to do) and you need an image to import. Otherwise it built fine locally for me

I'm curious about the Percona aspect of this doc. Is it worth unspooling in the intro?

LMK when you need a final review + approval next week!

PR #16011, opened 2024-01-31
chore(java): typo in compatibility page

meiao:
## Give us some context

Fixing a typo.


meiao:
Thought it was a typo. Might not be the case.

PR #16014, opened 2024-01-31
fix(java): cleaning up JVM compatibility table

meiao:
## Give us some context
The table was overcomplicated trying to list both the range of versions that support a given version of Java and the range that supports that version of Java but is no longer supported by NR.

So the third column was removed, and instead a note at the bottom of the table was added with a link to the list of supported versions.

Also, versions of Java that are not long term support were removed.


rhetoric101 (approved):
@meiao thanks for updating and simplifying this table!

PR #16029, opened 2024-02-01
Update view-system-limits.mdx - Removed unused limit

sskilbred:
The 5000 infra agent account limit is not being enforced. Having it documented creates unnecessary concerns.


rhetoric101 (approved):
@sskilbred thanks for removing this bullet point.

PR #16032, opened 2024-02-01
chore(nrql): clarify that average is mean average

ghost:
Clarify what `average` is in nrql docs


rhetoric101 (approved):
@ally-sassman thanks for updating this.

PR #16050, opened 2024-02-02
chore: add new config option

thezackm:
updating to add a new config option


rhetoric101 (approved):
@thezackm thanks for clarifying `drop_undefined`!

PR #16084, opened 2024-02-07
Accounts: Update group and user instructions for paging

rhetoric101:
This is for some doc updates because of a refactoring of the user and group interfaces. See NR-220651.


homelessbirds (approved):

src/content/docs/accounts/accounts-billing/new-relic-one-user-management/user-management-ui-and-tasks.mdx
```diff
@@ -100,9 +98,9 @@ Some important points about adding users:
 
   <Collapser
     id="edit-user-type"
-    title="Edit users' user type"
+    title="Edit a user's type"
   >
-Before changing your users' user type, we recommend you understand:
+Before changing the user type for a user, we recommend you understand:
```
jeff-colucci, line 103: The double use of user in this line is a little offputting, what about something like "before changing a user's type" or "a user's account type" or something along those lines? I know it's technically "user type", but I think the reader will know what we mean even if we don't use "user type" specifically.

src/content/docs/accounts/accounts-billing/new-relic-one-user-management/user-management-ui-and-tasks.mdx
```diff
@@ -118,13 +116,7 @@ You can also edit the user type and group of a specific user by clicking on that
...
 For how to use our API to manage user information, see [Manage users with NerdGraph](/docs/apis/nerdgraph/examples/nerdgraph-manage-users).
 
-    <img
-      title="Edit user type in the UI"
-      alt="New Relic user management UI - edit user type"
-      src={accountsUserMgmtEditUsers}
-    />
-
-    To manage how users upgrade their user type, see the [authentication domain settings](/docs/accounts/accounts-billing/new-relic-one-user-management/authentication-domains-saml-sso-scim-more/#user-upgrade).
+To manage how users upgrade their user type, see the [authentication domain settings](/docs/accounts/accounts-billing/new-relic-one-user-management/authentication-domains-saml-sso-scim-more/#user-upgrade).
```
jeff-colucci, line 119: Should this be "upgrade their user types" to match the case of the preceding "users"?

PR #16087, opened 2024-02-07
Update account-linking.mdx

maxlemieux:
NEW_RELIC_EXTENSION_SEND_FUNCTION_LOGS value must be lowercase

Docs have it in uppercase, which doesn't work. Please refer to this part of the config code: https://github.com/newrelic/newrelic-lambda-extension/blob/88b35fe1b82de965f88d23c4ea3ace700143fd7a/config/config.go#L133

Note that some other config booleans actually get converted to lowercase in the code, but not this one.

Impact is that it's difficult to collect logs for debugging, unless already aware that Golang requires lowercase for booleans.

Resolves https://github.com/newrelic/newrelic-lambda-extension/issues/129


homelessbirds (approved):

homelessbirds:
Thank you for this! I appreciate the attention to detail!

PR #16101, opened 2024-02-07
Logs: Add generic path to certificate

rhetoric101:
The original path looked like Ubuntu, but that's never mentioned in the doc.


bradleycamacho (dismissed):

rhetoric101:
@bradleycamacho could you do another review since I made some additional changes?

jeff-colucci (approved):
Looks good in the build!

PR #16102, opened 2024-02-07
chore(dashboards): add anchor names, clarify limitations around expor…

ghost:
Addresses customer feedback: NR-221295

The doc incorrectly says you can export a dashboard as a CSV. This is incorrect and only applies to table charts. I also added anchor links and screenshots and cleaned up some phrasing stuff.


jeff-colucci (approved):
Changes look good to me!

PR #16119, opened 2024-02-09
Remove <dd> and </dd> tags across site

rhetoric101:
We have a number of docs that have `<dd>` tags to indent content. This isn't the convention we're following on the site, and we can't guarantee how the translation engine will handle these (per development). So, I'm removing these.

This is for NR-220635.


jeff-colucci (approved):
They seem removed to me!

rhetoric101:
@jeff-colucci thanks for the review!

PR #16126, opened 2024-02-09
adding example of two joins in a subqery

akristen:
Addressing a customer feedback request about adding an example of multiple joins in one subquery.

### DIFF
--- a/src/content/docs/nrql/using-nrql/subquery-joins.mdx
+++ b/src/content/docs/nrql/using-nrql/subquery-joins.mdx
@@ -63,10 +63,26 @@ FROM Event [INNER|LEFT] JOIN (subquery) ON [key =] key SELECT ...
 ```
 
 There are a few simple rules to the syntax:
+
 * The JOIN clause must always follow immediately after the FROM clause.
 * The JOIN can be prefixed with the join type. INNER or LEFT is optional, and defaults to INNER when omitted.
 * Parentheses containing a subquery must immediately follow the JOIN clause.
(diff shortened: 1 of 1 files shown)


src/content/docs/nrql/using-nrql/subquery-joins.mdx
```diff
@@ -63,11 +63,27 @@ FROM Event [INNER|LEFT] JOIN (subquery) ON [key =] key SELECT ...
 ```
 
 There are a few simple rules to the syntax:
+
 * The JOIN clause must always follow immediately after the FROM clause.
 * The JOIN can be prefixed with the join type. INNER or LEFT is optional, and defaults to INNER when omitted.
 * Parentheses containing a subquery must immediately follow the JOIN clause.
 * The ON clause must immediately follow the subquery and has two forms (more details below).
 
+You can also have multiple joins in one query, too. For example, this query uses two JOINS within subqueries:
```
jeff-colucci, line 72: I don't think you need the "also" and the ",too" in the same sentence. Whichever one you like the most is the only one you need!

jeff-colucci (approved):
I saw one small thing but it looks great so I'm approving!

src/content/docs/nrql/using-nrql/subquery-joins.mdx
akristen: LOL fair! fixing~

homelessbirds (approved):

PR #16132, opened 2024-02-09
Apache OHI rewrite

bradleycamacho:
Local builds keep failing, trying out a Gatsby build


bradleycamacho:
@nbaenam no need to review this just yet

src/install/apache/install-infra.mdx
```diff
@@ -0,0 +1,6 @@
+---
+componentType: default
+headingText: Install the infrastructure agent
+---
+
+To use the apache integration, you need to also [install the infrastructure agent](/docs/infrastructure/install-infrastructure-agent/get-started/install-infrastructure-agent-new-relic/) on the same host. The infrastructure agent monitors the host itself, while the integration you'll install in the next step extends your monitoring with -specific data such as database and instance metrics.
```
akristen, line 6:
```suggestion
To use the apache integration, you need to also [install the infrastructure agent](/docs/infrastructure/install-infrastructure-agent/get-started/install-infrastructure-agent-new-relic/) on the same host. The infrastructure agent monitors the host itself, while the integration you'll install in the next step extends your monitoring with specific data such as database and instance metrics.
```

src/install/apache/whatsNext.mdx
```diff
@@ -0,0 +1,637 @@
+---
+componentType: default
+headingText: Metrics collected by the integration
+---
+
+### Apache configuration options [#config-options]
+
+The Apache integration collects both metrics and inventory information. This table provides a description for each config setting and whether it applies to metrics, inventory, or both.
+
+<table>
```
akristen, line 10: I'd chunk the "applies to" like with like. e.g., lead with 'metrics, inventory' at the top, then all the inventory, then all the metrics. 

The intrro section seems to bucket these configs by metrics/inventory/both, but the table lists the 'applies to' last. i wonder if we should have the first column be 'applies to'. I'll leave it up to you tho! just a thought

src/install/apache/whatsNext.mdx
```diff
@@ -0,0 +1,637 @@
...
+          role: load_balancer
+        inventory_source: config/apache
+    ```
+  </Collapser>
+
+  <Collapser
+    id="example3"
+    title="Custom Apache binary location"
+  >
+    By default the integration tries to detect the Apache binary on these 2 different locations: `/usr/sbin/httpd` and `/usr/sbin/apache2ctl`.<br/>If your Apache is installed on a custom location or you are using MS Windows, use the `BINARY_PATH` setting to specify the correct location and filename of your Apache binary:
```
akristen, line 273: i am not sure how i feel about the unofficial break `<br/>` no action here i am just expressing my skepticism. why not use a traditional ~ break

src/install/apache/whatsNext.mdx
```diff
@@ -0,0 +1,637 @@
...
+      <td style={{ 'text-align': 'center' }}/>
+    </tr>
+  </tbody>
+</table>
+
+## Example configurations [#examples]
+
+Here are some example YAML configurations: 
+
+<CollapserGroup>
```
akristen, line 206: I like how these are organized with collapsers. The descriptions are good, too. I'm not sure what level of knowledge we're assuming here. Is it possible to mix all these example configs together? Are there certain configs that are mutually exclusive? Do we need to describe when they might want to use these configs?

If it's assumed that iykyk, then disregard the comment. just wanted to call out it wasn't clear to me

src/install/apache/whatsNext.mdx
```diff
@@ -0,0 +1,637 @@
...
+  </Collapser>
+</CollapserGroup>
+
+### Labels [#labels]
+
+You can further decorate your metrics using labels. Labels allow you to add attributes (key/value pairs) to your metrics, which you can then use to query, filter, or group your metrics.
+
+Our default sample config file includes examples of labels but, because they're not mandatory, you can remove, modify, or add new ones of your choice.
+
+```yml
```
akristen, line 410: Should we preface this description with what this particular label is signifying? Is it worth adding some additional examples of what customeers typically would add that isn't out of the box. e.g., are there certain key/value pairs customers are expecting that we don't offer? should we include some of those?

src/install/apache/whatsNext.mdx
```diff
@@ -0,0 +1,637 @@
...
+      <td>
+        Current number of workers that are writing.
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+## Inventory data [#inventory]
+
+Inventory data captures the version numbers from running Apache and from all loaded Apache modules, and adds those version numbers under the `config/apache` namespace. For more about inventory data, see [Understand data](/docs/infrastructure/integrations-getting-started/getting-started/understand-integration-data-data-types#inventory-data).
```
akristen, line 578: small nitpick. the sentence is a bit wordy. can we break it up?

akristen (commented):
Left some comments about my own uncertainties, but I'm not the audience for the doc so I'll defer to your opinion about how much context to include/leave out. Overall I think the structure is solid and clear. 

- I found a couple possible? typos that I called out. 
- I made one suggestion about moving the 'applies to' column to the far left since the intro to this section buckets the config options by metrics/inventory/both.
- I'm also not sure what to do about the table formatting. I almost want the 'setting' column to be wider, and to let the description part wrap more than it does. I'll defer to you though.

src/install/apache/whatsNext.mdx
bradleycamacho: I think the column should stay at the end since folks tend to skim the through the left-most column to find things, but I chunked all the related metrics/inventory ones together!

src/install/apache/whatsNext.mdx
bradleycamacho: Labels are hyper-specific to user use-cases, but I think adding more context to the example is good!

nbaenam (commented):
I left some comments although I can't check the page because the build failed. Maybe you have here the problem I had with my OHI PRs if you opened it before Netlify.

PR #16157, opened 2024-02-13
style(nav): Surface NRQL getting started at L1 lvl

austin-schaefer:
Curious for your thoughts on this @jeff-colucci !


jeff-colucci (approved):

jeff-colucci:
Looks like we're removing the getting started section? Looks good to me, I want to redo the tutorials section next sprint when I start doing content rewrites for the whole section of the Doc, so I'll redo the tutorial navs then if that works for you!

PR #16172, opened 2024-02-14
style(sweep): Remove slash used as "or" or "and"

austin-schaefer:


src/install/java/frameworkInstallation/jboss.mdx
```diff
@@ -61,7 +61,7 @@ To pass the `-javaagent` argument on JBoss:
             <tbody>
                 <tr>
                 <td>
-                    Unix / macOS with 6.x EAP or 7.0.x AS and above
+                    Unix and macOS with 6.x EAP or 7.0.x AS and above
```
ally-sassman, line 64: @austin-schaefer I think these should be "or" because "and" implies you need both Unix and macOS

ally-sassman (dismissed):

src/install/java/frameworkInstallation/jboss.mdx
austin-schaefer: Thanks, swapped these around!

paperclypse (approved):
Solid. Thanks for doing this!

PR #16194, opened 2024-02-15
style(Lambda): Clarified that a value in config example is a placeholder

paperclypse:
Customer feedback NR-141784


akristen (approved):
Approving. Looks good in the preview, but seems like the CLA needs to be signed still.

PR #16209, opened 2024-02-19
fix(Apache): Fixed the navigation path

nbaenam:
Fixed a wrong URL in the navigation pane.

This is a request from the help-documentation channel (02/19/2024)


nbaenam:
I've checked it out and it's working now.

PR #16246, opened 2024-02-21
Accounts: Correct anchor link and fix add-on wording

rhetoric101:
This is a small follow-up fix to the recent Compute Add On docs.


paperclypse (approved):
Looks good

paperclypse (approved):
This one could go either way, I think

PR #16263, opened 2024-02-22
fix(Dashboards): Cleaned up some language

paperclypse:
Just wanted to clean up some language I came across.


akristen (approved):
Good fixes~

PR #16296, opened 2024-02-26
fix(IAST): Fixed urls for config file

nbaenam:
Fixed the URLs for the config file on the [Install IAST](https://docs.newrelic.com/docs/iast/install/) page.

Requested from Jayant (02/22/2024)

### DIFF
--- a/src/content/docs/iast/install.mdx
+++ b/src/content/docs/iast/install.mdx
@@ -50,12 +50,10 @@ To install New Relic IAST:
             title={<><InlineCode>newrelic.yml</InlineCode> config file</>}
 >
                 ```yml
-                host: staging-collector.newrelic.com
                 security:
                     enabled: true
-                agent:
-                    enabled: true
-                validator_service_url: wss://csec.nr-data.net
+                    agent:
+                        enabled: true
                 ```
         </Collapser>
         <Collapser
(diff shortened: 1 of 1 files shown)


src/content/docs/iast/install.mdx
```diff
@@ -50,14 +50,38 @@ To install New Relic IAST:
...
                 security:
                     enabled: true
-                agent:
+                    agent:
+                        enabled: true
+                ```
+        </Collapser>
+        <Collapser
+            id="config-file-example"
+            title={<><InlineCode>newrelic.yml</InlineCode> config file for EU</>}
```
bradleycamacho, line 61: Shouldn't these just be inline code wrapped like this: ``textHere` `?

I think using html code tags might mess up the translation workflow

bradleycamacho (approved):
Looks good! I left one comment about maybe replacing the `<inlinecode>` tags with markdown syntax instead.

Feel free to merge after changing (or not changing!) that

src/content/docs/iast/install.mdx
nbaenam: I've seen this structure in our doc site with filenames. I'm going to ask the question to our eng team. Thanks!

PR #16395, opened 2024-03-04
feat(OHI rabbitmq): Added branching doc structure

nbaenam:
Added the branching doc structure for this page: [RabbitMQ monitoring integration](https://docs.newrelic.com/docs/infrastructure/host-integrations/host-integrations-list/rabbitmq-monitoring-integration).

This is the ticket: NR-217448


src/install/config/rabbitmq.yaml
```diff
@@ -0,0 +1,94 @@
...
+        displayName: 'ECS'
+        logo: 'amazonECS'
+      - value: 'kubernetes'
+        displayName: 'Kubernetes'
+        logo: 'k8'
+      - value: 'ansible'
+        displayName: 'Ansible'
+        logo: 'linux'
+
+  - optionType: deployment
```
bradleycamacho, line 29: Screenshot 2024-03-12 at 4 04 33 PM

Why is there only Tarball as an option here? Looking at the exisitng doc, I think this should include the normal options of Yum, apt, etc.

bradleycamacho (changes requested):
Doc looks good other than the missing install options (tum,

src/install/config/rabbitmq.yaml
nbaenam: Added!

bradleycamacho (approved):

PR #16432, opened 2024-03-06
Logs: Correct SKU name of Live Archives

rhetoric101:
Per SME feedback, remove "Storage" from SKU name.


rhetoric101:
I just followed up with SMEs to see if this is still an issue.

homelessbirds (approved):

PR #16435, opened 2024-03-06
Note Fluent Bit does not support whitespace in key/value

bradleycamacho:
Add note that Fluent Bit does not support whitespace in key/value

### DIFF
--- a/src/content/docs/logs/forward-logs/forward-your-logs-using-infrastructure-agent.mdx
+++ b/src/content/docs/logs/forward-logs/forward-your-logs-using-infrastructure-agent.mdx
@@ -483,9 +483,9 @@ Although these configuration parameters aren't required, we still recommend you 
     ```
 
     <Callout variant="important">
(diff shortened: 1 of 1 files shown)


src/content/docs/logs/forward-logs/forward-your-logs-using-infrastructure-agent.mdx
```diff
@@ -485,6 +481,12 @@ Although these configuration parameters aren't required, we still recommend you
           region: example-us-02
           team: B-team
     ```
+
+    <Callout variant="important">
+      The `attributes` configuration parameter does not add custom attributes to logs forwarded via external Fluent Bit configuration (for example, using the `fluentbit` configuration parameter). In this scenario, you should refer to the `record_modifier` option in the [Fluent Bit documentation](https://docs.fluentbit.io/manual/).
```
jeff-colucci, line 486: does not > doesn't

src/content/docs/logs/forward-logs/forward-your-logs-using-infrastructure-agent.mdx
```diff
@@ -485,6 +481,12 @@ Although these configuration parameters aren't required, we still recommend you
...
+
+    Additionally, Fluent Bit does not support white space in keys or values.
```
jeff-colucci, line 488: doesn't

jeff-colucci (dismissed):
Comment looks good overall to me, just some missing contractions and I think you're set to go!

PR #16450, opened 2024-03-07
(Publish 03/13) Update background task docs

hmstepanek:
## Update background task docs

* We've had a lot of customers using the background task to wrap functions that run for too long or in some cases forever so  updated the docs to explain this.


src/content/docs/apm/agents/python-agent/python-agent-api/backgroundtask-python-agent-api.mdx
```diff
@@ -239,3 +239,6 @@ result = task(*args, **kwargs)
 ```
 
 In this example, if this call was made in a web transaction being monitored by an agent, the call will reclassify the web transaction as a background task. If, however, the call was made in a background thread or coroutine, then it would start the recording of a fresh background task transaction that tracks what occurs until the call returns.
+
+### Troubleshooting
+If the background task's transaction is not showing up in the UI. This could be because the transaction that was created by the background task was never exited or the background task ran for too long and was dropped by the server because it was considered to old to keep. Background tasks should not run for over 20 minutes and if they do they should be broken up into multiple background tasks.
```
hmstepanek, line 244:
```suggestion
If the background task's transaction is not showing up in the UI. This could be because the transaction that was created by the background task was never exited or the background task ran for too long and was dropped by the server because it was considered too old to keep. Background tasks should not run for over 20 minutes and if they do they should be broken up into multiple background tasks.
```

src/content/docs/apm/agents/python-agent/python-agent-api/backgroundtask-python-agent-api.mdx
```diff
@@ -239,3 +239,6 @@ result = task(*args, **kwargs)
 ```
 
 In this example, if this call was made in a web transaction being monitored by an agent, the call will reclassify the web transaction as a background task. If, however, the call was made in a background thread or coroutine, then it would start the recording of a fresh background task transaction that tracks what occurs until the call returns.
+
+### Troubleshooting
+If the background task's transaction is not showing up in the UI. This could be because the transaction that was created by the background task was never exited or the background task ran for too long and was dropped by the server because it was considered too old to keep. Background tasks should not run for over 20 minutes and if they do they should be broken up into multiple background tasks.
```
hmstepanek, line 244:
```suggestion
If the background task's transaction is not showing up in the UI, this could be because the transaction that was created by the background task was never exited or the task ran too long. If a background task runs for over 20 minutes it is dropped by the server because it was considered too old to keep. Background tasks should not run for over 20 minutes and if they do they should be broken up into multiple background tasks.
```

src/content/docs/apm/agents/python-agent/python-agent-api/backgroundtask-python-agent-api.mdx
```diff
@@ -239,3 +239,6 @@ result = task(*args, **kwargs)
 ```
 
 In this example, if this call was made in a web transaction being monitored by an agent, the call will reclassify the web transaction as a background task. If, however, the call was made in a background thread or coroutine, then it would start the recording of a fresh background task transaction that tracks what occurs until the call returns.
+
+### Troubleshooting
+If the background task's transaction is not showing up in the UI, this could be because the transaction that was created by the background task was never exited or the task ran too long. If a background task runs for over 20 minutes it is dropped by the server because it was considered too old to keep. Background tasks should not run for over 20 minutes and if they do they should be broken up into multiple background tasks.
```
hmstepanek, line 244:
```suggestion
If the background task's transaction is not showing up in the UI, this could be because the transaction that was created by the background task was never exited or the task ran too long. If a background task runs for over 20 minutes, it is dropped by the server because it was considered too old to keep. Background tasks should not run for over 20 minutes and if they do they should be broken up into multiple background tasks.
```

src/content/docs/apm/agents/python-agent/python-agent-api/backgroundtask-python-agent-api.mdx
```diff
@@ -239,3 +239,6 @@ result = task(*args, **kwargs)
 ```
 
 In this example, if this call was made in a web transaction being monitored by an agent, the call will reclassify the web transaction as a background task. If, however, the call was made in a background thread or coroutine, then it would start the recording of a fresh background task transaction that tracks what occurs until the call returns.
+
+### Troubleshooting
```
hmstepanek, line 243:
```suggestion
## Troubleshooting
```

bradleycamacho (approved):

rhetoric101:
Temporarily closing.

rhetoric101:
I have to do an override since there was no preview via GitHub. I looked at the preview in the cloud after a colleague did a cache reset.

PR #16473, opened 2024-03-11
Add missing redirect

bradleycamacho:


akristen (approved):
Confirmed the redirect works!

PR #16477, opened 2024-03-11
Improve wording in the doc about querying timeslice metrics

KevinCorcoran:


rhetoric101 (approved):
@KevinCorcoran thanks for updating these references to timeslice metrics!

PR #16485, opened 2024-03-11
Fix callout image sizing

bradleycamacho:
This doc had a weird issue with a callout containing an image within a side/side component. Played around with it a bit, and I couldn't come up with a good way to keep the image in there, so I decided to just cut the image and give directional info within the callout. 

Screenshot 2024-03-11 at 2 53 22 PM


rhetoric101 (approved):
@bradleycamacho thanks for addressing this image in the callout.

PR #16497, opened 2024-03-12
Various fixes for AI

bradleycamacho:
This doc is in rough shape and this is a small edit to fix a few egregious things. I struggled with writing a decent intro since I'm not too familiar with the product space, so I'd love some ideas there if you (reviewer) have any.

### DIFF
--- a/src/content/docs/alerts-applied-intelligence/applied-intelligence/incident-intelligence/Issues-and-Incident-management-and-response.mdx
+++ b/src/content/docs/alerts-applied-intelligence/applied-intelligence/incident-intelligence/Issues-and-Incident-management-and-response.mdx
@@ -1,5 +1,5 @@
 ---
-title: Issues and incident management and response
+title: Respond to issues
 tags:
   - Alerts and applied intelligence
   - Applied intelligence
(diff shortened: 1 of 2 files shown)


src/content/docs/alerts-applied-intelligence/applied-intelligence/incident-intelligence/Issues-and-Incident-management-and-response.mdx
```diff
@@ -28,9 +28,11 @@ import accountsIssueSummary from 'images/accounts_screenshot-full_issue-summary.
 import accountsIssueFeedFilterIcon from 'images/accounts_screenshot-crop_issue-feed-filter-icon.webp'
 
 
-## Issues feed [#issues-feed]
+Use New Relic to track and respond to your issues using issues feeds, postmortem reports, and more.
```
jeff-colucci, line 31: I think it would be helpful to flesh this out a bit, even if it's just basic or maybe obvious stuff. Just a sentence or two that gives value prop or goes over more specifics of the document just provides a smoother transition to the meat of the doc I think.

src/content/docs/alerts-applied-intelligence/applied-intelligence/incident-intelligence/Issues-and-Incident-management-and-response.mdx
```diff
@@ -44,35 +46,15 @@ feed."
...
-- Issue acknowledged(true or false)
-- Issue muted (true or false)
-- Issue correlated (true or false)
-- Issue priority (low, medium, high, or critical)
-- Issue source (NR alerts, NR anomalies, or REST API)
-- Entity
-- Alert policy
-- Alert condition
-- Tags (search for any tag related to the issue)
+You can search for any issue with the search bar free text. You can filter by any issue attributes or any related tag by clicking the filter icon.
```
jeff-colucci, line 49: Are h2s the best format here? I might consult with Austin but I think the visual problem I'm having with the doc is that the sections almost seem too short to support so many h2s as the main component. Just something to think about.

src/nav/alerts-applied-intelligence.yml
```diff
@@ -79,7 +79,7 @@ pages:
             path: /docs/alerts-applied-intelligence/new-relic-alerts/alert-notifications/muting-rules-suppress-notifications
   - title: Incident response
     pages:
-      - title: Issue and incident management and response
+      - title: Issues, incident management,and response
```
jeff-colucci, line 82: As far as the title goes, we don't really mention incidents in the doc outside of the specific UI elements, maybe the title should be reworked or we should go into incidents a little more? What I mean is if I come in to the doc looking for incidents specifically (for some reason), it doesn't appear in the right nav at all and if I'm not familiar with this section of the product, I feel like the lack of "incident" callouts could cause irritation for the reader.

PR #16502, opened 2024-03-12
Align with semantic conventions

keegoid-nr:
- Job manager releases always start with a lowercase `r`. https://hub.docker.com/r/newrelic/synthetics-job-manager/tags
- The releases page refers to the `Node browser runtime v2.2.22` https://docs.newrelic.com/docs/release-notes/synthetics-release-notes/node-browser-runtime-release-notes/node-browser-runtime-2.2.22/

cc @bpeckNR


rhetoric101 (approved):
@keegoid-nr thanks for making these entries consistent!

PR #16521, opened 2024-03-14
chore(infra-agent): updating supported SUSE/SLES versions

josemore:
## Give us some context

* Updating supported SUSE versions based on latest supported version by the manufacturer (https://endoflife.date/sles)


rhetoric101 (approved):
@josemore thanks for updating these supported versions!

PR #16524, opened 2024-03-14
feat(infra-agent): configurable integrations tmp dir

josemore:
## Give us some context

* infra-agent 1.50.0 added an additional setting for the agent to configure the temp directory for integrations sync.


rhetoric101 (approved):
@josemore thanks for adding this setting! I just had to make a small change to collapser tags to get this to build.

PR #16528, opened 2024-03-14
Update CLM and add link from Logs

planteater:
Two changes:

1. Add a query to Code-Level Metrics section so users can check their coverage.
2. Add a link to CS from Logs, as an additional way to view logs.


rhetoric101 (approved):
@planteater thanks for adding these clarifications!

src/content/docs/codestream/observability/code-level-metrics.mdx
```diff
@@ -29,12 +29,20 @@ Click the CodeLens to see charts visualizing each of the metrics. If the reposit
...
 
+Run the following NRQL query to identify where you might see CodeLenses with code-level metrics for one of your services. Just replace the example `entity.guid` value with the one for the given service. In the query results, look at the `code.*` attributes to see if any of them represent your code, and not framework code.
+
+`select * from Span where entity.guid = 'MXxBUE18QVBQTElDQVRJT058MjM' where code.function is not null since 30 minutes ago LIMIT MAX`
```
rhetoric101, line 40:
```suggestion
```
SELECT * 
FROM Span 
WHERE entity.guid = 'MXxBUE18QVBQTElDQVRJT058MjM' AND code.function is not null 
SINCE 30 minutes ago LIMIT MAX
```

rhetoric101 (commented):
@planteater thanks for making these latest changes! I just had one more question about the NRQL (see below).

rhetoric101 (approved):
@planteater thanks for making these updates!

PR #16530, opened 2024-03-14
Typo fix in PT

jmiraNR:
Typo fix in PT


rhetoric101 (approved):
@jmiraNR thanks for making this update!

PR #16537, opened 2024-03-14
Clarification of JP translation

jmiraNR:
Clarification of JP translation


rhetoric101 (approved):
@jmiraNR thanks for making these updates!

PR #16550, opened 2024-03-15
chore: update .NET agent verified compatible versions of core tech libraries

nr-ahemsath:
The .NET agent team has verified that several new versions of our instrumented libraries are compatible with our agent, and we are updating our docs accordingly.


rhetoric101:
@nr-ahemsath could you re-sign the CLA?

rhetoric101 (approved):
@nr-ahemsath thanks for updating these version numbers!

PR #16560, opened 2024-03-18
aws-private-link: add synthetics endpoint

josemore:
## Give us some context

* Adding the new supported endpoint for Synthetics in the AWS private link integration. It has the same limitations as the new endpoints for infrastructure monitoring.


homelessbirds:
Looks good! Thanks!

homelessbirds (approved):

PR #15830, opened 2024-01-16
AI monitoring GA (3/28) docs - DO NOT MERGE

akristen:
AI monitoring docs updates:

- All feature docs: intro, capat/reqs, config, UI doc, install doc, drop filters doc
- Updates to agent config files, compat tables
-  API docs for all agents for two new methods

PR #16586, opened 2024-03-19
Add Python Agent AI-related API docs. **DO NOT MERGE PLEASE!**

umaannamalai:
This PR adds API docs for new AI-related APIs in the Python Agent. It is not to be merged into develop.
This PR will eventually be closed and the contents will be moved into https://github.com/newrelic/docs-website/pull/15830.

### DIFF
--- a/src/content/docs/apm/agents/python-agent/python-agent-api/recordllmfeedbackevent-python-agent-api.mdx
+++ b/src/content/docs/apm/agents/python-agent/python-agent-api/recordllmfeedbackevent-python-agent-api.mdx
@@ -1,7 +1,7 @@
 ---
 title: record_llm_feedback_event (Python agent API)
 type: apiDoc
-shortDescription: Records an LLM feedback event
+shortDescription: Records LLM feedback event
 tags:
   - Agents
   - Python agent
--- a/src/content/docs/apm/agents/python-agent/python-agent-api/recordllmfeedbackevent-python-agent-api.mdx
+++ b/src/content/docs/apm/agents/python-agent/python-agent-api/recordllmfeedbackevent-python-agent-api.mdx
@@ -1,0 +1,129 @@
+---
+title: record_llm_feedback_event (Python agent API)
+type: apiDoc
+shortDescription: Records an LLM feedback event
+tags:
+  - Agents
+  - Python agent
+  - Python agent API
+metaDescription: 'Python API: This call records an LLM feedback event for querying in the AI Monitoring UI.'
+redirects:
+  - /docs/agents/python-agent/python-agent-api/recordllmfeedbackevent-python-agent-api
+  - /docs/agents/python-agent/python-agent-api/record_llm_feedback_event
+freshnessValidatedDate: never
+---
+
+## Syntax
+
+```py
+newrelic.agent.record_llm_feedback_event(trace_id, rating, category=None, message=None, metadata=None)
+```
+
+Records custom feedback events for LLM applications.
+
+## Requirements
+
+Python agent version 9.8.0 or higher.
+
+## Description
+
+This API records a feedback event of type `LlmFeedbackMessage` that can be viewed and queried in the New Relic UI. Feedback events correlate trace IDs between an AI-generated message and the feedback an end user submitted on it.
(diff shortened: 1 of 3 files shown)


src/content/docs/apm/agents/python-agent/python-agent-api/setllmtokencountcallback-python-agent-api.mdx
```diff
@@ -0,0 +1,160 @@
...
+    each model to support token counts appropriately.
+
+    Arguments:
+    model -- name of the model provided by or returned from the LLM
+    content -- context, input or output of the LLM
+    """
+    if "gpt" in model:
+        import tiktoken
+
+        enc = tiktoken.encoding_for_model(model)
```
hmstepanek, line 152: This raises a KeyError if the model is not found so we should probably include handling that in this example so customers can just copy-paste this into their code. Since we record a warning for <0 values it would be better to return None in Python's API instead of 0 as well. Instead of the if statement and all this stuff let's just simplify this to the following callback implementation:

```
import tiktoken

try:
    enc = tiktoken.encoding_for_model(model)
except KeyError:
    return None  # Unknown model.
return len(enc.encode(content))
```

src/content/docs/apm/agents/python-agent/python-agent-api/setllmtokencountcallback-python-agent-api.mdx
```diff
@@ -0,0 +1,160 @@
...
+        return len(enc.encode(content))
+
+    # for models that are not supported in the example
+    return 0
+
+# add this line where your application starts up
```
hmstepanek, line 158: Technically this can be added to the top of the customer's main file and it should be fine. (I just tested this locally to verify.) Let's just remove that comment because it's a bit confusing and non-specific.
```suggestion
```

src/content/docs/apm/agents/python-agent/python-agent-api/setllmtokencountcallback-python-agent-api.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+Python agent version 9.8.0 or higher.
+
+
+## Description
+
+This API registers a callback to calculate and store token counts on `LlmEmbedding` and `LlmChatCompletionMessage` events.
+This function will typically only be used if `ai_monitoring.record_content.enabled` is False as this prevents the agent from sending AI content to New Relic.
+If you would still like to capture token counts for Llm events in this instance, you can implement a callback in your app code to determine the token counts locally and send this information to New Relic.
+In most cases, this API will be called exactly once soon after the agent has initialized. However, you may make multiple calls to this API if you would like.
```
hmstepanek, line 35: This isn't a pre-requisite in Python's case. They can just put this call at the top of their main file.
```suggestion
In most cases, this API will be called exactly once. However, you may make multiple calls to this API if you would like.
```

hmstepanek (changes requested):

src/content/docs/apm/agents/python-agent/python-agent-api/setllmtokencountcallback-python-agent-api.mdx
hmstepanek: We could optionally provide a second example with passing in the application:
```
application = newrelic.agent.register_application(timeout=10.0)
newrelic.agent.set_llm_token_count_callback(token_count_callback, application)
```

src/content/docs/apm/agents/python-agent/python-agent-api/recordllmfeedbackevent-python-agent-api.mdx
```diff
@@ -0,0 +1,129 @@
+---
+title: record_llm_feedback_event (Python agent API)
+type: apiDoc
+shortDescription: Records an LLM feedback event
```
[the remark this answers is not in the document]
umaannamalai: Had the "a LLM" vs "an LLM" discussion with @akristen and the jury's still out on that one so I'm going to change this to "Record LLM feedback events" to minimize singular usage 😆

akristen:
https://github.com/newrelic/docs-website/pull/15830 please see

PR #16598, opened 2024-03-20
List issues and incidents pagination 5k limit

crshanks:
The listing of issues and incidents through pagination is limited to the NRQL LIMIT MAX, which is currently 5000. Further filter will be required if the pagination reaches 5000.


crshanks (commented):
@homelessbirds this looks good to me.

homelessbirds:
@crshanks I created the PR here: https://github.com/newrelic/docs-website/pull/16641. I can't merge this PR in because of some issues with the build. What do you think about the one I created?

PR #16603, opened 2024-03-20
Add edge function to load eu osano variant when in eu

LizBaker:
When users in the EU load the site this function will run and replace the osano script's variant 1 with variant 2 which is gdpr friendly. with htmlrewriter I wasn't totally sure if we wanted to use the [element](https://developers.cloudflare.com/workers/runtime-apis/html-rewriter/#element) over the [text](https://developers.cloudflare.com/workers/runtime-apis/html-rewriter/#text-chunks) option but even loading the site on slow 3g the script seems to come in with the correct variant from the start which seems adequate

to test:
set the ENVRIONMENT env var to production in your .env file to get osano to show up. download the netlify cli and run `netlify dev` to run the edge function locally. in osano-country.js you can add the US to the list of countries to get the variant=two to show up in the script, and remove it to see variant=one

### DIFF
--- a/netlify/edge-functions/osano-country.js
+++ b/netlify/edge-functions/osano-country.js
@@ -1,0 +1,67 @@
+import { HTMLRewriter } from 'https://ghuc.cc/worker-tools/html-rewriter/index.ts';
+
+export default async (request, context) => {
+  const response = await context.next();
+  const hasGdpr = [
+    'AT',
+    'BE',
+    'BG',
+    'HR',
+    'CY',
+    'CZ',
+    'DK',
+    'EE',
+    'FI',
+    'FR',
+    'DE',
+    'GR',
+    'HU',
+    'IE',
+    'IT',
+    'LV',
+    'LT',
+    'LU',
+    'MT',
+    'NL',
+    'PL',
+    'PT',
+    'RO',
+    'SK',
+    'SI',
+    'ES',
+    'SE',
+  ].includes(context.geo.country.code);
+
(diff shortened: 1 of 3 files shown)


netlify/edge-functions/osano-country.js
```diff
@@ -0,0 +1,53 @@
...
+    'MT',
+    'NL',
+    'PL',
+    'PT',
+    'RO',
+    'SK',
+    'SI',
+    'ES',
+    'SE',
+  ].includes(context.geo.country.code);
```
sunnyzanchi, line 33:
```suggestion
  const hasGdpr = new Set([
    'AT',
    'BE',
    'BG',
    'HR',
    'CY',
    'CZ',
    'DK',
    'EE',
    'FI',
    'FR',
    'DE',
    'GR',
    'HU',
    'IE',
    'IT',
    'LV',
    'LT',
    'LU',
    'MT',
    'NL',
    'PL',
    'PT',
    'RO',
    'SK',
    'SI',
    'ES',
    'SE',
  ]).has(context.geo.country.code);
```

normally this difference is trivial but since this function is going to run before every single response to every request, i think it's worth optimizing wherever we can

sunnyzanchi (approved):

LizBaker:
cache happy build: https://app.netlify.com/sites/docs-website-netlify/deploys/65fb23d981a2bd009f6a5747

clarkmcadoo (approved):

PR #16608, opened 2024-03-20
Serverless: Change AWS legacy file name to layerless in navigation.

rhetoric101:
I had previously changed the actual filename so it ends in "-layerless" instead of "-legacy" but evidently, I missed the related navigation link and label change in that previous commit.


rhetoric101:
netlify build

rhetoric101:
Since this PR has a broken preview check, I'm going to just close it and make a new one.

PR #16647, opened 2024-03-22
fix(Security bulletins): Replaced URL with relative path because it's…

paperclypse:
… an internal docs link.

Docs reviewer: please just approve and merge this one in.


bradleycamacho (approved):

homelessbirds (approved):
good catch!

PR #16681, opened 2024-03-26
fix(serverless): deleting lambda containerized images docs

homelessbirds:
Deleting these docs to be replaced by more accurate representations of our product.


akristen:
netlify build

akristen:
@homelessbirds hi hi. for some reason the build failed, but i reopened it and it built fine but didn't rebuild the deploy preview. 

i can see that netlify/docs-website-netlify/deploy-preview "built" from the details link in the GH checks, but i wanted to x2 check that you meant to delete the /es/[...]newrelic.ini files. I'm not familiar with those and they aren't deleted in the EN, JP, KR files... just ES/PT... Are these meant to be deleted when we remove docs from ES/PT directories?

clarkmcadoo (approved):

PR #16695, opened 2024-03-27
fix(links): Clarified custom module support and modified links

bpeckNR:
- Clarified custom modules with CPM are supported in both scripted API and scripted browser monitors
- Updated Selenium links
- Removed Selenium links where the content is no longer available (legacy API versions)


jeff-colucci (approved):
Fantastic as always!

PR #16716, opened 2024-03-28
SonarQube rewrite

bradleycamacho:
Rewrites the SonarQube integration


src/content/docs/infrastructure/host-integrations/host-integrations-list/sonarqube-monitoring-integration.mdx
```diff
@@ -9,7 +9,7 @@ freshnessValidatedDate: 2023-06-28
 ---
 import SonarQubeDashboard from 'images/infrastructure_screenshot_full-sonarqube-dashboard.webp'
 
-Our SonarQube integration monitors the performance of your SonarQube application, helping you diagnose and optimize the code. Our SonarQube integration makes use of our infrastructure agent, PosgreSQL agent, NRI-Prometheus, and NRI-JMX and gives you a pre-built dashboard with your most important SonarQube metrics.
+Our SonarQube integration monitors the performance of your SonarQube application, helping you diagnose and optimize the code. Our SonarQube integration makes use of our infrastructure agent, PosgreSQL integration, NRI-Prometheus, and NRI-JMX and gives you a pre-built dashboard with your most important SonarQube metrics.
```
akristen, line 12: Can we get a firmer verb than "makes use of". 

- Does it put data collected by these integrations in context of each other _through_ the dashboard?
- Maybe a one sentence explanation about why these other integrations are dependencies (am i using that right) for SQ? 
- Annnnd this might be out of scope but maybe identifying who finds this most useful can also be helpful. Mostly b/c asking customers to install 4 different technologies is a huge ask, so helping them decide if it's useful for their use case/persona can make the value prop more evident upfront

src/content/docs/infrastructure/host-integrations/host-integrations-list/sonarqube-monitoring-integration.mdx
```diff
@@ -21,314 +21,230 @@ After setting up our SonarQube integration, we give you a dashboard for your Son
 </figcaption>
 
 
-## Install the infrastructure agent [#infra-install]
+<Steps>
+    <Step>
+## Install the infrastructure agent [#infra-install]  
 
-You can install the infrastructure agent two different ways:  
+To use the SonarQube integration, you need to also [install the infrastructure agent](/docs/infrastructure/install-infrastructure-agent/get-started/install-infrastructure-agent-new-relic/) on the same host. The infrastructure agent monitors the host itself, while the integration you'll install in the next step extends your monitoring with SonarQube-specific data.
```
akristen, line 28: I'm wondering if the boilerplate language for these should be something like "All our integrations first need the infrastructure agent to connect your data to NR." At least for this one, reframing it as something all OHIs use might make customers less 🙄 about all the different installs they need to do

akristen: but written better obvs. not necessary but just something to think about

src/content/docs/infrastructure/host-integrations/host-integrations-list/sonarqube-monitoring-integration.mdx
```diff
@@ -21,314 +21,230 @@ After setting up our SonarQube integration, we give you a dashboard for your Son
...
+
+        # Whether the integration should skip TLS verification or not. Defaults to false
+        insecure_skip_verify: true
+      timeout: 10s
+  ```
+    </Step>
+    <Step>
+## Install and configure the JMX integration
+
+To use the SonarQube integration, you need to also [install our JMX monitoring integration](/docs/infrastructure/host-integrations/host-integrations-list/jmx-monitoring-install/). The JMX integration scrapes SonarQube data, which we will later turn into dashboards and queryable data.
```
akristen, line 102: this is good context and i think a small sentence for nri-prometh + prostgresql steps that parallel this intro might be helpful. 

i'd approach these sentences as a "mini narrative" that threads these separate install steps together and explains how the data from these diff install processes "come together" to tell a SQ story

src/content/docs/infrastructure/host-integrations/host-integrations-list/sonarqube-monitoring-integration.mdx
```diff
@@ -21,314 +21,230 @@ After setting up our SonarQube integration, we give you a dashboard for your Son
...
-* If you'd rather install our infrastructure agent manually, you can follow a tutorial for manual installation for [Linux](/docs/infrastructure/install-infrastructure-agent/linux-installation/install-infrastructure-monitoring-agent-linux), [Windows](/docs/infrastructure/install-infrastructure-agent/windows-installation/install-infrastructure-monitoring-agent-windows/), or [macOS](/docs/infrastructure/install-infrastructure-agent/macos-installation/install-infrastructure-monitoring-agent-macos/). 
+To use the SonarQube integration, you need to first install our PostgreSQL integration.
 
-##  Install the PostgreSQL
-
-1. Check out our PostgreSQL agent requirements in our [documentation](/docs/infrastructure/host-integrations/host-integrations-list/postgresql/postgresql-integration/)  before installing the agent.
+1. Check out our PostgreSQL integration requirements in our [documentation](/docs/infrastructure/host-integrations/host-integrations-list/postgresql/postgresql-integration/)  before installing the integration.
```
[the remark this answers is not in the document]
akristen: and maybe add a note "confirm your compatibility, then return to this doc." idk why but those little cues can sometimes help me remember to cmd+click lol. that just might be a me thing

src/content/docs/infrastructure/host-integrations/host-integrations-list/sonarqube-monitoring-integration.mdx
```diff
@@ -21,314 +21,230 @@ After setting up our SonarQube integration, we give you a dashboard for your Son
...
+    </Step>
+    <Step>
+## Configure NRI-Prometheus      
+
+1. Run the following command to create a NRI-Prometheus config file:
+
+  ```shell
+  touch /etc/newrelic-infra/integrations.d/nri-prometheus-config.yml
+  ```
+2. Paste the following snippet into the new config file. Be sure to update the `cluster_name` and `urls` fields:
```
akristen, line 48: update them with what? might be obvious, but do they only need to include cluster names for something related to SQ? do they need to include all cluster names? what urls (code comments say 'sq metrics lists' but why not just put that in the content part of the step) 

also, in your code comments, we should avoid using "Desired" but i'd follow up with A* about why. i can't remember but my takeaway was "don't"

src/content/docs/infrastructure/host-integrations/host-integrations-list/sonarqube-monitoring-integration.mdx
```diff
@@ -21,314 +21,230 @@ After setting up our SonarQube integration, we give you a dashboard for your Son
...
-    file: /opt/sonarqube/logs/es.log
-    attributes: 
-      logtype: sonar_es_logs
-  - name: web_logs
-    file: /opt/sonarqube/logs/web.log
-    attributes:
-      logtype: sonar_web_logs
-``` 
-
+Follow these steps to forward SonarQube logs to New Relic:
```
akristen, line 217: What level of detail do customers get by forwarding logs? Can they opt out of this if they have ingest concerns? In our other installs, logs tend to be optional but I noticed we have it as a required step here. We might want to explain how the ~fidelity of data gets degraded if they opt out (or maybe spin it positively to explain that logs are a key part oof the dashboard experience or w/e the reason is)

akristen: here + in the last OHI i reviewed

akristen (changes requested):
Copy + structure is great, so no asks there. I think a few steps should have a bit more detail on the _why_ customers are being asked to do this. I think we can get away with very barebones instructions in other OHIs, as they really have a limited number of installs to work through. Since this one is asking customers to do a little more upfront work, we might want to include how each discrete step relates to SonarQube (especially because there is no 'Install Sonarqube integration' step lol

PR #16725, opened 2024-03-29
fix DoNotTranslation deserialization

tabathadelane:
this PR does two main things:

- #### add upgraded versions of unified and related packages
  you can see these in code as `unified11` or `remark-parse10`, etc. it would be a huge undertaking to upgrade everywhere we use these packages, so this numbering lets us use the upgraded versions only in specific places. upgrading these packages in our deserialization code fixes the newline issue with DNT tags. there were some other minor changes necessary to get deserialization working with the upgraded packages
- #### add `uvu` and Sinon for ESM tests
  rather than desperately fight Jest to somehow get CJS and ESM tests working, this PR adds `uvu` specifically for running ESM tests. ESM tests are necessary as newer unified packages only have ESM support (no CJS), so our code has to be ESM as well. our test command now runs both Jest and uvu. one small note is that `uvu` does not support snapshot testing. i wrote a rudimentary module to accomplish this. i added Sinon for easier function mocking, as `jest.fn` can't be used outside of Jest

_serialization_ code is unchanged with this PR. i tried to keep the scope as tight as possible, so this only updates deserialization code and other files that needed to be updated to ESM out of necessity


scripts/actions/__esm-tests__/deserialize-html.test.mjs
```diff
@@ -23,6 +30,7 @@ test('deserializes mdx with DoNotTranslate', async () => {
 
   const mdx = await deserializeHTML(await serializeMDX(input));
 
+  // assert.is(mdx, input.trim());
```
clarkmcadoo, line 33: lets remove this

clarkmcadoo (approved):

austin-schaefer:
Hi @tabathadelane , is this PR still something we're actively working on?

PR #16748, opened 2024-04-01
feat: add new homepage banner and hide navigation

tabathadelane:
begining work on new homepage redesign
 - creates new HomePageLayout component that has no nav
 - new banner
 - remove any content we won't be using any longer
 - restyle SectionTitle

desktop/light mode
Screenshot 2024-04-01 at 5 32 53 PM
mobile/dark mode
Screenshot 2024-04-01 at 5 33 06 PM


src/layouts/HomePageLayout.js
```diff
@@ -0,0 +1,151 @@
...
+import { css } from '@emotion/react';
+import { scroller } from 'react-scroll';
+import { CSSTransition } from 'react-transition-group';
+import Layout from '../components/Layout';
+import SEO from '../components/SEO';
+import RootNavigation from '../components/RootNavigation';
+import { useLocation } from '@reach/router';
+import { MainLayoutContext } from '../components/MainLayoutContext';
+
+const MainLayout = ({ children, pageContext }) => {
```
clarkmcadoo, line 18: Do we still have a `MainLayout` template? If so let's rename this to be `HomePageLayout` and export it as such.

clarkmcadoo:
Mobile needs a little padding adjustment:
this is at 375px:
Screenshot 2024-04-03 at 11 04 28 AM

this is 500px:
Screenshot 2024-04-03 at 11 04 50 AM

tabathadelane:
> Do we still want to keep the mobile nav if we're getting rid of the nav on the desktop layout?
removed that and added some more responsiveness

LizBaker (approved):

PR #16825, opened 2024-04-05
fix(apm): updating definition of wall clock time

homelessbirds:


src/content/docs/apm/apm-ui-pages/monitoring/transactions-page-find-specific-performance-problems.mdx
```diff
@@ -48,9 +48,9 @@ The <DoNotTranslate>**Transactions**</DoNotTranslate> page also may include link
 
 ## Wall-clock time [#wall_clock_time]
 
-Wall-clock time is the amount of time recorded by the clock. New Relic uses wall-clock time for all the transactions and then sums that value across all of the transactions.
+Wall-clock time measures "real-time elapsed" during a specific transaction. For example, you're using a program to monitor transaction speeds. According to the program, the transaction took 30 seconds to complete, but one minute passed in the real world or on the wall clock due to discrepancies across the program. New Relic uses wall-clock time for all the transactions and then sums that value across all of the transactions.
```
austin-schaefer, line 51: > According to the program, the transaction took 30 seconds to complete, but one minute passed in the real world or on the wall clock due to discrepancies across the program

Wall clock time has never made sense to me, but to be honest I'm not sure if this update clarifies it. I think I'm getting stuck on "due to discrepancies across the program". What kind of discrepancies, and why? Why would something take 30 seconds according to my program, but a minute in "the real world"? 

Perhaps the challenge here is the focus on the wall-clock? Is the actual issue that people are using _programs that aren't accurate to real-world performance_? 

If so, perhaps we can reframe to be more explicit about that?

src/content/docs/apm/apm-ui-pages/monitoring/transactions-page-find-specific-performance-problems.mdx
```diff
@@ -48,9 +48,9 @@ The <DoNotTranslate>**Transactions**</DoNotTranslate> page also may include link
...
 
-The host can execute requests in parallel, so you may see percentages over 100. For example, 100% would indicate that the execution time across all selected transactions is equal to the time expended when recording wall-clock time.
+The host can execute requests in parallel so that you may see percentages over 100. For example, 100% would indicate that the execution time across all selected transactions equals the time expended when recording wall-clock time.
```
austin-schaefer, line 53: Not sure about the `,` > `that` change here; it seems to imply an intentionality. Perhaps instead `, which means` ? 

> The host can execute requests in parallel, which means you may see...

austin-schaefer (dismissed):

src/content/docs/apm/apm-ui-pages/monitoring/transactions-page-find-specific-performance-problems.mdx
austin-schaefer: One other thought—I think the round numbers here add a bit of confusion, particularly with the neat multiplying by two. It creates a sort of impression in my brain that `wall clock time = 2x other time`.

Perhaps do it with two unlike, non-round numbers? Like 22 seconds and 47 seconds.

src/content/docs/apm/apm-ui-pages/monitoring/transactions-page-find-specific-performance-problems.mdx
homelessbirds: Changed it back!

src/content/docs/apm/apm-ui-pages/monitoring/transactions-page-find-specific-performance-problems.mdx
homelessbirds: Fixed this!

src/content/docs/apm/apm-ui-pages/monitoring/transactions-page-find-specific-performance-problems.mdx
```diff
@@ -48,9 +48,9 @@ The <DoNotTranslate>**Transactions**</DoNotTranslate> page also may include link
 
 ## Wall-clock time [#wall_clock_time]
 
-Wall-clock time is the amount of time recorded by the clock. New Relic uses wall-clock time for all the transactions and then sums that value across all of the transactions.
+Wall-clock time measures "real-time elapsed" during a specific transaction. For example, let's say you're an engineer responsible for managing the checkout experience on an e-commerce site. You would like to understand how long it takes for a customer to add an item to their cart. The transaction took 15 seconds to complete but one minute in the real world or on the "wall clock" for the customer. This discrepancy in time could be because the function had to wait for network calls or other inputs and outputs, and that waiting time isn't considered in the system's transaction time. New Relic uses wall-clock time for all the transactions and then sums that value across all of the transactions.
```
akristen, line 51:
```suggestion
Wall-clock time measures "real-time elapsed" during a specific transaction. For example, let's say you're an engineer responsible for managing the checkout experience on an ecommerce site. You'd like to understand how long it takes for a customer to add an item to their cart. The transaction took 15 seconds to complete but one minute in the real world or on the "wall clock" for the customer. This discrepancy in time could be because the function had to wait for network calls or other inputs and outputs, and that waiting time isn't considered in the system's transaction time. New Relic uses wall-clock time for all the transactions and then sums that value across all of the transactions.
```

akristen: Great example and clarified things for me. I wonder if we should break up this chunk of text though, maybe introducing the time content with bullet points so it looks like:

> ...an item to their cart.
>
>  * The transaction took 15 seconds to complete, but one minute passed from the customer's point of view.
> * The time elapsed from the customer's point of view is called the "wall clock" for the customer.
>
> This discrepancy.... yadda yadda yadda

akristen (changes requested):
Made a suggestion to break up the text, and also a correction for ecommerce.

PR #16840, opened 2024-04-08
chore(java): removing security agent updates from security updates

meiao:
## Give us some context

When the Java agent 8.10.0 release notes were created, the security agent updates were added to the list of security updates.

Removing those as those show with a high priority on the update agent widget inside NR.


akristen (approved):
Looks good!

PR #16843, opened 2024-04-08
fix(sessionReplay): remove april 8 reference

ally-sassman:
Per the PM's request


akristen (approved):
@ally-sassman gave the 👍 that the date in this doc is an approved exception, so giving it an approval!!!!

PR #16847, opened 2024-04-08

bhecknr:
Per PMMs, revised "machine learning (ML) model performance monitoring" to "machine learning (ML) model monitoring."


akristen:
Hi @bhecknr ! I'm closing this ticket until product coordinates to update the UI to match this suggestion. 

- Your work won't be lost, and as soon as the change is reflected in the UI, we can get this re-opened and published
- When the PMs initiate this work to update language in product, please loop docs in at least 2 weeks ahead of the change so we can update the language in our docs as well

PR #16857, opened 2024-04-09
Update windows-linux-hybrid-cluster.mdx

bpschmitt:
Expand on explanation and other minor updates.

## Give us some context

Minor updates to existing docs.


akristen:
Giving this a review!

src/content/docs/kubernetes-pixie/kubernetes-integration/advanced-configuration/windows-linux-hybrid-cluster.mdx
```diff
@@ -28,9 +32,9 @@ Make sure you've completed the following before going further:
 
 ## Install the current chart bundle [#install-chart-bundle]
 
-The `newrelic/nri-bundle` chart is the regular one with all the latest updates for everything on Linux nodes. You can configure the `newrelic-logging` chart to install to Windows. This feature is reliable and supported in the logging chart.
+The `newrelic/nri-bundle` chart contains the latest updates for Linux-based nodes. The `newrelic-logging` chart can also be configured during this step to install to both Linux and Windows nodes. This feature for `newrelic-logging` is reliable and supported.
```
akristen, line 35: * How does the values.yaml file relate/influence/affect the install for Helm? 
* Can customers install helm first, then set up this file after? Or is the file necessary for directing certain behaviors when customers install helm? 

Some of this context seems out of place in this step and might be better placed during the actual Helm install. If the context needs to be here, can you explain how nri-bundle relates to the key/value pairs in this yaml file? For example, are the "charts" for nri-bundle these keys:

* nri-kube-events
* newrelic-prometheus-agent
* newrelic-infrastructure
* And others?

You also mention adding values for nodeSelector here in the case of a failed installation, but in the sample Yaml, nodeSelector is already added with some values. Are there additional values they should add? what are they?

src/content/docs/kubernetes-pixie/kubernetes-integration/advanced-configuration/windows-linux-hybrid-cluster.mdx
```diff
@@ -28,9 +32,9 @@ Make sure you've completed the following before going further:
...
 
-You can use other charts, but you must configure a `NodeSelector` to avoid scheduling to Windows. If an installation fails on a mixed cluster using default values from guided install, add the `nodeSelector` values.
+Other sub-charts in `nri-bundle` can be installed, but you must configure a `NodeSelector` to avoid scheduling to Windows. If an installation fails on a mixed cluster using default values from guided install, add the `nodeSelector` values.
 
 Create the values file with your favorite text editor, let's call it `values.yaml`. You can download a starter `values.yaml` file from the New Relic UI. Add the Kubernetes integration and choose **Helm** as the instrumentation method.
```
akristen, line 39: Are these last few steps mixing UI install procedures with manual install? We should focus this doc on manually installing this integration rather than throwing in UI install options, as it can lead to some confusion. It definitely confused me on a few rereads.

src/content/docs/kubernetes-pixie/kubernetes-integration/advanced-configuration/windows-linux-hybrid-cluster.mdx
bpschmitt: @akristen 

I think this doc was built assuming the user has some understanding of how our nri-bundle helm chart works (hence being nested under Advanced Configuration).

> How does the values.yaml file relate/influence/affect the install with Helm?

It is the primary configuration file that drives which charts are installed and how they are configured to run.

> Can customers install via helm first, then set up this file after? Or is the file necessary for directing certain behaviors when customers install helm?

It's necessary for a successful install / config regardless of __when__ the file is used.  It stores the configuration state.

> Some of this context seems out of place in this step and might be better placed during the actual Helm install.

There are two Helm install steps for this specific setup.  That is why I added the paragraph at the beginning to explain to users what they are about to step into as it is not our normal workflow.

> If the context needs to be here, can you explain how nri-bundle relates to the key/value pairs in this yaml file? For example, are the "charts" for nri-bundle these keys:

> nri-kube-events
newrelic-prometheus-agent
newrelic-infrastructure
And others?

These are sub-charts to the nri-bundle.  They are enabled/disabled and configured from the nri-bundle `values.yaml` file.

> You also mention adding values for nodeSelector here in the case of a failed installation, but in the sample Yaml, nodeSelector is already added with some values. Are there additional values they should add? what are they?

It depends on the environment and the cluster setup.  We could add some commented out examples but it would be up to the user to decide what is ultimately installed and how it fits inside the cluster (windows vs. linux).  One thing we could do for certain is to identify which components will not run on Windows no matter what.

src/content/docs/kubernetes-pixie/kubernetes-integration/advanced-configuration/windows-linux-hybrid-cluster.mdx
bpschmitt: @akristen 

> Are these last few steps mixing UI install procedures with manual install?

Not necessarily.  One place to get an example `values.yaml` file is from our UI.  The other is to get it from our Github repo, or potentially, this docs page.

src/content/docs/kubernetes-pixie/kubernetes-integration/advanced-configuration/windows-linux-hybrid-cluster.mdx
```diff
@@ -1,125 +1,154 @@
...
-When Helm returns, you should see `newrelic-windows-newrelic-infrastructure` pods on each Windows node.
-
-## Expected results [#expected-results]
-
-Once everything has been set, the New Relic UI should show your Windows nodes in the Kubernetes cluster. You'll see the infrastructure metrics, running pods, and all the related information. 
-
+To use this integration, you should:
+
+* Have a hybrid cluster for Linux and Windows, with the ability to pull container images to the Windows nodes.
+* Create a namespace:
```
akristen, line 21: not sure if we should just have 'create a namespace' as step 1 rather than a prerequisite

src/content/docs/kubernetes-pixie/kubernetes-integration/advanced-configuration/windows-linux-hybrid-cluster.mdx
```diff
@@ -1,125 +1,154 @@
...
+    helm repo update
+    kubectl create namespace newrelic
+    ```
+
+## Install the integration on a hybrid cluster [#install-hybrid]
+
+<Steps>
+    <Step>
+
+### Define your subcharts [#install-chart-bundle]
```
akristen, line 34: wanted to use different verbs for these steps. not sure if we have better titles to use, but wanted to avoid the 'install install install' repetition, since this step does something slightly different. my instinct could be a distinction w/o a difference though

src/content/docs/kubernetes-pixie/kubernetes-integration/advanced-configuration/windows-linux-hybrid-cluster.mdx
```diff
@@ -1,125 +1,154 @@
...
+    When Helm returns, you should see `newrelic-windows-newrelic-infrastructure` pods on each Windows node.
+    </Step>
+    <Step>
+    ### View your data [#view your data]
+
+    Once you generate some data, you can view it in New Relic. Go to <DoNotTranslate> **[one.newrelic.com](https://one.newrelic.com) > All capabilitiy > Kubernetes**</DoNotTranslate>, then select your cluster.
+    </Step>
+</Steps>
+
+## What's next? [#what's next?]
```
akristen, line 148: wanted to follow our pattern of having 'view your data' as part of a step procedure, and a what's next as a conclusion section. also not sure if i used the right pathing to view k8s data.

src/content/docs/kubernetes-pixie/kubernetes-integration/advanced-configuration/windows-linux-hybrid-cluster.mdx
nbaenam: I think we can keep it as a prerequisite.

src/content/docs/kubernetes-pixie/kubernetes-integration/advanced-configuration/windows-linux-hybrid-cluster.mdx
nbaenam: Yes, I agree with that.

nbaenam (approved):
I've added a commit with my edits.

PR #16875, opened 2024-04-10
fix: update misleading comment in kafka integration

sigilioso:

## Give us some context

This PR updates the title of a configuration example in the nri-kafka integration. The previous title `Zookeeper SSL discovery` suggest that the configuration example will set up the integration to connect to Zookeeper using SSL (which is not supported). The example actually configures the integration to connect to Zookeeper with [no authentication](https://docs.newrelic.com/docs/infrastructure/host-integrations/host-integrations-list/kafka/kafka-config/#zookeeper-autodiscovery) and to connect to JMX with SSL.


akristen:
Cache issues, so duplicated work here: https://github.com/newrelic/docs-website/pull/16881

PR #16898, opened 2024-04-11
chore: Add secure URL and custom headers auth to destination

nzur-newrelic:

## Give us some context

We added 2 new fields to the destination for security purposes. On this change we give more details on them, together with screenshots to make it easier to use for our users.


src/content/docs/alerts-applied-intelligence/notifications/notification-integrations.mdx
```diff
@@ -884,6 +888,28 @@ If you're building a new service to consume this payload you must test it locall
...
+  <img
+      width="80%;"
+      title="Webhook destination secure URL"
+      alt="Webhook destination secure URL"
+      src={accountsWebhookSecureUrl}
+    />
+
+### Custom Headers Authentication [#custom-headers-auth-webhook]
+
+Many users need to pass certain secrets in their headers, such as API keys and personal IDs. To prevent this sensitive information from being exposed, the custom headers authentication has been created. This will let you create headers that will be added to your notifications, while keeping their values encrypted and masked from both users and internal logs alike. In case a change in the value, the complete deletion and re-writing of the value will be needed. In case the same headers key will be used in both the Workflow channel and destination, the destination header will take precedence, and will be used instead of the channel header.
```
akristen, line 904:
```suggestion
In case the same headers key will be used in both the Workflow channel and destination, the destination header will take precedence, and will be used instead of the channel header. 
```

hi @nzur-newrelic ! Thanks for adding this content. Can you unspool this copy a little more?

- what are workflow channels?
- What are destinations in terms of header?
- How does a header key relate to the value assigned to sensitive information? 
- What is the "destination header taking precedence" over? 
- What are the differences between destination and channel headers? How do values relate to them?

akristen (changes requested):
hi @nzur-newrelic ! Thanks for adding this content. Can you unspool this copy a little more?

- what are workflow channels?
- What are destinations in terms of header?
- How does a header key relate to the value assigned to sensitive information?
- What is the "destination header taking precedence" over?
- What are the differences between destination and channel headers? How do values relate to them?

akristen (approved):

akristen (approved):

PR #16902, opened 2024-04-11
Licenses: Add EU reference to CCUs

rhetoric101:
The goal of this PR is to add language about regions that is similar to the GB_Month collapser.


src/content/docs/licenses/license-information/product-definitions/new-relic-one-pricing-definitions.mdx
```diff
@@ -37,6 +37,8 @@ This is a glossary of terms that appear in contracts for our [New Relic usage-ba
     title="Compute Capacity Unit (CCU)"
   >
     Compute Capacity Unit ("CCU") means the compute capacity consumed for a successfully-run action, including loading a page, executing a query, evaluating an alert condition, and invoking an API call. CCU includes the queries, usage, and processes necessary to retrieve your data, perform the user functions, or serve up New Relic features.
+
+    This unit of measure may indicate a data region, such as `EU`, associated with your CCU, as applicable.
```
jeff-colucci, line 41: Should it be unit of measurement? Honestly my gut instinct is to just replace "This unit of measure" with "CCUs" or "These units" or something like that to tie it in with the preceding paragraph

jeff-colucci (approved):

PR #17020, opened 2024-04-22
Update csec-go-agent support-matrix link

aayush-ap:
- Update csec-go-agent support-matrix link


rhetoric101:
@aayush-ap it looks like you need to sign our CLA before we can publish this. Can you sign that even if you've done this in the past?

rhetoric101 (approved):
@aayush-ap thanks for correcting this link!

PR #17024, opened 2024-04-22
feat(RuntimeUpgradeNerdGraph): Query examples

bpeckNR:
I added query examples for the runtime validation entities used by the runtime upgrades UI. This will allow users to interact with these runtime upgrade tests programmatically if they choose to do so.


rhetoric101 (approved):
@bpeckNR thanks for adding these two queries!

PR #17043, opened 2024-04-23
Fix jump link for compressed content encoding

lrafeei:
This PR fixes the jump link for the `compressed-content-encoding` setting within the Python Agent's configuration page.


rhetoric101 (approved):
@lrafeei thanks for updating this ID.

PR #17059, opened 2024-04-23
fix(K8s): Clarified K8s support

bpeckNR:
- Removed Windows as a supported deployment option for K8s
- Included that Linux containers can only be run on Linux K8s nodes


rhetoric101 (approved):
@bpeckNR sorry this took so long to approve! We had some technical issues. Thanks for  adding these comments.

PR #17081, opened 2024-04-24
chore(NRQL): Update reserved words links

austin-schaefer:


paperclypse (approved):
Checked all links in local build. Looks good!

PR #17088, opened 2024-04-25
chore: reformat classloading-issues-from-jboss-and-wildfly

kanderson250:
## Give us some context

This PR reformats a troubleshooting page to strip out the tabs and put everything on a single page. No actual text has been changed (except one typo). 

We determined this was necessary to make the content of the second tab more discoverable. Specifically, we want to be able to link to the Solution headers and for users to find everything on the page with a CONTROL-F.


rhetoric101 (approved):
@kanderson250 thanks for revamping this doc to remove tabs. This makes it more user-friendly!

PR #17093, opened 2024-04-25
update webdriver

tabathadelane:
update webdriver scripts after homepage redesign


sunnyzanchi (approved):
thanks for adding helpful comments/console.logs!

PR #17095, opened 2024-04-25
Update Config with Trim Stats Info

deleonenriqueta:
Update the agent config page to include trim_stats under *General Configuration Setting*

### DIFF
--- a/src/content/docs/apm/agents/java-agent/configuration/java-agent-configuration-config-file.mdx
+++ b/src/content/docs/apm/agents/java-agent/configuration/java-agent-configuration-config-file.mdx
@@ -1209,6 +1209,43 @@ Set these options in the `common` section. To [override](#System_Properties) any
       enabled: true
     ```
   </Collapser>
+
+  <Collapser
+    id="trim_stats"
+    title="trim_stats"
+  >
+    <table>
+      <tbody>
+      <tr>
+        <th>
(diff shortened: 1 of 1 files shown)


rhetoric101 (approved):
@deleonenriqueta thanks for adding this trim_stats collapser!

src/content/docs/apm/agents/java-agent/configuration/java-agent-configuration-config-file.mdx
```diff
@@ -1209,6 +1209,43 @@ Set these options in the `common` section. To [override](#System_Properties) any
...
+        <td>
+          `true`
+        </td>
+      </tr>
+      </tbody>
+    </table>
+
+    When the exclusive time of a segment is less than 2% of the total transaction time, the metric for that method call will be aggregated under `Java/other`.
+
+    Setting `trim_stats` to `false` will disable this behavior.
```
kford-newrelic, line 1243: Can we also add/referencing that disabling this behavior will cause the actual metric to be displayed or something like that?

Would like to make it clear to customers what the benefit/value of disabling it is.

kford-newrelic (commented):
Can you also check the Scala related config just above the `trim_stats` - I thought that the configs in this section were ordered alphabetically but that one doesn't appear to be

PR #17096, opened 2024-04-25
style(nav): `APM agents` instead of `Install`

austin-schaefer:
Inspired by frequent confusion when I watch people navigate docs, most recently @jeff-colucci earlier today. 

Would like input from @bradleycamacho / @rhetoric101 , one of whom I believe originally made this change away from calling it APM agents?


rhetoric101 (approved):
@austin-schaefer I approve of this change!

PR #17100, opened 2024-04-26
fix(Azure infra): Fixed a broken link

nbaenam:
Fixed a broken link.

This comes from this `docs feedback` ticket: NR-262347


paperclypse (approved):
Confirmed it's fixed in preview

PR #17105, opened 2024-04-26
fix(OMA): fixing OMA docs still in file system

jeff-colucci:


bradleycamacho (dismissed):
@jeff-colucci this looks good to me! 

The merge conflict is a bit of a pain, it's just confused since your PR deletes those files but they've been updated since you created this branch. Just make sure you don't "fix" the conflict the wrong way and keep the translated files on accident!

jeff-colucci:
Merging after prior approval (I only merged develop in as the change since the review)

PR #17120, opened 2024-04-29
fix(OptionalFields): Added optional fields

bpeckNR:
Added optional runtime related fields to broken link, cert check, and step monitors to match other monitor types. Using these fields allows you to create a monitor that utilizes newer synthetics runtimes and private job managers (if applicable).


homelessbirds (approved):

homelessbirds:
Thanks, Brian!

PR #17137, opened 2024-04-30
feat(Migration developer): Added account picker and button from SDK s…

nbaenam:
Migration of the SDK component library docs from the developer site to the docs site. 

This is the ticket: NR-261819


paperclypse:
@nbaenam I converted this to a draft, because it seems incomplete

paperclypse (approved):
Looks good to me, thanks!

PR #17139, opened 2024-04-30
fix: dynamically set text width based on parent width

clarkmcadoo:


tabathadelane (dismissed):
Looks good after resizing 🎉

tabathadelane:
Looks really good. I dont see the text issue. Though it does kind of look like the icon and the maybe the border radius is also being scaled. Is it possible to not scale the icon?  If not, it's better to have that than clipped text. Inspecting the border radius shows the same 10px border radius on hover, so maybe its just my eyes.

Screenshot 2024-05-01 at 4 56 01 PM

sunnyzanchi:
@tabathadelane i should be possible to do the inverse transform on the icon as well. as for the border radius, i see it too. i think we can avoid that if we put the border-radius on the product tile div instead of the SurfaceLink

sunnyzanchi (approved):

PR #17165, opened 2024-05-01
Page background color

tabathadelane:
- theme bump that updates site page background color
- override homepage background to old color to maintain contrast
- double checked other components using this variable, and everything seems fine. Though DocTiles have lost some contrast, but I think the drop shadow still makes them stand out enough.


clarkmcadoo:
Dark mode is displeased:
Screenshot 2024-05-06 at 1 28 41 PM

clarkmcadoo (approved):

PR #17175, opened 2024-05-02
Update JP nav file: jp.json

jmiraNR:
Update JP nav file: jp.json


homelessbirds:
Merging with @jmiraNR's blessing.

PR #17222, opened 2024-05-07
always link to /whats-new/feed.xml

sunnyzanchi:
i18n versions of this page were linking to `/<locale>/whats-new/feed.xml`, which doesn't exist, as the RSS feed isn't translated. this change makes sure we always properly link to the RSS feed


clarkmcadoo:
still getting the 404 on this page - https://deploy-preview-17222--docs-website-netlify.netlify.app/es/whats-new/

clarkmcadoo (approved):

PR #17253, opened 2024-05-09
generate RSS feed at i18n paths

sunnyzanchi:
trying pointing the link at `/whats-new/feed.xml` regardless of language didn't work. this updates the whats new RSS feed generation plugin to generate the feed at the appropriate i18n path for the non-English sites

i added this as a branch deploy branch for the JP site, it can be tested [here](https://sunny-rss-feed-fix--docs-website-jp.netlify.app/jp/whats-new/)


tabathadelane (approved):
spiffy ✨

PR #17271, opened 2024-05-10
test(NET): Methods in h2s instead of collapsers

austin-schaefer:
H2 variant of the NET API doc


austin-schaefer:
Builds aren't triggering, I think b/c this PR is against a branch besides develop. Closing in favor of https://github.com/newrelic/docs-website/pull/17274 which is identical

PR #17296, opened 2024-05-14
chore(errorsInbox): move version tracking content 

ghost:
The errors inbox PM requested to move version tracking content from the error tracking page to the version tracking page. I also included some misc. fixes found along the way.


rhetoric101 (approved):
@ally-sassman thanks for making these changes!

PR #17316, opened 2024-05-15
style(k8s): Make "What's next?" the last section of doc

austin-schaefer:
Feels weird to not have "What's next?" as the last section in a doc


jeff-colucci (approved):
Looks good!

PR #17330, opened 2024-05-16
Update azure-log-forwarding.mdx with size constraint

kmad-newrelic:
There is a size constraint on Azure Function uploads using the Node.js binding. This is a known issue on Azure's side. See here: https://github.com/Azure/azure-functions-host/issues/1361
This has caused a customer issue in the past and we're adding this section to make people aware of it.


kmad-newrelic:
Please don't merge this yet, I want some teammates to look over it as well.

jeff-colucci (dismissed):
Made some small style changes but otherwise it looks good to me!

jeff-colucci (approved):

PR #17368, opened 2024-05-20
Add new translated languages

bradleycamacho:
Adds Spanish and Portuguese to list of translated languages.


jeff-colucci (approved):
They look added to me!

PR #17376, opened 2024-05-20
fix(severless): adding diagram and environment variables

homelessbirds:
Further explaining environment variables for containerized images. Added diagram for context.


nbaenam (approved):
Hi @homelessbirds, I've added some minor TW edits

PR #17409, opened 2024-05-22
Clarify link, table syntax guidelines

austin-schaefer:
* Clarifies that we only use inline links (and why)
* Clarifies we don't use markdown tables (and why)
* Combine `Formatting` and `Structure` categories together
* Corrects various outdated and inaccurate info about tables
* Various minor wording and clarity fixes to various component docs


ally-sassman (approved):

paperclypse (dismissed):
Thanks for doing this!

paperclypse (approved):
better cat name

PR #17426, opened 2024-05-23
fix(Logs): Added missing header id

paperclypse:
Added missing header id to one you worked on


jeff-colucci (approved):
Seems added to me!

PR #17438, opened 2024-05-24
Update troubleshooting.mdx for certificate and firewall details

ankit-k2io:


nbaenam (approved):
Hi @ankit-k2io, thank you for the update! I've added a commit to update the UI path in the entire doc site.

nbaenam (approved):

PR #17431, opened 2024-05-23
docs: Update .NET agent install docs to reference the .NET agent samples repo

tippmar-nr:
Updates the .NET agent install docs in a few places to reference our recently-created code samples repository.

PR #17441, opened 2024-05-24
chore: Correct URLs for some references to the .NET agent samples repo.

tippmar-nr:
This PR updates two URLs that were incorrectly modified by the reviewer in #17431. We're using specific URLs on these pages to point to the examples relative to the docs page.


nbaenam (approved):
Hi @tippmar-nr, thank you for the update!! It looks good to me!

PR #17491, opened 2024-05-30
Logs: Add video and remove tab reference

rhetoric101:
We received a new video that explains this feature. Also, we need to remove the reference to the "Formatted" tab.


paperclypse (approved):
Video works, looks good to me

PR #17500, opened 2024-05-30
fix(OTel): Deleted developer OTel docs until we can decide what to do with them

paperclypse:
Moving these Developer site OTel docs at the request of the OTel team.

Hey @nbaenam, the OTel teams has some serious qualms about the OTel content we migrated from the Developer site and would prefer that we just delete it. That's what I've done in this PR.


paperclypse:
@nbaenam I'm deleting these files for now, thanks! I thought I had removed them from the yaml too.

nbaenam (approved):
LGTM!

PR #17506, opened 2024-05-31
chore(ruby-agent): Update CSP Nonce feature content

kaylareopelle:
`browser_monitoring.content_security_policy_nonce` is off by default. Our most recent release note and browser monitoring doc updates previously stated it was on by default.


rhetoric101 (approved):
@kaylareopelle thanks for adding these clarifications!

PR #17509, opened 2024-05-31
AI Monitoring: First attempt to fix squished images

rhetoric101:
We have two side-by-side images that are getting stretched to fit the view area.


rhetoric101:
Our developers implemented a temporary fix, so I am closing this.

PR #17528, opened 2024-06-03
Fixed mismatched header and rows in lookups API example

prodjito:
Fixed mismatched header and rows in lookups API example. Header has 6 values but rows only have 5 values. I removed the "text" in the header.


nbaenam (approved):
Hi @prodjito, thank you for the update!! It looks good to me!

PR #17542, opened 2024-06-04
.NET agent: add latest tested version of Microsoft.Data.SqlClient to compatibility and support docs

nr-ahemsath:
The .NET agent team has tested a new version of a supported library and are adding it to our compatibility and support docs.


nbaenam (approved):
Hi @nr-ahemsath, thank you for the update!! It looks good to me!

PR #17543, opened 2024-06-04
chore(ruby-agent): Add a section about turning off the agent 

kaylareopelle:
We want to document that `Rails::Command::RakeCommand` may not be desired in all contexts. Adding a section about turning off the agent in certain contexts to this existing page seemed like a good place to do that.

However, it does provide a lot of duplication with the existing content. There may be a better way to achieve this goal.


src/content/docs/apm/agents/ruby-agent/troubleshooting/control-when-ruby-agent-starts.mdx
```diff
@@ -128,3 +128,45 @@ To solve this problem:
       </table>
   </Collapser>
 </CollapserGroup>
+
+## Problem
+
+The Ruby agent is starting in contexts I don't want it to start.
```
kaylareopelle, line 134:
```suggestion
The Ruby agent is starts in contexts where I don't want it to start.
```

fallwith (approved):

jeff-colucci (approved):

PR #17545, opened 2024-06-05
Prometheus: Change title

rhetoric101:
Clarify that this integration is for host monitoring (per SME).


nbaenam (approved):
It looks good to me, although the nav title is `Node Exporter` and we should try to keep the consistency between nav titles and doc titles.

rhetoric101:
@nbaenam good point! Still, in this case, we're just adding this to help search engine optimization. I think it'll be OK to keep the shorter version in the left navigation because the longer version would be difficult to scan in the list.

PR #17554, opened 2024-06-05
fix intermittent webdriver failures

sunnyzanchi:
this adds a loop to redo the XPath selector for the release notes div if it comes back null. it also updates the headless argument per [this article](https://www.selenium.dev/blog/2023/headless-is-going-away/) and lowers the sleep time since it doesn't seem to need to be so long


clarkmcadoo (approved):

tabathadelane:
if there is actually an issue, will this eventually timeout and fail, or will it keep looping?

sunnyzanchi:
> if there is actually an issue with this xpath element, will this eventually timeout and fail, or will it keep looping?

good point. yes, it would continue looping. i'll add a max number of tries

clarkmcadoo (approved):

PR #17557, opened 2024-06-05
fix(OTEL): Move OTEL into its own tile

ghost:
Moved OTEL out of the **Other integrations** tile to match the current left nav. Waiting to merge so that docs-eng can create an OpenTelemetry icon.


paperclypse (dismissed):
Great idea!

PR #17559, opened 2024-06-05
java ai monitoring pr

akristen:


src/content/docs/apm/agents/java-agent/api-guides/ai-monitoring.mdx
```diff
@@ -13,9 +13,11 @@ redirects:
 
 When you've instrumented your app with AI monitoring, you can access some APIs for collecting token count and user feedback. To use the AI monitoring APIs, check that your Java agent is updated to version 8.12.0 or higher.
 
+This doc gives you high level procedures for updating your code to access the [token count and user feedback APIs](https://newrelic.github.io/java-agent-api/javadoc/com/newrelic/api/agent/AiMonitoring.html#recordLlmFeedbackEvent).
```
jeff-colucci, line 16: Not sure about the use of "high level" here

src/content/docs/apm/agents/java-agent/api-guides/ai-monitoring.mdx
```diff
@@ -37,20 +39,13 @@ If you've disabled the agent with `ai_monitoring.record_content.enabled=false`,
         NewRelic.getAgent().getAiMonitoring.setLlmTokenCountCallback(myTokenCountCallback);
     ```
 
-To use the callback, implement `LlmTokenCountCallback` so that:
-
-* It returns an integer that represents the number of tokens for a particular prompt, completion message, or embedding. 
-* If the values are less than or equal to 0, `LlmTokenCountCallbacks` will not be attached to an event.
-
-Keep in mind that this API should be called only once. Calling this API multiple times will replace each previous callback. 
+To use the callback, implement `LlmTokenCountCallback` so that it returns an integer that represents the number of tokens for a particular prompt, completion message, or embedding. If the values are less than or equal to `0`, `LlmTokenCountCallbacks` will not be attached to an event. Keep in mind that this API should be called only once. Calling this API multiple times will replace each previous callback.
```
jeff-colucci, line 42: will not > won't

jeff-colucci: Keep in mind that you should only call this API once

jeff-colucci (dismissed):
Looks good to me! Just found a couple things, once you fix them you should be good to go!

src/content/docs/apm/agents/java-agent/api-guides/ai-monitoring.mdx
akristen, line 16:
```suggestion
This doc provides procedures for updating your code to access the [token count and user feedback APIs](https://newrelic.github.io/java-agent-api/javadoc/com/newrelic/api/agent/AiMonitoring.html#recordLlmFeedbackEvent). 
```

src/content/docs/apm/agents/java-agent/api-guides/ai-monitoring.mdx
akristen, line 42:
```suggestion
To use the callback, implement `LlmTokenCountCallback` so that it returns an integer that represents the number of tokens for a particular prompt, completion message, or embedding. If the values are less than or equal to `0`, `LlmTokenCountCallbacks` won't attach to an event. Keep in mind that you should only call this API once. Calling this API multiple times will replace each previous callback. 
```

homelessbirds (approved):

PR #17566, opened 2024-06-06
fix(Docs site): Replaced callout note with important

nbaenam:
Replaced `<Callout variant="note">` with `<Callout variant="important">`.


paperclypse (approved):
Thank you for fixing these, @nbaenam!

PR #17604, opened 2024-06-10
fix(accounts): improve instructions for adding user groups

ghost:
This change is in response to customer feedback, where they said they couldn't find the information in the forum link. I reworked the section to improve findability.


homelessbirds (approved):
Looks good. Much more detailed and explicit. Thanks :)

PR #17629, opened 2024-06-12
fix(alerts): deleted old docs

homelessbirds:
I deleted two docs that are no longer relevant. Consulted with BG on the alerts team.


paperclypse (approved):
I assume you know what you're doing :) Looks like they're deleted. 

There's something funny going on with the vale linter check. You may need to check with our eng team on that one.

clarkmcadoo (approved):

PR #17636, opened 2024-06-13
Add back AddRemoveRedirects script with commander

LizBaker:
The script can now be run like so:

`yarn addRedirects src/content/docs/accounts/accounts-billing`
or
`yarn removeRedirects src/content/docs/accounts/accounts-billing`

this uses the relative path of a given docs content folder

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "@aws-sdk/client-s3": "^3.327.0",
     "@emotion/react": "^11.7.1",
     "@emotion/styled": "^11.3.0",
+    "@github-docs/frontmatter": "^1.3.1",
     "@mdx-js/mdx": "2.0.0-next.8",
     "@mdx-js/react": "2.0.0-next.8",
     "@newrelic/gatsby-theme-newrelic": "9.6.0",
@@ -175,6 +176,7 @@
   "license": "Apache-2",
   "scripts": {
     "add-files-to-translate": "node scripts/actions/add-files-to-translation-queue.js",
+    "addRedirects": "node scripts/addRemoveRedirects.js addRedirects",
     "build:production": "ENVIRONMENT=production yarn run build",
     "build": "node scripts/createNetlifyRedirects.mjs & NODE_OPTIONS='--max-old-space-size=5120' gatsby build --prefix-paths",
     "check-for-outdated-translations": "node scripts/actions/check-for-outdated-translations.js",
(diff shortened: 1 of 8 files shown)


scripts/handlers.js
```diff
@@ -0,0 +1,124 @@
+/* eslint-disable no-console */
+const { globSync } = require('glob10');
+const frontmatterGH = require('@github-docs/frontmatter');
```
sunnyzanchi: we have a `scripts/utils/frontmatter.js` module that can handle parsing frontmatter from a file. it doesn't have a stringify currently, but i'd rather we add it there than pull in another package

sunnyzanchi (commented):
right now the interface is `yarn addRedirects` or `yarn removeRedirects`. i think it might be a little nicer if it was `yarn redirects add` or `yarn redirects remove`. we'd leave it open for future operations, and `yarn redirects --help` would show the help for both. @LizBaker what do you think?

clarkmcadoo (dismissed):

scripts/handlers.js
LizBaker: ooh gotcha, this is just from the original script

scripts/README.md
```diff
@@ -66,3 +66,33 @@ It takes no arguments and will complete in about ~20min. It was created to be ru
...
+
+#### Example:
+
+`yarn redirects add src/contents/docs/taco`
+
+the file at path src/contents/docs/taco/cheese.mdx will now have a new redirect
+
+```
+redirects:
+- src/contents/docs/taco/cheese.mdx
```
sunnyzanchi, line 88: small note on these two redirects, redirects have slightly different paths
```suggestion
- /docs/taco/cheese
```

package.json
```diff
@@ -175,6 +175,7 @@
   "license": "Apache-2",
   "scripts": {
     "add-files-to-translate": "node scripts/actions/add-files-to-translation-queue.js",
+    "addRedirects": "node scripts/addRemoveRedirects addRedirects",
```
sunnyzanchi, line 178: did we still want to keep this?

sunnyzanchi (approved):

scripts/README.md
LizBaker: whoops!

package.json
LizBaker: def not

sunnyzanchi (approved):

PR #17688, opened 2024-06-18
fix(dotnet): Corrects an error in the TraceMetadata code sample.

tippmar-nr:
Fixes a minor error in the code sample for the `TraceMetadata` property in the .NET agent API.


akristen (approved):
Looks good! Thanks a bunch :)

PR #17698, opened 2024-06-18
fix: transaction_tracer.transaction_threshold config

brnhensley:
The details of this config were incorrect. After discussing with the node team I'm attempting to update it using their [recently updated internal documentation](https://github.com/newrelic/node-newrelic/blob/b1b5e3e54074cc8b535927f4edad07925618260c/lib/config/default.js#L521-L528).  

This PR is waiting on review from the Node team.

One point needs clarifying before approval. The internal docs state:
> When a transaction exceeds this threshold, a transaction trace will be recorded

Discussing this with the engineers my understanding was this did not determine if a transaction trace was recorded, only if it would be considered 'slow'.


jsumners-nr (approved):

jsumners-nr (approved):

rhetoric101 (approved):
@brnhensley thanks for tracking this down!

PR #17703, opened 2024-06-19
Migrate topics out of OTHER CAPABILITIES

rhetoric101:
This is work for NR-262987


src/nav/infrastructure.yml
```diff
@@ -142,6 +142,8 @@ pages:
             path: /docs/infrastructure/host-integrations/host-integrations-list/active-directory-integration
           - title: Aerospike integration
             path: /docs/infrastructure/host-integrations/host-integrations-list/aerospike-integration
+          - title: Monitor Apache Airflow with OpenTelemetry
```
paperclypse, line 145: Seems odd the way this is alphabetical order. Can we lose the "Monitor" here of shift it down the Ms?

src/nav/infrastructure.yml
```diff
@@ -176,6 +188,8 @@ pages:
             path: /docs/infrastructure/host-integrations/host-integrations-list/hashicorp-consul-monitoring-integration
           - title: IBM MQ monitoring integration
             path: /install/ibm-mq
+          - title: Monitor Jenkins with OpenTelemetry
```
paperclypse, line 191: Another one that seems oddly out of alphabetical

paperclypse (approved):
Looks good. Left a couple minor comments. Also, we'll need to update the homepage to match.

src/nav/infrastructure.yml
rhetoric101: @paperclypse I removed "Monitor."

src/nav/infrastructure.yml
rhetoric101: @paperclypse I removed "Monitor."

paperclypse (approved):
Looks good

paperclypse (approved):
Looks fine, as far as I can tell. I think you can ignore that failed vale linter check.

PR #17710, opened 2024-06-19
feat(.NET): Azure troubleshooting left nav move

jeff-colucci:
Moved Azure troubleshooting docs to live with the existing troubleshooting section for future rework


homelessbirds (dismissed):

homelessbirds (approved):

homelessbirds (changes requested):
Looks like the nav isn't quite right! Let me know if you need help with the translations

PR #17738, opened 2024-06-21
Android v7.4.0 release notes minor changes

ywang-nr:
* What problems does this PR solve? - minor changes due to PM request. The reason for taking the configuration out is this configuration will be controlled from UI config of NR1, so we wouldn't need the agent side config.


akristen (approved):
LGTM

PR #17767, opened 2024-06-25
Whats new 06 25 eol synthetics runtime cpm

louisleungk:
1. Add EOL post for legacy synthetics runtime
2. Updating previous post, as advised by docs team


rhetoric101 (approved):
@louisleungk thanks for making this update and adding the new post!

PR #17801, opened 2024-06-26
fix(alerts): removing outdated rest api documenation

homelessbirds:
Removing outdated references to the alerts REST API.


rhetoric101 (approved):
@homelessbirds thanks for removing this deprecated section.

PR #17815, opened 2024-06-27
Link check action

LizBaker:
This action will run once a week and create a github issue for any broken links it finds

Test runs:
Skipped: https://github.com/newrelic/docs-website/actions/runs/9769339799
Success: https://github.com/newrelic/docs-website/actions/runs/9769186972
	Issue created: https://github.com/newrelic/docs-website/issues/17884


.github/workflows/check-for-broken-links.yml
```diff
@@ -0,0 +1,67 @@
...
+    needs: check-links
+    if: ${{needs.check-links.outputs.broken-links}} == 'true'
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Create broken link issue
+        run: |
+          gh issue create \
+          --title "Broken link/s found" \
```
sunnyzanchi, line 63: i think this is a typo

```suggestion
          --title "Broken links found" \
```

scripts/actions/check-links/broken-links-filter.txt
sunnyzanchi, line 1: i don't think we should commit either of these txt files to the repo, since they're just intermediate steps for the script/action. if we added them to the gitignore, we could run the script locally without cluttering the working tree (we would have to do something instead of git diff in the action though)

sunnyzanchi (changes requested):

.github/workflows/check-for-broken-links.yml
LizBaker: i have it that way in case its a single link but i could also do link(s) or just call it plural because no one really cares 😆

scripts/actions/check-links/check-links.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+find ./src/content/docs/browser/new-relic-browser/troubleshooting \
```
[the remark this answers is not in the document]
LizBaker: i don't really foresee a need to run it on a particular directory once its up and running weekly

.github/workflows/check-for-broken-links.yml
sunnyzanchi: oooo. in that case i think link(s) would be more clear!

sunnyzanchi (approved):

PR #17843, opened 2024-06-28
Browser and OpenTelemetry: Fix indentation for redirects and add anchors.

rhetoric101:
This is a follow up to some previous PRs. The redirects in the OpenTelemetry file were never inserted by the redirect automation script (probably because the indentation was wrong). Also, I missed the anchors tags in the headings of two browser APIs. Note that very few of the existing API files have anchors. I think what happens is that contributors just copy an existing file and use it to model the next API.


paperclypse (approved):
Looks good, thanks!

PR #17885, opened 2024-07-02
chore(sweep): Remove all inline signup forms

austin-schaefer:


jeff-colucci (approved):
They all seem removed to me! The additions look fine to my eyes.

PR #17926, opened 2024-07-08
Docs: add in release notes for the iOS mobile app version 6.6.0

jvanmetre:
* Adds in latest iOS mobile app release notes
* Updates licenses to reflect the latest iOS mobile app


bradleycamacho:
Hi @jvanmetre - thanks for the update! Do we need to loop in legal for the license changes?  Looks great otherwise

bradleycamacho (approved):

jvanmetre:
Hi @bradleycamacho I don't believe so, the only slightly spicy one is the antler-4 bsd one, but that one is already in the java papers manifest repo, so that has previously been approved from legal... the other two are apache and MIT, which are approved.

bradleycamacho:
@jvanmetre gotcha - merging in now!

PR #17966, opened 2024-07-10
chore(Infra): added missing header anchors

jeff-colucci:


rhetoric101 (approved):
@jeff-colucci thanks for adding these anchors!

PR #17967, opened 2024-07-10
fix: update kr nav

clarkmcadoo:


LizBaker (approved):

gmanandhar-nr:
Hi @WriteMayur , the kr nav file we currently have in develop should be the updated one, and this may be a stale PR. hence, closing this one.

PR #17970, opened 2024-07-11
fix(NRQL Dashboards): Updated charts and dashboard page

cbehera-newrelic:
* [Updated charts and dashboard page](https://docs.newrelic.com/docs/nrql/get-started/charts-and-dashboards-with-nrql/)

* Ticket no. NR-266790


rhetoric101 (commented):
@cbehera-newrelic thanks for adding these improvements! Let's chat about a few more things we could add.

rhetoric101 (approved):
@cbehera-newrelic thanks for making these improvements!

PR #17987, opened 2024-07-11
Node.js agent: Add link from Node docker to main Node instructions

rhetoric101:
Jira NR-287339 said it would be helpful to have a link to the main Node.js instructions from the Docker instructions.


paperclypse (approved):
Seems fine. I'm not sure how I feel about "main", but I don't think it's worth holding this one up because of it.

PR #17988, opened 2024-07-11
Data APIs: Fix spurious link

rhetoric101:
Jira NR-287391 noted that the link was to a doc about custom attributes--not events. I switched the link to a local anchor.


jeff-colucci (approved):
It appears fixed to me!

PR #18055, opened 2024-07-17
Feat: new EOL Comms pages

tabathadelane:
opening for now for previews

page at `/eol`


src/pages/eol.js
```diff
@@ -0,0 +1,214 @@
...
+import EolTable from '../components/EolTable';
+import { DIRECTION } from '../components/EolTable/TableHeader';
+import PageTitle from '../components/PageTitle';
+import SEO from '../components/SEO';
+import ErrorBoundary from '../components/ErrorBoundary';
+import { TYPES } from '../utils/constants';
+
+// @tw: add any body content here.
+// it will go under the main heading and above the EOL timeline.
+const body = <p>howdy 🤠</p>;
```
paperclypse, line 20: @tabathadelane I think we want to remove this :)

paperclypse (changes requested):
Once you remove the howdy and get this added to the top-level nav, I'm good to approve it @tabathadelane

src/pages/eol.js
tabathadelane: Oh right. This was requested by Austin. Can we put some content in here?

tabathadelane: ^ I just commented this out for now.

paperclypse: Oh, I didn't realize we wanted some content to go there!

src/nav/root.yml
```diff
@@ -54,16 +54,18 @@ pages:
...
+    path: iast
   - title: Vulnerability Management
-    path: vuln-management 
+    path: vuln-management
   - title: Latest updates
   - title: Release notes
     path: release-notes
   - title: What's new?
     path: whats-new
+  - title: End-of-life updates
```
paperclypse, line 67:
```suggestion
  - title: End-of-life announcements
```

I was worried that this would be too many characters, but I think it'd be good to match the page title now that I see it in preview

paperclypse (approved):
Looks good to me now!

PR #18062, opened 2024-07-17
feat: add rss feed for eol posts

clarkmcadoo:
Adds RSS feed to for EOL pages. Go to `/eol` and click the button in the top right

Screenshot 2024-07-17 at 3 18 31 PM


plugins/gatsby-plugin-eol-rss/gatsby-node.js
```diff
@@ -0,0 +1,127 @@
+const preferDefault = (m) => (m && m.default) || m;
```
sunnyzanchi, line 1: just curious about this, does the parseISO require not work without it? why does format not need it?

clarkmcadoo: I'm not sure actually, I'm gonna get rid of it. Long term effects of a c/p

sunnyzanchi (approved):

PR #18063, opened 2024-07-17
Python agent: Move Airflow integration to Python agent as a third-party integration

rhetoric101:
In a previous PR, I had moved Apache Airflow from "other-integrations" to "on-host integrations." After thinking about it and chatting with the Python team, they said I could move this to a new "Third-party integrations" section in the Python agent.


paperclypse (approved):
The move makes sense to me

rhetoric101:
After more discussion, I moved Airflow into a new category called "Other infrastructure integrations." So, I am closing this since we're not going to move this to Python.

PR #18069, opened 2024-07-18
chore(accounts): remove RightScale topic

ubanerjeeNR:
* Deletes an old RightScale topic
* JIRA ticket: NR-289019


cbehera-newrelic (approved):
required file and the orphaned files (i18 files) got deleted successfully. Approving the PR.

PR #18109, opened 2024-07-23
Chore(infra logs): Updated docker image for log forwarding

cbehera-newrelic:
- [Updated with an "important message" to use the appropriate docker image  for log forwarding](https://docs.newrelic.com/docs/logs/forward-logs/forward-your-logs-using-infrastructure-agent/#infra)
- Ticket no: NR-218095


src/content/docs/logs/forward-logs/forward-your-logs-using-infrastructure-agent.mdx
```diff
@@ -137,6 +137,12 @@ To initiate your install, choose your deployment method:
   />
 </TechTileGrid>
 
+<Callout variant="important">
+Log forwarding capabilities for infrastructure agent installed from docker image is not supported. 
+If you want to forward logs for your docker container, use this [docker image](https://hub.docker.com/r/newrelic/newrelic-fluentbit-output) instead. 
+Alternatively, for log forwarding capabilities, deploy the agent on the host instead of deploying it directly on the docker container.
+</Callout>
```
ubanerjeeNR, line 144: The first sentence says that log forwarding is not supported for an infra agent that's installed using a docker image, whereas the second one tells user to install it from the given docker image.

Can you please check that?

Are we saying the following:

To forward docker container logs, install the infrastructure agent using this [docker image].... Or, deploy the agent on the host computer/server instead of deploying it directly on a docker container.

nbaenam: I think we should be a bit more flexible on this one. The ticket says that we don't support log forwarding installed from a Docker image, but I think we should avoid saying it like this. I think we should reword it like this:

```
To forward logs from a Docker container, you need to choose one of these options:
- Use this [Docker image](https://hub.docker.com/r/newrelic/newrelic-fluentbit-output). 
- Install an [on-host integration](/docs/infrastructure/host-integrations/get-started/introduction-host-integrations/)
```

I also saw we're using the old steps format on this page. Can you change it to these?
<Steps>
    <Step>
        
    </Step>
    <Step>
        
    </Step>
</Steps>

nbaenam (commented):
Hey @cbehera-newrelic. I added a comment. Let me know if you have questions ;-)

src/content/docs/logs/forward-logs/forward-your-logs-using-infrastructure-agent.mdx
nbaenam: I'd avoid the passive voice and write this: "`...base image to use with Kubernetes log integration...`"

nbaenam (approved):
Hi @cbehera-newrelic, I added a suggestion! Feel free to take it or leave it🙂

PR #18110, opened 2024-07-23
fix(infra-agent): Note about APM inventory data not being collected on the agent

josemore:
## Give us some context

* Adding a note based on customer feedback: inventory data for APM is not actually collected by the infrastructure agent but added in the backend once the same host is running both the infrastructure agent and APM agents.


nbaenam (approved):
Hi @josemore, thanks for adding this! It looks good to me!

PR #18138, opened 2024-07-24
fix(infra-agent): centOS stream 8 reached EOL

josemore:
## Give us some context

* Minor update: CentOS Stream 8 reached EOL so 9 or higher is recommended / officially supported.


rhetoric101 (approved):
@josemore thanks for updating this version.

PR #18140, opened 2024-07-24
fix(infra-agent): Explain status_server is enabled by default when using guided install (install recipe)

josemore:
## Give us some context

* Customer feedback requesting better explanation on why this setting is enabled by default in some scenarios.


rhetoric101 (approved):
@josemore thanks for adding this callout!

PR #18173, opened 2024-07-29
fix(browser): broken link

ally-sassman:
Fixed a broken browser link.


rhetoric101 (approved):
@ally-sassman thanks for updating this link and adding that redirect!

PR #18182, opened 2024-07-29
Java agent: Change category to "Related integrations"

rhetoric101:
The category "Third-party integrations" was too narrow since not all of these use the Java agent.


jeff-colucci (approved):
This is good, might see how "Other integrations" looks to your eyes as well but either one is fine.

PR #18202, opened 2024-07-30
fix(browser): add missing api pages to yml

ghost:
Two API pages are missing from the browser yml.

### DIFF
--- a/src/nav/browser.yml
+++ b/src/nav/browser.yml
@@ -124,6 +124,10 @@ pages:
         path: /docs/browser/new-relic-browser/browser-apis/log
       - title: noticeError
         path: /docs/browser/new-relic-browser/browser-apis/noticeerror
+      - title: pauseReplay
+        path: /docs/browser/new-relic-browser/browser-apis/pauseReplay/
+      - title: recordReplay
+        path: /docs/browser/new-relic-browser/browser-apis/recordReplay/
       - title: setApplicationVersion
         path: /docs/browser/new-relic-browser/browser-apis/setApplicationVersion
       - title: setCustomAttribute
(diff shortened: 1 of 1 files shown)


src/nav/browser.yml
```diff
@@ -124,6 +124,10 @@ pages:
         path: /docs/browser/new-relic-browser/browser-apis/log
       - title: noticeError
         path: /docs/browser/new-relic-browser/browser-apis/noticeerror
+      - title: pauseReplay
+        path: /docs/browser/new-relic-browser/browser-apis/pauseReplay/
```
rhetoric101, line 128: @ally-sassman I think the forward slash at the end of these two is preventing the navigation from keeping these in the left nav. When I click them in the left navigation pane, the correct file shows up, but the focus is lost in the left pane and it loads the generic list.

rhetoric101 (commented):
@ally-sassman I think you just need to remove some forward slashes to make this work in the UI.

src/nav/browser.yml
ally-sassman: @rhetoric101 I didn't think that the last `/` actually had an effect. It could be just a buggy build preview. But regardless, I'll add it anyway.

ally-sassman, line 128:
```suggestion
        path: /docs/browser/new-relic-browser/browser-apis/pauseReplay
```

src/nav/browser.yml
```diff
@@ -124,6 +124,10 @@ pages:
...
+      - title: recordReplay
+        path: /docs/browser/new-relic-browser/browser-apis/recordReplay/
```
ally-sassman, line 130:
```suggestion
        path: /docs/browser/new-relic-browser/browser-apis/recordReplay
```

rhetoric101 (approved):
@ally-sassman these docs are now working as expected! Thanks for fixing these URLs!

PR #18236, opened 2024-08-01
fix: resize images to ensure nothing is over 1920px

clarkmcadoo:
Resizes images to be less than 1920px


tabathadelane (approved):
🪄 BE SMOL!

I love that Vale decided to wake up today. I can't get the diff to load the so called comments, so gonna `dangerouslyAssume` it's not important.

sunnyzanchi (approved):

PR #18257, opened 2024-08-02
fix(licenses): remove outdated tile

ghost:
The old URL has been deprecated and redirects to this page, so the tile must be removed.


jeff-colucci (approved):
Looks removed to me!

PR #18266, opened 2024-08-05
chore(PHP): Corrects Dockerfile example

mfulb:
PHP team member here.

These instructions were recently modified and we have determined that this change needs to be made to be sure the instructions work on the Alpine Linux platform, which does not use GNU `grep` and requires using a different form of the `--only-match` option .  The change is compatible with GNU `grep` systems as well as these will accept the `-o` short form of the option.

Additionally, the `newrelic.daemon.collector_host` option was included in the original documentation.  After review we have determined this option should not be changed generally and so it was removed from the example.


jeff-colucci (approved):
Looks good to me!

PR #18271, opened 2024-08-05
feat: change page_action refs to generic_events as per new browser agent version change

metal-messiah:
The browser agent is updating to use the entry field `generic_events` instead of `page_action`.  This PR updates the refs that point at those entry points.  The `page_action` entry will continue to work in the agent for the foreseeable future, but the preferred method of implementation going forward should be `generic_events`.

Please reach out if we need to add more by way of callouts or other disclaimers.


jeff-colucci (approved):
Looks good!

PR #18278, opened 2024-08-05
feat(Docs): Added some guidance about whats new github action

paperclypse:


cbehera-newrelic (approved):
The content looks good to me, just curious, what is to be done if the ID generation fails or doesn't complete?

PR #18280, opened 2024-08-06
fix indentation and numbering

keegoid-nr:
- It might make sense to remove some of the itemized lists for clarity.
- fix for log lines to check in SJM
cc @jeff-colucci

### DIFF
--- a/src/content/docs/synthetics/synthetic-monitoring/private-locations/job-manager-configuration.mdx
+++ b/src/content/docs/synthetics/synthetic-monitoring/private-locations/job-manager-configuration.mdx
@@ -9,7 +9,7 @@ redirects:
 freshnessValidatedDate: 2024-07-29
 ---
 
-This doc will guide you through configuring your [synthetics job manager](/docs/synthetics/synthetic-monitoring/private-locations/install-job-manager) by showing you how to: 
+This doc will guide you through configuring your [synthetics job manager](/docs/synthetics/synthetic-monitoring/private-locations/install-job-manager) by showing you how to:
 
 * Use [environment variables](#environment-variables) to configure your synthetics job manager.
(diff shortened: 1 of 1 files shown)


src/content/docs/synthetics/synthetic-monitoring/private-locations/job-manager-configuration.mdx
```diff
@@ -1099,35 +1102,36 @@ To set permanent data storage on Docker:
...
-Example:
 
-```
-		docker run ... -v /sjm-volume:/var/lib/newrelic/synthetics:rw ...
-```
+    Example:
+
+    ```sh
+    docker run ... -v /sjm-volume:/var/lib/newrelic/synthetics:rw ...
+    ```
```
clarkmcadoo, line 1111:
```suggestion
Example:

```sh
  docker run ... -v /sjm-volume:/var/lib/newrelic/synthetics:rw ...
```
```

keegoid-nr: Granted, I have not previewed the change in VS Code (I still need to learn how to do that), but the goal here is to nest the Example under the itemized list. This becomes especially obvious when not nesting breaks the counter of the itemized list. Example:
image

jeff-colucci (approved):

keegoid-nr:
@jeff-colucci This should be one code block. The indentation is breaking the code block.
image

keegoid-nr:
If we're going to use itemized lists, it should start with 1 at the first step.
image

jeff-colucci:
@keegoid-nr Ah ok, let me take a crack at fixing this in a PR myself, I'll have it forwarded to you today or tomorrow, but it will be done this week.

PR #18297, opened 2024-08-07
chore(style guide): Delete docs library service info

austin-schaefer:
We used to use Swiftype to automatically surface relevant docs in the UI based on your location, but we decomissioned that capability over a year ago.


jeff-colucci (approved):
Made one small style fix

PR #18310, opened 2024-08-08
chore(PHP): Adds note for API call

mfulb:
I am a PHP team engineer.

This update is based on customer feedback.  The INI option `newrelic.error_collector.prioritze_api_errors` is intended to cause errors reported through the `newrelic_notice_error()` API to take precedence over other error types which occur.  However, if an uncaught error error also occurs it will have the highest priority.  This can be confusing if notice'd errors are not showing up and this setting is set to `true`.

I made it a "NOTE:" but perhaps a different formatting (callout?) would be better and I leave this to your discretion.


jeff-colucci (approved):
Looks great, just a small style change

PR #17314, opened 2024-05-15
fix: config.yaml indentation

crshanks:
The indentation appears incorrect in the docs page. The `cluster_name`, `transformation`, and `targets` keys all need to be at the same indentation to avoid errors. I'm unsure why the existing code block didn't preserve the indentation as the code block immediately before, using the same spacing etc., preserves it correctly.

The existing docs page displays the config.yaml example markdown code block with two spaces before the `transformation`, and `target` keys, i.e.:

```yaml
cluster_name: YOUR-CLUSTER-NAME

  transformations: # The attribute `mongodb_cluster_name` to all the MongoDB metrics is needed by the MongoDB dashboard.
    - {"add_attributes":[{"prefix":"mongodb_","attributes":{"mongodb_cluster_name": "YOUR-CLUSTER-NAME" }}]}

  targets:
    - urls:
      - http://exporter:9216
```

It should appear, for copying, as:
```yaml
cluster_name: YOUR-CLUSTER-NAME

transformations: # The attribute `mongodb_cluster_name` to all the MongoDB metrics is needed by the MongoDB dashboard.
  - {"add_attributes":[{"prefix":"mongodb_","attributes":{"mongodb_cluster_name": "YOUR-CLUSTER-NAME" }}]}

targets:
  - urls:
    - http://exporter:9216
```

PR #18336, opened 2024-08-12
fix(OHI Mongodb): Deleted files that were not deleted in the revamp p…

nbaenam:
Because of this PR: https://github.com/newrelic/docs-website/pull/17314, we realized that the former MongoDB docs were only deleted from the nav file.


jeff-colucci (approved):
Looks perfect to me!

PR #18338, opened 2024-08-12
fix: correct node-browser-runtime release notes version

kondracek-nr:
## Give us some context

Correct node-browser-runtime version in latest release notes


paperclypse (approved):
I assume this is the correct version number :)

PR #18366, opened 2024-08-13
Update manage-data-retention.mdx

nr-ksteinbach:
I'm the PM of the team that owns these downsamplers in our NR backend, and this change will fix the incorrect description that previously indicated that all infrastructure metrics "(including metrics from on-host integrations)" (sic) would be aggregated.


nbaenam (approved):
Hi @nr-ksteinbach, thank you for the update! I've added a minor TW edit

PR #18369, opened 2024-08-14
Instrumented view based on time picker

keegoid-nr:
- updating wording to account for new UI behavior


nbaenam (approved):
Hi @keegoid-nr, thank you for the update! I've added a minor TW edit

PR #18371, opened 2024-08-14
fix(Infra): reversing changes based on SME feedback

jeff-colucci:


rhetoric101 (approved):
@jeff-colucci it looks like you just needed to remove a `chmod` command. This looks good to me.

PR #18390, opened 2024-08-15
feat: update pageaction apis that mention limit

metal-messiah:
This PR adjusts the page action docs that mention a 120 event limit, which was changed in the latest browser agent release.


nbaenam (approved):
Hi @metal-messiah, thank you for the update!! It looks good to me!

PR #18394, opened 2024-08-15
docs: Correct wording on MacOS agent update steps

d3caf:
Noticed that the "update" guide for MacOS incorrectly said `Uninstall the agent` as one of the steps, when in fact the step was to upgrade the agent.


nbaenam (approved):
HI @d3caf, thank you for the update!! It looks good to me!

PR #18398, opened 2024-08-15
Update whats-new-template.mdx

aalfano7:
Updated path for adding images, I also recommend removing the screenshot as step 1 is actually to select the static folder, and image shows src folder being selected.


nbaenam (approved):
Hi @aalfano7, thank you for the update! I've added some TW edits

PR #18420, opened 2024-08-19
add step monitors to VSE doc

keegoid-nr:
- add step monitors
- update links


cbehera-newrelic (approved):
The updated links are fine.

PR #18422, opened 2024-08-19
fix code block

keegoid-nr:
- attempting to fix the code block and indentation once more


cbehera-newrelic (approved):
Done an editorial pass. LGTM

PR #18427, opened 2024-08-20
fix(infra-agent): Update supported OS

josemore:
## Give us some context

* Linux RedHat 7 still has extended support..
* Adding links to Docker (Container) and Kubernetes requirements to avoid data to be out of sync.


cbehera-newrelic (approved):
LGTM

PR #18432, opened 2024-08-20
Fix: Documentation about using environment variables to obfuscate sensitive values

lmbaschiera:

## Give us some context

Adds missing whitespaces between the curly brackets and the environment variable name. Without them, these are interpreted literally.
In the tests that fixed an issue with variable passthrough, they do have whitespaces https://github.com/newrelic/nri-kubernetes/blob/main/charts/newrelic-infrastructure/tests/configmap_integrations_test.yaml#L231

### DIFF
--- a/src/content/docs/infrastructure/host-integrations/installation/secrets-management.mdx
+++ b/src/content/docs/infrastructure/host-integrations/installation/secrets-management.mdx
@@ -63,7 +63,7 @@ integrations:
 
 ## Using environment variables [#using-variables]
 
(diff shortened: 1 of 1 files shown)


rhetoric101:
@lmbaschiera I spoke with one of our devlopers and they said they have tests that confirm this syntax should work with or without spaces inside the curly braces. Have you spoken with our support team about this?

lmbaschiera:
Hi @rhetoric101 - I have not reached out to the support, after some trial and error we figured it was because of the spaces when it worked after adding them. I don't think it was something else.

PR #18436, opened 2024-08-20
fix: broken in page link

brnhensley:
There is an in-page link rendering as plain text:

image

I didn't see a syntax issue causing this, but noticed it was the only in-page link in this doc that was broken and the only one on a `###` header, the rest were on `##` headers. So I'm trying to see if this fixes it.

 I also formatting some code blocks that I find tough to read


cbehera-newrelic (dismissed):
LGTM

cbehera-newrelic (approved):
LGTM

PR #18446, opened 2024-08-21
NR-304043: The support of browser and mobile logs brings a new set of reserved attributes (`instrumentation.*`)

luckslovez:
The support of browser and mobile logs brings a new set of reserved attributes (`instrumentation.*`)


cbehera-newrelic (approved):
Fixed the wrapping issue in the table. The  content looks good to me.

PR #18449, opened 2024-08-21
Update use-nerdgraph-manage-license-keys-user-keys.mdx

wildcardlinux:
The INGEST_KEY_ID needs to be in double quotes to work correctly. Confirmed this syntax myself today.

## Give us some context

Fixed the syntax for using Nerdgraphql to decode ingest license keys. The Key Id needs to be in double quotes to have valid syntax and for the API to respond.


cbehera-newrelic (approved):
LGTM

PR #18475, opened 2024-08-23
No new legacy monitors

bpeckNR:
- Updated all EOL callouts to make it obvious that no new legacy runtime monitors can be created as of Monday
- Added an example NerdGraph mutation that allows you to downgrade the runtime of an existing monitor

It will still be possible to downgrade monitors to the legacy runtime AND edit legacy runtime monitors until the EOL on October 22, 2024.

While we will roll out the no new legacy monitors restriction on Monday, it should be safe to merge this in at any time. It does not have to wait until Monday.


cbehera-newrelic (dismissed):
LGTM

PR #18501, opened 2024-08-27
fix: correct addToTrace documentation

patrickhousley:

## Give us some context

Fixing the `addToTrace` browser agent API documentation to remove reference to the `type` field. This field is not used in the agent. The documentation is misleading to customers that may think they can define a value for the `type` field. In reality the agent always set's the `type` field to a value of `api`.

https://new-relic.atlassian.net/browse/NR-302664


rhetoric101 (approved):
@patrickhousley this change looks good to me!

PR #18507, opened 2024-08-28
Updates to log partition guidance

JimHagan:
## This PR makes the following changes

* remove 1TB/day guidance
* focus on scanned events as limited factor for query performance
* explain the drivers of scanned events in NRQL queries and charts
* more nuanced guidance on how many partitions an org/account should have


JimHagan:
@rhetoric101 I found a couple of things that need to be changed.  

This query is not correct:

```
 FROM count(*) 
 FROM <partition_name> 
 SINCE <time window>
 ```
 
 It should be `SELECT count(*)` instead of `FROM count(*)`
 
 
 Also the title of the tip section current says `Finding events in your partitions` this doesn't really capture the essence of the tips which are `Finding the number of events in you partition` or `Analyzing the size of you partition`
 
 I don't have a strong preference, but I don't think `Finding...` captures the main focus of the tip.

rhetoric101 (approved):
@JimHagan thanks for adding this new information and fixing some errors.

PR #18519, opened 2024-08-29
Update log-event-data.mdx

luckslovez:
Browser and mobile integrations are using `instrumentation.{name,provider,version}` for their own purposes. Customers using this integration should not use them or expect them to be stored with logs.

NR-304043


danybmx (approved):
Nice!

rhetoric101:
To improve the table readability, we added some span tags with a class that prevents text wrapping (for the first column).

rhetoric101 (approved):
@luckslovez thanks for adding this warning!

PR #18537, opened 2024-08-30
add deprecation note

metal-messiah:
This PR adds a deprecation warning to the docs tied to browser's createTracer method.

### DIFF
--- a/src/content/docs/browser/new-relic-browser/browser-apis/createtracer.mdx
+++ b/src/content/docs/browser/new-relic-browser/browser-apis/createtracer.mdx
@@ -16,6 +16,10 @@ redirects:
 freshnessValidatedDate: never
 ---
 
+<Callout variant="important">
+  The createTracer browser API has been deprecated. The recommended way to trace the duration of a task is to capture a performance [mark](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark) and/or [measure](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure).  Future browser agent versions will capture marks and measures automatically, at which point support for createTracer will cease.   
+</Callout>
+
 ## Syntax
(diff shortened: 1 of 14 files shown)


src/content/docs/browser/new-relic-browser/browser-apis/createtracer.mdx
```diff
@@ -16,6 +16,10 @@ redirects:
 freshnessValidatedDate: never
 ---
 
+<Callout variant="important">
+  The createTracer browser API has been deprecated. The recommended way to trace the duration of a task is to capture a performance [mark](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark) and/or [measure](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure).  Future browser agent versions will capture marks and measures automatically, at which point support for createTracer will cease.
```
rhetoric101, line 20: @metal-messiah is `createTracer` an API itself or is it just a method in the SPA API? From perusing the browser docs, it seems like the terms "API" and "method" are almost used interchangeably.  I'm probably mistaken here, but this was what I was thinking:

The `createTracer` method in the SPA API has been deprecated. The recommended way to trace the duration of a task is to capture a performance [mark](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark) and/or [measure](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure).  Future browser agent versions will capture marks and measures automatically, at which point support for `createTracer` will cease.

rhetoric101 (commented):
@metal-messiah I just had one question, and I am almost certainly mistaken about it. 

I should mention that for non-English topics, you can skip those updates since we have a translation job that will add your changes. So, the next time you do this, you won't have to work so hard!

PR #18552, opened 2024-09-03
fix(.NET): add back migration guides

ghost:
A SME requested these docs to be added back (they were deleted two months ago).


tippmar-nr (changes requested):
Thanks, @ally-sassman  -- 

Unfortunately, these files were added to the go-agent; they should be added to the .NET agent. 

Additionally, there's a missing link that needs to be restored in https://docs.newrelic.com/docs/release-notes/agent-release-notes/net-release-notes/net-agent-10-0-0/#deprecationsremoved-features at "see the migration guide for details".

tippmar-nr (approved):
Thanks for taking care of this!

rhetoric101 (approved):
@ally-sassman I can see the two restored migration guide under dot-net, as well as the change to the 10.0.0 release notes.

PR #18577, opened 2024-09-05
fix(Style Guide): Fixed error in the navigation file

nbaenam:
Fixed the navigation file for the style guide. Removed the "/" at the end of a path. This is the file: [Frontmatter and page templates](https://docs.newrelic.com/docs/style-guide/writing-docs/processes-procedures/use-content-types-text-formats/).


rhetoric101 (approved):
@nbaenam thanks for fixing this path!

PR #18608, opened 2024-09-09
feat(SyntheticsFirefox): Added proxy support

bpeckNR:
Proxy support is available for Firefox monitors using Synthetics Private Locations. This requires the customer to use version 3.0.7 of the synthetics-node-browser-runtime container image or newer.


jeff-colucci (approved):
Looks good!

PR #18621, opened 2024-09-10
Accounts: Change max duration

rhetoric101:
Per SME, we needed to streamline the max duration description


paperclypse (dismissed):
I assume they know what they're talking about :)

paperclypse (approved):
Looks good

PR #18632, opened 2024-09-10
fix(Synthetics): Minor style guide edit

paperclypse:
@jeff-colucci Based on style guide, we use "higher or lower" and not "newer or older" when referring to versions.


jeff-colucci (approved):

jeff-colucci:
Thanks for catching that, I'll be sure to keep an eye out for that in the future!

PR #18675, opened 2024-09-16
chore(serverless): Renamed non-containerized with layered function

cbehera-newrelic:
* Renamed non-containerized function with layered function in the heading.
* Ticket: NR-305924


cbehera-newrelic:
The changes are already made in other PR and the updates are already available in the documentation . So closing this PR.

PR #18676, opened 2024-09-16
Update to aws-logs-ingestion Lambda function

AkhilReddyS-NR:
We have updated the default value of the NRLoggingEnabled parameter for the New Relic Lambda function from false to true. This change has been released and is now live. The release notes have been updated to reflect this change.


nbaenam (approved):
Hi @AkhilReddyS-NR, thank you for the update! I've added some TW edits

PR #18700, opened 2024-09-18
NR-307459 Added a section about watchOS compatibility

mbruin-NR:
Added a small section to the iOS agent compatibility page linking to the watchOS doc. https://new-relic.atlassian.net/browse/NR-307459


nbaenam (approved):
Hey @mbruin-NR, thank you for the update!! It looks good to me!

PR #18770, opened 2024-09-25
feat(ohi): ohi i18n puzzle

akristen:
This PR makes changes to the repo to avoid version drift:

- Legacy docs that have been adapted into OHI branching docs still exist in the repo
- These docs are not in the left nav, but data shows that users are still accessing them and in some cases adding updated information to these legacy docs
- This decision was likely made to preserve the i18n information, as branching docs cannot be translated
- However, this introduces two problems: version drift; leaving older, out of date docs that could lead non-English users into murky waters
- After speaking with Eng, I decided it is best that the legacy docs redirect to install and for any i18n users to translate in-browser. In-browser translation is the lesser evil against out-dated docs and version drift


akristen:
> > [!NOTE]
> 
> > Our Vale automated writing feedback tool has suggestions for this PR. You can view these suggestions in the _Files changed_ tab. If you're not a New Relic docs writer, we'll review any comments and you don't need to take any action (though feel free to make edits if you like!).

Error reads like this: 406 Sorry, the diff exceeded the maximum number of lines (20000) [{Resource:PullRequest Field:diff Code:too_large Message:}]

PR is just too big so vale failed

paperclypse (approved):
Redirects looks good

PR #18786, opened 2024-09-27
fix: new link name from ISpan.AddCustomAttribute()

brnhensley:
There were two identical in-page link ids for `"AddCustomAttribute"`. These are linked a lot in the docs so I only changed this one so I don't have to update docs everywhere. With both having the same ID this in page link doesn't work.

## Give us some context


rhetoric101 (approved):
@brnhensley thanks for pointing out this issue. I removed the links to the nested collapsers since they don't work for ITransaction and ISpan.

PR #18787, opened 2024-09-27
chore(styleGuide): add OpenTelemetry to product capitalization

ghost:
Clarified that OpenTelemetry Collector should always be upper cased. Also fixed some table entries that were incorrectly lowercased.


nbaenam (approved):
Thanks for improving this doc! LGTM!

PR #18788, opened 2024-09-27
Add `NewRelic.Labels` to the list of .NET agent configuration options that can be set in app settings

nr-ahemsath:
This has been an oversight in our docs for a while.  The .NET agent has a limited set of configuration options that can be set via standard Microsoft application configuration files (e.g. `appsetttings.json`).  One of the things that can be set is the `labels` setting, with an appsettings key of `NewRelic.Labels`.  This PR updates the docs to show that labels can be configured this way.


rhetoric101 (approved):
@nr-ahemsath thanks for adding these!

PR #18789, opened 2024-09-27
fix(on-host): Replace cp with copy for Windows instructions

rhetoric101:
For NR-313160.

In Windows instructions, replace any occurrences of the `cp` command with `copy`.

Fix one instance where the Windows restart commands use Linux commands.


jeff-colucci (approved):
Looks good to my eyes

paperclypse (approved):
Read through the full MDX files in the diff view in GitHub. Looks like you caught all the Windows commands. Thanks for fixing this!

PR #18796, opened 2024-09-30
fix(docs): update integrations & agents reference

ghost:
Found some references of an incorrect + after Integrations & Agents.


rhetoric101 (approved):
@ally-sassman thanks for catching these!

PR #18822, opened 2024-10-02
Add search results page

LizBaker:
You can test this out by going to this path `/search-results/?query=node%20agent&page=1`

### DIFF
--- a/src/pages/search-results.js
+++ b/src/pages/search-results.js
@@ -18,10 +18,10 @@ import { usePagination, DOTS } from '../hooks/usePagination';
 const SearchResultPageView = ({ pageContext }) => {
   const { queryParams } = useQueryParams();
   const query = queryParams.get('query');
-  const page = Number(queryParams.get('page'));
+  const page = Number(queryParams.get('page') ?? 1);
   const locale = useLocale();
-  const [results, setResults] = useState({});
-  const { records, pageCount } = results;
+  const [results, setResults] = useState({ loading: true });
+  const { records, pageCount, loading, error } = results;
   const { slug } = pageContext;
 
   const totalPages = pageCount;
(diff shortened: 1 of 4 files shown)


src/pages/search-results.js
```diff
@@ -0,0 +1,272 @@
...
+    (async () => {
+      const defaultSources = locale.isDefault
+        ? ['developer', 'docs', 'opensource', 'quickstarts']
+        : [
+            `developer-${locale.locale}`,
+            `docs-${locale.locale}`,
+            `opensource-${locale.locale}`,
+            `quickstarts`,
+          ];
+      const results = await search({
```
sunnyzanchi, line 50: we should handle the error case here. probably just an error message that says something like "unable to get search results, an error has occurred" is fine

src/pages/search-results.js
```diff
@@ -0,0 +1,272 @@
...
+          font-weight: 500;
+        `}
+        dangerouslySetInnerHTML={{ __html: result.highlight.title }}
+      />
+      <p dangerouslySetInnerHTML={{ __html: result.highlight.body }} />
+    </Surface>
+  );
+};
+
+SearchResultPageView.propTypes = {
```
sunnyzanchi, line 256: these proptypes don't match the component's implementation

src/pages/search-results.js
```diff
@@ -0,0 +1,272 @@
...
+  Surface,
+  useLocale,
+  useQueryParams,
+} from '@newrelic/gatsby-theme-newrelic';
+
+import { usePagination, DOTS } from '../hooks/usePagination';
+
+const SearchResultPageView = ({ pageContext }) => {
+  const { queryParams } = useQueryParams();
+  const query = queryParams.get('query');
```
sunnyzanchi, line 20: we should handle the case where `query` is `undefined` or an empty string. i think navigating to the home page would be fine

src/pages/search-results.js
```diff
@@ -0,0 +1,272 @@
...
+      to={result.url}
+      css={css`
+        em {
+          color: #00ac69;
+          font-style: normal;
+        }
+        margin-bottom: 1rem;
+        box-shadow: none;
+        color: var(--primary-font-color);
+        &:hover {
```
sunnyzanchi, line 230: after trying it out, i don't think hovering over a result that it's obvious enough you can click the entire result. what do you think about adding a `&:hover h3 { text-decoration: underline; }` here?

src/pages/search-results.js
sunnyzanchi, line 1: the designs have a uniform left and right padding around the whole page. but bc of the media query in MainLayout, the right padding gets smaller at larger screen sizes. i didn't think it was that big of a deal but i noticed that this also means the pagination is off center so i think we should try and fix that

sunnyzanchi (changes requested):
left a few comments but looks great overall! nice job!

src/pages/search-results.js
LizBaker: good point!

src/pages/search-results.js
LizBaker: poifect!

sunnyzanchi (approved):

PR #18838, opened 2024-10-03
docs(infra-agent): removing misleading note for Win 2012

josemore:
## Give us some context
* Remove a misleading note for Windows 2012
* Compatibility explanation for older versions (not officially supported past EOL but agent might still be compatible)


rhetoric101 (approved):
@josemore thanks for adding this clarification! I just made some minor edits.

PR #18839, opened 2024-10-03
feat(K8s): Added important note about billable events

nbaenam:
Added an important note about billable events on the [Query Kubernetes data](https://docs.newrelic.com/docs/kubernetes-pixie/kubernetes-integration/understand-use-data/find-use-your-kubernetes-data/#view-data) section.

Ticket: NR-317104

### DIFF
--- a/src/content/docs/kubernetes-pixie/kubernetes-integration/understand-use-data/find-use-your-kubernetes-data.mdx
+++ b/src/content/docs/kubernetes-pixie/kubernetes-integration/understand-use-data/find-use-your-kubernetes-data.mdx
@@ -325,7 +325,7 @@ Kubernetes data is attached to these [events](/docs/using-new-relic/data/underst
 
 
 <Callout variant="important">
-  `InternalK8sCompositeSample` is an event that New Relic generates and it's quite critical for some of our interfaces. While the other events listed out in the table are billable, `InternalK8sCompositeSample` is a non-billable event. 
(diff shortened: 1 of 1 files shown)


src/content/docs/kubernetes-pixie/kubernetes-integration/understand-use-data/find-use-your-kubernetes-data.mdx
```diff
@@ -323,6 +323,12 @@ Kubernetes data is attached to these [events](/docs/using-new-relic/data/underst
   </tbody>
 </table>
 
+
+<Callout variant="important">
+  `InternalK8sCompositeSample` is an event that New Relic generates and it's quite critical for some of our interfaces. While the other events listed out in the table are billable, `InternalK8sCompositeSample` is a non-billable event.
```
akristen, line 328: Is it important we describe why it's critical? Do we not bill because it's a critical event? What do you think about adding something like "...is a critical event that New Relic generates when [x occurs in y environment (OR) when x occurs because of y]. This event is the only non-billable event in the table."

akristen (dismissed):
Approving because it's billing related, but I do think we should add just a smidge more context! (See comment)

src/content/docs/kubernetes-pixie/kubernetes-integration/understand-use-data/find-use-your-kubernetes-data.mdx
nbaenam: Good catch! I've changed the entire callout.

akristen (approved):

PR #18873, opened 2024-10-06
Update infrastructure-agent-configuration-settings.mdx

nr-ksteinbach:
fix: explaining custom_attributes of the Infrastructure Agent better, along with the tip to prefix them with label.* so that these tags do not interact with Azure & GCP cloud monitoring

(Details are in NR-293494, NR-305957 and the linked page.)

### DIFF
--- a/src/content/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings.mdx
+++ b/src/content/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings.mdx
@@ -3236,7 +3236,7 @@ Metrics can also be enriched with extended cloud metadata (including custom reso
       </tbody>
     </table>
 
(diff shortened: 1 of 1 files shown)


src/content/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings.mdx
```diff
@@ -3236,27 +3236,27 @@ Metrics can also be enriched with extended cloud metadata (including custom reso
       </tbody>
     </table>
 
-    Use a list of custom attributes to annotate the data from this agent instance. Separate keys and values with colons `:`, as in `KEY: VALUE`, and separate each key-value pair with a line break. Keys can be any valid YAML except slashes `/`. Values can be any YAML string, including spaces.
+    Use a list of custom attributes to annotate the data from this agent instance. Separate keys and values with colons `:`, as in `KEY: VALUE`, and separate each key-value pair with a line break. Keys can be any valid YAML except slashes `/`. Values can be any YAML string, including spaces. Starting the key with `label.` will make sure it remains as is, even if that host might run in different cloud environments, where the cloud tags or cloud labels might interact with it in the combined decoration of Tags. In contrast, custom_labels migth be prefixed with `gcp.` or `azure.` when running in those environments and when combining data from the infrastructure agent with cloud monitoring together.
```
akristen, line 3239:
```suggestion
    Use a list of custom attributes to annotate the data from this agent instance. Separate keys and values with colons `:`, as in `KEY: VALUE`, and separate each key-value pair with a line break. Keys can be any valid YAML except slashes `/`. Values can be any YAML string, including spaces. Starting the key with `label.` will make sure it remains as is, even if that host might run in different cloud environments, where the cloud tags or cloud labels might interact with it in the combined decoration of tags. In contrast, `custom_labels` might be prefixed with `gcp.` or `azure.` when running in those environments and when combining data from the infrastructure agent with cloud monitoring together.
```

akristen (approved):

PR #18936, opened 2024-10-11
chore(Python): removing old python docs (288979)

jeff-colucci:
Removing old Python docs after confirming with the Python team that Stackato and WebFaction are no longer supported


paperclypse (approved):
I assume you have a good reason for this, but would you please add a note to the PR description that gives a little context? Thanks!

paperclypse (approved):

PR #18999, opened 2024-10-17
Adding RecordDatastoreSegment to .NET Agent API

chynesNR:
PR where the feature was added: https://github.com/newrelic/newrelic-dotnet-agent/pull/2320

### DIFF
--- a/src/content/docs/apm/agents/net-agent/net-agent-api/net-agent-api.mdx
+++ b/src/content/docs/apm/agents/net-agent/net-agent-api/net-agent-api.mdx
@@ -580,6 +580,16 @@ The following list contains the different calls you can make with the API, inclu
             Associates a user ID to the current transaction (see below for more details).
           </td>
         </tr>
+
+        <tr>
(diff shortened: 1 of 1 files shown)


chynesNR:
The page that was updated: https://deploy-preview-18999--docs-website-netlify.netlify.app/docs/apm/agents/net-agent/net-agent-api/net-agent-api/

jeff-colucci:
@chynesNR I see there's some "string" entries that have question marks after them, is this intended or are we waiting for confirmation that string is correct before we merge it? Just let me know how you'd like it handled and I'll get it taken care of!

chynesNR:
> @chynesNR I see there's some "string" entries that have question marks after them, is this intended or are we waiting for confirmation that string is correct before we merge it? Just let me know how you'd like it handled and I'll get it taken care of!

It's intentional. That does look a little funny if you aren't familiar with C# syntax. It means a "nullable string", i.e. it can be either a string value or null.

jeff-colucci (approved):

PR #19014, opened 2024-10-18
remove confusing link to self

keegoid-nr:
- remove link to the page the user is already on


jeff-colucci:
@keegoid-nr Should this go out at a later time or is it good to publish now?

jeff-colucci (approved):

PR #16030, opened 2024-02-01
Update php-agent-compatibility-requirements.mdx for Yii2

razvanphp:
This is related to and should be merged after this MR: https://github.com/newrelic/newrelic-php-agent/pull/823

PR #19018, opened 2024-10-20
Update php-agent-compatibility-requirements.mdx for Yii 1 & 2

razvanphp:
Yii 1.x support is still there, so for completion and given that this version is still widely used, I've udpated the docs to reflect that.

See: https://github.com/newrelic/newrelic-php-agent/issues/821

Related to #16030

### DIFF
--- a/src/content/docs/apm/agents/php-agent/getting-started/php-agent-compatibility-requirements.mdx
+++ b/src/content/docs/apm/agents/php-agent/getting-started/php-agent-compatibility-requirements.mdx
@@ -549,7 +549,7 @@ The following frameworks are supported:
       <td>
         Yii
       </td>
-      <td>2.0</td>
+      <td>1.x, 2.x</td>
       <td></td>
     </tr>
 


nbaenam:
Added a message in the php-agent channel to verify this info.

nbaenam:
Hi @razvanphp, the page is already updated and our engineers think this change is not needed. Thanks for chiming in!

razvanphp:
Can you please reconsider this?

If you look at the implementation PR for Yii2 (https://github.com/newrelic/newrelic-php-agent/pull/848), you can see yourself it is a separate implementation for Yii v1 and v2, as they are not compatible.

As a daily user of this framework, I know the community and I can tell you it's confusing if you specify just one on the documentation. Also, all the other frameworks have all historical versions specified in the documentation. Why have it different just for this one? Before this change, the documentation was specifying only 1.x, which still holds true today.

I don't see why NewRelic will not want to keep an up-to-date and accurate documentation....

nbaenam:
Hi @razvanphp, I've consulted with the eng team again. If they think this must be updated, I'll open a new PR and I'll let you know. Thank you for chiming in!

PR #19031, opened 2024-10-22
NR-307785 Update docs with latest bug fix if log message is empty

nr-raviraj:

## Give us some context
This PR fixes the issues when the log message is empty, customers were seeing our decorator values i.e NR-LINKING string which would be fixed by this PR and if the log message is empty they will see only empty message.
Related Jira ticket:- https://new-relic.atlassian.net/browse/NR-307785


nbaenam (approved):
Hi @nr-raviraj, thank you for the update! I've added some TW edits

PR #19024, opened 2024-10-21
chore: Add missing "to"

kaylareopelle:
## Give us some context

While reading the OSS Litmos training, I noticed a missing word in one of the screenshots. The related sentences are updated in this PR.

PR #19032, opened 2024-10-22
Revert "chore: Add missing "to""

nbaenam:
Reverts newrelic/docs-website#19024


akristen (approved):
I think grammatically we need the 'to' here but I'm not sure if there's some other context I'm missing so am approving + merging!

PR #19056, opened 2024-10-23

bhecknr:
Added New Relic Retail Solution to "When to use title case" section. This change is per VP, Product Marketing.


nbaenam (approved):
Hi @bhecknr, thanks for adding this! It looks good to me!

PR #19057, opened 2024-10-23
fix (NRQL): Adjust consumption queries to include SINCE clause

rhetoric101:
This came up as a suggestion from an SME.


paperclypse (approved):
Seems like a good idea. Thanks!

shuda-newrelic-com (approved):

PR #19062, opened 2024-10-24
Update flex-integration-tool-build-your-own-integration.mdx

nr-ksteinbach:
fix: adding nri-flex as a tag, hoping that this will make it easier to find this page when looking for this component with its full name


nbaenam (approved):
Hi @nr-ksteinbach, thank you for adding this! LGTM!

PR #19098, opened 2024-10-28
fix(K8s): Fixed capitalization and replaced "agents operator" with "A…

nbaenam:
Fixed capitalization and replaced "agents operator" with "APM auto-attach"


paperclypse (approved):
Thanks!

PR #19163, opened 2024-11-01
fix(aws): update incorrect code snippet

ghost:
This came from a hero request, who noticed that the CLI code snippet incorrectly referenced the lambda layers instead of cli. This PR fixes the code snippet.


paperclypse (approved):
Looks good, thanks!

PR #19185, opened 2024-11-04
fix(ruby): revert changes from #19152

ghost:
Reverts a phrasing change from #19152. This structure should match the rest of the collapsers (using incomplete sentences)


nbaenam (approved):
LGTM!

PR #19200, opened 2024-11-05
fix(accounts): Migrate archived pricing to public site

rhetoric101:
This PR is to insert archived pricing and product descriptions for the old Add On compute option.


paperclypse (approved):
I have to assume that all the info is technically correct. Looks good.

PR #19214, opened 2024-11-06
docs: Clean up the manual distributed tracing section of the .NET agent's main distributed tracing doc

nr-ahemsath:
1. Link to the correct APIs in the API guide doc
2. Add link to buildable/runnable example in our agent's examples repo
3. Add callout regarding the need to find a "carrier" for key-value pairs in whatever message object the customer is working with


nr-ahemsath:
I addressed some of the Vale linter comments, but I don't agree with the remaining two.  Whoever reviews this, please let me know if you have any questions.

paperclypse (approved):
@nr-ahemsath Good enough for something that needs to get out today. I think it'd be good to have a tech writer do a pass on this next week, though. Thanks!

PR #19255, opened 2024-11-10
Update install-nodejs-agent.mdx to add guide for pm2 deployment

tsq:
I think there are many Node.js apps deployed using PM2, so I added a guide for it.
 I used this command, and it worked for me.


rhetoric101:
@tsq thanks for your thoughtful suggestion! In this case, our developers prefer not to have multiple examples of starting Node.js with the agent. So, we're not going to add this PM2 reference.

PR #19265, opened 2024-11-11
add bullet point for advanced example

keegoid-nr:
Detail how/why we can use top level `await`.


rhetoric101 (approved):
@keegoid-nr thanks for adding this information! I just added a minor formatting change.

PR #19283, opened 2024-11-12
chore: Announce support for .NET 9 in .NET Agent compatibility document

nr-ahemsath:
.NET 9 just went GA: https://devblogs.microsoft.com/dotnet/announcing-dotnet-9/

This PR:

* Adds .NET 9 to the list of runtimes the .NET agent is compatible with
* Adds ASP.NET Core 9 to the list of web frameworks the .NET agent instruments
* Attempts to clarify our support stance on out-of-support versions of .NET.  The .NET agent actively refuses to profile any .NET application targeting a .NET Core runtime older than .NET Core 3.1.  For .NET Core 3.1 and newer, the agent will likely work.  However, Microsoft no longer supports .NET versions older than .NET 8.  We can only support (as in, provide tech support and bug fixes for) the agent instrumenting older versions so far; if the issue looks like a runtime problem we won't work on it.
* Removed a stray item about an unsupported .NET Framework version from the .NET Core versions support section.


cbehera-newrelic (approved):
lgtm

PR #19290, opened 2024-11-13
docs: add default information for Network Sample default filtered devices

alvarocabanas:
This PR adds add the default information for Network Sample default filtered devices.


cbehera-newrelic (approved):
lgtm

PR #19312, opened 2024-11-18
Update docs to include optional parameter.mdx

abhishuraina:
* Add optional parameter to run query to fetch DB instance size, the query can be slow that's why adding a param to fetch when needed.


adutta-newrelic (approved):
Made a minor change to the alignment of the collapser so that it could generate the Netlify link

PR #19313, opened 2024-11-18
fix: Update the indentation in callout tag so that file can be serial…

akashreddy-nr:
Resolved few Indentation Errors in metric-data-type.mdx file so the file can be serialized

Made these changes to fix "send content to machine translated" workflow


gmanandhar-nr (approved):
No content change. Just an indentation change to fix a dev job.

PR #19322, opened 2024-11-19
Fix: Change from sessionStorage to localStorage

steve-cunnew:
The docs still stated sessionStorage as being used, but it has been localStorage that has been used for quite some time now. Also corrected the mention of localStorage to start with lower case letter.


adutta-newrelic (approved):
Made some minor changes to the alignment of the document.

PR #19396, opened 2024-11-28
Updated the support date for Azure SQL database

jagan-k97:
New Relic will end support for the Azure SQL database integration polling API, because Azure is deprecating the SDK that the polling API uses. But this date has been changed from December 1st 2024 to July 31st 2025. So, we will be extending support for the same till 31st July 2025.


WriteMayur (approved):
Received the Approval in the following thread: https://newrelic.slack.com/archives/C35CCKT3L/p1732786234658849

sujitnewrelic:
Looks good to me

PR #19412, opened 2024-12-02
fix(mobile): Updated the link

WriteMayur:
Updated the links for the docs


sujitnewrelic (approved):
Verified. Links are correctly redirecting.

PR #19419, opened 2024-12-03
fix(Advance-config): Fixed the link directing to the entity definitions repo

sujitnewrelic:
Fixed the link that is directing to the entity definitions repo looking for all 'definition.yml' files inside the 'entity-types' repo.

Change in the page: /docs/network-performance-monitoring/advanced/advanced-config/#devices


WriteMayur (approved):
looks good!

PR #19429, opened 2024-12-03
Docs: Remove stale links from NR app dev guide

kennytrytek-wf:
The links in the app dev guide link to the developer site, which gives a 404.


WriteMayur (approved):
Thanks for making these updates

PR #19434, opened 2024-12-04
Add gems to the ruby agent list for support message brokers

tannalynn:
The ruby agent has added support for multiple new message broker gems but we haven't updated this page yet. This added the new gems we now support to the appropriate section


WriteMayur (approved):
looks good to me!

PR #19442, opened 2024-12-04
Streamline requirements

planteater:
Streamline/update requirements for code-level metrics, and also add requirements to Install article.


WriteMayur (approved):
looks good.

PR #19445, opened 2024-12-04
fix broken link to Prometheus agent install

keegoid-nr:
- broken links


WriteMayur (approved):
Thanks for fixing the link!

PR #19449, opened 2024-12-05
docs: Document sortType and flag sortBy as deprecated for entities query components

mrodriguez-nr:
The property `sortBy` was deprecated while ago in favor of `sortType`.

In the PR I also added some missing examples to show how `sortType` is used.

The deprecation is already over, the plan is to remove `sortBy` as soon as possible but first we have to verify that it is not used, so we are going to implement its use. In no case we are going to remove the property if it has use.


WriteMayur (approved):
Thanks for making these changes.

PR #19472, opened 2024-12-09
fix: JSON response returns success to be true for VM data integration

pthakral:
This PR is related to [NR-321768](https://new-relic.atlassian.net/browse/NR-321768)

One of the customers raised that the success value in JSON response is returned as false even after a successful import.
So, the changes have been made accordingly to return success value as true when the import is successful.


sujitnewrelic (approved):
LGTM

PR #19477, opened 2024-12-09
chore: Add step to update Jenkins entity.name WHERE statements

bpeckNR:
The quickstart dashboard needs updated if a customer uses anything other than jenkins for the service name. This isn't obvious in the current form of the doc, so I added an additional step.


sujitnewrelic (approved):
lgtm

PR #19478, opened 2024-12-09
docs: Update broken xcframework-installation link in mobile-monitoring.yml

shevaun-nr:
The name of this file was changed right after it was first created, but the link has never updated to reflect that. This PR should fix that link. 

The target file is: src/content/docs/mobile-monitoring/new-relic-mobile-ios/installation/xcframework-installation.mdx
The commit where the filename was changed is: https://github.com/newrelic/docs-website/commit/4410df08b7bf51a6abf332fabeea1deb8931045a


sujitnewrelic (approved):
lgtm

PR #19484, opened 2024-12-09
fix: Make it clear that we publicly just support `timestamp` as the log timestamp

bmcfeely:
## Give us some context

When looking into a support ticket, I noticed that the Logs API docs were confusing about which fields we'd use as a timestamp. This PR attempts to clarify that.

Note that `log_timestamp` is only used in built-in parsing rules, and we intentionally don't use that attribute as the log time, maybe because originally we weren't confident that the parsing rules could parse timestamps correctly?


arenjae (approved):

akashreddy-nr (dismissed):
Reran Netlify Checks, Build is successful and preview is ready , Approving the changes

sujitnewrelic (approved):
lgtm

PR #19498, opened 2024-12-10
feat: Add Cancel Query Nerdgraph Block

marcusperezNR:

## Give us some context

As the Dirac CAT team aims to add cancel async query to Nerdgraph, we are looking to add a block to explain to customers how to do so. We are adding onto the existing Nerdgraph async query document to keep everything cohesive and in one doc.


cbehera-newrelic (approved):
lgtm

PR #19516, opened 2024-12-12
Revert "Revert "docs(logs): Add new syslog mechanism to send the license key""

akashreddy-nr:
Reverts newrelic/docs-website#19515


sujitnewrelic (approved):
lgtm

PR #19544, opened 2024-12-16
feat(NRVM PHP): updated PHP package support table

sujitnewrelic:
Deleted the PHP package support table table and updated the content.


cbehera-newrelic (approved):
lgtm

PR #19569, opened 2024-12-18
Chore[S3-integration]: add information about metrics delays

morecar:
## Give us some context

S3 metrics (bucket) are notorious for being extremely late. Sometimes these metrics are over 48h late, which makes NR unable to ingest them, since they exceed the max age of NR's Metrics API. Customers often request support to understand why their metrics are missing. They should raise their concerns with AWS so they emit these metrics more timely. 
  
* Add any context that will help us review your changes such as testing notes, links to related docs, screenshots, etc.
The changes are documented in the documentation through links.


WriteMayur (approved):
looks good to me!

PR #19592, opened 2024-12-20
fix broken links

keegoid-nr:
Our Metric Streams install doc migrated to a new page with user-selectable options.


adutta-newrelic (approved):
LGTM

PR #19614, opened 2024-12-30
fix(grammar): Improved grammar from my previous PR

bpeckNR:
Fixed an issue with a sentence that describes how resource attributes that start with tags. are handled.


WriteMayur (approved):
looks good

PR #19674, opened 2025-01-09
wip

cbehera-newrelic:


src/content/docs/serverless-function-monitoring/overview.mdx
```diff
@@ -6,6 +6,9 @@ redirects:
 freshnessValidatedDate: never
 ---
 
+
+## Test review
```
vpayyapilly, line 10: This needs to be in title case

cbehera-newrelic: fixed

PR #19675, opened 2025-01-09
docs: updated mysql whatsNext.mdx to reflect Mysql 8.0 and above support

sairaj18:
- nri-mysql on host integration now also supports Mysql 8.0 and above.
- The query cache variables are removed from MySQL 8.0 [refer](https://dev.mysql.com/doc/refman/5.7/en/query-cache-status-and-maintenance.html) which need to be reflected in the docs after the feature gets released.
- https://github.com/newrelic/nri-mysql/commit/8a56a6331727a35a4e8e2d3931665e686dfedd9c

### DIFF
--- a/src/install/mysql/whatsNext.mdx
+++ b/src/install/mysql/whatsNext.mdx
@@ -238,7 +238,7 @@ The MySQL integration collects the following metrics:
           </td>
 
           <td>
-            Amount of free memory in bytes for the query cache.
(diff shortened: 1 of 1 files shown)


newrelic707195 (approved):
LGTM

nr-ksteinbach:
LGTM - thanks.

adutta-newrelic (approved):
LGTM

PR #19692, opened 2025-01-10
fix: poorly formatted code

brnhensley:
This code block was ugly and tough to read, so I put it in a codeblock

image


cbehera-newrelic (approved):
LGTM

PR #19699, opened 2025-01-13
Update live-archives.mdx

nedl86:
Added call-out relating to the RBAC permission required to query live archives


adutta-newrelic (approved):
LGTM

PR #19740, opened 2025-01-17
[NR-353267] Fix Action name and indentation

nr-raviraj:
This PR fixes the grok action name along with the indentation which was incorrect for one of the config parameter.


adutta-newrelic (approved):
LGTM

PR #19745, opened 2025-01-17
Whats new 01 17 cross account alerts

kimberlyburns:
This feature just went live, so this can go live at any point.  Since I don't do these often, I would appreciate a PMM review.  Jaime already reviewed the content, but just want to make sure I did the screenshots and formatting correctly.


adutta-newrelic (approved):
LGTM

PR #19766, opened 2025-01-22
State that agent supports PHP 8.4

mlocati:


sujitnewrelic (approved):
LGTM as per https://docs.newrelic.com/docs/release-notes/agent-release-notes/php-release-notes/php-agent-11-5-0-18/

PR #19787, opened 2025-01-23
Fix convention for environment variables

psimsa:
Currently used specification of environment variables to use in the yaml file is not correct. See https://opentelemetry.io/docs/collector/configuration/#environment-variables for details


sujitnewrelic (approved):
lgtm

sujitnewrelic (approved):
LGTM

PR #19802, opened 2025-01-24
Update opentelemetry-otlp.mdx to indicate support for OTLP 1.4.0

jack-berg:
cc @alanwest

### DIFF
--- a/src/content/docs/opentelemetry/best-practices/opentelemetry-otlp.mdx
+++ b/src/content/docs/opentelemetry/best-practices/opentelemetry-otlp.mdx
@@ -283,7 +283,7 @@ The mechanism to configure the endpoint will vary, but OpenTelemetry language SD
 
 ## OTLP protocol version [#otlp-version-support]
 
-New Relic uses [OTLP release [v1.4.0](https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v1.4.0). Later and earlier versions are supported but new features are not yet implemented. Experimental features which were removed after 0.18.0 are not supported. 
+New Relic uses OTLP release [v1.4.0](https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v1.4.0). Later and earlier versions are supported but new features are not yet implemented. Experimental features which were removed after 0.18.0 are not supported. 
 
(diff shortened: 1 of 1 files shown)


src/content/docs/opentelemetry/best-practices/opentelemetry-otlp.mdx
```diff
@@ -283,7 +283,9 @@ The mechanism to configure the endpoint will vary, but OpenTelemetry language SD
 
 ## OTLP protocol version [#otlp-version-support]
 
-New Relic uses [OTLP release v0.18.0](https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v0.18.0). Later versions are supported but new features are not yet implemented. Experimental features which were are no longer available in 0.18.0 are not supported.
+New Relic uses [OTLP release [v1.4.0](https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v1.4.0). Later and earlier versions are supported but new features are not yet implemented. Experimental features which were removed after 0.18.0 are not supported. 
+
+See [Traces](/docs/opentelemetry/best-practices/opentelemetry-best-practices-traces/), [Metrics](/docs/opentelemetry/best-practices/opentelemetry-best-practices-metrics/), and [Logs](/docs/opentelemetry/best-practices/opentelemetry-best-practices-logs/) for specific details on how data is mapped and which features are implemented.
```
alanwest, line 288: It could make sense to just delete this whole section. As we discussed, the version of the protos we use does not actually describe anything useful to customers. It'd be more useful to have a section articulating the list of things from the protos that we do not yet have any support for.

alanwest (approved):
Approving because I think this change is fine, but sharing an alternate thought...

src/content/docs/opentelemetry/best-practices/opentelemetry-otlp.mdx
jack-berg: People have asked what version of OTLP we support in the past. With backwards and forwards compatibility, and no guarantee that we implement all features of a particular version, this isn't an especially useful question to ask. But for someone who doesn't understand OTLP or protocol buffers deeply, I think its useful to maintain a section like this. Essentially, all it does is establish context for why the version of OTLP isn't very relevant, and link out to more relevant details.

sujitnewrelic (approved):
lgtm

PR #19959, opened 2025-02-12
NR-367015: Complex data types in body and attributes of OTLP Log Records

luckslovez:
We recently added support for complex data types (mainly KeyValueLists) in both OTLP LogRecord#attributes and LogRecord#body.

We think this doc update is required to let customers now about this extended support.


cbehera-newrelic (approved):
lgtm

PR #20015, opened 2025-02-19
Update Python DT page

lrafeei:
This PR updates the links and wording of the Python Agent's Distributed Tracing page.  The Python Agent has since deprecated the create/accept payload APIs and has moved towards insert/accept headers APIs.  The docs no longer link to nor reference the old APIs


WriteMayur (approved):
thanks for the update

PR #20022, opened 2025-02-20
Update view-system-limits.mdx

wildcardlinux:
We have a ticket to change the behavior in the UI so that violations only query the last 24 hours, as we recently discovered violation queries are not accurate outside of this time range: https://new-relic.atlassian.net/browse/NR-358126

The number is not accurate for 1 hour or time less than 24 hours.


WriteMayur (approved):
thanks

PR #20078, opened 2025-02-25
Update compatability requirement for Java lambda layers

obenkenobi:
I added a section in the compatibility docs for Java Lambda SDK and Layers. This is to guide customers if they are using an older version of `com.amazonaws:aws-lambda-java-events` and to tell them to upgrade to a comparable version.


cbehera-newrelic (approved):
LGTM

PR #20101, opened 2025-02-27
[NR-373715] add data endpoint settings

ywang-nr:
* What problems does this PR solve? - add data endpoint settings to match with iOS


adutta-newrelic (approved):
LGTM

PR #20143, opened 2025-03-05
chore: (NRQL Prediction): Updated the callout with the definition of gamma hyperparameter

sujitnewrelic:
Updated callouts with the definition of gamma hyperparameter.


adutta-newrelic (approved):
LGTM

PR #20149, opened 2025-03-06
fix(logs): (NR-371436) Clarify array support for logs

luckslovez:
We've never clarified the difference in array processing between opentelemetry and the rest of ingestion mechanisms.
This PR aims to fix that.

Bascially, we stringify all arrays except for Opentelemetry homogeneous (one type) arrays.

### DIFF
--- a/src/content/docs/logs/log-api/introduction-log-api.mdx
+++ b/src/content/docs/logs/log-api/introduction-log-api.mdx
@@ -461,7 +461,7 @@ Attributes may be of any of the following types:
       </td>
 
       <td>
-        Arrays will be _stringified_ and stored as strings. In the case of OpenTelemetry logs,
+        Arrays are converted to strings and then stored. OpenTelemetry logs support
         homogeneous arrays are supported. For more information see [arrays in OpenTelemetry](/docs/nrql/using-nrql/arrays-in-nrql/).
       </td>
     </tr>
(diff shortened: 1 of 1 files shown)


src/content/docs/logs/log-api/introduction-log-api.mdx
```diff
@@ -461,7 +461,8 @@ Attributes may be of any of the following types:
       </td>
 
       <td>
-        (unsupported)
+        Arrays will be _stringified_ and stored as strings. In the case of OpenTelemetry logs,
```
jdearana-nr, line 464: I don't think we should be using the verb _stringify_ in our pub docs, I'd rephrase it as: "Arrays will be **converted to text** ..."

WriteMayur, line 464: Suggestion:

```suggestion
        Arrays are converted to strings and stored as such. OpenTelemetry logs support
```

src/content/docs/logs/log-api/introduction-log-api.mdx
```diff
@@ -461,7 +461,8 @@ Attributes may be of any of the following types:
...
+        homogeneous arrays are supported. For more information see [arrays in OpenTelemetry](/docs/nrql/using-nrql/arrays-in-nrql/).
```
WriteMayur, line 465:
```suggestion
        homogeneous arrays. For more information see [arrays in OpenTelemetry](/docs/nrql/using-nrql/arrays-in-nrql/).
```

PR #20185, opened 2025-03-11
feat: (Alerts: Sliding window aggregation): Adding the associated pricing details

sujitnewrelic:
The users on Advanced and Core Compute pricing plans may incur additional CCU charges when utilizing sliding window aggregation. Adding a pricing section to that topic and callouts at other places where we have communicated about this feature.


adutta-newrelic (approved):
LGTM

PR #20192, opened 2025-03-11
remove support reference

keegoid-nr:
While still in pre-release status, no support is provided.


WriteMayur (approved):
thanks for the update

PR #20277, opened 2025-03-21
fix: fix typo in example query

metal-messiah:
This PR fixes a typo that was found by a customer that is present in an example query in our public preview of page resources capture.


sujitnewrelic (approved):
LGTM

PR #20318, opened 2025-03-27
(feat): Added mapping information from oTel metrics to New Relic K8S agent

WriteMayur:
As per the JIRA NR-365972, added mapping information from oTel metrics to New Relic K8S agent.


cbehera-newrelic (approved):
LGTM

PR #20323, opened 2025-03-27
chore: Update record-deployments.mdx

kaylareopelle:
This document doesn't reference the full scope of options to record deployments with the Ruby agent. 

Updating the bullet point to include our other options.


cbehera-newrelic (approved):
lgtm

PR #20324, opened 2025-03-27
chore: Update record-deployments-ruby-agent.mdx

kaylareopelle:
Update help options output to include information about required keys for v1 and v2 API options


cbehera-newrelic (approved):
lgtm

PR #20325, opened 2025-03-27
earliest wording

keegoid-nr:
The `earliest()` function returns the earliest value for each facet, not the most recent.


cbehera-newrelic (approved):
lgtm

PR #20334, opened 2025-03-28
Update acceptable-use-policy.mdx

NR-randymartin:
Added the last reviewed and updated date.


adutta-newrelic (approved):
Spoke with Randy Martin, and he mentioned that "even before him submitting these PR's he had a discussion with with Compliance and Product Legal.  We have customer's asking for the reviewed and updated dates for policies for their auditor's needs,This information needs to be publicly viewable for these policies." hence approving this PR

PR #20304, opened 2025-03-25
docs(dotnet): Update azure function environment variable configuration instructions

tippmar-nr:
Corrects formatting of environment variables for Azure Functions and adds a section to distinguish between environment variables required for "isolated" and "in process" mode functions.

PR #20341, opened 2025-03-31
**DRAFT** docs(dotnet): Remove recently added environment variable configuration for in-process Azure functions.

tippmar-nr:
Removes in-process Azure function environment variable configuration that was recently added with #20304. 

A [fix](https://github.com/newrelic/newrelic-dotnet-agent/pull/3068) to the .NET agent was applied that makes the separate configuration for in-process Azure functions unnecessary. 

Hold for .NET agent 10.40.0 release.


WriteMayur (approved):
thanks for the update

PR #20344, opened 2025-03-31
New feature: (API): Admin setup for user key creation and access

sujitnewrelic:
Added a new topic to configure Admin groups with a special role for reading API keys of other users while creating it using NerdGraph.


WriteMayur:
@sujitnewrelic  Do we still need the changes in this one? Since it was created nearly a year ago, I wanted to see if it’s still relevant or if we should close it out.

PR #20360, opened 2025-04-01
feat: french translation for licenses

akashreddy-nr:
- Removed Licenses from i18n exclusions
- Added translate - fr for Human french Translations


jmiraNR (approved):
Approved.

PR #20396, opened 2025-04-04
(chore): updated the IP range

WriteMayur:
As per this slack request, updated the IP range for the Synthetics.


adutta-newrelic (approved):
LGTM

PR #20414, opened 2025-04-07
update text to check in best practices example

keegoid-nr:
Update example to match changes in the page.


WriteMayur (approved):
thanks for making the update.

PR #20454, opened 2025-04-11
docs(logging): Add limitations of Log API to syslog page

danybmx:
This is to inform users that Log API limits would also apply for TPC syslog endpoints as it's the main entry point for logs.

### DIFF
--- a/src/content/docs/logs/log-api/use-tcp-endpoint-forward-logs-new-relic.mdx
+++ b/src/content/docs/logs/log-api/use-tcp-endpoint-forward-logs-new-relic.mdx
@@ -20,6 +20,10 @@ To forward logs to New Relic using a syslog client, you need:
 * A valid New Relic <InlinePopover type="licenseKey"/> for the account you want to send logs to
 * Some minor changes to the syslog client's configuration, as explained in this document
 
+## Limitations
+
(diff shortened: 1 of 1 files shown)


src/content/docs/logs/log-api/use-tcp-endpoint-forward-logs-new-relic.mdx
```diff
@@ -20,6 +20,10 @@ To forward logs to New Relic using a syslog client, you need:
 * A valid New Relic <InlinePopover type="licenseKey"/> for the account you want to send logs to
 * Some minor changes to the syslog client's configuration, as explained in this document
 
+## Limitations
+
+Log lines exceeding 1MB in compressed size will be dropped, and other limitations from the [Log API](/docs/logs/log-api/introduction-log-api/#limits) may also apply. Please ensure your log lines are within this size limit to avoid data loss.
```
WriteMayur, line 25:
```suggestion
Log lines that exceed 1 MB in compressed size will be dropped. Other limitations from the [Log API](/docs/logs/log-api/introduction-log-api/#limits) may also apply. To prevent data loss, ensure your log lines are within this size limit.
```

WriteMayur (approved):
thanks for making the update

PR #20477, opened 2025-04-15
fix examples

keegoid-nr:
- See the [nri-bundle values.yaml](https://github.com/newrelic/helm-charts/blob/8a36c205015dffc346561650284e9ee6516f9d0c/charts/nri-bundle/values.yaml#L52) for the most up-to-date  options.
- The `autoscaling/v2beta2` api version was deprecated in Kubernetes v1.23. Use `kubectl api-versions | grep autoscaling` to check your cluster's version.


adutta-newrelic (approved):
LGTM

PR #20484, opened 2025-04-16
chore: (Anomaly detection): Add a tip - The current solution is not supporting monthly and yearly seasonality options

sujitnewrelic:


adutta-newrelic (approved):
LGTM

PR #20493, opened 2025-04-16
feat(scorecards): Added the scorecards templates topic

ubanerjeeNR:
Added a topic describing the six templates that users could use to build their scorecards


vpayyapilly (approved):
Good to go

PR #20502, opened 2025-04-17
Update Python agent configuration settings page

lrafeei:
This PR updates the configuration settings page for the Python agent to include "environment variable" in settings that are also enabled through environment variables.


cbehera-newrelic (approved):
LGTM

PR #20506, opened 2025-04-19
docs: Update download.newrelic.com networks

albsga4:
We are updating the CDN that provides `download.newrelic.com`, therefore there is a need to update the docs with the new network block.


cbehera-newrelic (approved):
lgtm

PR #20556, opened 2025-04-28
fix: deserialisation

akashreddy-nr:
we are decoding the vfile here because in the processor pipeline,
 we have encoded the text nodes in the stringify as the workaround to handle unsafe characters.
 in that processor, we are using `htmlEncode` to encode the text nodes.
 however, we have found some nodes which are already encoded int rehype2remark
 (like `&gt`), and they are getting double encoded
  so decoding the vfile undos the double encoding.


gmanandhar-nr (approved):

gmanandhar-nr:
this fixes the issue apostrophe issue and '>', '<" issue. https://new-relic.atlassian.net/browse/NR-403313

gmanandhar-nr:
approved

PR #20558, opened 2025-04-28
Chore: [Do not merge]: Added NerdGraph and NRQL query references in REST API v2 examples

sujitnewrelic:
The REST API v2 are going to be EOLed and we are encouraging users to use NerdGraph and NRQL functions. So, added links callouts with references in the REST API v2 example pages, so that user can switch and use  NerdGraph and NRQL functions.


baurfamily:
Looks good!

adutta-newrelic (approved):
LGTM

PR #20633, opened 2025-05-06
Update support-arm-graviton-x86-64.mdx

nr-ksteinbach:
very important clarification, because some Windows IoT versions are also available as arm64, but neither the infra agent nor the OHI's are compiling to arm or arm64 binaries.


WriteMayur (approved):
looks good to me. Thanks for the update

PR #20688, opened 2025-05-15
Update requirements-infrastructure-agent.mdx

nr-ksteinbach:
Small clarification of supported Windows OS.
Windows 10 has a Microsoft EOL date, which will also the date we will stop supporting it.
Windows Server did not have any "Service Packs" anymore since Windows 2008 (long out of support already).


sujitnewrelic (approved):
LGTM

PR #20716, opened 2025-05-20
chor: New PM feedbacks in change tracking event doc

WriteMayur:
This PR address the feedback from the PMs to the change tracking event documentation.


gmanandhar-nr:
cla checks are failing across multiple teams. bypassing the checks and merging, since we know that Mayur has signed it before

PR #20718, opened 2025-05-20
Update synthetics nerdgraph examples to indicate optional fields more clearly

shevaun-nr:
## Update Nerdgraph API examples to correctly indicate which fields are optional for monitors

* The existing docs place several mandatory fields for monitor runtimes in the 'optional' section. Many of these are not optional, even though they have only one accepted value. 
* Specifically, the RUNTIME_* options are not optional; even though they have only one acceptable value, if they are not specified the defaults used by the backend are invalid (the legacy runtime options).
* The issue above is exacerbated by the Nerdgraph Explorer - the runtime values _are_ considered optional in the backend, so the explorer doesn't enforce them. However the request will fail without them due to invalid defaults being used. Users will refer to the docs for an explanation.
* To address this, this PR changes the section title and marks the fields which are actually optional.

A more ideal version of this would be to change the API to use more correct defaults, but this should help in the meantime.

### DIFF
--- a/src/content/docs/apis/nerdgraph/examples/nerdgraph-synthetics-tutorial.mdx
+++ b/src/content/docs/apis/nerdgraph/examples/nerdgraph-synthetics-tutorial.mdx
@@ -97,28 +97,28 @@ Configure simple browser, scripted browser, and step monitors to use one or more
 
   <Collapser
     id="optional-fields"
-    title="Define optional fields"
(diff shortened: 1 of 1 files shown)


WriteMayur:
@shevaun-nr, do we still need changes of these PR as it has been almost a year this PR got created?

PR #20729, opened 2025-05-20
.NET agent: add troubleshooting doc for app hang on startup issue

nr-ahemsath:
This PR adds a short document to the troubleshooting section of the .NET agent docs.  The doc lets customers know about a workaround for a potential hang in their .NET 8 apps on startup.


cbehera-newrelic (approved):
LGTM

PR #20789, opened 2025-05-29
fix: release notes duplication error for translation

akashreddy-nr:
**We currently have a bug in release notes page, where when we select the language in French, it displays both the English version and French Version**

- **This PR Resolves The Duplication Bug by filtering duplicated posts in src/templates/releaseNoteLandingPage.js**
- **This PR also addresses incorrect release-notes filenames in i18n folder by renaming them with same corresponding filenames in content/docs/release-notes folder** 

[Jira Ticket](https://new-relic.atlassian.net/browse/NR-418858)

image


package.json
```diff
@@ -238,5 +238,6 @@
   },
   "bugs": {
     "url": "https://github.com/newrelic/docs-website/issues"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
gmanandhar-nr: is this change needed?

gmanandhar-nr (approved):

gmanandhar-nr (dismissed):

package.json
akashreddy-nr: Removed

PR #20806, opened 2025-05-30
Draft: Update Azure Functions docs based on testing

nr-ahemsath:
This PR updates our Azure Functions monitoring docs (compatibility/support and install instructions) based on testing I've recently done on various combinations of function model (isolated vs. in-process), host platform (Windows vs. Linux), hosting plan type (consumption or dedicated), function runtime (.NET Framework vs. .NET), and agent install method (Azure Site Extension vs. Agent NuGet package).

Summary of changes:
* Called out that we do not support the combination of Linux OS, Consumption plan and In-Process function model
* Called out that we only support functions using version 4 of the functions runtime
* Changed the environment variable instructions for the Windows + Azure Site Extension install method to explicitly set all of the profiling env vars, similar to the instructions for using the NuGet package install method, except that the paths to the files/folders point to where the site extension places the agent (`C:\home\NewRelicAgent`).  I removed the line "The Azure websites extension automatically configures all other required environment variables" because that turns out to be false in some cases (particularly consumption-type hosting plans).


adutta-newrelic (approved):
LGTM

PR #20823, opened 2025-06-03
.NET agent: update compatibility docs based on latest testing

nr-ahemsath:
Update the .NET agent compatibility/support docs based on latest testing.


adutta-newrelic (approved):
LGTM

PR #20837, opened 2025-06-04
fix: we sample 10, not 10%

brnhensley:
Our APM agent DT pages all contain the same incorrect text:

> [Standard distributed tracing](#quick-start-apm) for APM agents captures up to 10% of your traces, but if you want us to analyze all your data and find the most relevant traces, you can set up [Infinite Tracing](/docs/understand-dependencies/distributed-tracing/infinite-tracing/introduction-infinite-tracing).

This is incorrect, see the [DT sampling details doc ](https://docs.newrelic.com/docs/distributed-tracing/concepts/how-new-relic-distributed-tracing-works/#trace-origin-sampling):

> The first service we monitor in a distributed trace is called the trace origin. The trace origin chooses requests at random to be traced. [...] The trace origin service samples 10 traces per minute by default. It attempts to spread out the collection of these 10 traces over a minute in order to get a representative sample over time.


adutta-newrelic (approved):
LGTM

PR #20873, opened 2025-06-09
Update nrql-syntax-clauses-functions.mdx

adutta-newrelic:
As discussed with @zyria removing the link as the course is currently hidden on learn.newrelic.com but it is still able to be accessed with a link. We’d like the link removed so users won’t access the out of date course.


sujitnewrelic (approved):
LGTM

PR #20910, opened 2025-06-13
Update rules-limits-alerts.mdx

qryuu:
The destination limits have been specified.

## Give us some context


relic-js:
The proposed doc update mentions "Slack" as a destination type but this particular limit is specifically in reference to the `SLACK_LEGACY` destination type, and we should clarify that.

Also, shouldn't the "minimum" number of destinations be `0` or `N/A`, rather than `1`? And do we want to include channels-per-destination-type limits as well?

(I think it would be useful to have this update reviewed by the Notifications PM)

sujitnewrelic (approved):
LGTM

PR #20950, opened 2025-06-17
fix: fix field names in examples

metal-messiah:
This PR fixes `entryName` field mistakes in the examples page for Marks and Measures.  in various places, `entryName` should have been `entryType`.  As well, the bottom example was updated to be more useful.


cbehera-newrelic (approved):
LGTM

PR #21002, opened 2025-06-24
.NET agent: update latest compatible libraries based on testing

nr-ahemsath:
Update the .NET agent's compatibility and support document with latest verified compatible versions based on our testing.


adutta-newrelic (approved):
LGTM

PR #21088, opened 2025-06-30
feat: Add Node v24 to the list of supported Node runtimes

amychisholm03:
The Node.js agent released support for the latest Node runtime, 24, in version 12.23.0 of the agent. We should reflect this change in the Node agent system requirements list.


bizob2828 (approved):

WriteMayur (approved):
lgtm

PR #21103, opened 2025-07-02
Minor text change in French.

jmiraNR:
Minor text change in French.


WriteMayur (approved):
thanks for making the update

PR #21107, opened 2025-07-02
.NET agent: Update gpg key import commands; remove sudo from commands in Dockerfiles

nr-ahemsath:
This PR:

1. Updates the syntax for the `gpg` command that imports the public signing key for an APT repo to be more robust (it will handle being run multiple times)
2. Alters those commands for `Dockerfile` examples to no longer prefix the command with `sudo` as this is both unnecessary (Dockerfile commands run as root by default) and will cause the instructions to fail (sudo isn't installed or enabled by default in most Docker images)


WriteMayur (approved):
looks good to me

PR #21267, opened 2025-07-17
Update Ruby Agent EOL policy to call out SemVer usage

chynesNR:
Adding some clarification to our Ruby Agent EOL policy regarding breaking changes.


adutta-newrelic (approved):
LGTM

PR #21268, opened 2025-07-17
Update .NET Agent EOL policy to call out SemVer usage

chynesNR:
Update .NET Agent EOL policy to call out SemVer usage


adutta-newrelic (approved):
LGTM

PR #21352, opened 2025-07-29
ktlo[PAYG price revise]: Update the per GB price

sujitnewrelic:
Updating the PAYG per GB price:
- Data plan from $0.35 to $0.40
- Data plus plan from $0.55 to $0.60


sujitnewrelic:
netlify build

PR #21354, opened 2025-07-29
chore(php): update supported alpine versions

hahuja2:
This PR adds `3.21` and `3.22` as supported alpine versions.


WriteMayur (approved):
thanks for the update

PR #21399, opened 2025-08-01
fix three link destination URLs

bibliobrisa:
Correcting three destination URLs that are out of date and added new line with new link to Github ReadMe.


WriteMayur (approved):
thanks for fixing it

PR #21408, opened 2025-08-04
Update vmware-tanzu-nginx-buildpack-integration.mdx

gsidhwani-nr:
The buildpack is mofiied for latest go build 1.23 and tested with latest OPS Manager 10.2.


sujitnewrelic (approved):
LGTM

PR #21412, opened 2025-08-04
feat(agent-control): private repositories configuration

sigilioso:

## Give us some context

This PR adds the documentation for configuring Agent Control with private helm repositories.

### DIFF
--- a/src/content/docs/new-relic-control/agent-control/configuration.mdx
+++ b/src/content/docs/new-relic-control/agent-control/configuration.mdx
@@ -379,3 +379,132 @@ For proxy setups using HTTPS authentication with self-signed certificates, you n
 </Callout>
 
 When using a proxy, you must also configure proxy settings for each managed agent individually. Refer to each agent's specific documentation for proxy configuration options.
+
+## Private repository configuration
+
+Agent Control supports configuring private Helm repositories to deploy both Agent Control itself and the managed agents. This enables environments where the New Relic Helm charts are not directly accessible.
+
+<Callout variant="caution">
+  When using private Helm repositories, the charts need to be compatible and the referenced images within the charts must be reachable. If not, the agents will not work as expected.
+</Callout>
+
+### 1. Enable private repositories for agents
+
+For security reasons, only explicitly enabled repositories will be allowed in remote configurations. To enable specific repositories, you need to update the Agent Control configuration:
+
+<CollapserGroup>
+  <Collapser
+    id="k8s-private-repository-config"
+    title="Enable private repositories"
+  >
+
+    ```yaml
+    # values-newrelic.yaml
+
+    global:
+      cluster: "YOUR_CLUSTER_NAME"
+      licenseKey: "YOUR_LICENSE_KEY"
+
+    # ...
+
+    agent-control-deployment:
+      config:
+        allowedChartRepositoryUrl:
+          - https://my-private-repository-1
+          - https://my-private-repository-2
+        # ...
+    ```
+
+  </Collapser>
+</CollapserGroup>
+
+The allowed repository configurations can then be used in your remote configurations within New Relic Control. Example:
+
+```yaml
+chart_version: "1.2.3"
+chart_repository:
+  url: "https://my-private-repository-1"
+  name: "my-chart-name" # Optional: use only if the chart name doesn't match New Relic's chart name
+```
+
+Additionally, you need to configure Agent Control's Helm installation to use your private repository if the `agent-control` chart itself is in a private repository. This is separate from the configuration for managed agents. Refer to the `agent-control` Helm chart [values.yaml](https://github.com/newrelic/helm-charts/blob/master/charts/agent-control/values.yaml) to set up the `installationJob` section. Specifically:
+
+- `chartRepositoryUrl` containing your repository URL
+- `name` if using a different chart name
+- `repositorySecretReferenceName` and `repositoryCertificateSecretReferenceName` for authentication (see the authentication section below for details)
+
+### 2. Set up authentication for private repositories
+
+You need to set up additional resources to enable authentication for accessing your private repository:
+
+<CollapserGroup>
+  <Collapser
+    id="k8s-private-repository-basic-auth"
+    title="Basic authentication"
+  >
+    To authenticate using basic authentication (username and password), you need to create a secret in the Agent Control namespace containing the expected values in `data.username` and `data.password`.
+
+    Example:
+
+    ```yaml
+    apiVersion: v1
+    kind: Secret
+    metadata:
+      name: my-secret
+    stringData:
+      username: "myUser"
+      password: "myPassword"
+    ```
+
+    For more details, see the [Flux documentation](https://fluxcd.io/flux/components/source/helmrepositories/#secret-reference).
+
+    When using basic authentication, the remote configuration should be configured as follows:
+
+    ```yaml
+    chart_repository:
+      url: "https://my-private-repository-1"
+      secret_reference:
+        name: my-secret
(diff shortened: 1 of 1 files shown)


src/content/docs/new-relic-control/agent-control/configuration.mdx
```diff
@@ -379,3 +379,124 @@ For proxy setups using HTTPS authentication with self-signed certificates, you n
 </Callout>
 
 When using a proxy, you must also configure proxy settings for each managed agent individually. Refer to each agent's specific documentation for proxy configuration options.
+
+## Private repository configuration
+
+Agent Control supports configuring private repositories to deploy both Agent Control itself and the managed agents. This enables environments where the New Relic helm-charts are not directly accessible.
```
MunhozThiago: nit: "...configuring private **Helm** repositories..." and "...New Relic **Helm** charts..."

src/content/docs/new-relic-control/agent-control/configuration.mdx
```diff
@@ -379,3 +379,124 @@ For proxy setups using HTTPS authentication with self-signed certificates, you n
...
+
+<Callout variant="caution">
+  When using private Helm repositories, the chart needs to be compatible and the referenced images must be reachable, otherwise the agents will not work as expected.
```
MunhozThiago: nit:

...Helm repositories, the **charts need** to be compatible and the referenced images **within the charts** must be reachable. **If not,** the agents will not work as expected.

src/content/docs/new-relic-control/agent-control/configuration.mdx
```diff
@@ -379,3 +379,124 @@ For proxy setups using HTTPS authentication with self-signed certificates, you n
...
+</Callout>
+
+### Enable private repositories for agents
```
MunhozThiago: Suggestion: 

### 1. Enable Private Repositories for Managed Agents

src/content/docs/new-relic-control/agent-control/configuration.mdx
```diff
@@ -379,3 +379,124 @@ For proxy setups using HTTPS authentication with self-signed certificates, you n
...
+
+For security reasons, only explicitly enabled repositories will be allowed in remote configurations. To enable specific repositories, you need to update the Agent Control configuration:
+
+<CollapserGroup>
+  <Collapser
+    id="k8s-private-repository-config"
+    title="Enable private repositories"
+  >
+
+    ```yaml
```
MunhozThiago, line 401: I would recommend adding this line here to add better context
    `# values-newrelic.yaml`

src/content/docs/new-relic-control/agent-control/configuration.mdx
```diff
@@ -379,3 +379,124 @@ For proxy setups using HTTPS authentication with self-signed certificates, you n
...
+        allowedChartRepositoryUrl:
+          - https://my-private-repository-1
+          - https://my-private-repository-2
+        # ...
+    ```
+
+  </Collapser>
+</CollapserGroup>
+
+The allowed repository configurations can then be used in remote configurations. Example:
```
MunhozThiago: nit:

"...be used **in your** remote configurations **within New Relic Control**..."

src/content/docs/new-relic-control/agent-control/configuration.mdx
```diff
@@ -379,3 +379,124 @@ For proxy setups using HTTPS authentication with self-signed certificates, you n
...
+
+```yaml
+chart_version: "*"
```
MunhozThiago: Should we explicitly refer to a version, something like this in the example?

"1.23.0" # Best practice: always use an explicit chart version.

src/content/docs/new-relic-control/agent-control/configuration.mdx
```diff
@@ -379,3 +379,124 @@ For proxy setups using HTTPS authentication with self-signed certificates, you n
...
+chart_repository:
+  url: "https://my-private-repository-1"
+  name: "my-chart-name" # Optional: use only if the chart name doesn't match New Relic's chart name
+```
+
+Additionally, you need to configure Agent Control installation to use your private repository. Refer to the `agent-control` Helm chart [values.yaml](https://github.com/newrelic/helm-charts/blob/master/charts/agent-control/values.yaml) to set up the `installationJob` section. Specifically:
```
MunhozThiago: Suggestion:

"..Agent Control**'s Helm** installation to use **a private repository** **if the `agent-control` chart itself is in a private repo**. **This is separate from the configuration for managed agents**..."

src/content/docs/new-relic-control/agent-control/configuration.mdx
```diff
@@ -379,3 +379,124 @@ For proxy setups using HTTPS authentication with self-signed certificates, you n
...
+
+- `chartRepositoryUrl` containing your repository URL
+- `name` if using a different chart name
+- `repositorySecretReferenceName` and `repositoryCertificateSecretReferenceName` for authentication (see the authentication section below for details)
+
+### Set up authentication for private repositories
```
MunhozThiago: ### 2. Set up Authentication for Private Repositories

src/content/docs/new-relic-control/agent-control/configuration.mdx
```diff
@@ -379,3 +379,124 @@ For proxy setups using HTTPS authentication with self-signed certificates, you n
...
+
+You need to set up additional resources to enable authentication for accessing your private repository:
+
+<CollapserGroup>
+  <Collapser
+    id="k8s-private-repository-basic-auth"
+    title="Basic authentication"
+  >
+    To authenticate using basic authentication (username and password), you need to create a secret in the Agent Control namespace containing the expected values in `data.username` and `data.password`.
```
MunhozThiago, line 445: Do you think there there a risk of exposing credentials in shell history? In that case we could recommend creating first a file for that.

If so, my suggestion would be something like this: Suggestion:

"...in the Agent Control namespace. 

To authenticate using basic authentication (username and password), you need to create a secret in the Agent Control namespace. The expected values are `data.username` and `data.password`. 
<Callout variant="important">
We recommend creating the secret from a file to avoid exposing credentials in your shell history."
</Callout>

And then, in the example, add a bash creation:
Example using a file:
```bash
$ echo '{"username": "myUser", "password": "myPassword"}' > auth.json
$ kubectl create secret generic my-secret --from-file=auth.json
```

WDYT?

MunhozThiago (commented):
Overall, all good. Take my suggestions with a grain of salt 😛

src/content/docs/new-relic-control/agent-control/configuration.mdx
sigilioso: Good point! Although, as it is merely an example, we could instead take the same approach that was taken for the certificates case and provide a yaml manifest instead. Something like:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: my-secret
stringData:
  username: "myUser"
  password: "myPassword"
```

The user can choose how to generate and apply it 🙂

sigilioso: The changes I've just pushed include the example in yaml-format in order to have it consistent with the other example. Let me know your thoughts.

MunhozThiago: You're right! The example in yaml format helps. All good!

MunhozThiago:
✅ All ok for me. Consider the PR approved from my side.
Sorry I still don't have the permissions to approve it.

adutta-newrelic (approved):
LGTM

PR #21418, opened 2025-08-04
remove attributions references

bibliobrisa:
Attributions not currently supported. Removing references suggesting otherwise.


WriteMayur (approved):
looks good

PR #21453, opened 2025-08-08
Update public docs to describe browser agent's harvest interval behavior

ellisong:
Adds a notice about the browser agent's harvest interval behavior. If the payload size is > 16KB or if it's been 30 seconds, we trigger a harvest for the data.

JIRA: https://new-relic.atlassian.net/browse/NR-419973

We reverted these changes before [in this PR](https://github.com/newrelic/docs-website/pull/21161) since the harvest internal behavior changes weren't released yet. They are released today so we'd like these documentation changes in to inform customers of the changes.


akashreddy-nr (commented):
netlify build fork

tejaswaroop-nr (approved):
LGTM

PR #21482, opened 2025-08-13
fix: agent env injection in k8s-agent-operator.mdx

hammady:
This fixes the schema for injecting environment variables in the agent runtime (spec.env -> spec.agent.env).


cbehera-newrelic (approved):
lgtm

PR #21486, opened 2025-08-13
feat: update to observability maturity series docsDocs maturity model

ricegi:
Update of Observability Maturity series to conform with new maturity model and the introduction of maturity scorecard templates.


ricegi:
closing pending html fix

PR #21570, opened 2025-08-26
ebpf : add new env variable, update description for drop data env var

anmol-ap:
In case where the linux headers cannot be downloaded, user can provide custom path where they can manually download the required linux headers through environment variable DOWNLOADED_PACKAGED_HEADERS_PATH. This is the path of the directory and not the headers itself. [PR](https://github.com/newrelic/newrelic-ebpf-agent/pull/574).

Description update for DROP_DATA_FOR_ENTITY


cbehera-newrelic (approved):
lgtm

PR #21620, opened 2025-09-03
feat(ac): adding release notes for components

paologallinaharbur:
## Give us some context

We are adding logs for two componenents of the agent control.
 - I added "chart" suffix since the versions refers to the chart, since in the future we might have also onHost support we need to differentiate it
 - I had to add "deployment" and "continuous delivery" since that is the actual name of the two components. 

"continuous delivery" could be called "cd" instead, I am opened to suggestions


WriteMayur:
netlify build fork

Aashirwadjain (approved):

Aashirwadjain:
netlify build

PR #21627, opened 2025-09-03
Update introduction-dashboards.mdx

johnathannewrelic:
I am a New Relic employee on the Accounts-Security team (Slack: #gts-accounts-security) and am updating behavior of dashboard asset and ownership expectations.

## Give us some context

### DIFF
--- a/src/content/docs/query-your-data/explore-query-data/dashboards/introduction-dashboards.mdx
+++ b/src/content/docs/query-your-data/explore-query-data/dashboards/introduction-dashboards.mdx
@@ -228,6 +228,12 @@ Here you can carry out the following actions:
 
 ## Dashboard permissions [#dashboards-permissions]
 
+Dashboards, as far as an asset, belong to the accoount ID they were created for.
+
+Ownership of of a dashboard belongs to the user that creates them. Ownership is what determine's who can manage a dashboard's permissions. 
+
+For example if the user that created a dashboard is then deleted, the dashboard still exists, still has the same permissions it had at the time the user was deleted, and belongs to the account ID it was created for.
+
 Dashboards have three types of permissions:
 
 * <DNT>**Edit - everyone in account**</DNT>: All users have full rights to the dashboard.
--- a/src/content/docs/query-your-data/explore-query-data/dashboards/introduction-dashboards.mdx
+++ b/src/content/docs/query-your-data/explore-query-data/dashboards/introduction-dashboards.mdx
@@ -228,11 +228,11 @@ Here you can carry out the following actions:
 
 ## Dashboard permissions [#dashboards-permissions]
 
-Dashboards, as far as an asset, belong to the accoount ID they were created for.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/query-your-data/explore-query-data/dashboards/introduction-dashboards.mdx
```diff
@@ -228,6 +228,12 @@ Here you can carry out the following actions:
 
 ## Dashboard permissions [#dashboards-permissions]
 
+Dashboards, as far as an asset, belong to the accoount ID they were created for.
+
+Ownership of of a dashboard belongs to the user that creates them. Ownership is what determine's who can manage a dashboard's permissions. 
+
+For example if the user that created a dashboard is then deleted, the dashboard still exists, still has the same permissions it had at the time the user was deleted, and belongs to the account ID it was created for.
+
```
WriteMayur, line 236: Hi @johnathannewrelic, I would suggest to write the information as follows. Please suggest if this is incorrect:

It's important to understand the distinction between where a dashboard belongs as an asset versus who owns it.

- Account Association: Every dashboard is a resource that belongs to the account in which it was created. This association does not change.
- User Ownership: The user who creates a dashboard is its owner. Ownership determines who has the right to manage the dashboard's permissions. 
For example, If an owner is deleted from the account, the dashboard is not removed. It persists with its last-known permissions, but because the owner is gone, those permissions can no longer be managed or changed.

WriteMayur: If you want to write in a single paragraph then you can write as:

All dashboards are resources that belong to the account where they are created. The user who creates a dashboard becomes its owner, which grants them the ability to manage its permissions. If an owner is deleted from the account, the dashboard is not removed. It persists with its last-known permissions, but because the owner is gone, those permissions can no longer be managed or changed.

johnathan-smith-GH: Hello @WriteMayur I do generally like the first suggestion you shared above. I have added some relevant details inspired by your suggestion, as well as made things a little more technically accurate:

It's important to understand the distinction between where a dashboard belongs as an asset versus who owns it.

Account Association: Every dashboard is a resource that belongs to the account in which it was created. This association does not change. If the account it belongs to is deleted, so is the dashboard.

User Ownership: The user who creates a dashboard is its owner. Ownership determines who has the right to manage the dashboard's permissions.

For example, If the owner user is deleted, the dashboard is not removed from the account it belongs to. It persists with its last-known permissions, but because the owner is gone, those permissions can no longer be managed or changed. The solution to this would be to [duplicate the dashboard](https://docs.newrelic.com/docs/query-your-data/explore-query-data/dashboards/introduction-dashboards/#dashboards-clone) and the user that does that will be the owner of the duplicate. The dashbaord that was duplicated can then [be deleted](https://docs.newrelic.com/docs/query-your-data/explore-query-data/dashboards/introduction-dashboards/#dashboards-delete).

adutta-newrelic (approved):
LGTM

PR #21705, opened 2025-09-15
.NET Agent: Update latest verified compatible libraries

nr-ahemsath:
This PR updates the .NET agent's list of compatible library versions based on our latest testing.


tejaswaroop-nr (approved):
LGTM!

PR #21723, opened 2025-09-17
Update nerdgraph-issues-api-via-github.mdx

dkoyano:
Updated the user of currentUser to actor in the Nerdgraph calls.

## Give us some context

The doc uses currentUser which is for internal use only. The update changes this to actor which will allow customers to make the Nerdgraph call.


tejaswaroop-nr (approved):
LGTM.

PR #21783, opened 2025-09-24
.NET Agent: Update all APT repo install instructions

nr-ahemsath:
This PR makes the following changes to all APT-based install instructions for the .NET agent:

1. Switch from `wget` to `curl`, as curl is more likely to be installed by default on most Linux distros.
2. Signing key location moved from `/usr/share/keyrings` to `/etc/apt/keyrings` based on new information about where these keyfiles should go by convention (keys managed "manually" by users go in /etc/apt/keyrings; /usr/share/keyrings is for keys managed by packages)
3. Using `gpg --dearmor` instead of what was there before allows these instructions to work on newer distros like Debian 13 where the APT command has been upgraded to >=v3 and it uses Sequoia instead of GNUPG for its OpenPGP key verification needs.


sujitnewrelic (approved):
LGTM

PR #21843, opened 2025-10-02
chore[live URLs]: live URLs can only be managed by NerdGraph API. Updated Doc

sujitnewrelic:
live URLs can only be managed by NerdGraph API. Updated the callout to reflect it correctly


PallaviWrite (approved):
lgtm

Issue #140, opened 2020-10-08
Decide on Open Source License for the new Docs Site


jpvajda:
In order to offer our site as opensource project we need to decide what license makes the most sense for this repository. Currently the OSS and Developer Sites use the [Apache 2.0 license](https://github.com/newrelic/developer-website/blob/develop/LICENSE) but for Docs this license make not be optimal.

PR #21857, opened 2025-10-06
Drop rules to Pipeline cloud rules

tejaswaroop-nr:
Migration docs for Pipeline cloud rules with Terraform

### DIFF
--- a/src/content/docs/infrastructure-as-code/terraform/migrate-drop-rules-to-pipeline-cloud-rules.mdx
+++ b/src/content/docs/infrastructure-as-code/terraform/migrate-drop-rules-to-pipeline-cloud-rules.mdx
@@ -23,8 +23,8 @@ If you're managing your NRQL drop rules through Terraform using the [`newrelic_n
 
 This guide covers two ways to migrate your NRQL drop rules to Pipeline cloud rules in Terraform:
 
-1. Native Terraform commands: The standard Terraform workflow using `terraform import`, `terraform plan`, and `terraform state rm` commands.
-2. New Relic CLI automation tools: New Relic provides CLI tools that automate the native Terraform workflow, available in two versions:
+- Native Terraform commands: The standard Terraform workflow using `terraform import`, `terraform plan`, and `terraform state rm` commands.
+- New Relic CLI automation tools: New Relic provides CLI tools that automate the native Terraform workflow, available in two versions:
    - GitOps (CI/CD): For teams using CI/CD tools where Terraform state is not directly accessible from the workspace (typically stored in remote backends).
    - Local Terraform: For teams executing Terraform commands where they can directly read the Terraform state file, whether it's local or in a remote backend.
 


tejaswaroop-nr:
Updated document preview link: https://pipeline-cloud-rules-migration--docs-website-netlify.netlify.app/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules/

tejaswaroop-nr:
Netlify build

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,993 @@
...
+
+    ```bash
+    terraform refresh
+    terraform apply
+    ```
+
+    After running the above commands, you'll get the Pipeline cloud rule entity ID for each drop rule. Keep the IDs handy for the next step.
+
+    ```hcl
+    # Example: Get the entity ID for a drop rule
```
pranav-new-relic:
```suggestion
    # Example: Get the ID of the Pipeline Cloud Rule counterpart of the Drop Rule, using `pipeline_cloud_rule_entity_id`
```

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,993 @@
...
+    Use the `terraform import` command with the Pipeline cloud rule entity IDs obtained in the previous step to import each rule as a `newrelic_pipeline_cloud_rule` resource.
+
+    ```bash
+    terraform import newrelic_pipeline_cloud_rule.foo <pipeline_cloud_rule_entity_id>
+    ```
+
+    <Callout title="Note">
+      The ID of a Pipeline cloud rule is not the same as the ID of the corresponding NRQL drop rule. You must use the `pipeline_cloud_rule_entity_id` attribute exported by the drop rule resource to get the correct Pipeline cloud rule ID.
+
+      You can pair the import with `terraform plan -generate-config-out` to automatically generate the corresponding resource configuration for the imported Pipeline cloud rules.
```
pranav-new-relic: @tejaswaroop-nr we might want to keep this outside the note, just so customers know that the preferred approach involves "importing" the resource and then generating it using `terraform plan -generate-config-out` 🤔

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,993 @@
...
+            ``` */}
+          </Step>
+
+          <Step>
+            #### Apply the changes [#apply-changes]
+
+            Once you've reviewed the plan and confirmed it looks correct, comment on your pull request:
+
+            ```
+            Terraform apply
```
pranav-new-relic:
```suggestion
            terraform apply
```

src/content/docs/infrastructure-as-code/terraform/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,993 @@
...
+  <Step>
+    #### Remove the old drop rule resources from Terraform state [#remove-old-drop-rules]
+
+    After successfully importing the Pipeline cloud rules, you should remove all references to `newrelic_nrql_drop_rule` resources from your Terraform state using the `terraform state rm` command.
+
+    ```bash
+    terraform state rm newrelic_nrql_drop_rule.foo
+    ```
+
+    This removes the drop rules from Terraform state without deleting them from New Relic.
```
pranav-new-relic, line 113: At this point, we might want to drop a tiny reference to the `removed{}` block in Terraform, as a configuration-internal alternative to running `terraform state rm` on the drop rule resources 

https://developer.hashicorp.com/terraform/language/block/removed#lifecycle

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,993 @@
...
+    terraform state rm newrelic_nrql_drop_rule.foo
+    ```
+
+    This removes the drop rules from Terraform state without deleting them from New Relic.
+  </Step>
+
+  <Step>
+    #### Clean up your Terraform configuration [#clean-up-terraform-configuration]
+
+    Run the `terraform plan` command to remove the old `newrelic_nrql_drop_rule` resource definitions from your Terraform configuration files and verify the migration.
```
pranav-new-relic: We'll need to correct this by a bit - since it currently looks like the recommendation is to perform a `terraform plan` to clean up existing Drop Rule configuration; however, let's structure this to make clear that the customer (after the above step) will need to clean up/comment all `newrelic_nrql_drop_rule` resources in their configuration (and any auxiliary resources depending on these). Post this, they would need to run a `terraform plan`, and if it says "no changes", we should be good

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
pranav-new-relic: + let's make this look like (include commands from) Step 3 in this guide
https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/guides/drop_rules_eol_guide#manual-migration-process-overview

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,993 @@
...
+<Steps>
+  <Step>
+    #### Get the Pipeline cloud rule IDs [#get-pipeline-cloud-rule-ids]
+
+    Upgrade the New Relic Terraform Provider to v3.68.0 or greater, then refresh and apply your configuration with the existing `newrelic_nrql_drop_rule` resources. This allows the `pipeline_cloud_rule_entity_id` attribute to be exported by each drop rule resource.
+
+    ```bash
+    terraform refresh
+    terraform apply
+    ```
```
pranav-new-relic: As discussed, let's have this segment look like Step 2 here: https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/guides/drop_rules_eol_guide#manual-migration-process-overview

(so we can talk about `terraform apply -refresh-only`, `terraform apply -refresh-only -target=newrelic_nrql_drop_rule.foo`, etc)

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,993 @@
...
+            In your CI/CD environment or locally, run the `terraform state rm` commands for each old drop rule:
+
+            ```bash
+            terraform state rm newrelic_nrql_drop_rule.drop_debug_logs
+            terraform state rm newrelic_nrql_drop_rule.drop_health_checks
+            terraform state rm newrelic_nrql_drop_rule.drop_pii_data
+            ```
+
+            <Callout variant="Note">
+              These commands remove the resources from Terraform state without deleting them from New Relic. The old drop rules remain in New Relic but are no longer managed by Terraform.
```
pranav-new-relic: "The old drop rules remain in New Relic but are no longer managed by Terraform."

As discussed, we'll need to make the first part of this a li'l more clear, to not let customers assume that drop rules will not cease to exist after January 7th - more along the lines of "they'll be eventually removed in the backend at the EOL"

tejaswaroop-nr:
Netlify build

src/content/docs/infrastructure-as-code/terraform/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,922 @@
+---
+title: Migrate from drop rules to Pipeline cloud rules
```
pranav-new-relic, line 2: @tejaswaroop-nr A tiny nitpick: I guess the title should be something like "Migrate from drop rules to Pipeline cloud rules in Terraform", since the current title, if found by someone browsing through the docs would be misleading; causing customers (who have never managed Drop Rules via Terraform) to assume that they would actually need to "migrate" their drop rules to Pipeline cloud rules; though the "migration" on the "backend" is already done for such customers, meaning they'd need to do nothing.

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,922 @@
...
+freshnessValidatedDate: never
+---
+
+<Callout variant="important">
+As of May 21, 2025, new customers can no longer use NRQL drop rules. [NRQL drop rules will be end-of-life](/eol/2025/05/drop-rule-filter) on January 7, 2026.
+
+Migrate to [Pipeline cloud rules](/docs/new-relic-control/pipeline-control/cloud-rules-api) to continue managing your data dropping rules.
+</Callout>
+
+Terraform users must manually migrate their `newrelic_nrql_drop_rule` resources to `newrelic_pipeline_cloud_rule` resources. Although New Relic has already migrated the underlying drop rules to Pipeline cloud rules in the backend, Terraform maintains its own state file and only knows about the resources you explicitly configured. Since Terraform treats these as two completely different resource types, you'll need to import the Pipeline cloud rules into Terraform's state and remove the old drop rule resources.
```
pranav-new-relic: About

"Terraform users must manually migrate their `newrelic_nrql_drop_rule` resources to `newrelic_pipeline_cloud_rule` resources"

I'd like to get your opinion on the following (not a mere suggestion, I'd genuinely like to hear your thoughts, and then, we can correct this if these questions make sense) -
- instead of "Terraform users must manually migrate their `newrelic_nrql_drop_rule` resources to `newrelic_pipeline_cloud_rule` resources", should we instead make this a li'l more clear; something along the lines of "Users consuming NRQL drop rules via the Terraform resource `newrelic_nrql_drop_rule` must manually migrate their...."? Maybe this is a bit too elaborate (and not needed, maybe?) but I wanted to make sure I get your opinion on this :) 
- Wherever we reference `newrelic_nrql_drop_rule` and `newrelic_pipeline_cloud_rule` in this paragraph (or rather, in this entire document, if this proposal makes sense): I opine that it might be useful to add a hyperlink to https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/nrql_drop_rule and https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/pipeline_cloud_rule respectively; do you think this could be useful

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,922 @@
...
+
+## Migration approaches [#migration-approaches]
+
+This guide covers two ways to migrate your NRQL drop rules to Pipeline cloud rules in Terraform:
+
+1. Native Terraform commands: The standard Terraform workflow using `terraform import`, `terraform plan`, and `terraform state rm` commands.
+2. New Relic CLI automation tools: New Relic provides CLI tools that automate the native Terraform workflow, available in two versions:
+   - GitOps (CI/CD): For teams using CI/CD tools.
+   - Local Terraform: For teams executing Terraform commands on their local machines.
```
pranav-new-relic: Again, here's an ask for your opinion on my suggestion, @tejaswaroop-nr :)
Would it make sense to make this a little more elaborate, just so the customer knows most parameters needed to pick the right tool 🤔 

(something along the lines of)

```suggestion
   - GitOps (CI/CD): For teams using CI/CD tools, i.e., Terraform tooling in which the state file is not readily accessible within workspaces, and is often, hosted in an undisclosed remote location 
   - Local Terraform: For teams executing Terraform commands on their local machines, or any sort of Terraform tooling where the state file is readily accessible within workspaces, allowing resources in the state to be read
```

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,922 @@
...
+  - `NEW_RELIC_REGION` (optional) - Set to 'US' or 'EU' based on your account region (defaults to 'US').
+
+### Choose your approach [#choose-approach]
+
+<CollapserGroup>
+  <Collapser
+    id="gitops-cicd"
+    title="Using GitOps (CI/CD)"
+  >
+    This is a three-phase migration process for CI/CD environments.
```
pranav-new-relic: @tejaswaroop-nr I think it might be useful to share a very brief gist of the workflow this process would follow, alongside this existing line; the gist could be a combination of things you've documented under each phase (or) can be something [similar to the paragraph under this header in the guide we have in the registry](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/guides/drop_rules_eol_guide#automated-migration-for-cicd-environments) - let me know your thoughts about this

src/content/docs/infrastructure-as-code/terraform/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,922 @@
...
+          </Step>
+        </Steps>
+      </Collapser>
+    </CollapserGroup>
+  </Collapser>
+
+  <Collapser
+    id="local-terraform"
+    title="Using local Terraform"
+  >
```
pranav-new-relic, line 763: Similar to this comment under the CI/CD section: https://github.com/newrelic/docs-website/pull/21857/files#r2493310120

@tejaswaroop-nr I think it might be useful to share a very brief gist of the workflow this process would follow, alongside this existing line; the gist could be a combination of things you've documented under each step (or) can be something [similar to the paragraph under this header in the guide we have in the registry](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/guides/drop_rules_eol_guide#interactive-automated-migration-for-local-terraform-workspaces) - let me know your thoughts about this

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,922 @@
...
+
+### Old drop rules still showing in Terraform state
+
+If `terraform state list` still shows `newrelic_nrql_drop_rule` resources after migration:
+
+**Solution:** Use the `terraform state rm` command to manually remove them from state without deleting them from New Relic. For example:
+```bash
+terraform state rm newrelic_nrql_drop_rule.rule_name
+```
+Follow the recommendations provided in the CLI output from Phase 2 for the complete list of resources to remove. */}
```
pranav-new-relic: @tejaswaroop-nr this sounds like a comment you've added, so this can be added to a different guide in future; I don't think we need to document this for inclusion somewhere else, as I would suggest adding a "References" section that would lead to
- https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/guides/drop_rules_eol_guide#nrql-drop-rules-eol-implications-and-actions-needed
- https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/pipeline_cloud_rule
- https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/nrql_drop_rule
- https://github.com/newrelic/terraform-provider-newrelic/tree/main/examples/drop_rule_migration_ci#troubleshooting
- https://github.com/newrelic/newrelic-cli/blob/main/internal/migrate/tf_importgen_ci_guide.md
- https://github.com/newrelic/newrelic-cli/blob/main/internal/migrate/tf_importgen_guide.md

with each of these shown with relevant labels such as Troubleshooting for the CI workflow, Troubleshooting for the local workflow, etc - do you think this would make sense

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,922 @@
...
+We recommend using New Relic's automation tools as they significantly reduce manual work and potential for errors. However, if you prefer complete control or have specific requirements, you can follow the native Terraform command approach.
+
+## Migrate with native Terraform commands [#native-terraform]
+
+This section describes the standard Terraform workflow for migrating from `newrelic_nrql_drop_rule` resources to `newrelic_pipeline_cloud_rule` resources using native Terraform commands.
+
+### Prerequisites [#native-prerequisites]
+
+- **Terraform or OpenTofu v1.5 or above:** This version introduced the `import` block feature, which makes the migration process more efficient by eliminating the need to manually write resource configurations.
+- **New Relic Terraform Provider v3.73.0 or above:** This version adds support for the `pipeline_cloud_rule_entity_id` attribute on `newrelic_nrql_drop_rule` resources and introduces the `newrelic_pipeline_cloud_rule` resource.
```
pranav-new-relic: @tejaswaroop-nr an ask I'd like to get your thoughts on: like we discussed last time,
- `3.68.0` is the version with which the resource `newrelic_pipeline_cloud_rule` was introduced, and the `newrelic_nrql_drop_rule` resource was officially deprecated in the New Relic Terraform Provider
- `3.73.0` is the version which we'd _like_ customers to use who would start with Pipeline Cloud Rules - though `3.68.0` should work too, `3.73.0` includes some fixes we've made to ensure a better performance

Thank you for changing all references of `3.68.0` to `3.73.0` based on this conversation we've had previously - I wonder if we can change the "version adds support for" here, to something like "extends support for the ..... added in a previous version released in September 2025", so we don't let the customer feel like a **very** recent, 3.73.0 version offers support for Pipeline Cloud Rules (meaning they've been given this version very late), but we want them to use 3.73.0 and not 3.68.0 instead - what do you think 🤔

src/content/docs/infrastructure-as-code/terraform/migrate-drop-rules-to-pipeline-cloud-rules.mdx
tejaswaroop-nr: @pranav-new-relic, This document will be housed within the "Terraform guides" section of our docs site, I feel that the readers who land on it will have that context that it's specific to Terraform users.

That said, I'll also be updating the first line of the document to make it even clearer that this guide is specifically for Terraform users who need to manually migrate their resources.

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
tejaswaroop-nr: **First point:** Thanks for pointing this, I agree the first sentence sounds rather weak. I hope I've made it clear (in the recent commit) so it reads as specific to Terraform users only.

**Second point:** I've hyperlinked the resource names in the first paragraph. My opinion is that hyperlinking them everywhere would prompt users to constantly open links and it becomes a distraction rather than allowing them to read and follow through the guide.

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
tejaswaroop-nr: @pranav-new-relic Thanks for the suggestion! My opinion is this:

Yes, even though experienced users might implicitly know these details, I think adding the technical constraint serves explaining the "why" behind two separate tools. The natural user question would be: "Why does the tool need two separate versions? Why can't one tool handle both workflows?" I believe the state accessibility explanation answers this clearly. The two tools exist because they handle different constraints on retrieving state data.

- **CI/CD tool:** Cannot directly access the state file, so it requires a multi-phase process.
- **Local tool:** Can directly read the Terraform state file to get the necessary Pipeline cloud rule IDs.

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
tejaswaroop-nr: @pranav-new-relic I agree. The current wording could make it feel like Pipeline cloud rule support was added very recently in v3.73.0. I'll update it to the following:

- **New Relic Terraform Provider v3.73.0 or above:** This version extends support for Pipeline cloud rules and the `pipeline_cloud_rule_entity_id` attribute on `newrelic_nrql_drop_rule` resources (added in v3.68.0 released in September 2025) with important bug fixes to ensure a smooth migration process.

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
tejaswaroop-nr: @pranav-new-relic while I believe that the collapser headers effectively list the steps, I also believe that providing a brief overview at the start of each automation section will improve the reader's experience.

Although this risks repetition, which we usually avoid in documentation, I believe the repetition is necessary and beneficial for two reasons:

- Context reinforcement: Many might readers will skip directly to their relevant automation section (Local or CI/CD). An upfront summary ensures they immediately understand the flow and purpose of that specific tool without needing to expand the collapsers or read the entire guide first.

- Justifying tool design: The core difference between the Local and CI/CD tools is the accessibility of the Terraform state file. The brief summary states this technical constraint, which validates why the CLI offers two separate tools and helps the user confirm they are using the correct one for their environment.

src/content/docs/infrastructure-as-code/terraform/migrate-drop-rules-to-pipeline-cloud-rules.mdx
tejaswaroop-nr: @pranav-new-relic I'll make the required changes here as well.

tejaswaroop-nr:
Netlify build

src/content/docs/data-apis/manage-data/migrate-drop-rules-to-pipeline-cloud-rules.mdx
tejaswaroop-nr: @pranav-new-relic, as discussed, I've added text under both the automated approaches and hyperlinked it to cross-reference the troubleshooting steps covered in your Github files.

src/content/docs/infrastructure-as-code/terraform/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,904 @@
...
+</Steps>
+
+## Migrate with New Relic's automation tools [#automation-tools]
+
+To simplify the native Terraform migration workflow described above, New Relic provides CLI automation tools that handle the import and state management for you. These tools automate the same underlying Terraform operations (import, plan, apply, and state removal) but with significantly less manual work.
+
+### Prerequisites [#automation-prerequisites]
+
+- **Terraform or OpenTofu v1.5 or above:** Required for the automated migration approaches.
+- **New Relic Terraform Provider v3.73.0 or above:** This version adds support for the `pipeline_cloud_rule_entity_id` attribute and the `newrelic_pipeline_cloud_rule` resource.
```
pranav-new-relic: @tejaswaroop-nr a tiny nitpick: we might need to change this too, to make it similar to the change we've added above as well

https://github.com/newrelic/docs-website/pull/21857/files#diff-0fd20000d752ff574a13ff4250f780ed69b3d2bd0e90e28d4da83fc62f0033bcR40

pranav-new-relic: Also, with regard to the line we've edited above

- **New Relic Terraform Provider v3.73.0 or above:** This version extends support for Pipeline cloud rules and the `pipeline_cloud_rule_entity_id` attribute on `newrelic_nrql_drop_rule` resources (added in v3.68.0 released in September 2025) with important bug fixes to ensure a smooth migration process.

do you think it would be a bit more clear (or would it be redundant, instead) if we were to specify that the version 

"adds support to manage Pipeline cloud rules via Terraform using the new `newrelic_pipeline_cloud_rule` resource, and the ...." 

as opposed to

"extends support for Pipeline cloud rules and the.."

since the addition of the new `newrelic_pipeline_cloud_rule` resource is kinda the primary feature of this new version released, and the addition of a `pipeline_cloud_rule_entity_id` attribute to the resource `newrelic_nrql_drop_rule` is more of a consequential change caused by the addition of the new Pipeline cloud rule resource?

I'm sorry to have spotted this so late (should've done this earlier :( ) - but yeah, please feel free to let me know if this is a change that's unnecessary from a docs semantics/comprehensibility standpoint.

pranav-new-relic (commented):
~

src/content/docs/infrastructure-as-code/terraform/migrate-drop-rules-to-pipeline-cloud-rules.mdx
tejaswaroop-nr: Hey @pranav-new-relic, good catch! I agree with line #140 matching line #40.

Also, as per my understanding and please correct me if I'm wrong, with v3.68.0 we've added the `newrelic_pipeline_cloud_rule` resource and the `pipeline_cloud_rule_entity_id` attribute on `newrelic_nrql_drop_rule` resources. I believe with v3.73.0 we've extended/improved with bug fixes.

So how about we add this:

- **New Relic Terraform Provider v3.73.0 or above:** This recommended version extends support for managing Pipeline cloud rules via Terraform using the `newrelic_pipeline_cloud_rule` resource and the `pipeline_cloud_rule_entity_id` attribute on `newrelic_nrql_drop_rule` resources (added in v3.68.0 released in September 2025) with important bug fixes to ensure a smooth migration process.

pranav-new-relic: Yup, this makes complete sense. Thank you @tejaswaroop-nr :)

src/content/docs/infrastructure-as-code/terraform/migrate-drop-rules-to-pipeline-cloud-rules.mdx
```diff
@@ -0,0 +1,907 @@
...
+Migrate to [Pipeline cloud rules](/docs/new-relic-control/pipeline-control/cloud-rules-api) to continue managing your data dropping rules.
+</Callout>
+
+If you're managing your NRQL drop rules through Terraform using the [`newrelic_nrql_drop_rule`](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/nrql_drop_rule) resource, you'll need to migrate them to the [`newrelic_pipeline_cloud_rule`](https://registry.terraform.io/providers/newrelic/newrelic/latest/docs/resources/pipeline_cloud_rule) resource. Although New Relic already migrated the underlying drop rules, Terraform maintains its own state file and only knows about the resources you explicitly configured. Since Terraform treats these as two completely different resource types, you'll need to import the Pipeline cloud rules into Terraform's state and remove the old drop rule resources.
+
+## Migration approaches [#migration-approaches]
+
+This guide covers two ways to migrate your NRQL drop rules to Pipeline cloud rules in Terraform:
+
+1. Native Terraform commands: The standard Terraform workflow using `terraform import`, `terraform plan`, and `terraform state rm` commands.
```
vpayyapilly: We use numbered lists only when there is a sequence. When it is an unordered list, use bullets.

vpayyapilly (approved):

PR #21886, opened 2025-10-09
fix: update with latest changes to actions and APIs

Julien4218:
Change list:

- Updated action names with current naming, for example `utils.http.post` is now `http.post`
- Also updated newrelic.ingest.sendLogs with new input format
- Replace old naming of `autoflows` with `workflowAutomation`
- Remove references to timezone api, and replace with simpler example using NewRelic status API and `Hello World` workflow
- Update APIs with latest schema (dropping reference to accountId in some APIs)


cbehera-newrelic (approved):
lgtm

PR #21934, opened 2025-10-16
chore: Update CodeStream licensing and third party notices information directly in docs

CalvinAllen:
The CodeStream code repository is private, so we can no longer reference it via public links and instead need to embed the content directly for the license and third party notices.


adutta-newrelic:
netlify build

PR #21967, opened 2025-10-21
Revise job manager configuration equations and terms

keegoid-nr:
Updated equations and variables for calculating required CPU cores and job metrics in synthetic monitoring documentation.

- unify duration variables and specify units as m or s for minutes or seconds
- clarify rates versus counts with R_ or N_
- standardizing LaTeX using `$\mathrm{...}$` (roman font) for multi-letter variable names like "Parallelism" to distinguish them from single-letter variables.
- use `period` instead of `frequency`
- add to description for ping runtime scaling


WriteMayur:
netlify build fork

PR #21982, opened 2025-10-23
chore(ruby-agent): Add a section for log frameworks

kaylareopelle:
Previously, it was difficult to tell what logging libraries had APM logs in context support. Now, it has its own table. We plan to add more logging libraries this quarter.


kaylareopelle:
moving back to a draft so I can preview locally tomorrow

WriteMayur:
netlify build fork

PR #22071, opened 2025-11-03
fix: build errors

akashreddy-nr:
Released date is specified as '2025-06-31' June 31st doesn't exist 

hence the build is failing


akashreddy-nr:
netlify build

PR #22108, opened 2025-11-05
Update netty release notes for 8.20/8.21

obenkenobi:
## Give us some context

* A fix in Java Agent in 8.20 is bugged and was corrected in 8.21. The release notes is updated so the messaging reflects that. My intention is to strikethrough part of the 8.20 release and add a Callout section warning customers about it.


adutta-newrelic (approved):
LGTM

PR #22133, opened 2025-11-07
chore: test PR, build break test with 31st June: DO NOT MERGE

pranav-new-relic:
⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️
⚠️⚠️ DO NOT MERGE ⚠️⚠️
⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️

PR #22137, opened 2025-11-07
chore(workflows): enhance netlify build CI workflow to reflect status of triggered builds and furnish build details

pranav-new-relic:
### ✨ Summary

Enhances the Netlify deployment preview workflow with deployment tracking, and comprehensive error handling.

#### Why do (I think) we need this? 

The current workflow that is triggered upon commenting `netlify build` on the PR only "triggers" a build on Netlify, but does not report the status of the build back - implying that the CI check triggered by the `netlify build` ends at triggering the build, and succeeds all the time. This leads to an incomplete check, and the onus is on the reviewers of the PR to go to Netlify to find the status of the build, as a preview link is returned by the current workflow if it is successful; else, it returns nothing.

While this is the current process, this leaves a lot of room for erroenous considerations to be made; it is highly probable that 
- a reviewer gets the PR accidentally merged if they miss checking Netlify to find the latest build status on the PR, or 
- a reviewer goes by an optimal scenario in which the build worked many commits ago, after which a bunch of trivial commits were added, causing a notion that a re-check is not needed (though it could be likely that the smallest of discrepancies in the latest commits could be leading to a build failure - the example with the 31st June case :)) 

and all of this did occur multiple times in the past; neither does the CI check fail (it succeeds all the time) so we do not obstruct a merge, nor do we notify the maintainers in the PR's comments that the build is screwed - leading to the check getting ignored at times, causing a major debacle later on in the daily release PRs.

Obviously, all of this hassle is now seen because we no longer use Netlify's Deploy Preview feature (which would handle the entire process of building on a commit, and notifying maintainers) - and when we do not do so, the only way to get the status of a build, is to:
- run a build in the CI's environment, i.e. instead of using Netlify's runtime, build Netlify here in the GitHub CI's runners (modify the workflow to do so), and then, run the exact same build directly in the runner
- continue using the approach in the existing workflow (to trigger a build on Netlify, via the Netlify API), and then, keep polling to map ongoing builds associated with the current branch to the commit using which the build was triggered - and continue to keep track of the build's status to keep reporting the latest on the build to the PR (and affect the status of the CI check accordingly).

I did try (1) with #22133, only to realise that GitHub CI's runners are too lightweight to run a heavy job like scaffolding Netlify and dependencies via npm, and then run a build - most builds kept running for over 1hr 50 minutes and eventually timed out; implying, this method is infeasible - so (2) seemed to be the only route. This is the approach proposed by the workflow modified in this PR.

In simpler words, the workflow's trigger remains the same - a `netlify build` comment, and the immediate step after this remains the same too - triggering a build on Netlify using the Netlify API. Post this, a bunch of additional steps have been added. 
- Netlify's API is polled to find out all ongoing builds associated with the current branch, and these are filtered to find the build associated with the latest commit on the branch (after which the `netlify build` comment was made to trigger the build) - after which a build is identified, and this build is monitored by the workflow. 
- As the build is kept track of, updates are posted as PR comments. Firstly, after the build is found, details of the deployment on Netlify (URL, etc) are posted to the PR comment; and similar updates are made to the same comment when it is in progress. Based on the status of the build towards the end - if the build is successful, the CI check succeeds and a comment is added confirming the success status; but if it fails, the CI check fails, and a similar comment is added in vibrant red to let the maintainer know that the PR should not be merged.
A few other changes do exist too, e.g. checking for any existing builds on the branch to prevent "concurrent" builds running (to save resources), using branch based filtering if a commit-based build is not found, checking if the build times out and adding comments accordingly, checking for credential expiry, and a bunch of edge cases I could think of.

The success comment [looks like this](https://github.com/newrelic/docs-website/pull/22137#issuecomment-3577357596).
The failure comment [looks like this](https://github.com/newrelic/docs-website/pull/22137#issuecomment-3577019781).
Upon commenting `netlify build` , if it is found that there is a build associated with this branch that is already in progress, [this is the comment shown](https://github.com/newrelic/docs-website/pull/22137#issuecomment-3577319541).
### Netlify Workflow Improvements

**Deployment Discovery & Monitoring**
- Commit-based deployment matching with branch-based fallback
- 45-second initial wait, then 10 polling attempts with 15s intervals
- Status monitoring polls every 90s for up to 30 minutes
- Real-time PR comment updates with IST timestamps

**Outcome Handling**
- **Success**: Posts preview URL, a link to the deployment on Netlify and other details
- **Failure**: Posts critical DO-NOT-MERGE warning with brief error details and a link to the failing deployment on Netlify
- **Timeout**: Provides manual check instructions

**Pre-deployment Validation**
- Netlify API token validation before triggering builds
- Concurrent deployment detection prevents conflicts
- Posts actionable error comments for token expiration or ongoing builds


WriteMayur:
@pranav-new-relic as this is a test PR can we close this?

PR #22157, opened 2025-11-11
Update Java agent configuration with new options

jtduffy:
Added configuration options for datastore multihost preference and exec call SQL regex disabling.


adutta-newrelic (approved):
LGTM

PR #22182, opened 2025-11-12
Php/update trixie apt installation

mfulb:
The PHP agent team needs to update the installation instructions for Debian Trixie when using `apt`.


adutta-newrelic (approved):
LGTM

PR #22234, opened 2025-11-18
chore(apm/php agent): PHPUnit updates

lavarou:
As of 9/30/2025 New Relic PHP Agent does not support any version of PHPUnit package. Therefore PHPUnit needs to be removed from list of supported libraries available on New Relic PHP Agent's compatibility page. Additionally all documentation pertaining how PHPUnit support looked like has to be removed, and information about troubleshooting the incompatibility between New Relic PHP Agent and PHPUnit version >=11 must be moved to a dedicated troubleshooting document.

### DIFF
--- a/src/content/docs/apm/agents/php-agent/troubleshooting/phpunit-incompatibility.mdx
+++ b/src/content/docs/apm/agents/php-agent/troubleshooting/phpunit-incompatibility.mdx
@@ -11,23 +11,35 @@ freshnessValidatedDate: never
 
 ## Problem
 
-When you use [PHPUnit](https://phpunit.de/) version 11 or newer to manage and run your unit tests, and have New Relic PHP Agent installed and enabled, 
-executing `phpunit` script (the PHPUnit command-line test runner) to run unit tests will cause the script to use all memory available on the system,
-before actually running the test.
+When you use [PHPUnit](https://phpunit.de/) version 11 or newer to manage and run your unit tests, and have New Relic PHP agent installed and enabled,
+executing the `phpunit` script (the PHPUnit command-line test runner) will cause it to consume all available system memory before actually running any tests.
 
 ## Solution
 
-To prevent out-of-memory errors you must <DNT>**explicitly disable**</DNT> the agent by setting `newrelic.enabled` to `false`. You can disable the it while using `phpunit` like this:
+To prevent out-of-memory errors you must <DNT>**explicitly disable**</DNT> the agent by setting `newrelic.enabled` to `false`. You can disable it while using `phpunit` like this:
+```bash
+php -d newrelic.enabled=false vendor/bin/phpunit tests/
 ```
-php -d newrelic.enabled=false phpunit ...remaining arguments...
+
+For example, to run a specific test file:
+```bash
+php -d newrelic.enabled=false vendor/bin/phpunit tests/Unit/ExampleTest.php
 ```
-This workaround disables all instrumentation of unit tests.
+
+<Callout variant="important">
+This workaround disables all New Relic instrumentation while running unit tests, meaning no APM data will be collected during test execution.
+</Callout>
+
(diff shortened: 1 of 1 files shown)


src/content/docs/apm/agents/php-agent/troubleshooting/phpunit-incompatibility.mdx
```diff
@@ -0,0 +1,33 @@
...
+  - PHP agent
+  - Troubleshooting
+metaDescription: Troubleshooting PHPUnit out-of-memory errors
+freshnessValidatedDate: never
+---
+
+## Problem
+
+When you use [PHPUnit](https://phpunit.de/) version 11 or newer to manage and run your unit tests, and have New Relic PHP Agent installed and enabled, 
+executing `phpunit` script (the PHPUnit command-line test runner) to run the unit test will cause the script to use all memory available on the system,
```
zsistla:
```suggestion
executing `phpunit` script (the PHPUnit command-line test runner) to run unit tests will cause the script to use all memory available on the system,
```

src/content/docs/apm/agents/php-agent/troubleshooting/phpunit-incompatibility.mdx
```diff
@@ -0,0 +1,33 @@
...
+```
+This workaround disables all instrumentation of unit tests.
+
+## Cause
+
+The reason for this incompatibility is the new code that was added in PHPUnit 11.x: 
+[Reapply "Check and restore error/exception global handlers" · sebastianbergmann/phpunit@0214cf8](https://github.com/sebastianbergmann/phpunit/commit/0214cf81d8b12a1191932e6096d22c8496943911).
+This new code retrieves the list of active exception handlers using a method that requires popping each excaption handler off the exception handlers' stack.
+However, the default behavior of New Relic PHP Agent is to install its own exception handler and not allowing it to ever being removed from the exception handlers' stack -
+New Relic PHP Agent detects that its exception handler is being removed from the exception handlers' stack and puts it back on. This causes infinite loop in `phpunit` script:
```
zsistla:
```suggestion
New Relic PHP Agent detects that its exception handler is being removed from the exception handler stack and puts it back on. This causes an infinite loop in `phpunit` script:
```

src/content/docs/apm/agents/php-agent/troubleshooting/phpunit-incompatibility.mdx
```diff
@@ -0,0 +1,33 @@
...
+```
+php -d newrelic.enabled=false phpunit ...remaining arguments...
+```
+This workaround disables all instrumentation of unit tests.
+
+## Cause
+
+The reason for this incompatibility is the new code that was added in PHPUnit 11.x: 
+[Reapply "Check and restore error/exception global handlers" · sebastianbergmann/phpunit@0214cf8](https://github.com/sebastianbergmann/phpunit/commit/0214cf81d8b12a1191932e6096d22c8496943911).
+This new code retrieves the list of active exception handlers using a method that requires popping each excaption handler off the exception handlers' stack.
```
zsistla:
```suggestion
This new code retrieves the list of active exception handlers using a method that requires popping each exception handler off the exception handler stack.
```

src/content/docs/apm/agents/php-agent/troubleshooting/phpunit-incompatibility.mdx
```diff
@@ -0,0 +1,33 @@
...
+However, the default behavior of New Relic PHP Agent is to install its own exception handler and not allowing it to ever being removed from the exception handlers' stack -
```
zsistla:
```suggestion
However, the default behavior of New Relic PHP Agent is to install its own exception handler and then prevent it from being removed from the exception handler stack. 
```

PR #22258, opened 2025-11-21
fix: Incorrect Link to Before You Begin Section

amayer5125:
Fixes links to an anchor that was changed from `#prereq` to `#before-you-begin` in d451c5c27fbc2885e31ddcce0e75999ffe8e356b.


sujitnewrelic (approved):
LGTM

PR #22317, opened 2025-12-01
docs(dotnet): Update latest supported library versions for .NET agent

tippmar-nr:
Updates latest supported versions for libraries instrumented by the .NET agent. 

Also updates some references to .NET 10 that were missed in a previous PR.


WriteMayur (approved):
Thanks for making the updates

PR #22331, opened 2025-12-03
Add note about hash collisions for uniqueCount below 256

mgmur:

## Give us some context

Add a brief note about the potential for hash collisions when using uniqueCount() nrql function for less than 256 values


src/content/docs/nrql/nrql-syntax-clauses-functions.mdx
```diff
@@ -2120,6 +2120,8 @@ SELECT histogram(duration, 10, 20) FROM PageView SINCE 1 week ago
...
 
+    **How exact counting works under 256 values**: When counting fewer than 256 unique string values, `uniqueCount()` uses a hash-based counting method to determine uniqueness. This method generates a 32-bit integer hash code from the characters in each string. While this approach efficiently distributes hash values for most strings, hash collisions can occasionally occur—meaning two different strings may generate the same hash code. This is more likely to happen when counting very short strings (like single characters) or strings that are extremely similar to each other. In rare cases where a hash collision occurs, the function may slightly undercount the true number of unique values.
```
WriteMayur:
```suggestion
**How exact counting works under 256 values:** When counting fewer than 256 unique string values, `uniqueCount()` uses a hash-based counting method to determine uniqueness. This method generates a 32-bit integer hash code from the characters in each string.This approach efficiently distributes hash values for most strings. However, a hash collision can occasionally occur, meaning two different strings may generate the same hash code. This is more likely to happen when counting very short strings (such as single characters) or strings that are extremely similar to each other. In these rare cases, the function may slightly undercount the true number of unique values.
```

src/content/docs/nrql/nrql-syntax-clauses-functions.mdx
```diff
@@ -2120,6 +2120,8 @@ SELECT histogram(duration, 10, 20) FROM PageView SINCE 1 week ago
...
 
+    **How exact counting works under 256 values**: When counting fewer than 256 unique string values, `uniqueCount()` uses a hash-based counting method to determine uniqueness. This method generates a 32-bit integer hash code from the characters in each string. While this approach efficiently distributes hash values for most strings, hash collisions can occasionally occur—meaning two different strings may generate the same hash code. This is more likely to happen when counting strings that are extremely similar to each other. In rare cases where a hash collision occurs, the function may slightly undercount the true number of unique values.
```
amohammed-2: Nit: Should we wrap this with a `<Callout variant="tip" title="Note">` ?

mgmur: Good catch, fixed in [f9d8341](https://github.com/newrelic/docs-website/pull/22331/commits/f9d83417327a9d6a8b48380c02ea76192920be3e)

PR #22333, opened 2025-12-03
chore: Added AI monitoring namespace retention values

Tarunuppu:
## Give us some context

    Added AI Monitoring namespace retention values in the table


WriteMayur:
netlify build fork

PR #22394, opened 2025-12-09
chore[Alert]: Updated the facet limit number for an alert condition to 20K

sujitnewrelic:
Updated the facet limit number for an alert condition to 20K From PM: https://newrelic.slack.com/archives/D084A2ND36J/p1765218389176169 From #wizards-alerts-incident-management: https://newrelic.slack.com/archives/CTMRYF51R/p1765146625672319


PallaviWrite (approved):
LGTM

PR #22411, opened 2025-12-10
chor: Update browser-app-settings-page.mdx

WriteMayur:
This PR deletes the following line from the doc:

We strongly encourage all users to adopt this new feature


adutta-newrelic (approved):
LGTM

PR #22447, opened 2025-12-11
changed 10e9 to 10e8

PallaviWrite:
Fixed typo in usage query for GB conversion issue: Customer reported that the NRQL query for converting bytes to GB was using 10e9 (which equals 10 GB) instead of 1e9 or 10e8 (which both equal 1 GB). This caused the query to return the number of 10 GB units rather than GB. 

**Resolution**: Updated the divisor from 10e9 to 10e8 in the bytecountestimate queries to correctly convert bytes to gigabytes. 

Source: Customer feedback via @XiXiaPdx, who explained the math:
1e9 = 1,000,000,000 = 1 GB ✓
10e8 = 10 × 100,000,000 = 1,000,000,000 = 1 GB ✓
10e9 = 10,000,000,000 = 10 GB ✗


PallaviWrite:
netlify build

PR #22468, opened 2025-12-12
chore(PHP): Add compatibility for Drupal 11.1 and 11.2

mfulb:
The PHP team would like to update its compatibility page to include support for Drupal 11.1 and 11.2.


tejaswaroop-nr:
Netlify build fork

PR #22491, opened 2025-12-15
correction to env var

keegoid-nr:
Package reporting should be disabled for Lambda monitoring to improve cold start times.


tejaswaroop-nr:
Netlify build fork

PR #22619, opened 2026-01-05
Add NR_TAGS alternative

keegoid-nr:
Added instruction to set NR_TAGS environment variable for tags synchronization if not using a cloud integration.


sujitnewrelic (approved):
LGTM

PR #22635, opened 2026-01-07
doc update[Alerts]: Add clarification for OR operator behavior in workflow filters

sujitnewrelic:
This update addresses a common customer confusion regarding how the OR operator works with exclusion filters (DOES_NOT_CONTAIN, DOES_NOT_EQUAL, IS_NOT) in workflow filter queries.

Added an important callout after step 4 that:
- Explains the counter-intuitive behavior of OR with exclusions
- Provides clear "Incorrect query" vs "Correct query" examples
- Shows why multiple exclusions should use AND instead of OR
- References ticket NR-481918 where this pattern was identified

Reference: https://new-relic.atlassian.net/browse/NR-481918?focusedCommentId=1468817


sujitnewrelic:
netlify build

PR #22663, opened 2026-01-08
right navigation menu fixes

keegoid-nr:
- Add explicit ID [#support-hours] to the `Support hours` heading. This aligns with the convention used in the rest of the file and ensures the TOC links correctly.
- Add a spacer element (empty div with height) at the end of the file. This is an attempt to allow the sections near the bottom (OpenTelemetry support, Support videos) to be scrolled to the top of the viewport properly, ensuring that clicking their TOC links highlights the correct section instead of defaulting to the very bottom (RSS).


sujitnewrelic:
netlify build fork

PR #22715, opened 2026-01-13
docs: .NET agent, document new application name exclusion config and env vars

nr-ahemsath:
The .NET agent has a new feature in version 10.48.0, which lets users configure a list of application names to ignore.

See https://github.com/newrelic/newrelic-dotnet-agent/pull/3387 for more details.


adutta-newrelic (approved):
LGTM

PR #22722, opened 2026-01-14
feat: Register docs

metal-messiah:
This adds documentation for a new product we are launching in the future.  We need this to be exposed on the preview domain for the docs site. TYIA

### DIFF
--- a/src/content/docs/browser/new-relic-browser/troubleshooting/micro-frontend-monitoring-limitations.mdx
+++ b/src/content/docs/browser/new-relic-browser/troubleshooting/micro-frontend-monitoring-limitations.mdx
@@ -113,7 +113,7 @@ Unlike full browser agents, micro frontend monitoring requires manual instrument
 
 Entity relationships have constraints:
 
-Entities and relationships are synthesized at data harvest time.  If your microfrontend registration does not harvest any data, it will not synthesize these primatives.
+Entities and relationships are automatically created only when data is collected.  Even after registration, if your micro frontend does not harvest any data to New Relic, no entity or relationship will be created.
 
 ---
 


src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
```diff
@@ -0,0 +1,113 @@
...
+tags:
+  - Browser
+  - Browser monitoring
+  - npm package
+  - Micro frontends
+metaDescription: "Learn how to use the RegisteredEntity interface to monitor individual micro frontends in your JavaScript applications."
+freshnessValidatedDate: never
+---
+
+Use the RegisteredEntity interface to monitor individual micro frontends within your JavaScript application. After installing the browser agent in your container application, each micro frontend can register itself to report data under its own browser entity in New Relic.
```
ErikGrijzen: Do people know what a "container application" is? Anyway, can't we just say here "web application"?

ErikGrijzen: Shouldn't this be something like:

> ... report data under its own browser micro frontend entity in New Relic.

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
```diff
@@ -0,0 +1,113 @@
...
+  If you haven't installed the browser agent in your container application yet, see [Install browser monitoring using NPM](/docs/browser/browser-monitoring/installation/install-browser-monitoring-agent/#npm) for complete installation instructions.
+</Callout>
+
+## How the RegisteredEntity class works [#how-it-works]
+
+The RegisteredEntity interface allows micro frontends to:
+
+1. Conduct a safety check to ensure the container agent exists on the page before trying to execute business logic
+2. Register with the main browser agent running in the container application
+3. Report data under their own browser entity in New Relic
```
ErikGrijzen: Here as well, should we say "browser micro frontend entity"?

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
```diff
@@ -0,0 +1,113 @@
...
+4. Maintain their own custom attributes and context
+5. Track lifecycle events and performance independently from the main page
+
+## Register a micro frontend [#register]
+
+Once the main browser agent is initialized in your container application, each micro frontend can register itself using the `register()` method:
```
ErikGrijzen: The example below is not showing the `register()` method.

This actually brings up an interesting question. Should we align both register and deregister APIs from the `window` object and the NPM package?

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+```js
+  import { RegisteredEntity } from '@newrelic/browser-agent/interfaces/registered-entity'
+
+  const mfeAgent = new RegisteredEntity({
+    id: '550e8400-e29b-41d4-a716-446655440000',
+    name: 'Shopping Cart MFE',
+    tags: ['Critical']
```
ErikGrijzen: This example looks incorrect. Tags should always be a tuple of `key` and `value` right? Something like:

```js
{
  tags: [ { key: 'tag-key', value: 'tag-value' }]
}
```

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
```diff
@@ -0,0 +1,113 @@
...
+5. Track lifecycle events and performance independently from the main page
+
+## Register a micro frontend [#register]
+
+Once the main browser agent is initialized in your container application, each micro frontend can register itself using the `register()` method:
+
+```js
+  import { RegisteredEntity } from '@newrelic/browser-agent/interfaces/registered-entity'
+
+  const mfeAgent = new RegisteredEntity({
```
ErikGrijzen: This makes me wonder if the semantic naming is actually correct here...

I would expect this to be either:
```js
const mfeAgent = new MicroFrontendAgent{ ... });
```
or
```js
const registeredEntity = new RegisteredEntity({ ... });
```

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
```diff
@@ -0,0 +1,113 @@
...
+- [`addRelease()`](/docs/browser/new-relic-browser/browser-apis/addrelease) - Track deployments and releases for your micro frontend
+- [`deregister()`](/docs/browser/new-relic-browser/browser-apis/deregister) - Unregister the micro frontend when it unmounts or is no longer needed. Timings describing the lifecycle of your micro frontend will be automatically tabulated based on the usage of this API.
+- [`interaction()`](/docs/browser/new-relic-browser/browser-apis/interaction) - Access the SPA interaction API for custom timing
+- [`log()`](/docs/browser/new-relic-browser/browser-apis/log) - Capture log messages scoped to your micro frontend
+- [`measure()`](/docs/browser/new-relic-browser/browser-apis/measure) - Create custom performance measurements
+- [`noticeError()`](/docs/browser/new-relic-browser/browser-apis/noticeerror) - Manually capture errors with micro frontend attribution
+- [`recordCustomEvent()`](/docs/browser/new-relic-browser/browser-apis/recordcustomevent) - Record custom events for analytics
+- [`register()`](/docs/browser/new-relic-browser/browser-apis/register) - Register an entity for scoped data observability
```
ErikGrijzen, line 75: Maybe it's worth to put `register()` on the top of the list together with `deregister()`?

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
```diff
@@ -0,0 +1,113 @@
...
+- [`setApplicationVersion()`](/docs/browser/new-relic-browser/browser-apis/setapplicationversion) - Track the specific version of your micro frontend
+- [`setCustomAttribute()`](/docs/browser/new-relic-browser/browser-apis/setcustomattribute) - Add custom attributes that apply to all data from this micro frontend
+- [`setUserId()`](/docs/browser/new-relic-browser/browser-apis/setuserid) - Associate a user ID with this micro frontend's data
```
ErikGrijzen, line 78: Two questions here:
1. Does it actually make sense to expose this API on the MFE agent?
2. Does this inherit the user ID if it's being set on the container agent?

The only use case where I think this might make sense, is if the MFE is loaded in an iframe. But even there, this would work if we load the container agent in there or setup a bridge, since the ingest would flow through the bridge into the parent frame's container agent.

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+Each of these methods works identically to the standard browser agent APIs, but automatically scopes all collected data to your micro frontend entity.
+
+## Best practices [#best-practices]
+
+- **Initialize early**: Register micro frontends as soon as they mount to capture all events and best represent [lifecycle timings](/docs/browser/browser-monitoring/npm-package/micro-frontend-performance-timing)
+- **Use consistent IDs**: Keep micro frontend IDs stable across deployments for consistent tracking
```
ErikGrijzen: Should we also say that IDs should be unique?

src/content/docs/browser/new-relic-browser/browser-apis/register.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+This API call registers a micro frontend or component with the container browser agent, enabling independent monitoring and data collection. The registration process:
+
+1. **Connects the micro frontend** to the New Relic platform for entity synthesis and sampling
+2. **Creates relationships** between the container application and the micro frontend entity
+3. **Returns a scoped API interface** with methods pre-configured for the registered micro frontend
+4. **Enables data isolation** so each micro frontend's telemetry is tracked separately
+5. **Tracks basic lifecycle timings** so that usage from import to deregistration can be quantified
+
+After registration, all data captured through the returned API interface will be attributed to the micro frontend entity rather than the container application.
```
ErikGrijzen: I thought  the data is duplicated in both the browser application and the micro frontend? Or did we change that approach...?

src/content/docs/browser/new-relic-browser/browser-apis/register.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+<Callout variant="important">
+The micro frontend browser entity must be created before calling `register`. Entities can be created through the New Relic UI or programmatically via NerdGraph.
```
ErikGrijzen: This is incorrect. The entity will automatically be created through entity synthesis, right? Or is this a typo and you are talking about the browser application entity?

src/content/docs/browser/new-relic-browser/browser-apis/register.mdx
```diff
@@ -0,0 +1,111 @@
...
+- [`measure()`](/docs/browser/new-relic-browser/browser-apis/measure) - Create custom performance measurements
+- [`noticeError()`](/docs/browser/new-relic-browser/browser-apis/noticeerror) - Manually capture errors with micro frontend attribution
+- [`recordCustomEvent()`](/docs/browser/new-relic-browser/browser-apis/recordcustomevent) - Record custom events for analytics
+- [`register()`](/docs/browser/new-relic-browser/browser-apis/register) - Register an entity for scoped data observability
+- [`setApplicationVersion()`](/docs/browser/new-relic-browser/browser-apis/setapplicationversion) - Track the specific version of your micro frontend
+- [`setCustomAttribute()`](/docs/browser/new-relic-browser/browser-apis/setcustomattribute) - Add custom attributes that apply to all data from this micro frontend
+- [`setUserId()`](/docs/browser/new-relic-browser/browser-apis/setuserid) - Associate a user ID with this micro frontend's data
+
+Each of these methods works identically to the standard browser agent APIs, but automatically scopes all collected data to your micro frontend entity.
```
ErikGrijzen, line 94: Another reasons to maybe seperate `register()` and `deregister()` is we cannot say:
> Each of these methods works identically to the standard browser agent APIs

Because these two methods didn't exist before...

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+title: Monitor micro frontends with Browser monitoring
+tags:
+  - Browser
+  - Browser monitoring
+  - Micro frontends
+metaDescription: "Monitor the performance and user experience of individual micro frontends within your web application using New Relic Browser monitoring."
+freshnessValidatedDate: never
+---
+
+Micro frontend (MFE) monitoring lets you track the performance and user experience of individual micro frontends within your web application. Instead of seeing all data aggregated at the page level, you can now isolate telemetry data for each micro frontend component into distinct entities in New Relic, giving you precise visibility into how each part of your application performs.
```
ErikGrijzen: Regarding "page level". With standard Browser monitoring, they see things at the "application" level, right? It's true that you can filter by page (although that is very limited), but I believe you see the telemetry and insights for all pages at once, right?

ErikGrijzen: I'm saying this as well, because a very common way (and a Micro Frontend recommended practice) is to split a web application up is that micro frontends become pages or a set of pages that belong together. So mentioning: "Instead of page level aggregation" might be confusing...

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+
+## What are micro frontends? [#what-are-mfes]
+
+Micro frontends are an architectural pattern where large web applications are broken down into smaller, independently deployable components. Each micro frontend can be:
+
+- Owned and maintained by separate team(s)
+- Developed using independent technologies or frameworks
```
ErikGrijzen: This is considered quite an anti pattern for some time (2020). See:
https://www.thoughtworks.com/en-de/radar/techniques/micro-frontend-anarchy

Maybe it's best not to mention it. It might make us look a bit outdated...

A valid case for this would be a temporary migration project from 1 framework to another. But that's not the same as what is stated here.

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+- Deployed independently of other components
+- Representative of a specific business domain or feature
+
+Common micro frontend patterns include:
```
ErikGrijzen: I would maybe remove the list below. I'm not sure if this brings any value to the reader, unless they link to a deeper explanation somewhere in the docs or externally...

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+## How New Relic's micro frontend monitoring works [#how-it-works]
+
+New Relic's approach uses a **central agent architecture** where:
+
+1. **One browser agent** runs in the container application, which continues to capture page-wide data. This agent would typically be used by the container-level or platform teams, interested in monitoring holistic application performance.
+2. **Each micro frontend registers** with the central agent using the `register` API
+3. **Data is scoped and harvested** separately for each registered micro frontend when using the scoped APIs.
+4. **New MicroFrontend entities automatically appear** in New Relic when utilized properly.
+5. **Relationships are automatically created** between the container and micro frontends.
+6. **Entity-level attibution is supported** to facilitate assignment of teams, repositories and tags.
```
ErikGrijzen:
```suggestion
6. **Entity-level attribution is supported** to facilitate assignment of teams, repositories and tags.
```

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+This architecture pattern provides several benefits:
+
+- **Lightweight implementation**: No duplicate agent code is ever loaded on the page
+- **Shared infrastructure**: All micro frontends benefit from the container's feature set
+- **Consistent harvesting**: Data collection follows the same patterns across all components -- Collect data by using the APIs you are already familiar with.
+- **Simplified configuration**: Container team is still in control of what happens on a given page.
+
+## Architecture patterns and implementation [#architecture-patterns]
+
+### How registration works [#how-registration-works]
```
ErikGrijzen: Shouldn't we just reference to the register docs, so it's not duplicated?

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+- [`recordCustomEvent()`](/docs/browser/new-relic-browser/browser-apis/recordcustomevent) - Record custom events for analytics
+- [`setApplicationVersion()`](/docs/browser/new-relic-browser/browser-apis/setapplicationversion) - Track the specific version of your micro frontend
+- [`setCustomAttribute()`](/docs/browser/new-relic-browser/browser-apis/setcustomattribute) - Add custom attributes that apply to all data from this micro frontend
+- [`setUserId()`](/docs/browser/new-relic-browser/browser-apis/setuserid) - Associate a user ID with this micro frontend's data
+
+Each of these methods works identically to the standard browser agent APIs, but automatically scopes all collected data to your micro frontend entity.
+
+### Key Requirements
+1. IDs supplied should be generated and managed according to your organization's internal processes and conventions. It serves as a permanent, stable reference to uniquely identify this specific documentation entity across your system. Ensure this value remains consistent and is not duplicated across other documents or entities.
```
ErikGrijzen:
```suggestion
1. IDs supplied should be generated and managed according to your organization's internal processes and conventions. It serves as a permanent, stable reference to uniquely identify this specific entity across your system. Ensure this value remains consistent and is not duplicated across other entities.
```

Not sure what is ment with "documentation" and "document" here?

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+  });
+}
+
+// Track custom logs specific to this module
+catalogAgent.log('Catalog items loaded successfully', { 
+  level: 'debug', customAttributes: {
+    itemCount: 150,
+    loadTime: 245 
+  }
+});
```
ErikGrijzen:
```suggestion
catalogAgent.log('Catalog items loaded successfully', { 
    level: 'debug', 
    customAttributes: {
      itemCount: 150,
      loadTime: 245 
    }
  }
);
```

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+catalogAgent.measure('catalog-rerender', {
+  start: 1234,
+  end: 5678
+});
+
+// Cleanup when the micro frontend unmounts and is no longer needed
+catalogAgent.deregister();
+```
+
+## Generic component monitoring [#generic-components]
```
ErikGrijzen, line 135: While technically possible... do you think we should document this?

By trying to provide a solution for all cases, we might risk creating product that is overloaded with use cases. In New Relic we have a tendency to do this... By trying to fulfil a minority use case, we are degrading the experience of the product on how the majority of the users use it. A product that becomes cluttered, overwhelming for the user and hard to understand.

We might run the risk that the Micro Frontend solution need to be able to monitor everything, since everything you can think of can be scoped some way or another. I'm worried that this will make it really difficult to make a strong product solution. For example, you can "scope" the implementation/mechanism of the routing and navigation, but making that a separate entity seems a bit weird to me.

Like I said before in different forums, customers can do strange things with our agents and they most likely will, we cannot stop them. But I always thought it would be a good idea to have a strong thought leadership opinion on the "normal" use case, so we can optimise our product solution around that.

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+onboardingAgent.addPageAction('onboardingStepCompleted', { step: 3 });
+```
+
+## Data relationships and entity synthesis [#relationships]
+
+When you register a micro frontend, New Relic automatically:
+
+1. **Creates a browser entity** for the micro frontend if it doesn't exist
+2. **Establishes relationships** between the container and micro frontend entities
+3. **Tracks various timings** describing the import and execution times of your micro frontend scripts
```
ErikGrijzen: When I think about "execution times of frontend scripts", I'm thinking of the JavaScript bundle source code evaluation. 

Maybe this can all be described as "micro frontend lifecycle timings"? What do you think?

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+4. **Enables distributed tracing** connections between backend services and micro frontends when available
+5. **Adds entity tagging and metadata** to help with organization and discovery
+6. **Decorates telemetry** with the appropriate attributes necessary for cross-linking between the container and the MFE.
```
ErikGrijzen: Isn't this already covered with point #2?
> 2. **Establishes relationships** between the container and micro frontend entities

Sounds like more of an internal technical detail on how #2 is done...

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+### Enablement
+The container browser agent acts as the gatekeeper for the page and controls what any registered children can report. As such, it can be configured to block incoming registration data:
+
+To configure this, edit the application settings in New Relic for your container agent app.  To do so with NPM, ensure that the container agent is supplied with:
+```js
+{
+  init: {
+    // ... other configs
+    api: {
+      allow_registered_children: true|false
```
ErikGrijzen: Is there a default value?

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+    }
+  }
+}
+```
+
+## Data duplication
+
+The container browser agent can be configured to duplicate data reported by registered micro frontends to its own app, thereby ensuring that _all_ data can be observed from the top-level application if desired.
+This can be further broken down, as the agent will automatically add `parent.id` and `child.id` attributes to the relevant data reported by each, enabling easy cross-linking as needed.
```
ErikGrijzen, line 219: Will this parent and child ID only be added if duplicated data is enabled or is it always present?

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+
+To configure this, edit the application settings in New Relic for your container agent app.  To do so with NPM, ensure that the container agent is supplied with:
```
ErikGrijzen:
```suggestion
To configure this, edit the application settings in New Relic for your container agent app. To do so with the Browser Agent NPM package, ensure that the container agent is supplied with:
```

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+  }
+}
+```
+
+## Performance timing and measurements [#performance-timing]
+
+Micro frontend monitoring can capture the same performance data as standard browser monitoring, but scoped to each component:
+
+- **Performance timings** (when applicable to the micro frontend's initialization) - reported as `MicroFrontEndTiming` events
+- **AJAX requests** made from within the micro frontend context
```
ErikGrijzen: How are AJAX request being captured so they are scoped to the MFE? I believe this is still missing, right?

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+}
+```
+
+## Performance timing and measurements [#performance-timing]
+
+Micro frontend monitoring can capture the same performance data as standard browser monitoring, but scoped to each component:
+
+- **Performance timings** (when applicable to the micro frontend's initialization) - reported as `MicroFrontEndTiming` events
+- **AJAX requests** made from within the micro frontend context  
+- **JavaScript errors** that occur in micro frontend code
```
ErikGrijzen: This might be misleading, because only errors that are captured with the `noticeError()` API will be scoped. Others will will be reported to the container browser application entity.

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+- **Custom performance marks and measures** specific to the component
+
+## Getting started [#getting-started]
+
+To start monitoring your micro frontends:
+
+1. **Ensure you have a browser agent** (Pro or Pro+SPA) running on your container application
+2. **Register each micro frontend** within your application using the `register` API
+3. **Use the returned APIs** to capture micro frontend-specific data
+4. **Find your new entities** in New Relic under "Browser" > "Micro Front Ends"
```
ErikGrijzen:
```suggestion
4. **Find your new entities** in New Relic under "Browser" > "Micro-Frontends"
```
The tab shows now:
image

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+5. **Query your data and set up teams, repos, dashboards and alerts** for your individual micro frontend entities
```
ErikGrijzen:
```suggestion
5. **Query your data and set up dashboards, alerts, service levels and link your teams and repositories** for your individual micro frontend entities
```

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-performance-timing.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+**Description**: The time from the beginning of loading the MFE until the time the MFE is registered to report, reported in MS.
+
+**Measurement**:
+* **Start time**: `fetchStart` (or Page origin `0` if no remote resource timing is found)
+* **End time**: `.register()` call time
+* **Duration**: End - Start
+
+<Callout variant="tip">
+  If no resource timing is found (for example, the registered entity was pre-bundled with the main app code), this metric reports as `0`.
```
ErikGrijzen, line 27: Is this correct? Wouldn't the `timeToLoad` not take the bundle of the main application in this case?

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-performance-timing.mdx
```diff
@@ -0,0 +1,78 @@
...
+</Callout>
+
+### timeToBeRequested
+
+**Description**: The time from the page origin time until the MFE was requested to be lazy-loaded, reported in MS.
+
+**Measurement**:
+* **Start time**: Page origin (`0`)
+* **End time**: `fetchEnd` (or `0` if no resource timing is found)
```
ErikGrijzen: Taking this in mind:
> The time from the page origin time until the MFE was requested to be lazy-loaded

Should the end time be the fetch start of the MFE? That's the moment it's requested to be lazy loaded, right?

Maybe I find the word "requested" hard to understand...

src/content/docs/browser/new-relic-browser/troubleshooting/micro-frontend-monitoring-limitations.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+```js
+// This will do nothing when used with the Lite agent
+const mfeAgent = window.newrelic.register({
+  id: 'my-component',
+  name: 'My Component'
+});
+```
+
+### Cross-origin limitations [#cross-origin]
```
ErikGrijzen, line 33: Can customers add the container agent to the iframe to make this work?

src/content/docs/browser/new-relic-browser/configuration/configure-micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+For full coverage, ensure all features are enabled in your container agent that will be shared with micro frontends:
+
+<Steps>
+<Step>
+
+If using NPM, enable these features as needed (which are enabled by default):
+- **JavaScript Errors**
+- **Generic Events**
+- **Logs**
```
ErikGrijzen, line 65: Should we add a section how to enable Logs for micro frontends? I believe it needs to be enabled on the container browser application entity, right?

src/content/docs/browser/new-relic-browser/troubleshooting/micro-frontend-monitoring-limitations.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+### Limited automatic instrumentation [#limited-auto-instrumentation]
+
+Unlike full browser agents, micro frontend monitoring requires manual instrumentation
+
+### Relationship mapping limitations [#relationship-limitations]
+
+Entity relationships have constraints:
+
+Entities and relationships are synthesized at data harvest time.  If your microfrontend registration does not harvest any data, it will not synthesize these primatives.
```
ErikGrijzen: This sounds very very technical. Should we maybe rephrase this in a more non-technical way?

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
metal-messiah:
```suggestion
Use the RegisteredEntity interface to monitor individual micro frontends within your JavaScript application. After [installing the browser agent in your container application](/docs/browser/new-relic-browser/browser-pro-features), each micro frontend can register itself to report data under its own micro frontend entity in New Relic.
```

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
metal-messiah:
```suggestion
3. Report data under their own browser micro frontend entity in New Relic
```

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
metal-messiah:
```suggestion
Once the main browser agent is initialized in your container application, each micro frontend can register itself using the `new RegisteredEntity()` constructor:
```

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
metal-messiah: This is being updated in the agent to match.  the new syntax will be
```js
{
    tags: {foo: 'bar', foo2: 'bar2'} // tags of foo: bar, and foo2: bar2 are created
}
```

metal-messiah:
```suggestion
    tags: {isParent: true, nested: true, dataTypes: 'all'}
```

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
metal-messiah:
```suggestion
  const registeredEntity = new RegisteredEntity({
```

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
metal-messiah: The list is currently alphabetized to match formatting from other docs lists

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
metal-messiah: ```Does it actually make sense to expose this API on the MFE agent?```
yes, as the MFE agent supports custom attributes (which is how enduser.id is reported)

```Does this inherit the user ID if it's being set on the container agent?```
No, as currently constructed, MFE agents do not inherit container application custom attributes, and maintain their own.  This was by design to keep them "isolated"

File: src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx (hunk 8 above)
Comment (metal-messiah) [reply src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx]:
```suggestion
- **Use consistent IDs**: Keep micro frontend IDs unique and stable across deployments for consistent tracking
```

src/content/docs/browser/new-relic-browser/browser-apis/register.mdx
metal-messiah: Theres a config to control that.  I'll note that here

metal-messiah:
```suggestion
After registration, all data captured through the returned API interface will be attributed to the micro frontend entity rather than the container application, unless explicitly configured to duplicate across applications.
```

src/content/docs/browser/new-relic-browser/browser-apis/register.mdx
metal-messiah:
```suggestion
The micro frontend entity will be synthesized automatically upon the receipt of new events, using the ID and name provided during registration.
```

src/content/docs/browser/new-relic-browser/browser-apis/register.mdx
metal-messiah: These are describing the return object, in that the APIs behave identically to the window APIs, not that they existed "before". It wouldn't make sense to separate them, because they are part of the return object.

src/content/docs/browser/new-relic-browser/browser-apis/register.mdx
```diff
@@ -0,0 +1,111 @@
...
+- [`addRelease()`](/docs/browser/new-relic-browser/browser-apis/addrelease) - Track deployments and releases for your micro frontend
+- [`deregister()`](/docs/browser/new-relic-browser/browser-apis/deregister) - Unregister the micro frontend when it unmounts or is no longer needed. Timings describing the lifecycle of your micro frontend will be automatically tabulated based on the usage of this API.
+- [`interaction()`](/docs/browser/new-relic-browser/browser-apis/interaction) - Access the SPA interaction API for custom timing
+- [`log()`](/docs/browser/new-relic-browser/browser-apis/log) - Capture log messages scoped to your micro frontend
+- [`measure()`](/docs/browser/new-relic-browser/browser-apis/measure) - Create custom performance measurements
+- [`noticeError()`](/docs/browser/new-relic-browser/browser-apis/noticeerror) - Manually capture errors with micro frontend attribution
+- [`recordCustomEvent()`](/docs/browser/new-relic-browser/browser-apis/recordcustomevent) - Record custom events for analytics
+- [`register()`](/docs/browser/new-relic-browser/browser-apis/register) - Register an entity for scoped data observability
```
metal-messiah:
```suggestion
- [`register()`](/docs/browser/new-relic-browser/browser-apis/register) - Register an entity for scoped data observability. The registered entity will be related to the API it is called from. This means that when `.register()` is called from the global `newrelic` API, it will become a "child" of that container browser agent.  When it is called from another registered entity's API, it will become a "child" of that entity.
```

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
metal-messiah:
```suggestion
Micro frontend (MFE) monitoring lets you track the performance and user experience of individual micro frontends within your web application. Instead of seeing all data aggregated at the application level, you can now isolate telemetry data for each micro frontend component into distinct entities in New Relic, giving you precise visibility into how each part of your application performs.
```

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
metal-messiah:
```suggestion
```

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+- Developed using independent technologies or frameworks
+- Deployed independently of other components
+- Representative of a specific business domain or feature
+
+Common micro frontend patterns include:
+
+- **Component-based architecture**: Individual UI components that render within a container application
+- **Route-based splitting**: Different application routes handled by separate micro frontends  
+- **Server-side composition**: Micro frontends assembled on the server before being sent to the client
+- **Client-side composition**: Micro frontends loaded and rendered dynamically in the browser
```
metal-messiah:
```suggestion
```

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,296 @@
...
+- [`measure()`](/docs/browser/new-relic-browser/browser-apis/measure) - Create custom performance measurements
+- [`noticeError()`](/docs/browser/new-relic-browser/browser-apis/noticeerror) - Manually capture errors with micro frontend attribution
+- [`register()`](/docs/browser/new-relic-browser/browser-apis/register) - Register an entity for scoped data observability
+- [`recordCustomEvent()`](/docs/browser/new-relic-browser/browser-apis/recordcustomevent) - Record custom events for analytics
+- [`setApplicationVersion()`](/docs/browser/new-relic-browser/browser-apis/setapplicationversion) - Track the specific version of your micro frontend
+- [`setCustomAttribute()`](/docs/browser/new-relic-browser/browser-apis/setcustomattribute) - Add custom attributes that apply to all data from this micro frontend
+- [`setUserId()`](/docs/browser/new-relic-browser/browser-apis/setuserid) - Associate a user ID with this micro frontend's data
+
+Each of these methods works identically to the standard browser agent APIs, but automatically scopes all collected data to your micro frontend entity.
```
metal-messiah:
```suggestion
### How registration works.
All micro frontend activity is facilitated by using the `register` API on the container browser agent. Please refer to the [register API](/docs/browser/new-relic-browser/browser-apis/register) documentation for more information.

File: src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx (hunk 19 above)
Comment (metal-messiah) [reply src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx:135]: This was asked for already from customers and by our product leaders already which is why its included in here.  It may come to be removed before becoming public

File: src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx (hunk 21 above)
Comment (metal-messiah) [reply src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx]: i think I meant there are two forms of linkage here -- entity relationships as well as telemetry attributes.  I can simplify still though

File: src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx (hunk 20 above)
Comment (metal-messiah) [reply src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx]:
```suggestion
3. **Tracks various timings** describing the lifecycle timings of your micro frontend scripts
```

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
metal-messiah:
```suggestion
```

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
metal-messiah:
```suggestion
      allow_registered_children: true|false // defaults to `false`
```

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
metal-messiah: it will always be added (but only on container agent data if duplicated, since thats the only time that data is created)

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
metal-messiah: yeah this needs to be omitted for now, ty

metal-messiah:
```suggestion
```

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-monitoring.mdx
metal-messiah:
```suggestion
- **JavaScript errors** that occur in micro frontend code and captured using the scoped `.noticeError()` API
```

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-performance-timing.mdx
metal-messiah: it assumes 0 because it does not know what "main" bundle means. An application can load with 20 scripts. Beyond that, the most likely case that we cannot find the resource timing is that it was called from an inline script, which would its origin as `0`.

src/content/docs/browser/new-relic-browser/browser-pro-features/micro-frontend-performance-timing.mdx
metal-messiah: Yeah this is wrong. it should be `fetchStart`

metal-messiah:
```suggestion
* **End time**: `fetchStart` (or `0` if no resource timing is found)
```

src/content/docs/browser/new-relic-browser/troubleshooting/micro-frontend-monitoring-limitations.mdx
metal-messiah: yes, that is true.  extending list

src/content/docs/browser/new-relic-browser/troubleshooting/micro-frontend-monitoring-limitations.mdx
```diff
@@ -0,0 +1,120 @@
...
+    id: 'cross-origin-widget',
+    name: 'External Widget'
+  });
+} catch (error) {
+  // SecurityError: Blocked a frame with origin "https://widget.external.com" 
+  // from accessing a cross-origin frame.
+  console.error('Cross-origin access blocked:', error);
+}
+```
+
```
metal-messiah:
```suggestion
A container agent can be added to each iframe to help facilitate these constraints.
```

src/content/docs/browser/new-relic-browser/configuration/configure-micro-frontend-monitoring.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+For all other installations, the Pro or Pro+Spa builds of the agent will have the features needed for micro frontend observability by default.
+
```
metal-messiah:
```suggestion
<Callout variant="important">
If capturing micro frontend logs, you must ensure the container agent is configured to sample and collect logs as desired.  The log settings applied to the container agent will be inherited by any registered micro frontend entities. See [getting started with logs](https://docs.newrelic.com/docs/browser/browser-monitoring/browser-pro-features/browser-logs/get-started/) for more information.
</Callout>
```

src/content/docs/browser/new-relic-browser/troubleshooting/micro-frontend-monitoring-limitations.mdx
metal-messiah:
```suggestion
Entities and relationships are automatically created only when data is collected.  Even after registration, if your micro frontend does not harvest any data to New Relic, no entity or relationship will be created.
```

src/content/docs/browser/browser-monitoring/npm-package/micro-frontend-npm-usage.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+```js
+  import { RegisteredEntity } from '@newrelic/browser-agent/interfaces/registered-entity'
+
+  const mfeAgent = new RegisteredEntity({
+    id: '550e8400-e29b-41d4-a716-446655440000',
+    name: 'Shopping Cart MFE'
+  });
+  
+  // If the container agent was not available at registeration time, all API methods will safely no-op without throwing errors
```
OscarNR, line 101:
```suggestion
  // If the container agent was not available at registration time, all API methods will safely no-op without throwing errors
```

WriteMayur:
Closing this PR as this are LP features. @adutta-newrelic has created relative PR in beta repo.

PR #22890, opened 2026-01-30
Refactor K8s deployment guidance

keegoid-nr:
Focus on practical guidance and less on equations that are often not accurate once deployed to a real cluster.


adutta-newrelic:
netlify build fork

PR #22992, opened 2026-02-11
Update prerequisites-to-share-dashboards-charts.mdx

pati04g:
In the Add permissions to a custom role section, the description of the Modify capability is not accurate and customers believe they only need Modify to change the properties of the live URLs, when this capability is also needed to Create live urls. I have updated so it reads: 
 Modify: Can create live URLs and/or update the properties of the live urls


adutta-newrelic (approved):
LGTM

PR #23158, opened 2026-02-23
Update WebLogic compatibility range to 14.1.x

jasonjkeller:
Java agent instrumentation should apply to versions beyond 14.1.1.


PallaviWrite (approved):
LGTM

PR #23180, opened 2026-02-24
chore: added date when we dropped support for Node 18. Also preemptively added EOL date of Node 24

bizob2828:
## Give us some context

 * We dropped support for Node 18 and forgot to include the date in this table.
 * We also want to broadcast future EOL dates ahead of the release. This adds the EOL date for node 24 as well


PallaviWrite (approved):
LGTM

PR #23195, opened 2026-02-25
Update flex-integration-tool-build-your-own-integration.mdx

nr-ksteinbach:
It was causing customer confusion that this doc was calling the "account" to be compatible with macOS in the previous point 2 (now removed), when in the fourth (now third) point the bundling of nri-flex was correctly specified as NOT available for macOS. In a discussion with other infra wizards, we concluded that the fluff about the account's compatibility is not required.


PallaviWrite:
netlify build fork

PR #23304, opened 2026-03-05
Add callout for AWS region-specific sink setup

wildcardlinux:
Added important callout about region-specific sinks for AWS monitoring.

We should link out to this documentation and add a caution banner so customers know before they start doing initial setup.

https://docs.newrelic.com/docs/infrastructure/amazon-integrations/manage-aws-data/aws-multi-account/

## Give us some context

* Confusion with sending multiple metric streams to New Relic.


tejaswaroop-nr:
netlify build fork

PR #23306, opened 2026-03-06
feat(agent control): minor edit

ubanerjeeNR:


WriteMayur:
@ubanerjeeNR, could you please confirm if we still need these changes?

PR #23433, opened 2026-03-16
Removed the callout tip

adutta-newrelic:


WriteMayur:
@adutta-newrelic, could you please confirm if this PR is still needed or we can close it?

PR #23470, opened 2026-03-19
feat(srx): added dotnet srx integration details to overview

AustinNr:

## Give us some context

    -  This PR adds information about SRX's .NET integration that customers will soon have access to.
    - Question:  We're handling preview through a feature flag rather than an actual process, is the preview callout acceptable for this use case?


adutta-newrelic:
Hello @OhTinter could you please sign the CLA so that we can review the PR

PR #23528, opened 2026-03-25
Whats new dashboards 0326

pati04g:
This is a what's new for new Dashboard features. Ideally to be published on the 27th of March. 
I don't believe it needs product marketing manager review as it has already been reviewed by our Product Manager.


WriteMayur (approved):
looks good to me

PR #23728, opened 2026-04-14
fix: updated the grant command 

WriteMayur:
This PR updated the Oracle DB integration access grant commands as per [this slack conversation](https://newrelic.slack.com/archives/D091Y4A0UNR/p1776161392578789)


PallaviWrite (approved):
LGTM

PR #21863, opened 2025-10-06
Python release v11.1.0

lrafeei:
This PR provides release notes for v11.1.0 of the Python agent

PR #23815, opened 2026-04-22
Revert verbiage to 'yanked' in Python Agent release notes

lrafeei:
A previous release of the Python Agent's v11.0.0 notes had noted the status as "withdrawn" and this PR changes it back to the term "yanked"

It seems as though [this comment](https://github.com/newrelic/docs-website/pull/21863#issuecomment-3377794941) may have been missed before the previous merge.  The reason for this is because "yanked" is a [specific industry term](https://docs.pypi.org/project-management/yanking/) that denotes the continued presence of the specified version despite no longer being shown.  In other words, it is now hidden.  There is also the concept of deleting a release, but this is reserved for more extreme cases.  In this case, using this specific term is necessary for customers to be able to differentiate the two.


hmstepanek (approved):

adutta-newrelic (approved):
LGTM

PR #23883, opened 2026-04-28
docs(otel-apm): fix misaligned table headers

kb-newrelic:
## Give us some context
- Merges Unit and UCUM into a single column (as they are intended to be a single column, see other tables) as it breaks the formatting of the table.


adutta-newrelic (approved):
LGTM

PR #23927, opened 2026-05-04
Infra k8s pixie orphaned files

PallaviWrite:


PallaviWrite:
This PR is closed since @cbehera-newrelic has already addressed.

PR #23962, opened 2026-05-06
Fix otlp hostname

jeffalder:
This is the correct hostname or customers will misconfigure their systems. Note that this hostname is available on the linked document immediately below.


WriteMayur (approved):
thanks for correcting it.

PR #23963, opened 2026-05-07
Fix: Updated the GTS Benefits

WriteMayur:
As per request received on the help-docs channel, updated the Designated Support Customer Experience Manager to Designated Support Account Manager.


adutta-newrelic (approved):
LGTM

PR #24101, opened 2026-05-15
Update set-up-session-replay.mdx

CherylFrankenfield:
Updated GA agent versions

* These are the agent versions when we went GA vs an earlier version


PallaviWrite:
netlify build fork

PallaviWrite (approved):
LGTM

PR #24187, opened 2026-05-22
docs(nodejs-agent): clarify which Node.js VM metrics are collected by default vs require native module

gauravagrawal01:

## Give us some context

**What problems does this PR solve?**

The `## What you need` [section](https://docs.newrelic.com/docs/apm/agents/nodejs-agent/extend-your-instrumentation/nodejs-vm-measurements/#requirements) currently states the native module is required for "garbage collection, memory, and CPU" metrics. This is partially incorrect — memory and CPU are collected by default via built-in Node.js APIs with no native module needed. Only GC and precise event loop metrics require `@newrelic/native-metrics`.

This causes real confusion: developers see flat GC and event loop panels in the Node VMs UI and cannot tell from the docs whether it is a traffic issue or a missing dependency.

**Changes made**

- Updated the opening of `## What you need` to clearly distinguish between default samplers and native module samplers
- Added inline `<Callout>` labels inside each measurement collapser so users can immediately see whether a metric requires the native module or not
- Updated `freshnessValidatedDate` from `never` to `2026-05-22`

No existing content was removed.

**Verification**

Confirmed against agent source code:  
https://github.com/newrelic/node-newrelic/blob/main/lib/system-metrics-sampler.js

- `sampleMemory`, `checkEvents`, and `sampleCpu` always register regardless of native module availability
- `sampleGc` and `sampleLoop` only register when `@newrelic/native-metrics` loads successfully under `agent.config.plugins.native_metrics.enabled`

image

image

### DIFF
--- a/src/content/docs/apm/agents/nodejs-agent/extend-your-instrumentation/nodejs-vm-measurements.mdx
+++ b/src/content/docs/apm/agents/nodejs-agent/extend-your-instrumentation/nodejs-vm-measurements.mdx
@@ -9,14 +9,17 @@ redirects:
   - /docs/agents/nodejs-agent/extend-your-instrumentation/nodejs-vm-measurements
   - /docs/agents/nodejs-agent/supported-features/node-vm-measurements
   - /docs/agents/nodejs-agent/supported-features/nodejs-vm-measurements
-freshnessValidatedDate: never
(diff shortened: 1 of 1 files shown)


sujitnewrelic:
netlify build fork

sujitnewrelic (approved):
LGTM

PR #24225, opened 2026-05-26
chore: Remove reference to `newrelic.instrumentLoadedModule`

amychisholm03:
We plan to remove the `instrumentLoadedModule` API in the next major version of the Node agent. This is because we have moved most of our instrumentation support to a subscriber-based model which strongly recommends using `node -r newrelic` only for loading the agent.

## Related Issues
https://github.com/newrelic/node-newrelic/issues/4031


sujitnewrelic:
netlify build fork

sujitnewrelic (approved):
LGTM

PR #24245, opened 2026-05-27
docs: add section on handling delayed data and older timestamps for outlier detection

tomtupy-nr:
**Description of the change:**

Added a new "Handling delayed data and older timestamps" section to the Outlier Detection page.  https://docs.newrelic.com/docs/alerts/create-alert/set-thresholds/outlier-detection/

This update includes:

An explanation of why signals with older timestamps miss the evaluated window and are processed as forced inliers.

A breakdown of common architectural causes (Cloud-provider polling, long-running transactions, and buffered data).

A NRQL query using the NrAiSignal event to help users identify excluded entities and their outlierProcessingSkippedTimeDelta.

Actionable resolution steps (Splitting the condition or increasing the aggregation window).

**Why it’s necessary:**

Customers frequently experience excluded signals in outlier detection due to standard architectural behaviors (like AWS CloudWatch polling intervals or long-running transactions). This change empowers users to understand why their data is being excluded, investigate the impacted entities themselves using NRQL, and confidently resolve the issue by adjusting their condition configurations.

### DIFF
--- a/src/content/docs/alerts/create-alert/set-thresholds/outlier-detection.mdx
+++ b/src/content/docs/alerts/create-alert/set-thresholds/outlier-detection.mdx
@@ -78,6 +78,64 @@ Follow these steps to create an alert condition with outlier detection:
 - Avoid over-segmentation which can reduce effectiveness.
 - Consider seasonality and business patterns when grouping.
 
+### Handling delayed data and older timestamps [#handling-delayed-data-and-older-timestamps]
+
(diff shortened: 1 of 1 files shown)


sujitnewrelic:
netlify build fork

ubanerjeeNR:
netlify build fork

cbehera-newrelic (approved):
lgtm

Issue #1385, opened 2021-03-17
Gatsby CreatePages error: Non-deterministic routing danger


jpvajda:
After make a change and saving, on hot reload we see this error which runs for like 5 mins through every page it seems.  It makes me think it has something to do with our / change to our routes.  I've done some research on it, and haven't found anything definitive as to what the problem may be. But we've tried to do all the standard troubleshooting techniques, and it still persists

```
warn Non-deterministic routing danger: Attempting to create page:
"/jp/docs/using-new-relic/welcome-new-relic/get-started/glossary", but page
"/jp/docs/using-new-relic/welcome-new-relic/get-started/glossary/" already exists

```
This error will spit out in the terminal for about 5 mins

We reached out to Dustin at Gatsby and he pointed to this [code](https://github.com/newrelic/gatsby-theme-newrelic/blob/3b5d97e2e8bc557108a68515b146649862c77894/packages/gatsby-theme-newrelic/gatsby-node.js#L288)

and said 

```
There’s no guarantee that that’s deterministic and may differ. I bet if you log before that line that’s the code path that is triggered and is causing the difference.

```

### Steps to repro

- after running yarn start and viewing the site locally
- make a change locally and save, watch your terminal
- go make a coffee, come back. 

### Expected 

- This error is resolved so it doesn't appear when reloading a local change to the site.

DSchau:
Hi folks :) I solve this by console logging, e.g. something like:

```
console.log('\n==================================='n)
console.log(JSON.stringify(page, null, 2))
```

and I _predict_ (but I've been wrong before!) you'll see a different page path in the highlighted context. You likely just need to normalize in the same way when you create the page in the first place, albeit that is a touch strange.

In any scenario -- the warning is not a red herring, definitely on content update, the page is being re-created with a different (non-trailing) path, so that's the most likely spot!

jpvajda:
@DSchau thanks for the pro tip here. We appreciate it!

PR #24253, opened 2026-05-28
Update domain validation docs: real TLD detection replaces structural regex

ccedacero-nr:
## Summary

Updates the name field validation documentation to reflect an improvement to domain detection:

- The domain check now uses [Mozilla's Public Suffix List](https://publicsuffix.org/) to identify **real** top-level domains, rather than blocking any dot-separated string that structurally resembles a domain.
- The domain check now applies uniformly to `name`, `given_name`, and `family_name` (previously `given_name`/`family_name` skipped this check entirely).
- Dot-separated names like `firstname.lastname` are explicitly called out as allowed.

image

## Context

The previous regex-based approach blocked legitimate non-English names that use dots for clarity (e.g., Japanese romanized names). This caused provisioning failures for real users. The improved detection only blocks strings ending in registered TLDs (`.com`, `.net`, `.co.uk`, `.app`, etc.), eliminating false positives on surnames.

Related: user-service PR [#1385](https://source.datanerd.us/account-auth-and-access/user-service/pull/1385)

## Changes

- `user-management-ui-and-tasks.mdx`: Updated "No bare domain names" row to "No real domain names" with examples, changed `given_name`/`family_name` column from "Allowed" to "Blocked", expanded the tip callout to explain that dot-separated names are valid when the suffix is not a real TLD.

### DIFF
--- a/src/content/docs/accounts/accounts-billing/new-relic-one-user-management/user-management-ui-and-tasks.mdx
+++ b/src/content/docs/accounts/accounts-billing/new-relic-one-user-management/user-management-ui-and-tasks.mdx
@@ -288,7 +288,7 @@ When creating or updating a user, the `name`, `given_name`, and `family_name` fi
 
     <tr>
       <td>
-        No bare domain names
+        No real domain names (for example, `evil.com`, `phish.co.uk`)
       </td>
 
       <td>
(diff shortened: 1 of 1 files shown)


sujitnewrelic:
netlify build fork

sujitnewrelic (approved):
LGTM

PR #24354, opened 2026-06-08
feat(ac): jp support

paologallinaharbur:
Updating a couple of comments and removing duplicated content


adutta-newrelic:
Hey @paologallinaharbur I believe that @PallaviWrite is working on updating the JP documents, she has informed that she will be updating the document with the current information and will make the document public with the JP release

PR #24458, opened 2026-06-16
docs: Added a section in Node.js agent for setting up instrumentation for Next.js

bizob2828:
## Give us some context

Provides the recommended way to setup the Node.js agent with Next.js. We did some investigation and found that we can support native Next.js otel spans with 14.1.0 of the agent as well as deploy to vercel. Historically this has been an issue for customers, and discovered this while working on https://github.com/newrelic/node-newrelic/issues/3188

### DIFF
--- a/src/content/docs/apm/agents/nodejs-agent/extend-your-instrumentation/nextjs-instrumentation.mdx
+++ b/src/content/docs/apm/agents/nodejs-agent/extend-your-instrumentation/nextjs-instrumentation.mdx
@@ -8,17 +8,17 @@ metaDescription: The New Relic Node.js agent can intercept native Next.js OpenTe
 freshnessValidatedDate: never
 ---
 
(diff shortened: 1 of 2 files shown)


WriteMayur:
netlify build

WriteMayur:
netlify build fork

bizob2828:
I'm looking at the preview but have to navigate directly. I thought pages were automatically aded to the side bar, where do i have to add it?

bizob2828:
> I'm looking at the preview but have to navigate directly. I thought pages were automatically aded to the side bar, where do i have to add it?

never mind, i see a yml file that has the nav bar

vpayyapilly (dismissed):

PallaviWrite:
netlify build fork

tejaswaroop-nr (approved):
LGTM!

PR #24588, opened 2026-06-26
feat(search): migrate /search-results to SearchGPT

akashreddy-nr:
## Summary

Migrates the `/search-results` page from **Swiftype** to **SearchGPT** (New Relic's internal search service, REST API v2), consuming the `search()` / `highlightText()` / `useLocale()` / `localeToSearchLanguage()` exports from `@newrelic/gatsby-theme-newrelic`.

Depends on theme **9.17.0** (published with the SearchGPT v2 client + `localeToSearchLanguage` + `useLocale` re-export); this PR bumps the dep from 9.15.0.

## Changes

- **`src/pages/search-results.js`** — rewritten against the SearchGPT `search()` contract:
  - `search({ searchTerm, cursor, limit, language })` replaces the old Swiftype `{ searchTerm, defaultSources, filters, page, perPage }`.
  - Reads `results` / `totalCount` / `nextCursor` from the normalized hybrid-search shape.
  - **Cursor-based Prev/Next pagination.** SearchGPT paginates with opaque seek cursors (`search_after`), so there is no random page access. The page tracks the current cursor plus a stack of previously-visited cursors to support "Previous". Because the walk is seek-based, the full result set is reachable — it is **not** clamped to the 100-results-per-source window.
  - **Upfront "Page N of ~M" estimate.** Current page = cursor-stack depth + 1; total = `Math.ceil(totalCount / LIMIT)`, shown as approximate (`~M`) since the hit count is an estimate and the walk terminates when `nextCursor` goes null.
  - **Language scoping.** `useLocale()` + `localeToSearchLanguage()` scope results to the site language (jp→ja, kr→ko, pt→pt-br; English site = en). An unmapped locale yields `undefined`, which omits the param and searches all languages. Pagination resets when the query changes.
  - **Result rendering.** `result.title` highlighted client-side via `highlightText()`; `result.summary` is server-highlighted. Emphasis markup renders as `<span class='highlight'>` (styled green).
  - **Rate-limit handling.** A `RateLimitError` surfaces "Rate limited. Try again in Ns." from `err.rateLimit.resetInSeconds`; other failures show a generic error.
  - Pagination controls use the theme `Button` (`variant=OUTLINE`, `size=SMALL`) with native `disabled`, avoiding a disabled-`<Link>` a11y regression.
- **`package.json` / `yarn.lock`** — `@newrelic/gatsby-theme-newrelic` 9.15.0 → 9.17.0.

## Verification

- Confirmed the installed 9.17.0 `search()` return shape (`{ results, nextCursor, prevCursor, totalCount }`; results carry `url` / `title` / `summary`) and `highlightText()` markup (`<span class='highlight'>`) match the page's expectations.
- Lint passes.
- ⚠️ Runtime search can't be exercised from `localhost` — the SearchGPT service whitelists `docs.newrelic.com` (CORS), not localhost. **Verify exact `totalCount`, highlighted titles + summaries, Prev/Next cursor walk (including the reset-on-new-query behavior), language scoping on localized sites, and error/empty/rate-limited states on the deploy preview / production.**

## Env vars (build env + local `.env`)

```
GATSBY_SEARCHGPT_API_KEY=NRAK-...                                     # required
GATSBY_SEARCHGPT_BASE_URL=https://support-search.service.newrelic.com # optional (default)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "@emotion/styled": "^11.3.0",
     "@mdx-js/mdx": "2.0.0-next.8",
     "@mdx-js/react": "2.0.0-next.8",
-    "@newrelic/gatsby-theme-newrelic": "9.15.0",
+    "@newrelic/gatsby-theme-newrelic": "9.16.0",
     "@splitsoftware/splitio-react": "^1.2.4",
     "ansi-colors": "^4.1.3",
     "cockatiel": "^3.0.0-beta.0",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "@emotion/styled": "^11.3.0",
     "@mdx-js/mdx": "2.0.0-next.8",
     "@mdx-js/react": "2.0.0-next.8",
-    "@newrelic/gatsby-theme-newrelic": "9.16.0",
+    "@newrelic/gatsby-theme-newrelic": "9.17.0",
     "@splitsoftware/splitio-react": "^1.2.4",
     "ansi-colors": "^4.1.3",
     "cockatiel": "^3.0.0-beta.0",
--- a/src/pages/search-results.js
+++ b/src/pages/search-results.js
@@ -9,72 +9,78 @@ import {
   search,
   Spinner,
   Surface,
-  useLocale,
+  highlightText,
   useQueryParams,
 } from '@newrelic/gatsby-theme-newrelic';
 import { navigate } from '@reach/router';
 
-import { usePagination, DOTS } from '../hooks/usePagination';
-
-const SearchResultPageView = ({ pageContext }) => {
+const SearchResultPageView = () => {
   const { queryParams } = useQueryParams();
   const query = queryParams.get('query');
-  const page = Number(queryParams.get('page') ?? 1);
-  const locale = useLocale();
-  const [results, setResults] = useState({ loading: true });
-  const { records, pageCount, loading, error } = results;
-  const { slug } = pageContext;
 
-  const totalPages = pageCount;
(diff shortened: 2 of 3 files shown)


akashreddy-nr:
netlify build

akashreddy-nr:
netlify build

akashreddy-nr:
https://searchgpt-search-results--docs-website-netlify.netlify.app/

Nandu-pns (approved):

Nandu-pns:
One thing I want to flag on the two effects at the top — I think there's a subtle race between them.

`cursor`/`cursorStack` get reset in their own effect keyed on `query`, and the fetch runs in a separate effect keyed on `[query, cursor, language]`. The problem is these are two independent effects, so when `query` changes while someone's sitting on page 2+, they both fire in the same render pass — but the fetch effect still closes over whatever `cursor` was *before* the reset, since `setCursor(undefined)` hasn't actually committed yet at that point. So in practice: user is on page 3 of "foo", types "bar" into search, and we fire off a request for "bar" using page-3's cursor before immediately re-firing with the correct `cursor: undefined` a beat later. It self-corrects, but there's a wasted (and possibly weird/rejected) request in between, which feels worth avoiding given we're already handling rate limits explicitly elsewhere in this same change.

Simplest way I can think to close this: keep a ref to the previous `query` and, at the top of the fetch effect, use `query !== prevQueryRef.current ? undefined : cursor` as the cursor for that request rather than trusting the `cursor` state value directly whenever the query just changed. That way the reset and the fetch stay in sync within a single effect run instead of depending on the reset effect having already committed.

PR #24668, opened 2026-07-02
chore(Navigation): Remove overlay APIs from SDK navigation docs

mrodriguez-nr:
This PR removes the overlay methods from the `navigation` platform API page because they will be removed from the SDK after all use cases have been migrated to the new `openNerdlet` method.

I made the same removals in the French translation of the page so both stay in sync. All other navigation methods and type definitions are untouched.

The methods will continue working but only for current consumers, we have a EOL allowlist so it's not possible for new consumers to use those methods.


sujitnewrelic:
netlify build fork

sujitnewrelic (approved):
LGTM

PR #24672, opened 2026-07-02
chore: Raise the minimum Ruby version

kaylareopelle:
The initial hybrid agent release for Ruby didn't include support for Span Links or Span Events. Now that we have support for those, raise the displayed minimum version to that value.


sujitnewrelic (approved):
LGTM

PR #24736, opened 2026-07-07
chore: Added new supported node versions to table as well as specified when node 20 was dropped

bizob2828:
## Give us some context

Node.js agent just released support for node 26. We also forgot to update table when we dropped support for node 20. Lastly, I added node 26 to EOL table with a future date of april 2029


jsumners-nr (approved):

tejaswaroop-nr:
Netlify build fork

tejaswaroop-nr (approved):
LGTM!

PR #18484, opened 2024-08-26
Updating Support Doc

mikayla-mcgraw-newrelic:
Please follow [conventional commit standards](https://www.conventionalcommits.org/en/v1.0.0/)
in your commit messages and pull request title.

## Give us some context

* What problems does this PR solve?
 Updating support doc with latest screenshot of home page and removing reference to support portal because it has been EOL'd
* Add any context that will help us review your changes such as testing notes,
  links to related docs, screenshots, etc.
* If your issue relates to an existing GitHub issue, please link to it.

PR #24792, opened 2026-07-13
fix(nav): point find-help nav entry at canonical slug

tejaswaroop-nr:
Updates the 'Our support portal' nav entry from the old redirected slug find-help-use-support-portal to the canonical find-help-get-support (renamed in PR #18484 but nav was never updated). Nav-only. NR-582852.

Note: the OTel Kubernetes collector intro was dropped from this PR — it's an intentional orphan (retained but not added to nav).


tejaswaroop-nr:
Netlify build

WriteMayur (approved):
lgtm

PR #24799, opened 2026-07-13
fix: Remove contradictory, irrelevant callout in alert event attributes doc

relic-js:
We should remove the **IMPORTANT** callout on this page, as the two sentences contained therein do not make any sense:

> The alert event is a concept used to determine alerting features.

I have no idea what this could possibly mean. That is certainly not the definition of an alert event — which, I should note, is *correctly* defined under the **What is an alert event?** header).

> While you can query some of its associated attributes via NerdGraph, you cannot directly query the alert event.

This is untrue and contradictory, especially in the context of a document that gives you tips on querying alert events.

Given the above, I have just removed the callout altogether. I think it may be a vestige of a much older version of this doc.


WriteMayur (approved):
thanks for updating it.

PR #24857, opened 2026-07-17
chore: agent control release notes

danielorihuela:
## Give us some context

Adds release notes for Agent Control.


akashreddy-nr (dismissed):

src/content/docs/release-notes/agent-control-release-notes/agent-control-2026-07-06.mdx
```diff
@@ -0,0 +1,28 @@
...
+- On-host agents: removed command-based version checking. Agent version is now determined from OCI package metadata, eliminating the need for `deployment.version` configuration in agent type definitions.
+- Replace filesystem in on-host agent-type definitions with an explicit, recursive, tagged-kind tree: every entry declares `kind: file | dir | dir_content_from_map`, and `dir` entries nest via `entries:`.
+- On-host filesystem entries now accept a `persistent` flag (default `false`): ephemeral entries are deleted on sub-agent stop, persistent entries survive until the agent is removed from the fleet. Reconciliation across writes is driven by a `.ac-managed-paths.json` manifest (reserved filename — agent types must not declare it) so paths Agent Control no longer owns are deleted while sub-agent-created files are preserved.
+- Include new 'shared-filesystem-dir` variable for Agent Types.
+- Extend internal `fs` crate with a copy operation.
+- Removed the filesystem manifest-based prune mechanism; Agent Control no longer deletes previously-declared paths that a config later drops. Agent-managed files are reclaimed only when the agent is removed from the fleet.
```
WriteMayur:
```suggestion
- Added exponential backoff and jitter retries to OCI artifact fetches using a new `BackoffPolicy` (configured under `self_update.download_retry`), replacing the previous `with_retries(usize, Duration)` API.
- Added post-download script support for OCI packages.
- Hardened service restart policies on Linux and Windows platforms (implementing `systemd` rate limiting: maximum of 5 restarts within 60 seconds) to prevent crash-looping from saturating CPU capacity.
- Added support for retrieving remote agent type definitions directly from OCI registries.
- Configured agent type definitions to tolerate unknown fields to ensure forward compatibility.
- Added support for creating on-host agents without defining a health check, which disables health reporting for that agent.
- Streamlined version tracking for on-host agents by deriving versions directly from OCI package metadata. This eliminates command-based version checking and removes the need for the `deployment.version` configuration in agent type definitions.
- Replaced the `filesystem` configuration in on-host agent type definitions with an explicit, recursive, tagged-kind tree structure where every entry declares `kind: file | dir | dir_content_from_map` and `dir` entries nest using `entries:`.
- Introduced a `persistent` flag (default value is set to `false`) for on-host filesystem entries; ephemeral entries are deleted when a sub-agent stops, while persistent entries remain until the agent is removed from the `fleet.Reconciliation` across writes is driven by a `.ac-managed-paths.json` manifest (reserved filename — agent types must not declare it) so paths Agent Control no longer owns are deleted while sub-agent-created files are preserved.
- Introduced the `shared-filesystem-dir `variable for agent type configurations.
- Extended the internal `fs` crate to support file copy operations.
- Removed the filesystem manifest-based prune mechanism; Agent Control no longer deletes previously-declared paths dropped by subsequent configurations; agent-managed files are now reclaimed only when the agent is removed from the fleet.
```

src/content/docs/release-notes/agent-control-release-notes/agent-control-2026-07-06.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+## Fixes
+
+- On-host self-update: an empty `version: ""` pushed from Fleet Control now behaves the same as an absent `version` field (no-op, no update attempted). Previously it silently triggered a pull of the `:latest` OCI tag.
```
WriteMayur:
```suggestion
- Fixed an on-host self-update issue where pushing an empty `version: ""` string from Fleet Control incorrectly triggered an image pull of the `:latest` OCI tag. It now correctly treats an empty string as a no-op (), matching the behavior of an absent version field.
```

WriteMayur:
netlify build fork

WriteMayur (approved):
looks good

PR #24893, opened 2026-07-20
Add 10.x min Ruby version to table

kaylareopelle:
Now, the agent only supports Ruby versions 2.6 and above. We missed this in our documentation updates released as part of 10.0.0.


adutta-newrelic:
netlify build fork

adutta-newrelic (approved):
LGTM

PR #24915, opened 2026-07-21
chore: Add a note that Node agent only supports Hybrid Agent for NextJS >=16

amychisholm03:
- Updates casing for the Hybrid Agent
- Adds a note that this approach is only for Next.js versions 16 and above.


adutta-newrelic (approved):
LGTM

PR #24920, opened 2026-07-22
fix(dnt): fix DNT tagging in AWS manual integration doc

sujitnewrelic:
### Removed `<DNT>` from proper nouns (brand/product names globally recognized, should not carry a DNT tag):
- `AWS` (all standalone occurrences)
- `CloudFormation` (all standalone occurrences)
- `Terraform`
- `Firehose` (standalone in "Kinesis Data Firehose")

### Added `<DNT>` to AWS parameter values (exact UI strings that must not be translated):
- `Disabled` — value for **Data transformation** and **Record format conversion** fields
- `Failed data only` — value for **S3 backup mode** field
- `Create or update IAM role` — option under **Service access**


adutta-newrelic (approved):
LGTM

PR #24954, opened 2026-07-23
add best practices

keegoid-nr:
Add missing best practices so customers can self-serve.
These best practices stem from knowledge gained from real customer cases.


cbehera-newrelic:
Netlify build fork

cbehera-newrelic (approved):
lgtm

PR #24982, opened 2026-07-27
Revert sharp bump to 0.30.7 (fixes flaky dependency install failures)

akashreddy-nr:
## Summary
- Reverts #24958, which bumped the top-level `sharp` dependency from `^0.30.7` to `^0.35.0` via Dependabot.
- `gatsby-plugin-sharp`, `gatsby-transformer-sharp`, `gatsby-plugin-manifest`, and `gatsby-sharp` all still require `sharp@^0.30.7`. Bumping only the top-level pin broke yarn's ability to hoist a single shared `sharp` install, forcing 5 separate nested `sharp@0.30.7` installs instead of 1.
- Each of those nested installs independently downloads and checksums the same `libvips-8.12.2-linux-x64.tar.br` binary from GitHub during `yarn install`, multiplying exposure to a transient corrupted-download failure ("Integrity check failed for linux-x64") that broke both the `develop` and production (`main`) builds.
- Reverting restores the single hoisted `sharp@0.30.7` install (one libvips download instead of five), which should make the dependency-install step reliable again.

## Follow-up
- The `sharp` bump can be revisited once `gatsby-plugin-sharp`/`gatsby-transformer-sharp`/`gatsby-plugin-manifest`/`gatsby-sharp` support `sharp@0.35.x`.

## Test plan
- [ ] Confirm `yarn install` succeeds cleanly with a single hoisted `sharp@0.30.7`
- [ ] Confirm Netlify build passes on this branch

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "remark-rehype": "^8.0.0",
     "rss": "^1.2.2",
     "sequelize": "^6.37.8",
-    "sharp": "^0.35.0",
+    "sharp": "^0.30.7",
     "superstruct": "^0.16.5",
     "turndown": "^6.0.0",
     "unist-util-filter": "^2.0.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2236,13 +2236,6 @@
   dependencies:
     "@jridgewell/trace-mapping" "0.3.9"
 
-"@emnapi/runtime@^1.11.0":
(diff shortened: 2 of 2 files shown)


akashreddy-nr:
netlify build

WriteMayur (approved):
lgtm

PR #24990, opened 2026-07-28
retire: remove Mobile Crash REST API v1 doc and add redirects

cbehera-newrelic:
Removes the deprecated Mobile Crash REST API v1 documentation page as requested. The v1 API is 6+ years old and has been superseded by v2. Adds redirects for both the primary URL and the existing /docs/apis/... alias to the mobile monitoring intro page.


WriteMayur (approved):
lgtm

PR #25138, opened 2026-08-05
feat(search): same-origin Netlify Function proxy for SearchGPT (v2)

akashreddy-nr:
## Summary

Adds `netlify/functions/searchgpt-proxy.mjs`, bound to `/v2/*` via `config.path`. It injects the `api-key` from the `SEARCHGPT_API_KEY` env var **server-side** and forwards to `support-search.service.newrelic.com`, so the key never reaches the browser. The client points at it with `GATSBY_SEARCHGPT_BASE_URL="/"`.

## Why

`GATSBY_`-prefixed vars are inlined into the public client bundle at build time. Routing search through a same-origin proxy means there is nothing key-shaped in the client bundle to leak — the durable fix for the earlier key exposure.

## Hardening

- **GET-only** — 405 otherwise (read-only API surface; can't be turned into a write path).
- **Fail closed** — 500 if `SEARCHGPT_API_KEY` is unset, rather than forwarding unauthenticated.
- **Path allowlist** — only `/v2/*` is forwarded (404 otherwise); any client-sent `api-key` is dropped.
- **Shared CDN cache** — `public, max-age=60, stale-while-revalidate=300` collapses identical typeahead/result GETs into far fewer upstream calls, which also keeps the shared service key under its per-principal rate limit.

## Deploy prerequisites (Netlify env, not committed)

- Set `SEARCHGPT_API_KEY` to a **valid non-personal scoped** service key in the Netlify UI.
- Set `GATSBY_SEARCHGPT_BASE_URL="/"`.
- Leave `GATSBY_SEARCHGPT_API_KEY` **unset** in production.

## Companion PR

Requires the proxy-aware client: newrelic/gatsby-theme-newrelic#1178 (released via a theme bump).

## Notes

- No key literals committed (repo's check-for-keys CI applies).
- Swiftype remains untouched.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -4,6 +4,11 @@ package = "@netlify/plugin-gatsby"
 [functions]
   included_files = ["!.cache/data/datastore/data.mdb","!.cache/query-engine"]
 
+# SearchGPT search proxy: netlify/functions/searchgpt-proxy.mjs serves /v2/* (its
+# route is declared via `config.path` in the function). It injects the api-key
+# from the SEARCHGPT_API_KEY env var server-side so the key never ships to the
+# browser. The client points at it via GATSBY_SEARCHGPT_BASE_URL="/".
+
 [[headers]]
 for = "/*"
 
--- a/netlify/functions/searchgpt-proxy.mjs
+++ b/netlify/functions/searchgpt-proxy.mjs
@@ -1,0 +1,77 @@
+// Same-origin proxy for the SearchGPT REST API (v2).
+//
+// The docs search client (gatsby-theme-newrelic/src/utils/searchGPT.js) is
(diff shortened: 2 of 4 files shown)


pranav-new-relic (dismissed):
(let us unset `GATSBY_SEARCHGPT_API_KEY` before merging this)

⚠️⚠️⚠️⚠️

akashreddy-nr:
netlify build

akashreddy-nr:
netlify build

pranav-new-relic (approved):

akashreddy-nr:
netlify build

pranav-new-relic:
netlify build

pranav-new-relic (approved):

PR #25139, opened 2026-08-05
feat(search): same-origin Netlify Function proxy for SearchGPT (v2)

akashreddy-nr:
## Summary
- Adds `netlify/functions/searchgpt-proxy.mjs`, bound to `/v2/*` via `config.path`. It injects the `api-key` from the server-side `SEARCHGPT_API_KEY` env var and forwards to `support-search.service.newrelic.com`, so the key never ships to the browser. The client points at it via `GATSBY_SEARCHGPT_BASE_URL="/"`.
- Hardening: GET-only (405 otherwise), fail-closed 500 if the key is unset, only `/v2/*` is forwarded (404 otherwise), any client-sent `api-key` header is dropped. Responses carry a 60s shared CDN cache (`stale-while-revalidate=300`) to collapse repeated typeahead/result GETs and stay under the per-principal rate limit on the shared key.
- Bumps `@newrelic/gatsby-theme-newrelic` to `9.18.0` to pick up the matching proxy-aware client (`GATSBY_SEARCHGPT_BASE_URL` resolution + omitting `api-key` when unset) and the debounced `suggest()`.

Verified live against the `searchgpt-search-results` deploy preview: browser calls same-origin `/v2/search` and `/v2/search/suggest` with no `api-key` header, and the function proxies to the real service and returns real results (see PR #25138).

## Test plan
- [ ] Confirm `SEARCHGPT_API_KEY` is set in Netlify's Production env (server-side only, never `GATSBY_`-prefixed)
- [ ] After deploy, search on the live site and confirm requests go to same-origin `/v2/*` with no `api-key` header in the browser
- [ ] Confirm the public JS bundle contains no API key (`grep -r NRAK public/`)

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -4,6 +4,11 @@ package = "@netlify/plugin-gatsby"
 [functions]
   included_files = ["!.cache/data/datastore/data.mdb","!.cache/query-engine"]
 
+# SearchGPT search proxy: netlify/functions/searchgpt-proxy.mjs serves /v2/* (its
+# route is declared via `config.path` in the function). It injects the api-key
+# from the SEARCHGPT_API_KEY env var server-side so the key never ships to the
+# browser. The client points at it via GATSBY_SEARCHGPT_BASE_URL="/".
+
 [[headers]]
 for = "/*"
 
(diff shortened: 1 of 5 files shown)


pranav-new-relic (approved):

pranav-new-relic:
Nice

PR #25154, opened 2026-08-06
docs: list metrics-only GCP resource types in GCP v2 (WIF) integration docs

cnivyanthsai:
## Problem

The GCP v2 integration (Workload Identity Federation) **polls more resource types than it synthesizes entities for**. Those extra resource types ingest dimensional metrics into the `Metric` event type, but never appear in the entity explorer — so today there is no way for a reader to discover the data exists.

For example, Compute Engine also ingests `gcp.compute.*` metrics for Storage Pools, Node Groups, Service Attachments, VPC Networks and 12 more. None of them are documented anywhere.

## Change

Each affected integration doc gains a `#### Metrics-only resource types` subsection inside its existing `## Workload Identity Federation` section, listing **Resource | Resource type | Metric namespace | Launch stage**, plus an example NRQL query and a launch-stage caveat.

The existing `#### Entities` tables are **not touched** — this is additions only (0 deletions).

| Doc | Metrics-only rows |
|---|---|
| `google-compute-engine-monitoring-integration.mdx` | 16 |
| `google-bigquery-monitoring-integration.mdx` | 8 |
| `google-vertexai-monitoring-integration.mdx` | 6 |
| `google-kubernetes-engine-monitoring-integration.mdx` | 5 |
| `google-cloud-storage-monitoring-integration.mdx` | 3 |
| `google-cloud-dataproc-monitoring-integration.mdx` | 3 |
| `google-cloud-bigtable-monitoring-integration.mdx` | 2 |
| `google-cloud-load-balancing-monitoring-integration.mdx` | 1 |
| `google-cloud-firestore-monitoring-integration.mdx` | 1 |
| `google-cloud-router-monitoring-integration.mdx` | 1 |

The other 20 GCP integration docs are unchanged — every resource type in their datasource already synthesizes an entity, so their `#### Entities` table is already the complete list.

## How the data was derived

Cross-referenced `gcp-polling-v2/src/main/resources/datasource-mappings.yml` (every polled resource type) against `gcp-entity-definitions.yml`, then **re-confirmed against the `newrelic/entity-definitions` repo** that each type genuinely has no registered entity type.

- **Excluded:** `compute.googleapis.com/BackendService` and `RegionBackendService` — flagged `auxiliary: true`, fetched only to resolve a URL map's `loadBalancingScheme`. They emit no metrics and no entities.
- **Launch stage** comes from the per-datasource BETA annotations in `datasource-mappings.yml`, with an explicit "as of the time of writing" caveat in each doc.

## Reviewer notes

**1. Metric namespace is not always the sibling entity's namespace.** Metrics-only types have no entity definition, so no `metric_prefix` override applies and the namespace derives from the resource type itself. Two places where this is counter-intuitive, and where I added inline callouts:

- Cloud Router → Network Region is `gcp.compute.*`, **not** `gcp.router.*` (that override applies only to `compute.googleapis.com/Router`)
- Cloud Load Balancing → Subnetwork is `gcp.compute.*`, not `gcp.loadbalancing.*`

**2. BigQuery** has two `Project` rows (BI Engine, Storage API). They share the `bigquery_project` monitored resource with the existing BigQuery Project *entity*, separated only by metric namespace. The table shows their full asset types to avoid reading as duplicates, with a callout explaining the split.

**3. GKE — possible pre-existing inaccuracy, deliberately left alone.** The doc's callout states that under WIF the integration "collects GKE metrics but does not create entities". However `newrelic/entity-definitions` registers `GCPKUBERNETESCONTAINER` and `GCPKUBERNETESNODE`, so Container and Node do appear to synthesize. Separately, `GCPKUBERNETESPOD` is declared in gcp-polling-v2 but has no entity type in `entity-definitions`, so Pod does not synthesize.

I did **not** reword that callout — it reads as a deliberate product recommendation (steering users to the New Relic Kubernetes integration), and the GKE table is written to be consistent with it. **Flagging for the GKE/Kubernetes team to confirm** whether the callout needs updating.

## Not covered — follow-up

Three metrics-only datasources have no integration doc at all today:

| Datasource | Resource type | Metric namespace |
|---|---|---|
| `gcp_istio` | `istio.io/Service` (on `cloud_run_revision`) | `gcp.istio.io.*` |
| `gcp_firebase_app_hosting` | `firebaseapphosting.googleapis.com/Backend` | `gcp.firebaseapphosting.*` |
| `gcp_firebase_vertex_ai` | `firebasevertexai.googleapis.com/Model` | `gcp.firebasevertexai.*` |

## Testing

- Structural validation across all 10 files: JSX tag balance, 4 cells per row, no empty cells (the repo has zero `<td />` / `<td></td>` under `src/content/docs/infrastructure/`), launch stage ∈ {GA, BETA} — all pass.
- Automated cross-check of every committed row against the source YAMLs: counts and namespaces match exactly for the 9 non-GKE docs.
- `git diff --numstat`: 10 files, 808 insertions, **0 deletions**.
- ⚠️ `yarn verify-mdx` was **not** run locally (no `node_modules` in my checkout) — relying on CI for the MDX build and render check.


PallaviWrite:
@cnivyanthsai can you please sign the CLA

sarvetiNr (approved):

tejaswaroop-nr:
Netlify build fork

tejaswaroop-nr (dismissed):
LGTM!

tejaswaroop-nr:
Netlify build fork

tejaswaroop-nr (approved):
LGTM!

PR #25161, opened 2026-08-07
docs(scorecards): Clarify accounts and joinAccounts intent

nr-gmd:
Clarifies the accounts and joinAccounts attributes intent so that scorecards users are able to deal with different scenarios and get the expected results when the rule is executed.


PallaviWrite:
netlify build fork

PallaviWrite (approved):
lgtm

PR #25183, opened 2026-08-10
[chore] Update the default Apdex score value for the apdex(attribute, t:) function

sujitnewrelic:
The default Apdex score value was documented as 0.5 seconds, where as actually it is 0.4 seconds. Doing this change as per this request: https://newrelic.slack.com/archives/C0DSGL3FZ/p1786132497595989


sujitnewrelic:
netlify build

tejaswaroop-nr (approved):
LGTM!

PR #25262, opened 2026-08-13
OtelConfig interactive component

WriteMayur:
This PR adds a mechanism similar to AgentConfig. This will help us to document dynamic configuration for Otel.


akashreddy-nr:
## ⚠️ Blocking: this breaks the Java install page's generated config

Tested against the deploy preview: https://mssql-phase2--docs-website-netlify.netlify.app/install/java/

`config.replace(/^\s{2}([a-z][a-z0-9_]*):(\s*)$/gm, '$1:$2')` in [`AgentConfig.js`](https://github.com/newrelic/docs-website/blob/5b6c78081cee14a511ee94f9ceefc51d97a57d66/src/components/AgentConfig.js#L67) strips 2-space indentation from *any* line matching `  key:` with nothing else on the line. That pattern isn't unique to the "extra whitespace" bug it's meant to fix — it also matches legitimately-nested YAML keys that just happen to have no inline value (i.e. they open a nested block).

In [`src/install/config/agent-config/java-newrelic.txt`](https://github.com/newrelic/docs-website/blob/develop/src/install/config/agent-config/java-newrelic.txt), 13 keys are nested one level under `common: &default_settings` with exactly this shape: `application_logging:`, `attributes:`, `transaction_tracer:`, `error_collector:`, `transaction_events:`, `distributed_tracing:`, `span_events:`, `cross_application_tracer:`, `thread_profiler:`, `browser_monitoring:`, `class_transformer:`, `jfr:`, `labels:`.

On the deploy preview, step 4 of the Java install walkthrough renders all 13 of these flush at column 0, while their child keys (e.g. `enabled: true` under `application_logging:`) stay indented deeper. I pulled the rendered text out of `#agent-config-codeblock` and confirmed it — e.g.:

```yaml
  #proxy_scheme: https

  # Limits the number of lines to capture for each stack trace. 
  # Default is 30
  max_stack_trace_lines: 30

  # Provides the ability to configure the attributes sent to New Relic. These
  # attributes can be found in transaction traces, traced errors,  
  # transaction events, and page views.
attributes:
  
    # When true, attributes will be sent to New Relic. The default is true.
    enabled: true
```

`attributes:` is no longer nested under `common:` — it's now a sibling of `common`, `development`, `test`, `production`, `staging`. Running the rendered output through `js-yaml` confirms it's invalid YAML:

```
PARSE ERROR: bad indentation of a mapping entry at line 138, column 3:
      max_stack_trace_lines: 30
      ^
```

So anyone downloading/copying the `newrelic.yml` from the guided Java install gets a config file the agent can't parse.

The regex can't distinguish "erroneously double-indented key" from "correctly-nested key with no inline value" — both look identical as `^\s{2}key:$`. Could you share a repro of the original bug (screenshot/steps of where the extra 2 spaces showed up)? That'll help find a fix that doesn't blanket-strip indentation from every bare nested key in the config.

src/components/MDXContainer.js
```diff
@@ -27,6 +27,7 @@ import MDXButtonGroup from './MDXButtonGroup';
 import ContributorNote from './ContributorNote';
 import MDXTechTileGrid from './MDXTechTileGrid';
 import OptionReference from './OptionReference';
+import OtelConfig from './OtelConfig';
```
akashreddy-nr, line 30: `OtelConfig` import won't resolve on a case-sensitive filesystem


`src/components/OTelConfig.js` (capital "T") is imported in `MDXContainer.js`

Rename the file to `OtelConfig.js` to match the import and the existing `AgentConfig.js`/`AgentConfig` convention.

src/components/OTelConfig.js
```diff
@@ -0,0 +1,112 @@
+import React, { useState } from 'react';
+import { css } from '@emotion/react';
+import PropTypes from 'prop-types';
+import {
+  CustomTextInput,
+  InteractiveOutput,
+} from '@newrelic/gatsby-theme-newrelic';
+import MDXContainer from './MDXContainer';
+
+const OtelConfig = ({ inputOptions, config, tipMdx, onChange, fileName }) => {
```
akashreddy-nr: ## Suggestion: share the implementation, keep the components separate

Makes sense to keep `OtelConfig` as its own component, Right now it's a full copy of `AgentConfig.js` though — same props, JSX, and styles, just a different `containerId` and a config transform. Rather than maintaining two near-identical files, could this be a shared `ConfigBuilder` that both `AgentConfig` and `OtelConfig` wrap?

​```jsx
// AgentConfig.js
const AgentConfig = (props) => <ConfigBuilder {...props} containerId="agent-config-codeblock" />;

// OtelConfig.js
const OtelConfig = (props) => (
  <ConfigBuilder {...props} containerId="otel-config-codeblock" configTransform={otelTransform} />
);
​```

Each still gets used wherever it's fit for, and a future third config type (or a fix to the shared layout/styles) doesn't need to be copy-pasted a third time.

WriteMayur:
netlify build

akashreddy-nr (approved):

akashreddy-nr:
verified the configuration in  https://agent-configu-udpates--docs-website-netlify.netlify.app/docs/opentelemetry/database/mssql/linux-hosted/

PR #25252, opened 2026-08-12
Update wildfly with logmanager scenario

obenkenobi:
This is just an addition to the wildfly doc to add another troubleshooting section.

PR #25268, opened 2026-08-13
Update wildfly with logmanager scenario

sujitnewrelic:
## Summary

- Recreates the contribution from #25252 by @obenkenobi (a new troubleshooting scenario for a JBoss logmanager classloading failure), since we can't push maintainer edits directly to that fork's protected `develop` branch.
- Restructures the "Application fails to start or has other classloading issues" section so each problem's title is its own heading, with dedicated Problem/Solution subheadings underneath — matching the pattern already used in the doc's first section — instead of numbered "Problem 1"/"Problem 2" labels.
- Adds a lead-in sentence noting the app-runtime failure can stem from either problem.
- Fixes a stray quote in the "Fix 2" heading.

Supersedes #25252, which will be closed in favor of this PR.


sujitnewrelic:
netlify build

adutta-newrelic (approved):
LGTM

PR #25302, opened 2026-08-17
Clarify navigation for switching between organizations with the same email

sujitnewrelic:
## Summary
- Rewrites the "Switch between user records" section on `multiple-user-records.mdx` to describe the actual UI path (user menu > **Other users**) instead of the previous generic wording.
- Links first mention of organization to the New Relic account structure doc, notes the option is hidden when a user has access to only one organization, and moves the cross-email log-out/log-in note into a tip callout.

PR #25459, opened 2026-08-28
[chore] Fold in follow-up feedback on switching between user records

sujitnewrelic:
## Summary
- Follow-up to #25302. Hunter Beezley flagged that after the screenshot was replaced with descriptive text, a few details it conveyed visually were missing from the "Switch between user records" section on `multiple-user-records.mdx`.
- Adds that each entry in the **Other users** list is labeled with the organization name and authentication domain.
- Clarifies that a user can have multiple user records for the same organization if it has more than one authentication domain (one record per domain).
- Reorders the procedural sentence to purpose → place → action for clarity.

## Context
See Slack thread with Hunter: https://newrelic.slack.com/archives/C0DSGL3FZ/p1786727648860959


sujitnewrelic:
netlify build

cbehera-newrelic (approved):
lgtm

PR #25622, opened 2026-09-09
Fix api-Key casing to api-key in Claude Code MCP setup config

ubanerjeeNR:
Fixes incorrect header casing in the Claude Code API key JSON config example. `api-Key` should be `api-key` (all lowercase).


WriteMayur (approved):
lgtm

Issue #254, opened 2020-11-10
Create a Docs Site feedback widget and hook it into the Slack #docs-customer-feedback channel


jpvajda:
### User Stories 
- As a user I would like to submit feedback from using the Docs site so Relics can see it and take action. 
- As a Docs writer I'd would like to see this feedback in a new slack channel called : `#docs-customer-feedback`
- As a PM or engineer I'd like to know what our customer think about our new Docs Site.

### AC 
- Push feedback posts into main  #docs-customer-feedback
- Auto generate a JIRA issue based on source of issue

### Implementation & Design details
-  Gather Design input to figure out where this goes on the site. Perhaps we follow the current design pattern in the product. 
- Offer some kind of holistic reporting to see trends across the site (might be a STRETCH item, but Jira offers a lot of reporting options, so it seems we can use that for the time being)

### In Product Experience

Style and interactions should follow the in-product expeirence.  

<img width="395" alt="Screen Shot 2020-11-10 at 1 00 06 PM" src="https://user-images.githubusercontent.com/1395158/98757292-7546d480-2389-11eb-82c8-ac341835bf42.png">

jpvajda:
This will be detailed out in an MMF document as it's large enough to warrant a dedicated block of time to the development o this feature.

PR #25630, opened 2026-09-09
Add Agent Control guide: automating installation at scale without a 12h re-issue

MunhozThiago:
## Summary
- Adds a new page, `new-relic-control/agent-control/automated-installation.mdx`, covering two ways to automate Agent Control installs (Terraform, Ansible, CI/CD) without a person manually re-issuing the guided install's 12-hour system identity credential every 12 hours.
- Links the existing 12h-expiry error note in `setup.mdx` to the new page, since that's the exact error a customer sees today with no pointer to a fix.
- The durable-parent section uses `newrelic-auth-cli` directly for both steps (create the parent, get a per-install token), rather than a bespoke script and raw curl calls, since `create-bootstrap-identity key` already grants the new identity permission to create children in the same call (confirmed in its own source).

## Dependency
This assumes [newrelic-auth-rs#254](https://github.com/newrelic/newrelic-auth-rs/pull/254) (adds `NEW_RELIC_API_KEY` env var support to `create-bootstrap-identity`, so the API key never has to be a CLI argument) lands and ships in a release. That PR is open, not yet merged. The version pin in Step 1's download command has a comment flagging it needs to be bumped to whichever release actually includes it. Until then, the example as written won't work against an older `newrelic-auth-cli` binary, worth holding merge of this PR until #254 ships, or reverting that one command back to `--api-key` if this needs to go out sooner.

## Test plan
- [x] `node scripts/verify_mdx.js` passes on all changed files
- [x] The embedded script in the "installing occasionally" section (and the removed durable-parent script, before it was replaced by direct `newrelic-auth-cli` calls) was verified this session: mocked NerdGraph responses, exit-code correctness, a stdin-hang fix, JP region support
- [x] Parent-token flag confirmed against real source in `newrelic-agent-control` and `open-install-library` (not guessed): `NEW_RELIC_AUTH_TOKEN` → `--auth-parent-token` (Linux/K8s), `-AuthParentToken` (Windows)
- [ ] Confirm with Paolo/Agent Control engineering that the above still matches their read, source-verification is strong but isn't the same as an engineer signing off on public docs
- [ ] Docs team review for house style
- [ ] Legal/Compliance review (Kyle Apfeld asked for this specifically, since it's customer-facing communication, not just a security question)
- [ ] Decide on the #254 dependency above before merging

### DIFF
--- a/src/content/docs/new-relic-control/agent-control/automated-installation.mdx
+++ b/src/content/docs/new-relic-control/agent-control/automated-installation.mdx
@@ -1,0 +1,395 @@
+---
+title: "Automate Agent Control installation at scale"
+metaDescription: "Install Agent Control from Terraform, Ansible, or a CI/CD pipeline without a person manually re-issuing a system identity every 12 hours."
+freshnessValidatedDate: never
+---
+
+When you install Agent Control through [Guided Install](/docs/new-relic-control/agent-control/setup), it creates a system identity for you and shows you a client ID and client secret. That secret expires 12 hours after it's issued. For a single install, that's not a problem. For automating installs with Terraform, Ansible, or a CI/CD pipeline, it is: any run that happens more than 12 hours after the credential was issued fails, and someone has to go back into the UI for a new one.
+
+This page covers two ways to automate installation without that manual step. Both use the same [system identity creation capability](/docs/apis/nerdgraph/get-started/introduction-new-relic-nerdgraph/) Guided Install itself uses, called directly with a New Relic [User API key](/docs/apis/intro-apis/new-relic-api-keys/).
+
+Two rules apply to both approaches below:
+
+- **Each host or install target still gets its own identity**, minted fresh for it.
+- **A private key never travels off the machine it was created on.** If you set up one durable identity that authorizes creating others (the second approach below), its private key stays on whichever machine controls your automation. Only a short-lived access token derived from it, never the key itself, is passed anywhere else.
+
+## Installing occasionally: mint a short-lived credential per run [#per-run]
+
+If you're not installing often enough to justify managing a durable identity, this is the simplest path. Your pipeline authenticates with your API key to mint a throwaway credential, uses it once to run the normal `newrelic install` command, and the installer generates the long-lived identity that Agent Control actually runs on. You never see or store that long-lived identity yourself, the installer writes it straight to the host.
+
+### Linux
+
+```bash
+#!/bin/bash
+set -e
+
+# Set these from your CI secrets store. NEW_RELIC_API_KEY should be a User API
+# key (NRAK-...). It's used once below and never written to disk.
+: "${NEW_RELIC_API_KEY:?}" "${NEW_RELIC_ACCOUNT_ID:?}" "${NEW_RELIC_ORGANIZATION:?}" \
+  "${NEW_RELIC_REGION:?}" "${NR_CLI_FLEET_ID:?}"
+
+NEWRELIC_AUTH_VERSION="0.5.0"
+
+# 1. Download the auth CLI.
+curl -sS -L \
+  "https://github.com/newrelic/newrelic-auth-rs/releases/download/${NEWRELIC_AUTH_VERSION}/newrelic-auth-cli_amd64.tar.gz" \
+  -o /tmp/newrelic-auth-cli.tar.gz
+tar -xzf /tmp/newrelic-auth-cli.tar.gz -C /tmp
+
+# 2. Mint a fresh, throwaway credential for this install only.
+BOOTSTRAP=$(/tmp/newrelic-auth-cli create-bootstrap-identity secret \
+  --name "bootstrap-$(hostname)-$(date +%s)" \
+  --organization-id "$NEW_RELIC_ORGANIZATION" \
+  --environment "$NEW_RELIC_REGION" \
+  --api-key "$NEW_RELIC_API_KEY")
+
+NEW_RELIC_AUTH_CLIENT_ID=$(echo "$BOOTSTRAP" | jq -r '.client_id')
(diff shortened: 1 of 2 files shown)


MunhozThiago:
Added Chef and Puppet, and restructured the "installing occasionally" section into tabs (Linux/Ansible/Chef/Puppet/Windows) for easier navigation.

While researching this, replaced the hand-rolled Ansible example with the real official `newrelic.newrelic_install` role, confirmed `agent-control` is a real target for all three (Ansible, Chef, Puppet), each in their own official New Relic-maintained cookbook/module, just under-documented as such in Ansible's own README.

One honest gap: the Puppet manifest is the one snippet on this page I couldn't run through a real parser (no working `puppet` binary here, system Ruby too old to install one fresh). Written by close comparison against `puppet-install`'s own real `install.pp` class, but worth an actual `puppet parser validate` before this merges. Everything else (4 bash blocks, the Ansible YAML, the Chef recipe) was independently extracted from this exact file and verified: `bash -n`, YAML parse, and `ruby -c` respectively.

MunhozThiago:
Closed the one gap from my last comment: got a real `puppet parser validate` working via Docker (`puppet/puppet-agent` image; worth noting for anyone else on Rancher Desktop, it doesn't share `/tmp` into its VM, files have to be staged under `/Users` for volume mounts to actually work).

Parser validate passed on the original manifest, but I didn't stop there, actually compiled the class with sample values and ran `puppet apply --noop` against it. That caught a real bug: Puppet's `exec` resource doesn't inherit `PATH`, so the bare `test` command in the `unless` guard failed with "not qualified and no path was specified." Fixed by adding `path => ['/usr/bin', '/bin', '/usr/local/bin']`. Re-ran the same compile + noop-apply after the fix, clean exit.

Also tried to get the Chef example through a real `chef-client --local-mode` run (built a minimal cookbook wrapping the actual `chef-install` resource file), but hit a container packaging issue unrelated to the recipe itself (missing shared lib in the image, both native arm64 and forced amd64). Didn't chase it further: Chef's compile-time model is structurally lower-risk for this pattern than Puppet's (no separated compile/apply parameter resolution to get wrong), and it's already `ruby -c` clean plus written by close comparison against the cookbook's own real example in `recipes/default.rb`. Flagging that as the one remaining piece verified by pattern-matching rather than full execution, if anyone reviewing has a working Chef environment handy, worth a quick real run.

MunhozThiago:
Switched all three download examples (Linux, Ansible, Chef — Puppet reuses the Linux script) from a hardcoded `newrelic-auth-cli` version to GitHub's `/releases/latest/download/` redirect, so there's no version to manually bump every time newrelic-auth-rs ships a release. Verified the redirect resolves to the exact same asset as the pinned 0.5.3 (identical content-length), not just that it returns 200.

Added an explicit `--version` print (or, in Ansible's case, a `debug` task) right after download in each example. Worth calling out as a deliberate tradeoff, not just an improvement: always pulling latest means a future newrelic-auth-rs release with a breaking CLI change could break a customer's pipeline with no warning, right after it worked yesterday. The version print doesn't prevent that, but it means a support ticket has something concrete to go on instead of "it stopped working."

One thing this interacts with: Step 1 of the durable-identity section still depends on newrelic-auth-rs#254, which isn't merged yet. Latest today doesn't have that fix, so that step will still fail until #254 ships — reworded the comment to say so plainly instead of the old "bump this tag" instruction, which no longer applies now that there's no pinned tag.

Re-verified after the edit: `bash -n` on all 4 bash blocks, `ruby -c` on the Chef block, YAML parse plus a real `ansible-playbook --syntax-check` via Docker on the Ansible block (parsed cleanly through everything I touched; only failed later on the Galaxy role not being installed locally, which is an environment limitation, not a syntax issue), and the repo's own `verify_mdx.js` on the whole file.

MunhozThiago:
Reverted Step 1 of the durable-identity section back to \`--api-key\` directly. \`newrelic-auth-rs#254\` (the env-var alternative this was waiting on) was closed per a product decision: the flag already works, and no customer asked for the env-var version specifically. Named the ps/shell-history exposure explicitly in the prose now instead of framing it as a temporary gap pending a fix — it's an accepted, deprioritized trade-off, not a bug waiting on a merge.

Also relevant: the second draft PR from this same effort, \`newrelic-auth-rs#259\` (audit logging + advisory rate limit for the parent-mints-child guardrails), was closed too — that work is being redirected to be an ask of IAM rather than something built client-side in that repo. Doesn't change anything on this page directly, flagging for the record since it's the same initiative.

src/content/docs/new-relic-control/agent-control/automated-installation.mdx
```diff
@@ -0,0 +1,316 @@
...
+  -o /tmp/newrelic-auth-cli.tar.gz
+tar -xzf /tmp/newrelic-auth-cli.tar.gz -C /tmp
+/tmp/newrelic-auth-cli --version >&2
+
+# 2. Mint a fresh, throwaway credential for this install only.
+BOOTSTRAP=$(/tmp/newrelic-auth-cli create-bootstrap-identity secret \
+  --name "bootstrap-$(hostname)-$(date +%s)" \
+  --organization-id "$NEW_RELIC_ORGANIZATION" \
+  --environment "$NEW_RELIC_REGION" \
+  --api-key "$NEW_RELIC_API_KEY")
```
sigilioso, line 70: Because `newrelic-auth-cli` includes environment variables support in the [latest release](https://github.com/newrelic/newrelic-auth-rs/releases), we can use `NR_AUTH_API_KEY` as environment variable instead of the `--api-key` parameter. Check https://github.com/newrelic/newrelic-auth-rs/pull/262 for details

WriteMayur:
netlify build fork

PR #25660, opened 2026-09-10
Update VS Code MCP API key config: remove inputs block

ubanerjeeNR:
Simplifies the VS Code API key config by removing the promptString inputs block and using a plain API key placeholder, per updated guidance from engineering.


WriteMayur (approved):
look good to me

PR #25663, opened 2026-09-10
endpoint fix

Philip-R-Beckwith:

## Give us some context
Fixes the domain on for the translation endpoints. 

https://prometheus-api.newrelic.com is for querying NRDB using PromQL syntax. 
https://promql-gateway.service.newrelic.com is for translating Grafana Dashboards into NR dashboards. 

The page being updated is for translating Grafana dashboards, so should use https://promql-gateway.service.newrelic.com instead of https://prometheus-api.newrelic.com


WriteMayur (approved):
thanks for updating and letting us know about it.

PR #25586, opened 2026-09-07
Revert "docs(gcp): document GCPSERVICEATTACHMENT and GCPPRIVATESERVICECONNECTENDPOINT entities"

nr-sparsi:
Reverts newrelic/docs-website#25551

PR #25551, opened 2026-09-03
docs(gcp): document GCPSERVICEATTACHMENT and GCPPRIVATESERVICECONNECTENDPOINT entities

nr-sparsi:
## Relevant information

Graduates GCP Private Service Connect **Service Attachment** and **Private Service Connect Endpoint** from metrics-only to full entities in New Relic, following the precedent set by #25416 (Spanner Database restoring `GCPSPANNERDATABASE`).

Companion PR (NR-605048): entity synthesis in `beyond/gcp-polling-v2#280` (internal) and platform registration in `newrelic/entity-definitions#3218`.

## What changed

`google-compute-engine-monitoring-integration.mdx` (the "VMs" / Compute Engine integration page):

- Added **Service Attachment** (`GCPSERVICEATTACHMENT`, resource type `gce_service_attachment`) and **Private Service Connect Endpoint** (`GCPPRIVATESERVICECONNECTENDPOINT`, resource type `compute.googleapis.com/PrivateServiceConnectEndpoint`) rows to the Compute Engine entities table.
- Removed their two `Metrics-only resource types` collapsers (`metrics-only-service-attachment`, `metrics-only-private-service-connect-endpoint`).
- Relocated their metric tables — content unchanged, 12 producer metrics + 9 consumer metrics — into new `#### Key metrics — Service Attachment` / `#### Key metrics — Private Service Connect Endpoint` sections, matching the existing per-entity `Key metrics` layout used for VM Instance / VM Disk / Interconnect / Interconnect Attachment on the same page.

`src/i18n/content/**` left untouched (translation-managed).

## Test plan

- [x] Verified no content was lost in the collapser → table relocation (metric names/units/descriptions copied verbatim).
- [x] Manual tag-balance check across the whole file (`<table>`/`<tr>`/`<td>`/`<tbody>`/`<thead>` all balanced, pre- and post-edit).
- [x] Cross-checked resource types, metric name NR-prefix derivation, and unit presence against `gcp-polling-v2`'s entity definitions.
- [ ] `verify-mdx` / `vale-lint` CI (not run locally — Node/Yarn toolchain setup wasn't available in this environment; relying on CI).

PR #25685, opened 2026-09-14
docs(gcp): document GCPSERVICEATTACHMENT and GCPPRIVATESERVICECONNECTENDPOINT entities

nr-sparsi:
## Relevant information

Re-introduces the GCP Private Service Connect entity docs that #25586 reverted (the revert of #25551). #25551 was reverted only because the entities weren't live yet — they are **now deployed and verified synthesizing** in New Relic, so this restores the exact approved content of #25551.

Graduates **Service Attachment** and **Private Service Connect Endpoint** from metrics-only to full entities:
- Adds both to the Compute Engine **entity types** table:
  - Service Attachment → `GCPSERVICEATTACHMENT` / `gce_service_attachment`
  - Private Service Connect Endpoint → `GCPPRIVATESERVICECONNECTENDPOINT` / `compute.googleapis.com/PrivateServiceConnectEndpoint`
- Moves their metrics out of "Metrics-only resource types" into **Key metrics** tables under the entities section (12 producer / Service Attachment metrics, 9 consumer / PSC Endpoint metrics).

## Validation

- Entity types + monitored resource types match the running integration's `gcp-entity-definitions.yml`.
- All metric names + units verified against live data: producer set = 12 metrics, consumer set = 9 metrics (exact match).
- Entities confirmed synthesizing in the target account (`entitySearch` returns the GCPSERVICEATTACHMENT + GCPPRIVATESERVICECONNECTENDPOINT entities).

File content is byte-identical to the previously-approved #25551.

Re-applies #25551 (reverted by #25586).


PallaviWrite:
netlify build fork

PallaviWrite (approved):
lgtm

PR #25747, opened 2026-09-16
Add Ruby 4.0 as supported runtime

hannahramadan:
Ruby 4.0 is now supported for lambda.

I also noticed that [newrelic-lambda-extension](https://github.com/newrelic/newrelic-lambda-extension) has been archived and moved to [lambda-extension-rust](https://github.com/newrelic/newrelic-lambda-extension-rust). I updated a few of those references, however there may be additional places we should consider updating those links.


PallaviWrite:
netlify build fork

PallaviWrite (approved):
LGTM

PR #25772, opened 2026-09-17
Fall back to English markdown for Copy for LLM/View as Markdown on translated pages

akashreddy-nr:
## Summary
- `MdxLlmTools` (the "Copy for LLM" / "View as Markdown" toolbar) built its markdown URL from the page's own pathname regardless of locale, but the llms-txt build query explicitly excludes i18n content - only English pages get a generated `.md` file. Clicking either control on a translated page (e.g. `/jp/docs/...`) 404'd.
- `docPage.js` already computes an English-equivalent path for `MachineTranslationCallout` (stripping the locale prefix via `pageContext.locale`). Reused that same computation for `MdxLlmTools` on a translated page instead of the page's own pathname, and added an `isEnglishFallback` flag so both controls read `(English)` with an explanatory tooltip - not a silent language switch.
- Only `docPage.js` needed the change - What's New and EOL announcements are never localized in this repo, so they can't hit this bug.

## Test plan
- [x] Confirmed the actual bug live: `/jp/docs/apm/agents/java-agent/troubleshooting/all-hosts-appear-localhost/` returns 200, but its `.md` counterpart returned 404 before this fix.
- [x] Verified `pageContext.locale`'s value convention (`'en'` for English, real locale code otherwise) by tracing it to the theme's own `addLocale` page-transform.
- [x] Rendered the real component directly (`react-dom/server`) for both an English and a translated page and inspected the output HTML - confirmed the `(English)` label, the tooltip, and that the link points at the real, existing English `.md` path rather than the broken locale-prefixed one.
- [x] Lint clean on both changed files.
- [ ] Manual click-through on a live deploy preview (not yet done in this session - recommend before merging).


akashreddy-nr:
netlify build

PR #25777, opened 2026-09-17
fix: inline standalone <code> tag breaking FR/KR/JP builds

akashreddy-nr:
## Summary

Same fix as #25776, ported to \`develop\` so it doesn't drift back out of \`main\` on the next promotion.

\`main-fr\`/\`main-kr\` builds were crashing (\`children.trim is not a function\` in \`CodeBlock.js\`) because the translated Security RX Agent overview page has \`<code>@claude fix</code>\` on its own line with blank lines around it, which makes remark compile its children to a \`<p>\` element instead of a string. Fixed by putting it back inline, matching the EN source.

## Test plan
- [x] Same change already verified via local repro + direct MDX-compile inspection for #25776.


akashreddy-nr:
netlify build

PR #25778, opened 2026-09-17
fix:  Remove CCU definition link from aCCU section. 

DataNerdRichard:
aCCU shouldn't link to the CCU definition. This looks like it was a find/replace. But, as it was in the aCCU section Advnaced Compute Capacity Units had Compute Capacity Units linked to the Compute Capacity Units. This also implies a non-extant realationship.


PallaviWrite:
@DataNerdRichard can you please sign the CLA

